PR #2, opened 2016-04-09
Adds arbitrary date parsing

tkroman:
Inspired by a recent discussion @ slack channel, where a user was complaining that he can't compare dates stored as strings in a way other than lexicographical.
This PR adds a procedure that accepts a (string) value and an optional `DateFormat`-compatible pattern and returns a unixtime of date parsed from value with the given format. null input => null output.

Real-life example:

```
MATCH (p:Person) 
WITH p 
CALL apoc.date.unixtime(p.born) 
YIELD value AS dob
RETURN p 
ORDER BY dob
```

will order people by dates given that their 'born' field are compatible with the default `EEE MMM dd HH:mm:ss zzz yyyy` format.

Since procedure methods can't be overloaded, there's a difference in names: `unixtime(date)` vs `unixtime_f(date, format)`, which is totally disputable.

### DIFF
--- a/src/main/java/apoc/date/Date.java
+++ b/src/main/java/apoc/date/Date.java
@@ -1,0 +1,139 @@
+package apoc.date;
+
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoField;
+import java.time.temporal.TemporalAccessor;
+import java.time.temporal.TemporalQueries;
+import java.time.temporal.TemporalQuery;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.TimeZone;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+import apoc.result.LongResult;
+import apoc.result.MapResult;
+import apoc.result.StringResult;
+import org.neo4j.procedure.Name;
+import org.neo4j.procedure.Procedure;
+
+
+/**
+ * @author tkroman
+ * @since 9.04.2016
+ */
+public class Date {
+	private static final String DEFAULT_FORMAT = "yyyy-MM-dd HH:mm:ss";
+	private static final int MILLIS_IN_SECOND = 1000;
+	private static final String UTC_ZONE_ID = "UTC";
+	private static final List<TemporalQuery<Consumer<Map<String, Object>>>> DT_FIELDS_SELECTORS = Arrays.asList(
+			temporalQuery(ChronoField.YEAR),
+			temporalQuery(ChronoField.MONTH_OF_YEAR),
+			temporalQuery(ChronoField.DAY_OF_MONTH),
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/date/Date.java
```diff
@@ -0,0 +1,43 @@
...
+		try {
+			return Stream.of(new LongResult(getFormat(pattern).parse(dateField).getTime() / 1000));
+		} catch (ParseException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	private static DateFormat getFormat(String pattern) {
+		if (pattern == null) {
+			return new SimpleDateFormat("EEE MMM dd HH:mm:ss zzz yyyy");
```
jexp: Should we use the ISO8601 time format as default? or sth like `"yyyy-mm-dd HH:MM:SS"` ?

src/test/java/apoc/date/DateTest.java
```diff
@@ -0,0 +1,119 @@
...
+		TestUtil.registerProcedure(db, Date.class);
+	}
+
+	@After
+	public void tearDown() {
+		db.shutdown();
+	}
+
+	@Test public void testDateAccessor() throws Exception {
+		try (Transaction tx = db.beginTx()) {
```
jexp: you can simplify the test by just passing in the string value a parameter value. No need to create nodes and such.

jexp:
It would be cool to have the inverse function (unix-time->long) too.

And perhaps something that splits either date type into a map format with entries for all parts.

Could you add the section / function(s) to the readme too?

tkroman:
@jexp, I'll simplify tests and add docs.
On account of default format, "yyyy-mm-dd HH:MM:SS" seems reasonable, although I choose the current one on account that it's the one produced by `Date#toString()`, so I'm not 100% sure.
On account of the 

> inverse function (unix-time->long) too.

do you mean unixtime to date?

jexp:
we shouldn't care about Java internal formattings :) neither in Cypher nor Neo 

yes unixdate to string date

tkroman:
@jexp I addressed your comments, all except the one about parts. I'm not sure how to approach it in a universal way (looks like there's no nice way to iterate through `Calendar`'s values), so I guess, I'll have to think on that a little.

jexp:
I'd like to merge it, but love to have nicer names.
(Also the names for procedures are camel-case not lower-case-underscore).

I love if the names flow, like `apoc.date.parse` or `apoc.date.toString`, `apoc.date.from/toUnix`,`apoc.date.from/toMillis, or `apoc.date.unixToString`.

I'm also not sure if we should put "unix" in the name? Perhaps rather seconds/millis (and document that they are since Jan 1st 1970). Btw. what for parse / format we need to set a locale and timezone too. I'm not sure that it defaults to UTC, I think it defaults to the local TZ.

tkroman:
@jexp, I've addressed all the comments.
The only conceptually questionable thing is the `apoc.date.listFields` now. I decided not to anchor pattern names to values (i.e. `yyyy` => `2016`, `MM` => `Jan`), but rather use a universal `Months`, `Seconds` etc.

jexp:
Amazing thanks so much. Looks great. I'll change listFields to just fields and lowercase the field names.

jexp:
merged manually

PR #5, opened 2016-04-11
[WIP] JS evaluation

tkroman:
My take at JS evaluation. 
Some questions I'd like to discuss, though.
I'm not sure about the security implications (file access etc).
Another question: should we populate the global script engine context with shortcuts for methods like `n.foo(bar)` for nodes/rels etc. so users could do 

```
CALL evalWithParams('n.x', {n: someNode})
```

instead of

```
CALL evalWithParams('n.getProperty("x")', {n: someNode})
```
?

Next possible thing on the TODO-list would be adding compiled script support, something like this:

```
CALL apoc.js.compile('function id(n) { return n.getId() } ')
CALL apoc.js.call(id(n), {n: personNode})
```
This can potentially allow for 'unlimited' extensibility in runtime without restarts at all.

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -290,6 +290,10 @@ call apoc.path.expand(p,"ACTED_IN>|PRODUCED<|FOLLOWS<","+Movie|Person",1,2) yiel
 return p, pp 
 ----
 
+=== Javascript evaluation
+call apoc.js.eval('1 + 1') => 2
+call apoc.js.evalWithParams('x + y', {x: 1, y: 2}) => 3
+call apoc.js.evalWithParams('n.getProperty("x")', {n: someNode}) // equivalent to n.x
 
 
 == Plans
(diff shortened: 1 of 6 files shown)


tkroman:
TIL evaluating JS became _so_ easy :)

jexp:
Did you see my JS procedures? https://github.com/jexp/neo4j-javascript-procedures

I kept it separate for now due to the security implications

we can use 

```
Bindings b = engine.createBindings();
b.put("db",db);
engine.eval(script,b);
```

In mine I added these globally:

https://github.com/jexp/neo4j-javascript-procedures/blob/master/src/main/java/scripts/Scripts.java#L56

```
                    "function label(s) org.neo4j.graphdb.DynamicLabel.label(s)",
                    "function type(s) org.neo4j.graphdb.DynamicRelationshipType.withName(s)",
                    "function collection(it) { r=[]; while (it.hasNext()) r.push(it.next());  return Java.to(r); }");
```

jexp:
Btw for basic evaluation another idea I have is to use Cypher itself. 

See https://github.com/jexp/neo4j-apoc-procedures/issues/6

tkroman:
@jexp oh no, I haven't. I see you've got the persistence thing done. I've read some posts on securing/sandboxing the Nashorn engine this evening, and the idea of having a full-functional language still seems worthy. If you have preferences toward Cypher, I guess I should close this one and focus on improving the 'stdlib' then.

jexp:
I'm too wary that people will dismiss apoc entirely if we offer "arbitrary" scripting.

We discussed of having a separate repo/project "apoc-scripting" which not only supports JS but also all the other script engines possible.

jexp:
@tkroman please feel free to help us improving: https://github.com/neo4j-contrib/neo4j-script-procedures

PR #24, opened 2016-05-06
spatial - sort collection of paths by distance

ikwattro:
Example : 

```
MATCH (a:City {name:'bruges'}), (b:City {name:'dresden'})
MATCH p=(a)-[*]->(b)
WITH collect(p) as paths
CALL apoc.spatial.sortPathsByDistance(paths) YIELD path, distance
RETURN path, distance
```

### DIFF
--- a/docs/spatial.adoc
+++ b/docs/spatial.adoc
@@ -1,7 +1,10 @@
 = Spatial Functions
 
 The spatial procedures are intended to enable geographic capabilities on your data.
-Currently there is only one procedure _geocode_ which will convert a textual address
+
+=== geocode
+
+The first procedure _geocode_ which will convert a textual address
 into a location containing _latitude_, _longitude_ and _description_. Despite being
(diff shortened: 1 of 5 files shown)


craigtaverner:
Is this procedure required considering the fact that Cypher in 3.0 has a built-in distance function:

```
MATCH (a:City {name:'bruges'}), (b:City {name:'dresden'})
MATCH p=(a)-[*]->(b)
WITH nodes(p) as cities
WITH [x IN range(1,size(cities)-1) | [cities[x-1],cities[x]] ] as pairs, [c in cities | c.name] as names
WITH [pair in pairs | distance(point(pair[0]),point(pair[1])) ] as distances, names
RETURN reduce(distance = 0.0, d in distances | distance + d) as distance, names ORDER by distance
```

Admittedly this is not a super pretty query.....

ikwattro:
Yeah the purpose was a shortcut of the reduce query

PR #25, opened 2016-05-06
Added warmup procedure and documented it in the readme file

SaschaPeukert:
Hey there,

I added a quick warmup procedure using the Kernel API.

Cheers,
Sascha

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -412,13 +412,20 @@ RETURN path, weight
 MATCH (n:Person)
 ----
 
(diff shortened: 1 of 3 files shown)


jexp:
Thanks a lot @SaschaPeukert 
I realized that in 3.0 we can do all of that efficiently in cypher (e.g. using the counts-store, so I rewrote it while merging)

It now also returns statistics about what it did.


Merged manually

PR #31, opened 2016-05-10
Flatmap

sarmbruster:


src/main/java/apoc/map/Maps.java
```diff
@@ -78,4 +76,23 @@
...
+    @Description("apoc.map.flatten(map) yield map - flattens nested items in map using dot notation")
+    public Stream<MapResult> flatten(@Name("map") Map<String, Object> map) {
+        Map<String, Object> flattenedMap = new HashMap<>();
+        flattenMapRecursively(flattenedMap, map, "");
+        return Stream.of(new MapResult(flattenedMap));
+    }
+
+    private void flattenMapRecursively(Map<String, Object> flattenedMap, Map<String, Object> map, String prefix) {
+        for (Map.Entry<String, Object> entry : map.entrySet()) {
+             if (entry.getValue() instanceof Map) {
```
tkroman: I'd stick a `null` check here, `Map.Entry#getValue` is a notorious source of NPEs :)

sarmbruster: @tkroman don't see where this can cause an NPE. According to https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.20.2 `x instanceof SomeClass` returns false if x is null.

PR #35, opened 2016-05-12
Added parallel node search apoc.search.node*

kvegter:
Michael,

Sorry for the extra readme.adoc commits. I first updated the readme.adoc locally. do you have a good adoc editor client?


jexp:
I just use any text editor, IntelliJ also has an AsciiDoctor plugin and Atom and Sublime too.

PR #47, opened 2016-05-22
Fix the partitioning logic in mapParallel partitionSubList

ADTC:
The original partitioning logic is wrong and causes `AssertionError` when doing `testMapParallel`. The fix is to get ceiling of the quotient so that the partitions will always cover the total number of items in the list. In addition, the `subList` call had to include another `Math.min` call to avoid `IndexOutOfBoundsException`.

**Example:**
Total: 1000
Partitions: 400

_Incorrect batch size logic:_ 1000/400 = (int)2
With each batch having 2 list elements, the `subList` can only reach up to (2*400=) 800 elements.

_Correct batch size logic:_ 1000/400 = 2.5; ceiling of 2.5 = 3
With each batch having 3 list elements, the `subList` can reach up to (3*400=) 1200 elements, enough to cover the total of 1000 elements in the list.

### DIFF
--- a/src/main/java/apoc/cypher/Cypher.java
+++ b/src/main/java/apoc/cypher/Cypher.java
@@ -186,9 +186,9 @@ public class Cypher {
     public Stream<List<Object>> partitionSubList(@Name("list") List<Object> data, int partitions) {
         List<Object> list = new ArrayList<>(data);
         int total = list.size();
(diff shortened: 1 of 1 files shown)


ADTC:
Yay! Build success!

jexp:
Merged manually, thanks

PR #49, opened 2016-05-25
Added new Gradle task to run asciidoctor and build the index.html file

ADTC:
The additional copy task is necessary as asciidoctor cannot directly output to the root of the project in a Windows system due to [file locking](https://discuss.gradle.org/t/error-with-a-copy-task-on-windows/1803/4). Having it output to a build folder then copying from there ensures that the task will run in any system.

*Removed the shell script, as it's no longer needed.*

Note that the index.html is 100% generated by asciidoctor

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,21 +9,21 @@ description = """neo4j-apoc-procedures"""
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
-
-
 repositories {
-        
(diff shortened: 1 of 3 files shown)


ADTC:
Note that the task name is `asciidoctor`, under the `documentation` task group. On Windows: `gradlew.bat asciidoctor` and on Linux: `./gradlew asciidoctor`

PR #50, opened 2016-05-30
Wcc general

tommichiels:
Hi, 

This is my first attempt to create a neo4j stored procedure. I recycled some work from @SaschaPeukert . This is not the multithreaded approach and I only implemented the Weak Connected Component version very simplified. 
Hope this can help some people. Remarks are welcome? Are there any style guides we need to follow.


ikwattro:
@tommichiels Thanks a lot for your contribution.

a) I think in general it makes sense to return `NodeResult` objects rather than simply `Long`s, in this we can reuse the nodes in further parts of the query without doing another lookup by id.

b) please check build failure due to the newly added dependency.

Thanks again !

tommichiels:
Ok I will fix the return type.

jexp:
Why not multithreaded?

For adding a high-performance collection library, we should make a conscious decision.

Neo4j also has it's own primitive-collections, which I would want to try first.

src/main/java/apoc/algo/WeaklyConnectedComponents.java
```diff
@@ -0,0 +1,105 @@
...
+
+	@Context
+	public Log log;
+
+	@Procedure("apoc.algo.wcc")
+	@Description("CALL apoc.algo.wcc() YIELD number of weakly connected components")
+	public Stream<LongResult> wcc() {
+		try {
+			long componentID = 0;
+			ResourceIterator<Node> nodes = db.getAllNodes().iterator();
```
jexp: I would probably prefer kernel-API here

src/main/java/apoc/algo/WeaklyConnectedComponents.java
```diff
@@ -0,0 +1,105 @@
...
+			String errMsg = "Error encountered while calculating weakly connected components";
+			log.error(errMsg, e);
+			throw new RuntimeException(errMsg, e);
+		}
+	}
+
+	private LongHashSet go(Node node, Direction direction) {
+
+		LongHashSet visitedIDs = new LongHashSet();
+		List<Node> frontierList = new LinkedList<>();
```
jexp: why not a LongHashset too?

src/main/java/apoc/algo/WeaklyConnectedComponents.java
```diff
@@ -0,0 +1,105 @@
...
+					continue;
+				}
+				visitedIDs.add(child.getId());
+				frontierList.add(child);
+			}
+		}
+		return visitedIDs;
+	}
+
+	private List<Node> getConnectedNodeIDs(Node node, Direction dir) {
```
jexp: inline this loop, it doesn't make sense to create a list first and then iterate over that again

src/main/java/apoc/algo/WeaklyConnectedComponents.java
```diff
@@ -0,0 +1,105 @@
...
+		List<Node> it = new LinkedList<Node>();
+		RelationshipType types = RelationshipType.withName("LINK");
```
jexp: this rel-type should not be hard-coded, but either free, or configurable, see PathExplorer and Dijkstra/A*

src/main/java/apoc/algo/WeaklyConnectedComponents.java
jexp: should it do all nodes, or should the labels be configurable?

src/main/java/apoc/algo/WeaklyConnectedComponents.java
```diff
@@ -0,0 +1,105 @@
...
+					componentID++;
+
+				} catch (NoSuchElementException e) {
+					break;
+				}
+				it = allNodes.iterator();
+			}
+			return Stream.of(new LongResult(componentID));
+		} catch (Exception e) {
+			String errMsg = "Error encountered while calculating weakly connected components";
```
jexp: should probably also provide the inputs, which exceptions can actually occur?

src/main/java/apoc/algo/WeaklyConnectedComponents.java
```diff
@@ -0,0 +1,105 @@
...
+					LongHashSet reachableIDs = go(db.getNodeById(n), Direction.BOTH);
+					allNodes.removeAll(reachableIDs);
+					componentID++;
+
+				} catch (NoSuchElementException e) {
+					break;
+				}
+				it = allNodes.iterator();
+			}
+			return Stream.of(new LongResult(componentID));
```
jexp: this result doesn't make sense.
it should return a a stream of result with either
* a component and a list of nodes for this component
* a component and a list of node-ids for this component
* a node and it's component-id
* a node-id and its component-id

jexp: Alternatively it could directly set the component-id as a property (configurable) on those nodes. then it would need `@PerformsWrites` annotation

src/test/java/apoc/algo/WeaklyConnectedComponentsTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class WeaklyConnectedComponentsTest {
+	
+	private static GraphDatabaseService db;
+	// TODO refactor in multiple execute statements dirty workaround
+	public static final String CC_GRAPH =
+            "CREATE (:Node:Record { name : 'A' })  "+
+            "WITH count(*) as dummy " +
```
jexp: you don't need the WITH's
FOREACH (name IN ['A',....'Z'] | CREATE (:Node:Record {name:name}))

src/test/java/apoc/algo/WeaklyConnectedComponentsTest.java
```diff
@@ -0,0 +1,126 @@
...
+	// ==========================================================================================
+
+	@Test
+	public void shouldReturnExpectedResultCountWhenUsingWeaklyConnectedWithEmptyDb() {
+		assertExpectedResult(0L, "CALL apoc.algo.wcc()");
+	}
+
+	@Test
+	public void shouldReturnExpectedResultCountWhenUsingWeaklyConnected() {
+		db.execute(CC_GRAPH).close();
```
jexp: this should happen in setup, as the setup is static, you probably want to make the setup non-static here

src/test/java/apoc/algo/WeaklyConnectedComponentsTest.java
```diff
@@ -0,0 +1,126 @@
...
+		assertExpectedResult(5L, "CALL apoc.algo.wcc()");
+	}
+
+	private void assertExpectedResult(Long expectedResultCount, String query) {
+		TestUtil.testResult(db, query, (result) -> {
+			Object value = result.next().get("value");
```
jexp: you can just assertEquals it with the long value

src/test/java/apoc/algo/WeaklyConnectedComponentsTest.java
```diff
@@ -0,0 +1,126 @@
...
+            "WITH count(*) as dummy " +
+            "CREATE (:Node:Record { name : 'M' })  "+
+            "WITH count(*) as dummy " +
+            "CREATE (:Node:Record { name : 'N' })  "+
+            "WITH count(*) as dummy " +
+            "CREATE (:Node:Record { name : 'Z' })  "+
+            "WITH count(*) as dummy " +
+            "MATCH (a:Node),(b:Node),(c:Node)  "+
+            "WHERE a.name = 'C' AND b.name = 'D' AND c.name = 'E'  "+
+            "CREATE (a)-[:LINK]->(b)-[:LINK]->(c)-[:LINK]->(a)  "+
```
jexp: You can just spell your relationship-creates out
CREATE (c:Node {name:'C'})
CREATE (c)-[:LINK]->(d:Node {name:'D'})-[:LINK]->(e:Node {name:'E'})-[:LINK]->(c)
CREATE (f:Node {name:'F'})
CREATE (f)-[:LINK]->(g:Node {name:'G')-[:LINK]->(f)
... etc

src/main/java/apoc/algo/WeaklyConnectedComponents.java
tommichiels: This was a piece of code still left for our internal use case. My first attempt was to keep all relationships. I will make this type configurable.

src/main/java/apoc/algo/WeaklyConnectedComponents.java
tommichiels: About the return type -> if I return a list of all components with all the nodes I'm in sense returning the whole graph. For further processing this is maybe not so convenient? Is it maybe better to introduce a new CC node with an id and relationships with all its connected components?  If I only set a property it is hard to get the count. What do you think?

jexp: We can set the property (configurable) and still return the counts per component.

We can also turn each component into a virtual graph and then create something that creates a nice summary of that graph? i.e. which label(counts) and rel-type-counts it contains and perhaps some graph metrics?

I don't know if the cc node with relationships works so well. It just a massively dense node.

jexp:
Hey @tommichiels would love if you found time to continue on this 👍 
Thanks a lot.

src/main/java/apoc/algo/WeaklyConnectedComponents.java
tommichiels: A virtual graph sounds nice. This is actual what we are trying to build a summary like this 

CC A has x number of label A and y numbers of label B.

tommichiels:
Hi,

I tried to process all the feedback I got. Hope it is ok. New feedback always welcome.

ikwattro:
@jexp @tommichiels what's the status of this one ?

tommichiels:
Waiting on feedback if it is ok right now. Multithreaded approach is not implemented right now.

PR #52, opened 2016-06-13
Use Neo4j Stored Procedure compiler

fbiville:
This Pull Request is meant as a discussion starter and needs
some clarifications before it can be merged.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,12 +10,14 @@ sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
 repositories {
+     maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
      maven { url "http://repo.maven.apache.org/maven2" }
 }
 
 dependencies {
     compile group: 'it.unimi.dsi', name: 'fastutil', version:'7.0.12'
     compile group: 'commons-codec', name: 'commons-codec', version:'1.9'
+    compileOnly group: 'net.biville.florent', name: 'neo4j-sproc-compiler', version:'1.0-SNAPSHOT'
     
     testCompile group: 'org.apache.derby', name: 'derby', version:'10.12.1.1'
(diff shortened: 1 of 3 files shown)


fbiville:
I would like to continue the conversation I started with @jexp here.
There are mainly two kinds of error that are reported by the plugin:

1. undocumented procedure input type: `Node`, `Relationship` and `Path` are not listed as supported procedure types, that's why the compiler complains about it
1. Map erasure: it seems the runtime accepts `Map` instances other than the documented `Map<String,Object>`

I'll send a documentation update PR this week and update the compiler accordingly.
About point 2, I'm still unclear about what to do. In my opinion, the compiler is strict and should be. 

Usage of raw types such as `Map` should definitely be discouraged (and people don't have to use the compiler if they don't want compile-time checks).
For the other map errors, I'm not sure. The docs clearly state that people should only declare `Map<String,Object>`.

Shall I keep the compiler strict? Shall I issue warnings instead of errors?
Shall Neo4j documentation be updated to have a `Map` definition similar to `List`?

fbiville:
The main point of this PR is to clarify what should be documented and what to do in case something is not.

jexp:
Perhaps for 2) send a PR to make APOC compliant.

fbiville:
After some thinking, I changed the compiler so that:

1. Node, Relationship and Path are valid input type (whether directly or in a collection)
2. Map value types are processed the same as List value types

In the end, neo4j-apoc failed to compile with only 2 errors: `Map` rawtype was used.

I'll send a PR to Neo4j docs in order to reflect the changes I did.

jexp:
Can you add it in the pom.xml too? Thx :)

jexp:
Really cool, thanks a lot @fbiville !!!

fbiville:
Added to `pom.xml` here: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/53 :)

PR #62, opened 2016-07-04
Fixing a failing test

bachmanm:
... and cheekily adding myself to the list of DEVs, there will be more stuff coming from GraphAware as we will retire some of our modules and migrate the functionality to APOC


ikwattro:
Thanks @bachmanm 

Build pass locally, merging..

PR #72, opened 2016-07-26
update readme.adoc

anuragknoldus:
I  just want to add this change because when we place jar into $Neo4j_Home/plugins than it is very useful to find the count of procedure into any specific package. So we can know that we have latest apoc-procedure or not.

Thanks.


readme.adoc
```diff
@@ -81,6 +81,16 @@ RETURN name AS undocumented
 
 // end::help[]
 
+=== Builtin Pakage and Procedure count
+
+// tag::procedurecount[]
+
+To find the procedure count with the pakage in database:
```
jexp: typo in package

jexp: also add the Cypher statement in a [source,cypher] block not just the image

readme.adoc
```diff
@@ -81,6 +81,16 @@ RETURN name AS undocumented
 
 // end::help[]
 
+=== Builtin Pakage and Procedure count
```
jexp: typo in Pa_c_kage

anuragknoldus:
Sorry Michael for the typing error.
I will update it and than again send the pull request.

PR #82, opened 2016-08-10
Make tests pass (even on Windows), support for Elastic basic auth when sending requests

bachmanm:


src/test/java/apoc/spatial/GeocodeTest.java
```diff
@@ -100,10 +100,10 @@ private void testGeocodeAddress(String address, double lat, double lon) {
         testCall(db, "CALL apoc.spatial.geocodeOnce({url})", map("url", address),
                 (row) -> {
                     Map value = (Map) row.get("location");
-                    assertEquals("Incorrect latitude found", lat, Double.parseDouble(value.get("latitude").toString()),
-                            0.0005);
-                    assertEquals("Incorrect longitude found", lon, Double.parseDouble(value.get("longitude").toString()),
-                            0.0005);
+                    assertEquals("Incorrect latitude found for address "+address, lat, Double.parseDouble(value.get("latitude").toString()),
+                            0.1);
+                    assertEquals("Incorrect latitude found for address "+address, lon, Double.parseDouble(value.get("longitude").toString()),
```
sarmbruster: shouldn't this be longitude in the message string?

bachmanm: yeah, well spotted!

jexp:
yo @bachmanm 
can we split this into 3 PRs ?

for ES auth, I want it to be part of the URL and pull it from there, so that it can go into the ES host-or-key and so into neo4j.conf

I think the export test was already covered by Viksits' PR

PR #83, opened 2016-08-10
Pagerank procedure with cypher queries as inputs

atuljangra:
I've implemented a separate pagerank procedure that takes cypher query as input and does the pageranking on the subgraph thus created.

Example query: 
`        Query = "MATCH (b:Company {name:'b'})\n" +
                "CALL apoc.algo.pageRankWithCypher([b],{node_cypher:\"MATCH (node:Company) return id(node)\", " +
                "rel_cypher:\"\", write:true, iterations:20}) " +
                "YIELD node, score\n" +
                "RETURN node.name AS name, score\n" +
                "ORDER BY score DESC";
`

Parameters allowed:
1. "node_cypher" : Cypher query to return the id of nodes that should be considered for pagerank.
Eg: `MATCH (node:Company) return id(node)`

2. "rel_cypher" : Cypher query to return the "source", "target" and "weight" of a relationship.
Eg: `MATCH (s)-[r]->(t) RETURN id(s) as source, id(t) as target, 1 as weight`.

3. "write": true|false, Default false
Should we write the results back to the database.
4. "iterations": how many iterations should be ran for calculating pagerank. 

### Implementation Details:

- The program is divided into three parts:

1. Reading the input: Reads the input into data structures.
2. Computing: Computes pagerank on the basis of local data structures only.
3. Writing/Returning the output: Returning or writing of output to db.

- Currently we have a mapping from prog-id to node-id and reverse mapping is done by binary search on the same array. This is useful for reading and writing only.
- Reading is currently single threaded. It executes rel_cypher and node_cypher twice each.
- Computation is also single threaded and works on the prog-id. 
- Writing back is done in parallel.

Current memory usage is : 
1. 6 arrays of the size of node out of which two can be easily eradicated by computing them on the fly. 
2. 2 arrays of the size of the relationships to store targets and weights. We can have an argument called "hasWeights" and then if it is set to false, we can eradicate the weights array.  

### Future Goals
- Make the computation concurrent.
- Make the reading concurrent wherever possible. This would depend upon the batch size and the queries that the user has specified.
- Merge this procedure with the existing procedure so that only difference would be in the input reading part. Computation and writeback would be shared.


src/main/java/apoc/algo/PageRank.java
```diff
@@ -55,6 +69,34 @@
...
+        if (relCypher.equals("") || relCypher.isEmpty()) {
+            relCypher = DEFAULT_PAGE_RANK_CYPHER_REL;
+        }
+
+        PageRankArrayStorageParallelCypher pageRank = new PageRankArrayStorageParallelCypher(dbAPI, pool, nodeCypher, relCypher);
+        pageRank.compute(iterations.intValue());
+        if (shouldWrite) {
+            pageRank.writeResultsToDB();
+        }
+        return nodes.stream().map(node -> new NodeScore(node, pageRank.getResult(node.getId())));
```
jexp: We shouldn't reurn nodes. Just statistics info like runtimes for the steps and total, nodes and rels processed, iterations

src/main/java/apoc/algo/PageRank.java
```diff
@@ -55,6 +69,34 @@
...
+    public Stream<NodeScore> pageRankWithCypher(
+            @Name("nodes") List<Node> nodes,
+            @Name("config") Map<String, Object> config) {
+        Long iterations = (Long) config.getOrDefault(SETTING_PAGE_RANK_ITERATIONS, DEFAULT_PAGE_RANK_ITERATIONS);
+        String relCypher = (String)config.getOrDefault(SETTING_PAGE_RANK_CYPHER_REL, DEFAULT_PAGE_RANK_CYPHER_REL);
+        String nodeCypher = (String)config.getOrDefault(SETTING_PAGE_RANK_CYPHER_NODE, DEFAULT_PAGE_RANK_CYPHER_NODE);
+
+        boolean shouldWrite = (boolean)config.getOrDefault(SETTING_PAGE_RANK_WRITE, DEFAULT_PAGE_RANK_WRITE);
+
+        if (nodeCypher.equals("") || nodeCypher.isEmpty()) {
```
jexp: Wrap retrieving the cypher in a method and reuse

jexp:
Please add your detailed commit message as docs too.

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        double val = 0;
+        int logicalIndex = getNodeIndex((int)node);
+
+        if (logicalIndex >= 0 && pageRanks.length >= logicalIndex) {
+            val = toFloat(pageRanks[logicalIndex]);
+        }
+        return val;
+    }
+
+    public String getPropertyName()
```
jexp: Pass into write back method as parsm

jexp: Param

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+                continue;
+            currentIndex += sourceDegreeData[i];
+        }
+    }
+
+    private void readDataIntoArray(String relCypher, String nodeCypher) {
+        Result nodeResult = db.execute(nodeCypher);
+        Result nodeCountResult = db.execute(nodeCypher);
+
+        // TODO Do this lazily.
```
jexp: Yes. Just create array chunks. And fill each chunk

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        int totalNodes = (int)nodeCountResult.stream().count();
+        this.nodeCount = totalNodes;
+        nodeMapping = new int[totalNodes];
+        sourceDegreeData = new int[totalNodes];
+        sourceWeightData = new int[totalNodes];
+        sourceChunkStartingIndex = new int[totalNodes];
+
+        String columnName = nodeResult.columns().get(0);
```
jexp: Use result.columnAs("id")

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        this.nodeCypher = nodeCypher;
+        readDataIntoArray(this.relCypher, this.nodeCypher);
+    }
+
+    private int getNodeIndex(int node) {
+        int index = Arrays.binarySearch(nodeMapping, node);
+        return index;
+    }
+
+    private int getDegree(int node) {
```
jexp: This should mostly use the algo-node-id, or?

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        int degree = 0;
+        int index = Arrays.binarySearch(nodeMapping, node);
```
jexp: Reuse getNodeIndex

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        if (index >= 0) {
+            degree = sourceDegreeData[index];
+        }
+        return degree;
+    }
+
+    private void setDegree(int node, int degree) {
```
jexp: This should happen while iterating over the mappingarray so there should be no id-lookup needed.

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        int index = Arrays.binarySearch(nodeMapping, node);
+        sourceDegreeData[index] = degree;
+    }
+
+    private int getWeight(int node) {
+        int degree = 0;
+        int index = Arrays.binarySearch(nodeMapping, node);
```
jexp: Same here, during the algo we shouldn't need id lookups. And the initial population happens while iterating over the mapping or rel-data which is also mapped

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        if (index >= 0) {
+            degree = sourceWeightData[index];
+        }
+        return degree;
+    }
+
+    private void setWeight(int node, int weight) {
+        int index = Arrays.binarySearch(nodeMapping, node);
```
jexp: Same here

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        sourceWeightData[index] = weight;
+    }
+
+    private void calculateChunkIndices() {
```
jexp: ?? What is this needed for?

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        sourceWeightData = new int[totalNodes];
+        sourceChunkStartingIndex = new int[totalNodes];
+
+        String columnName = nodeResult.columns().get(0);
+        int index = 0;
+        while(nodeResult.hasNext()) {
+            Map<String, Object> res = nodeResult.next();
+            int node = ((Long)res.get(columnName)).intValue();
+            nodeMapping[index] = node;
+            sourceDegreeData[index] = 0;
```
jexp: They should be zero initially. No needvto set to zero

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+        pageRanks = new int [totalNodes];
+        previousPageRanks = new int[totalNodes];
+
+        Result result = db.execute(relCypher);
+        int totalRelationships = 0;
+        while(result.hasNext()) {
+            Map<String, Object> res = result.next();
+            int source = ((Long) res.get("source")).intValue();
+            int weight = ((Long) res.getOrDefault("weight", 1)).intValue();
+
```
jexp: Do mode id resolution only once and then only use the algo-node-id

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+                setWeight(source, storedWeight + weight);
+            }
+            totalRelationships++;
+        }
+
+        result.close();
+        relationshipTarget = new int[totalRelationships];
+        relationshipWeight = new int[totalRelationships];
+
+        for (int i = 0; i < totalRelationships; i++) {
```
jexp: There is Arrays.fill

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+            relationshipTarget[i] = -1;
+            relationshipWeight[i] = -1;
+        }
+        this.relCount = totalRelationships;
+        calculateChunkIndices();
+
+        // We have degrees for all the nodes at the point.
+        result = db.execute(relCypher);
+        while(result.hasNext()) {
```
jexp: This should bevdone in one iteration over the cypher results, not twi

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
```diff
@@ -0,0 +1,324 @@
...
+                continue;
+            }
+            int prevRank = pageRanks[node];
+            previousPageRanks[node] =  toInt(ALPHA * toFloat(prevRank) / weightedDegree);
+            pageRanks[node] = ONE_MINUS_ALPHA_INT;
+        }
+    }
+
+    public void writeResultsToDB() {
+        writeBackResults(db, this);
```
jexp: Pass in properties and two arrays

src/main/java/apoc/algo/PageRank.java
atuljangra: In case when writeback is false and the user just want to see the result, shouldn't we be returning nodes?
I've tried to follow the other pagerank implementation approach where we can project the nodes whose pagerank we want to see.

I can completely get rid of the projection and return information as you asked, runtimes for reading, computation and writing back.
But what about the case when "write" is false?

src/main/java/apoc/algo/PageRank.java
atuljangra: Yes

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
atuljangra: Yes makes sense.

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
atuljangra: Yes. I can use getNodeIndex here to get the index for the source and then write/read from that index.

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
atuljangra: This was one doubt that I had earlier. As far as I think, unless the user has actually sorted the rel_cypher according to the source id, we might have relationships that are random:
source target weight
3           4           1
1           4           1
3          1            3

So in this case, when I get the first relationship at 3, I should be knowing where in the array I should be storing the targetid and weight. So for that I'm iterating over the relationship cypher twice. In the first iteration, I calculate the degrees of all the nodes. Then this function calculated the projected starting chunkIndices where the chunk for a given node would be stored and then in the next iteration over the cypher results, I actually store the target and store.

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
atuljangra: Added the reason above to why I'm doing two iterations.

src/main/java/apoc/algo/pagerank/PageRankArrayStorageParallelCypher.java
atuljangra: Yep. Would make this static too.
Initially I thought I should move this function to PagerankUtils.java. Would that be okay?

neo4j-oss-build: Yes

Von meinem iPhone gesendet

src/main/java/apoc/algo/PageRank.java
jexp: Ya ordering is probably too expensive. 
We could attempt to put them lazily into buckets. 

Optionally we could have an option that the node-query also returns degrees.

Von meinem iPhone gesendet

atuljangra: `Optionally we could have an option that the node-query also returns degrees.`
Was this for the running of relationship query twice comment?

neo4j-oss-build: Yes

Von meinem iPhone gesendet

PR #88, opened 2016-08-12
decrease the precision of Geospatial tests significantly in order to …

bachmanm:
…make the build green


sarmbruster:
thanks @bachmanm

PR #89, opened 2016-08-13
Random Graph Generators

bachmanm:


jexp:
Thanks a lot @bachmanm !!

Would you mind moving the detailed description into the docs in a separate generate.adoc file and only keep the overview in the readme?

Did you see the build failure? Dependency issue with faker?

bachmanm:
will do - sorry about the build, I was using Maven, will fix gradle dependencies

bachmanm:
OK, I think it should be ready. Thanks!

jexp:
Looks good, thanks so much !

PR #96, opened 2016-08-18
Improvements to pagerank: Making Computation parallel and reducing memory

atuljangra:
We have now added an assumption that the relationship cypher should return the relationships in the order of the source ids.
This version 

- does parallel computation for the pagerank algorithm. 
- Runs both the cypher just once.
- Gives an option to writeback in parallel.

Further ahead: 
- Adding docs.
- Making buckets of arrays instead of lazy loading.
- Adding more reading methods.


jexp:
Merged manually, please check.

PR #110, opened 2016-08-18
get all of the relationship properties for gephi stream

korostelevm:
Just a single line, the graph streamer only mapped the type property and ignored the weights set to relationships in the neo graph


jexp:
it's not quite correct because you now lost the type + LABEL information

jexp:
You also have to fix the test.

PR #120, opened 2016-08-23
couchbase integration

inserpio:


src/test/java/apoc/couchbase/CouchbaseIT.java
```diff
@@ -0,0 +1,210 @@
...
+    CouchbaseAbstractTest.tearDown();
+    if (graphDB != null) {
+      graphDB.shutdown();
+    }
+  }
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void testGetViaCall() throws Exception {
+    assumeTrue(couchbaseRunning);
```
sarmbruster: all of your test method use `assumeTrue(coubebaseRunning)`. Isn't this a common aspect that should be refactore into a method using `@Before/@BeforeClass` ?

inserpio: @sarmbruster I took inspiration from the [MongoDBTest.java](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/master/src/test/java/apoc/mongodb/MongoDBTest.java).
Let me know how you'd prefer to refactor them.

jexp: Is there something like assume on the global level @sarmbruster ?

inserpio: Just tested this way (+ removed from every @test method) and it works fine:
```  @BeforeClass
  public static void setUp() throws Exception {
    CouchbaseAbstractTest.setUp();
    assumeTrue(couchbaseRunning);
  }
```
If you like I can commit this way.

inserpio:
Moved to #122 (as I created a specific branch on my fork for Couchbase and continue working on other apoc)... Sorry, I didn't expected the PR would have been automatically closed...

Issue #91, opened 2016-08-15
Sugestion: add a neo4j function for FORMAT or TRANSFORM


JavierSJ:
Being new with neo4j, I can't found a integrated function that transform the output of numbers or strings.
Is "ugly" to see something like 12345678.901234, instead of clearest 12,345,678.90.
The same for dates, or strings. In fact, one of the most usefull and powerfull functions in Excel is, the ability to format a cell value.

Hope it helps
regards
Javier

jexp:
Kinda apoc.number.format, apoc.number.parse similar to apoc.date.*

Should be straightforward to implement using Java's DecimalFormat plus some extra config that can hold thousand and decimal-separator.

If you have any java knowledge feel free to take a stab at it and send a PR.

inserpio:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/123

PR #123, opened 2016-08-25
number formatting functions - #91

inserpio:


src/main/java/apoc/number/Number.java
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * 
+ * @since 25.8.2016
+ * @author inserpio
+ */
+public class Number {
+
+  @Procedure
+  @Description("apoc.number.formatLong(number, pattern, locale) yield value | format a long using a (optional) pattern and for a (optional) locale to produce a string")
+  public Stream<StringResult> formatLong(final @Name("number") long number, @Name("pattern") String pattern, @Name("language") String language) {
```
jexp: we could have one `format` method that either takes a long or double i.e. an object and checks at runtime.
also the simple variant should just take a pattern and use the locale of the server
or even simpler use a default sensible pattern
like two decimal places for doubles and thousands grouping

src/main/java/apoc/number/Number.java
```diff
@@ -0,0 +1,126 @@
...
+
+  @Procedure
+  @Description("apoc.number.negativePrefix(pattern, locale) yield value | get the negative prefix for the given (optional) pattern and (optional) locale")
+  public Stream<StringResult> negativePrefix(@Name("pattern") String pattern, @Name("language") String language) {
+    DecimalFormat format = buildFormatter(pattern, language);
+    return Stream.of(new StringResult(format.getNegativePrefix()));
+  }
+  
+  @Procedure
+  @Description("apoc.number.positivePrefix(pattern, locale) yield value | get the positive prefix for the given (optional) pattern and (optional) locale")
```
jexp: not sure we need all these helper methods. I think if someone needs this information they can just try it out? With a concrete negative number or number > 1000 ?

jexp:
really cool @inserpio thanks soo much for working on this!

you're on an APOC roll !

can you have a look at my comments?

In general I'm happy to merge but I think a convenience function could be nice and the addtional methods don't add big value?

inserpio:
@jexp my one was just an early draft proposal. I'm super happy to define the "interface" for these procedures with you folks. I'll start changing it accordingly and if in the meantime anybody has more ideas then let me know.

src/main/java/apoc/number/Number.java
inserpio: I can merge the two formatLong and formatDouble as you're suggesting.

if you pass pattern=null and/or language=null than it uses default server values, but ok, we can override the function and remove the two params.

src/main/java/apoc/number/Number.java
inserpio: ok, I'll remove them... I was blown by a wrapping mania ;-)

inserpio:
@jexp I would go with:
- `format(Object number)`
- `formatByPattern(Object number, String pattern)`
- `formatByLanguage(Object number, String language)`
- `formatByPatternAndLanguage(Object number, String pattern, String language)`

ok?

jexp:
Not bad, what do you think about?

apoc.number.format
apoc.number.format.pattern
apoc.number.format.lang
apoc.number.format.pattern.lang

inserpio:
Better with dots: names are shorter...
An idem for parse then:
apoc.number.parse
apoc.number.parse.pattern
apoc.number.parse.lang
apoc.number.parse.pattern.lang

inserpio:
Or we could make procedures overload-able ;-)
Would it be possible to look at params when you register procs also?

src/main/java/apoc/number/Number.java
```diff
@@ -22,76 +22,97 @@
...
+    validateNumberParam(number);
+    DecimalFormat format = buildFormatter(pattern, lang);
+    return Stream.of(new StringResult(format.format(number)));
+  }
+  
+  // Parse As Long
+  
+  @Procedure("apoc.number.parseAsLong")
+  @Description("apoc.number.parseAsLong(text) yield value | parse a text using the default system pattern and language to produce a long")
+  public Stream<LongResult> parseAsLong(final @Name("text") String text) throws ParseException {
```
inserpio: @jexp should I return a `ResultObject` so that we can merge `parseAsLong` and `parseAsDouble` into a more generic `parse`?

inserpio: `ResultObject` implies that the caller will have to manage the "cast" to `long` or `double`

neo4j-oss-build: Yes we need to keep parseInt and parseDouble for the result types 

Von meinem iPhone gesendet

neo4j-oss-build: No keep them

Von meinem iPhone gesendet

jexp:
They are not overloadable in Neo4j

Von meinem iPhone gesendet

jexp:
Yep

Probably parseInt parseDouble

Von meinem iPhone gesendet

inserpio:
Ok.
I'd keep parse*Long* not parse*Int* to have the effective return type in the name.
What's for the other apocs? Is *Int* used?

inserpio:
looking for some alternatives to `apoc.number.format.*.pattern.lang`, but I lack a bit of imagination :-/
`apoc.number.format.*.full` ?
`apoc.number.format.*.patternLang`
`apoc.number.format.*.patternAndLang` ?

`full`is fast to write but if in the future we'll add new apocs with more parameters then it'd become an incorrect name..

jexp:
In 3.1 there are default params :)

Naming is hard.

inserpio:
just decide for keeping `parseLong` or changing it to `parseInt` and from my side it should be all ok for merging... unless you want to add something else...

jexp:
I would align with cypher's `toInt` and do `parseInt`
and `toFloat` -> `parseFloat`

jexp:
Cypher has only the type CTInteger and CTFloat

inserpio:
@jexp done.

jexp:
Thanks so much @inserpio for this great contribution!

PR #129, opened 2016-08-30
Create manual-indexes.adoc

rohankharwar:
Kindly review and let me know your comments.

### DIFF
--- a/docs/manual-indexes.adoc
+++ b/docs/manual-indexes.adoc
@@ -2,51 +2,110 @@
 
 == Data Used
 
-In the below examples I used the flight data setup from the following link:
+In the below examples I used the flight data setup from the following link. The link guides with loading the complete data set.
 
 https://github.com/nicolewhite/neo4j-flights
 
-
-== Enable Manual Indexes on Node and Relationship Property
-
-Below are the steps to setup manual index on the nodes and relationship.
-In order to do so, one needs to first enable auto index and identify the property on the nodes and relationships
-that will be indexed.
-Once the property is identified set enable the auto index and set the property as shown in the configuration file.
-Below is the configurations setting to enable auto index in versions 3.0 and above.
-Here we are adding manual index on property `taxi_time` on Relationship `DESTINATION` and on property `flight_num` on Node with label `Flight`.
-
-
-In conf/neo4j.conf file
+However I have pasted a sample subset of the data that can be load to try the example queries below:
 
 ----
-dbms.auto_index.relationships.enabled=true
-dbms.auto_index.relationships.keys=taxi_time, flight_num
+begin
+CREATE (:`Airport` {`abbr`:"SLC", `id`:14869, `name`:"SALT LAKE CITY INTERNATIONAL"});
+CREATE (:`Airport` {`abbr`:"OAK", `id`:13796, `name`:"METROPOLITAN OAKLAND INTERNATIONAL"});
+CREATE (:`Airport` {`abbr`:"BUR", `id`:10800, `name`:"BOB HOPE"});
+CREATE (:`Flight`:`UNIQUE IMPORT LABEL` {`flight_num`:6147, `day`:2, `month`:1, `weekday`:6, `year`:2016, `UNIQUE IMPORT ID`:233903});
+CREATE (:`Flight`:`UNIQUE IMPORT LABEL` {`flight_num`:6147, `day`:2, `month`:1, `weekday`:6, `year`:2016, `UNIQUE IMPORT ID`:235869});
+CREATE (:`Flight`:`UNIQUE IMPORT LABEL` {`flight_num`:6147, `day`:9, `month`:1, `weekday`:6, `year`:2016, `UNIQUE IMPORT ID`:265364});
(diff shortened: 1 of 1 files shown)


docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
+= Manual Indexes
+
+== Data Used
+
+In the below examples I used the flight data setup from the following link:
+
+https://github.com/nicolewhite/neo4j-flights
+
+
+== Enable Manual Indexes on Node and Relationship Property
```
jexp: Actually those are about auto-indexes, which are a kind of legacy-index but not the ones I was focusing on in apoc. Apoc has it's own procedures for adding things to the indexes which are meant to be called in batch post-creation or better as par of the create statements.

I'd rather prefer not to have legacy-autoindexes.

docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+Once the auto indexes are enabled.
+
+One will have to add the relationship property to the index.
+This can be achieved using the APOC stored procedure - `apoc.index.addRelationship`.
+
+----
+match (f:Flight)-[r:DESTINATION]->(a:Airport)
+with r
+call apoc.index.addRelationship(r,["taxi_time"])
```
jexp: also this works differently than the auto-index b/c the index-name is the relationship-type and here you only add one property whereas in the auto-index you had two rel-properties indexed

docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+
+Once the relationship index is created we can start using it.
+Here is one example of using this without Stored Procedure.
+
+
+This query will give me the count of the relationship `DESTINATION` with the property `taxi_time = 11`.
+
+----
+START r=relationship:DESTINATION(taxi_time="11")
```
jexp: I wouldn't want to use start anymore. It's meant to go away. And the index procedures do the same but better.

docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
...
+return node
+----
+
+To Remove the manual index `DESTINATION` created -
+
+----
+CALL apoc.index.remove('DESTINATION') YIELD type,name,config
+----
+
+== Adding Manual Index on Node Property
```
jexp: I would start with nodes and then discuss relationships next. Node indexes are easier to understand.

docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+Once the auto indexes are enabled.
+
+One will have to load the nodes to the index.
+This can be achieved using one of the APOC stored procedure - `apoc.index.addNode`.
+
+----
+match (f:Flight)
```
jexp: perhaps promote to run the apoc.index.addNode as part of the create or load csv script.

docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
...
+call apoc.index.addNode(f,["flight_num"])
+return count(f)
+----
+
+The above statement will create the auto index with the same name as the Label name in this case `Flight` and add the relationships to the index.
```
jexp: add the _nodes_

docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
...
+----
+CALL apoc.index.list() YIELD type,name,config
+----
+
+Output:
+
+image::https://github.com/neo4j-contrib/neo4j-apoc-procedures/tree/master/docs/img/apoc-manual-indexes-nodes.png[width=500]
+
+
+Once the relationship index is created we can start using it.
```
jexp: _node_ index

docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
...
+----
+call apoc.index.nodes('Flight','flight_num:43') YIELD node
+match (node)-[:DESTINATION]->(a:Airport)-[:IN_CITY]->(c:City)
+return distinct a.name, c.name limit 10
+----
+
+To Remove the manual index `Flight` created -
+
+----
+CALL apoc.index.remove('Flight') YIELD type,name,config
```
jexp: you don't need yield for a *plain* call

docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
...
+----
+match (a:Flight {flight_num:43}) with a
+call apoc.index.out(a,"DESTINATION","taxi_time:11") YIELD node
+return node
+----
+
+To Remove the manual index `DESTINATION` created -
+
+----
+CALL apoc.index.remove('DESTINATION') YIELD type,name,config
```
jexp: you don't need yield here.

docs/manual-indexes.adoc
```diff
@@ -0,0 +1,166 @@
+= Manual Indexes
+
+== Data Used
+
+In the below examples I used the flight data setup from the following link:
+
+https://github.com/nicolewhite/neo4j-flights
```
jexp: perhaps we can include a subset of the load script as we plan to turn this manual into individual browser guides, then they should be self-contained.

rohankharwar:
Hi @jexp: Updated the images and the implemented the changes to the doc.

jexp:
Merged manually, thanks a lot @rohankharwar

Issue #131, opened 2016-09-02
apoc.warmup.run does not actually load all nodes and relationships


fpavageau:
See [that SO question](https://stackoverflow.com/q/39286740/1350869) for an example.

The problem is that the procedure retrieves the number of nodes (resp. relationships) and then tries to load all nodes (resp. relationships) [with ids between 1 and that number](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.0.4.1/src/main/java/apoc/warmup/Warmup.java#L44), as if they were all sequentially numbered. It's only true for a graph where no node or relationship has been deleted (and its id not reused).

I believe a better way of forcing the load of all nodes would be to ask for the maximum id: either that's enough, because that requires scanning all nodes, or it's not and the current query can then be used to iterate through them all, using that maximum id instead of the number of items.

The first option seems to be enough for the moment: executing
```
PROFILE MATCH (n) RETURN count(n) AS count, max(id(n)) AS maxId
```
shows about the same number of DB hits as I have nodes, and takes between 650 ms and 850 ms for 1.4 million nodes on my machine.
```
+-------------------+
| count   | maxId   |
+-------------------+
| 1398103 | 1398103 |
+-------------------+
1 row
644 ms

Compiler CYPHER 3.0

Planner COST

Runtime INTERPRETED

+-------------------+----------------+---------+---------+--------------+--------------+
| Operator          | Estimated Rows | Rows    | DB Hits | Variables    | Other        |
+-------------------+----------------+---------+---------+--------------+--------------+
| +ProduceResults   |           1182 |       1 |       0 | count, maxId | count, maxId |
| |                 +----------------+---------+---------+--------------+--------------+
| +EagerAggregation |           1182 |       1 |       0 | count, maxId |              |
| |                 +----------------+---------+---------+--------------+--------------+
| +AllNodesScan     |        1398103 | 1398103 | 1398104 | n            |              |
+-------------------+----------------+---------+---------+--------------+--------------+

Total database accesses: 1398104
```

jexp:
Good point. 

We can ask the store for the high id. 


Von meinem iPhone gesendet

fpavageau:
I guess this can be closed since the PR has been merged.

PR #132, opened 2016-09-02
Use their max ids to warm up all the nodes/rels

fpavageau:
As reported in #131, instead of the node (or relationship) count, the maximum id needs to be used to try and load all pages: if some data has been deleted, the id space will be sparsed instead of sequential, and the count less than the actual maximum id.

I've also set the locale for the tests in the Gradle configuration, as some tests would fail on my machine due to number formats being different that the asserted ones using the French locale.

### DIFF
--- a/src/main/java/apoc/warmup/Warmup.java
+++ b/src/main/java/apoc/warmup/Warmup.java
@@ -2,15 +2,23 @@ package apoc.warmup;
 
 import apoc.Description;
 import apoc.util.Util;
+import org.neo4j.cursor.Cursor;
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.kernel.api.ReadOperations;
+import org.neo4j.kernel.impl.core.ThreadToStatementContextBridge;
+import org.neo4j.kernel.impl.storageengine.impl.recordstorage.RecordStorageEngine;
+import org.neo4j.kernel.impl.store.NeoStores;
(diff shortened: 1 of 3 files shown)


src/main/java/apoc/warmup/Warmup.java
```diff
@@ -39,13 +39,23 @@ public Warmup() {
         int relsPerPage = pageSize / RelationshipRecordFormat.RECORD_SIZE;
         long nodesTotal = Util.nodeCount(db);
         long relsTotal = Util.relCount(db);
+        long maxNodeId = Util.runNumericQuery(db, "MATCH (n) RETURN max(id(n)) AS result", null);
```
jexp: I'd rather use this again, like the original implementation.

https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/25/files#diff-2a0b8c8ad695da883dab5f279def2112R50

This is instant. No query required.
Perhaps we should just revert to @SaschaPeukert's original implementation which is much more efficient on huge stores.

fpavageau: Oh, OK, I only looked at the merged version, not at the original PR. Sure, with the store providing that information, it's obviously faster.

I can create a version closer to the original, while still replacing the hard-coded values with constants.

jexp: Thanks a ton!

Von meinem iPhone gesendet

PR #133, opened 2016-09-03
Adding Unweighted Betweenness Centrality

atuljangra:
This pull requests add unweighted betweenness centrality.

Similar to  pagerank, we take nodeCypher and relCypher and store them in local Ds.
Using the stored locations, we calculate the betweenness centrality in parallel. Computation and memory perf are good as compared to iGraph. Example, for a graph that took ~one day in igraph,we were able to do that in approximately 75 minutes.

I've used double array for precision to store the results.
I think memory can be improved some more.

### DIFF
--- a/src/main/java/apoc/algo/algorithms/Algorithm.java
+++ b/src/main/java/apoc/algo/algorithms/Algorithm.java
@@ -164,4 +164,62 @@ public class Algorithm {
         long after = System.currentTimeMillis();
         log.info("Time to read relationship data " + (after - before) + " ms");
     }
+
+    private boolean readRelationshipMetadataAndNodes(String relCypher, String nodeCypher, boolean weighted) {
+        Result nodeResult = db.execute(nodeCypher);
+
+        long before = System.currentTimeMillis();
+        ResourceIterator<Object> resultIterator = nodeResult.columnAs("id");
+        int index = 0;
+        int totalNodes = 0;
+        nodeMapping = new int[INITIAL_ARRAY_SIZE];
+        int currentSize = INITIAL_ARRAY_SIZE;
+        while(resultIterator.hasNext()) {
+            int node  = ((Long)resultIterator.next()).intValue();
+
+            if (index >= currentSize) {
+                if (log.isDebugEnabled()) log.debug("Node Doubling size " + currentSize);
+                nodeMapping = doubleSize(nodeMapping, currentSize);
+                currentSize = currentSize * 2;
+            }
+            nodeMapping[index] = node;
+            index++;
+            totalNodes++;
+        }
+
+        this.nodeCount = totalNodes;
+        Arrays.sort(nodeMapping, 0, nodeCount);
+        long after = System.currentTimeMillis();
+        readNodeMillis = (after - before);
+        log.info("Time to make nodes structure = " + readNodeMillis + " millis");
+        before = System.currentTimeMillis();
+
+        sourceDegreeData = new int[totalNodes];
+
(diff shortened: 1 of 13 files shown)


ikwattro:
Thanks @atuljangra for this contribution, good refactoring.

Before @jexp has a look to it, can you make sure tests are fixed, especially division by zero exceptions.

Thanks !

atuljangra:
Yes. Lemme fix those tests and resend it. Thanks :)

src/main/java/apoc/algo/algorithms/Algorithm.java
```diff
@@ -0,0 +1,205 @@
...
+        Result nodeResult = db.execute(nodeCypher);
+
+        long before = System.currentTimeMillis();
+        ResourceIterator<Object> resultIterator = nodeResult.columnAs("id");
+        int index = 0;
+        int totalNodes = 0;
+        nodeMapping = new int[INITIAL_ARRAY_SIZE];
+        int currentSize = INITIAL_ARRAY_SIZE;
+        while(resultIterator.hasNext()) {
+            int node  = ((Long)resultIterator.next()).intValue();
```
tkroman: Why are node ids coerced to ints here? Shouldn't the array be of `long[]` type?

src/main/java/apoc/algo/algorithms/BetweennessCentrality.java
```diff
@@ -0,0 +1,268 @@
...
+                                     int start,
+                                     int end,
+                                     int [] sourceDegreeData,
+                                     int [] sourceChunkStartingIndex,
+                                     int [] relationshipTarget) {
+        Stack<Integer> stack = new Stack<>(); // S
+        Queue<Integer> queue = new LinkedList<>();
+
+        log.info("Thread: " + Thread.currentThread().getName() + " processing " + start + " " + end);
+        Map<Integer, ArrayList<Integer>>predecessors = new HashMap<Integer, ArrayList<Integer>>(); // Pw
```
tkroman: Memory profile can be improved here by using primitive collections:

`PrimitiveIntObjectMap m = Primitive.intObjectMap();`

src/main/java/apoc/algo/algorithms/Algorithm.java
atuljangra: Actually I was trying to save on the memory by using ints for ints. For these algorithms, we were aiming at graphs with a few hundred million nodes.

If I have to get the int out of the long array, then I will have to do `intValue()`

src/main/java/apoc/algo/algorithms/BetweennessCentrality.java
atuljangra: Oh! I was looking for something like this. This would improve memory a lot.
I didn't know neo4j code had this implementation.

atuljangra: There are few more places where this change would help a lot! Going from Integer to int is good.

jexp:
Yes ints are fine here, we're not aiming at graphs with more than 1 billion nodes with these algorithms. Also we're mapping node-ids from the store to internal id's for the algorithm, so you can select a subset of a larger store.

Using the primitive collections makes sense, that's true. @atuljangra

tkroman:
@jexp the possibility of nodes being selected as a subset of a larger graph was the reason behind my advice to use longs. But I guess that would be easy to change in case someone faces it in the future.

atuljangra:
Sending another commit with the fix.

PR #137, opened 2016-09-06
extract credentials and make them a header when passed in url

ikwattro:


jexp:
Can you squash the commits?

jexp:
No need to pass in headers. Just return a new map, something like `return MapUtil.map(key,value)`

jexp:
No need to pass in headers. Just return a new map, something like `return MapUtil.map(key,value)`

jexp:
No need to pass in headers. Just return a new map, something like `return MapUtil.map(key,value)`

jexp:
No need to pass in headers. Just return a new map, something like `return MapUtil.map(key,value)`

jexp:
No need to pass in headers. Just return a new map, something like `return MapUtil.map(key,value)`

jexp:
No need to pass in headers. Just return a new map, something like `return MapUtil.map(key,value)`

ikwattro:
ok !

ikwattro:
ok !

ikwattro:
Thanks will do

ikwattro:
Thanks will do

jexp:
No need to pass in headers. Just return a new map, something like `return MapUtil.map(key,value)`

jexp:
Sorry, Safari + shit internet.

jexp:
Build failed, probably needs to check if headers is null, I think there is a map.merge Function in utils somewhere that takes care of that.

ikwattro:
There is but doesn't seem to handle `null` instances. Test fixed

jexp:
Thanks a lot

PR #140, opened 2016-09-08
Update load jdbc doc

anuragknoldus:


jexp:
Hey, thanks a lot for the contribution, some notes on things that I'd love to be fixed before merging it.

I prefer *textual output over images* where it makes sense. Makes it easier to format and also for search engines to find. Just copy the text (e.g. from neo4j-shell, psql or sqlplus) and put it into code blocks surrounded by `----`.

The only place where it makes sense otherwise would be after creating of the data from load-jdbc to `return *` and show the graph in the image.

The `:img:` attribute is already set globally correctly you can remove the individual ones again

Why do you set an additional attribute for each file ? Like `:loadjdbc:` but which is then never used?

Please add a `[source,cypher]` or `[source,sql]` annotation to all code blocks where that makes sense.

Also please squash your commits into one if possible.

Thanks so much !!

anuragknoldus:
Hey, thanks for your reviews. 

Yeah sure, I will make changes as you mention here.

The :img: attribute I have set in each file because it was not working properly in files so I have mention it in each file.

anuragknoldus:
Hey Michael,

I have made changes as you mention. I have replaced all the images with the text data.

I was not able to fetch the image which were presented in the files so I kept :img: attribute.

docs/loadjdbc.adoc
```diff
@@ -1,7 +1,11 @@
 = Load JDBC
 
+:img: https://raw.githubusercontent.com/neo4j-contrib/neo4j-apoc-procedures/master/docs/img
```
jexp: you don't need that :img: variable it's in the top document.

docs/loadjdbc.adoc
```diff
@@ -224,3 +228,266 @@ CREATE (t)-[:GENRE]->(g);
...
+.Return row from products table
+[source,cypher]
+----
+with "jdbc:postgresql://localhost:5432/testdb?user=postgres&password=postgres" as url
+CALL apoc.load.jdbc(url,"employee_details") YIELD row
+RETURN row LIMIT 4;
+----
+
+----
+╒══════════════════════════════╕
```
jexp: use a wide variant of the table, there is a little slider on the bottom right to pull the table wide

docs/loadjdbc.adoc
```diff
@@ -224,3 +228,266 @@ CREATE (t)-[:GENRE]->(g);
...
+├──────┼───────┼──────┼───────┤
+│Detail│id     │true  │CREATED│
+└──────┴───────┴──────┴───────┘
+----
+
+.Create Nodes in Neo4j
+[source,cypher]
+----
+CALL apoc.load.jdbc('jdbc:postgresql://localhost:5432/testdb?user=postgres&password=postgres','employee_details') yield row
+CREATE (t:Detail {id:toString(row.id), name:row.name, age:toString(row.age), address:row.address, salary:toString(row.salary)});
```
jexp: you don't need toString. They should have the right types.

docs/loadjdbc.adoc
```diff
@@ -224,3 +228,266 @@ CREATE (t)-[:GENRE]->(g);
...
++----------+
+----
+[source,cypher]
+----
+CALL apoc.load.jdbc('oracle_url',"employee_details") YIELD row
+RETURN row LIMIT 5;
+----
+----
+╒══════════════════════════════╕
+│row                           │
```
jexp: same here, pull it wide.

docs/loadjdbc.adoc
```diff
@@ -224,3 +228,266 @@ CREATE (t)-[:GENRE]->(g);
...
+│ME: MANISH, AGE: 26, SALARY: 2│
+│0000}                         │
+├──────────────────────────────┤
+│{ADDRESS: NEW DELHI, ID: 5, NA│
+│ME: HARMEET, AGE: 26, SALARY: │
+│21000}                        │
+└──────────────────────────────┘
+----
+
+Let’s create some indexes and constraints, note that other indexes and constraints will be dropped by this.
```
jexp: I don't think, we have to do the index / constraint for all the examples, the docs should be focused on one procedure at a time. We can ink to `apoc.schema.assert` though.

docs/loadjson.adoc
```diff
@@ -1,5 +1,6 @@
 = Load JSON
 
+:img: https://raw.githubusercontent.com/neo4j-contrib/neo4j-apoc-procedures/master/docs/img
```
jexp: don't need the `:img:` variable

docs/overview.adoc
```diff
@@ -1,5 +1,7 @@
 = Included Procedures Overview
 
+:img: https://raw.githubusercontent.com/neo4j-contrib/neo4j-apoc-procedures/master/docs/img
```
jexp: don't need the `:img:` variable

anuragknoldus:
@jexp thaks for your review again and I have made all the changes which you proposed.

jexp:
Hey @anuragknoldus I thought about your PR and am not sure the different database examples add so much over the general docs and the drivers table here:

https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.0/docs/loadjdbc.adoc#load-jdbc---resources

Would you be ok, to close this?

PR #158, opened 2016-09-26
Additional option parameters for apoc.index.addAllNodes

dkrizic:
e.g.

`CALL apoc.index.addAllNodes('name',{ Supplier: ["companyName", "description","address"], Product:  ["productName","address"] }, {to_lower_case: false, analyzer: "org.apache.lucene.analysis.core.WhitespaceAnalyzer", autoUpdate: true, queue: "asciifolding" } )`

Next steps: Add to trigger if autoUpdate is true, @sarmbruster had some ideas about this queue stuff.

### DIFF
--- a/src/main/java/apoc/index/FreeTextSearch.java
+++ b/src/main/java/apoc/index/FreeTextSearch.java
@@ -52,11 +52,12 @@ public class FreeTextSearch {
      *
      * @param index     The name of the index to create.
      * @param structure The labels of nodes to index, and the properties to index for each label.
+     * @param options   A list of options, e.g. "to_lower_case: false" or "analyzer: own.analyzer.Implementation"
      * @return a stream containing a single element that describes the created index.
      */
     @Procedure
     @PerformsWrites
-    @Description("apoc.index.addAllNodes('name',{label1:['prop1',...],...}) YIELD type, name, config - create a free text search index")
+    @Description("apoc.index.addAllNodes('name',{label1:['prop1',...],...}, {config}) YIELD type, name, config - create a free text search index")
     public Stream<IndexStats> addAllNodes(@Name("index") String index, @Name("structure") Map<String, List<String>> structure, @Name("options") Map<String,Object> options ) {
         if (structure.isEmpty()) {
             throw new IllegalArgumentException("No structure given.");
--- a/src/main/java/apoc/index/FreeTextSearch.java
+++ b/src/main/java/apoc/index/FreeTextSearch.java
@@ -52,13 +52,32 @@ public class FreeTextSearch {
      *
      * @param index     The name of the index to create.
      * @param structure The labels of nodes to index, and the properties to index for each label.
-     * @param options   A list of options, e.g. "to_lower_case: false" or "analyzer: own.analyzer.Implementation"
      * @return a stream containing a single element that describes the created index.
      */
     @Procedure
     @PerformsWrites
-    @Description("apoc.index.addAllNodes('name',{label1:['prop1',...],...}, {options}) YIELD type, name, config - create a free text search index")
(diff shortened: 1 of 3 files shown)


jexp:
Looks good to me. I'd like to ask you for 2 things:

1. as this changes/breaks the API we need to change the method name (for now until we have 3.1 with default parameter values) to e.g. `addAllNodesWithConfig` which should be deprecated immediately and used with default param in the 3.1 branch
2. squash your commits (but I can do that too)

dkrizic:
@jexp Please take a look now, obviously my new commits are added to new existing Pull Request.

src/main/java/apoc/index/FreeTextSearch.java
```diff
@@ -57,12 +57,32 @@ private IndexStats(String label, String property, long nodeCount) {
     @Procedure
     @PerformsWrites
     @Description("apoc.index.addAllNodes('name',{label1:['prop1',...],...}) YIELD type, name, config - create a free text search index")
-    public Stream<IndexStats> addAllNodes(@Name("index") String index, @Name("structure") Map<String, List<String>> structure) {
+    @Deprecated
```
jexp: actually the new method should be deprecated, as in 3.1. we'll have default parameters, and don't need the extra name anymore. :)

src/main/java/apoc/index/FreeTextSearch.java
```diff
@@ -176,9 +196,18 @@ private void populate(Index<Node> index, Map<String, List<String>> config, Consu
...
+
+            /* add options to the parameters */
+            for( String key : options.keySet() ) {
+                config.put( key, "" + options.get( key ) ); // explicit conversion to String
+            }
+            log.info("Creating or updating index '%s' with config '%s'", index, config );
             Index<Node> nodeIndex = db.index().forNodes(index, config);
             tx.success();
             return nodeIndex;
+        } catch( Exception e ) {
```
jexp: should we really hide exceptions?

src/main/java/apoc/index/FreeTextSearch.java
```diff
@@ -176,9 +196,18 @@ private void populate(Index<Node> index, Map<String, List<String>> config, Consu
         }
         try (Transaction tx = db.beginTx()) {
             updateConfigFromParameters(config, structure);
+
+            /* add options to the parameters */
+            for( String key : options.keySet() ) {
```
jexp: options.forEach() or entrySet()
probably rather use value.toString()

src/main/java/apoc/index/FreeTextSearch.java
dkrizic: Ok

src/main/java/apoc/index/FreeTextSearch.java
dkrizic: Oops, was for testing

src/main/java/apoc/index/FreeTextSearch.java
dkrizic: Ok

dkrizic:
Can you take a look now, please?

src/main/java/apoc/index/FreeTextSearch.java
```diff
@@ -57,12 +57,34 @@ private IndexStats(String label, String property, long nodeCount) {
     @Procedure
     @PerformsWrites
     @Description("apoc.index.addAllNodes('name',{label1:['prop1',...],...}) YIELD type, name, config - create a free text search index")
-    public Stream<IndexStats> addAllNodes(@Name("index") String index, @Name("structure") Map<String, List<String>> structure) {
+    public Stream<IndexStats> addAllNodes(@Name("index") String index, @Name("structure") Map<String, List<String>> structure ) {
+        return addAllNodesExtended( index, structure, new HashMap<String,Object>(0) );
```
jexp: Collections.emptyMap()

PR #168, opened 2016-10-10
Upgrade procedure compiler

fbiville:
It now supports user-defined functions and allows CSV export.


fbiville:
Could we try a rebuild on this?

fbiville:
I'll leave the SNAPSHOT version of the compiler until the CSV output is deemed acceptable (cc @jexp)

fbiville:
This will definitely need a rebuild because I screwed up the SNAPSHOT deployment and I had not noticed until now.

fbiville:
Closing for now, as that'll need some rework

PR #184, opened 2016-10-21
apoc.warmup.run() output is not well understood

Dcanzano:
Using Neo4j 3.0.6 I have my conf/neo4j.conf defined with

----
Neo4j, and is heuristically set to 50% of RAM minus the max Java heap size.
dbms.memory.pagecache.size=2G
----

and the results of my du -hc data/databases/graph.db/*store.db* reports a tail of 

8.0K    neostore.schemastore.db
4.0K    neostore.schemastore.db.id
6.2M    total

so clearly my graph.db should fit entirely in the cache.

But upon restart if I run

call apoc.warmup.run()

the results report

pageSize,nodesPerPage,nodesTotal,nodesLoaded,nodesTime,relsPerPage,relsTotal,relsLoaded,relsTime,totalTime
8192,546,50183,94,0,240,10256,43,0,0

to which this would report nodeTotal = 50183  (and this is correct for match (n) return count(n) returns this value).   But why does nodesLoaded report 94.   Is not this the number of nodes loaded into the cache?   The same can be said about relsTotal and relsLoaded.

apoc output reports relsTotal = 10256

match (n)-[r]-() return count(distinct r)   = 10256

but yet apoc reports relsLoaded = 43 ????

in the end either I dont understand what the output is trying to convey and had the output done what I expected I was then hopeing to use the output to produce output similar to 'All of graph.db is in the page cache' which might be reported if nodesLoaded=nodesTotal and relsLoaded=relsTotal.

is nodesLoaded actually number of pages of Nodes.  and since nodesPerPage=546 then if nodesLoaded*nodesPerPage  (i.e. 94*546 = 51324) > nodesTotal (i.e. 50183) means it is all in graph.   i.e. we have 94 pages of nodes an each page can support up to 546 nodes and therefor we could support up to 51324 nodes into these 94 pages and since this number is greater than nodesTotal (i.e. all nodes should be able to fit into the 94 pages) then this is now all in cache?    Is that how it is to be read?   if that is the case then can we rename nodesLoaded to PagesNodeLoaded?  or at a minimum leave the name as is but provide documentation on what its meaning is.     Also maybe we could add a new return value which simply reports whether the entire graph.db was loaded into cache or whether the cache was too small

further if I delete all my nodes via a 

match (n) detach delete n;

and then run 

call apoc.warmup.run();

the results are then

pageSize,nodesPerPage,nodesTotal,nodesLoaded,nodesTime,relsPerPage,relsTotal,relsLoaded,relsTime,totalTime
8192,546,0,94,0,240,0,43,0,0

which does report nodesTotal = 0  which is correct since I just deleted all nodes/all relationships, but
nodesLoaded=94 ?   ???


jexp:
Something is totally off with your PR ?
You have to rebase it.

Basically nodes/rels-loaded is how many page-cache pages were touched to trigger loading each page.

jexp:
We can rename the columns to "nodePagesLoaded" and "relPagesLoaded" ? Would that make more sense?

Dcanzano:
Yes.   Better as a first start.    Also when I delete all nodes all rels it
still reports it is loading nodes / rels which is odd until u understand
that deleted nodes / rels are not really deleted

PR #203, opened 2016-11-06
Add `apoc.script.engines` and `apoc.script.eval`.

### LINKED ISSUE #202: Allow to easily call a java utility method.
From [#163](https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/163#issuecomment-258652873)

I'd be nice to have a way to easily call a POJO method without having to create a server plugin just for it. For example, some java libs could be useful on neo4j, like generating a [geohash](https://github.com/davidmoten/geo) from lat,lon. Or calling simple one-shot standard library utils.

vic:
This branch implements a couple of user-functions to let you
call any JSR 223 (javax.script) language from within cypher.

By default Oracle's JDK includes the Nashorn Javascript engine,
so if you are running neo4j with it, you can at least eval
some javascript code.

You can also drop other languages jars into the plugin/
directory (be sure they have or you provide a JSR-223 factory)

`apoc.script` can be useful for calling one-shot java utilities
without having to create full server plugins for them.

Currently there is no way to load scripts from fileystem, and
perhaps never will, as the idea is that scripts should be just
simple one-liners.

This branch is based on the `3.1` branch.

Closes #202

### DIFF
--- a/src/main/java/apoc/script/EnginesResult.java
+++ b/src/main/java/apoc/script/EnginesResult.java
@@ -1,0 +1,70 @@
+package apoc.script;
+
+import javax.script.ScriptEngineFactory;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * @author Victor Borja vborja@apache.org
+ * @since 12.06.16
+ */
+public class EnginesResult {
+
+    public final String name;
+    public final String version;
+    public final String languageName;
+    public final String languageVersion;
+    public final String names;
+
+    public EnginesResult(ScriptEngineFactory factory) {
+        this.name = factory.getEngineName();
+        this.version = factory.getEngineVersion();
+        this.languageName = factory.getLanguageName();
+        this.languageVersion = factory.getLanguageVersion();
+        this.names = factory.getNames().stream().collect(Collectors.joining(","));
+    }
+
+    public EnginesResult(Map<String, Object> row) {
+        this.name = (String) row.get("name");
+        this.version = (String) row.get("version");
+        this.languageName = (String) row.get("languageName");
+        this.languageVersion = (String) row.get("languageVersion");
+        this.names = (String) row.get("names");
+    }
+
+    @Override
+    public String toString() {
+        return "EnginesResult{" +
+                "name='" + name + '\'' +
+                ", version='" + version + '\'' +
+                ", languageName='" + languageName + '\'' +
+                ", languageVersion='" + languageVersion + '\'' +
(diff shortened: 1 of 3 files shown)


jexp:
Hey @vic 
This looks good to me, but for security reasons, I'm reluctant to put it in apoc.

I was planning to add a "scripting" procedure library, that also integrates my work from here: 
https://github.com/jexp/neo4j-javascript-procedures

So that people who want to use this kind of dynamic, less safe approach (e.g for development) can do it without compromising the applicability of APOC in a general way.

vic:
@jexp yeah I understand your point and sounds right to me. is `neo4j-script-producedures` focused on running scripts with `nashorn` only ? or you will be able to drop any JSR-223 language jar and use that. Would you be interested in me moving this patch to `neo4j-script-procedures`or shall it be a completely separate plugin?

vic:
As a sidenote: Another example of a DB that allows scripting is Elasticsearch, the way they handle
the security concerns is by only allowing you to invoke scripts by name (instead of passing in the code)
and those scripts have to been already registered (in ES case, as a file under a `scripts/` directory).

How about only allowing to invoke scripts that have been registered by name (say on neo4j.conf) would that be reasonable?

jexp:
Yes, configured/registered scripts in a script directory are better.

And invoking them by name, perhaps with a procedure / function that also has different variants for different return types. e.g. apoc.run.function.boolean(name, params).

One question would be if one passes params as a list or map, or just pass in an "Object" which could be any of those.

so there would have to be a configuration syntax for registering functions, e.g.

neo4j.conf:

```
apoc.register.function.time=java.lang.System::currentTimeMillis
apoc.register.function.pareto=pareto.js::pareto
```

I would probably apply your patch to `neo4j-script-procedures` after I now moved it over to neo4j-contrib.

jexp:
@vic would you mind retargeting this PR to https://github.com/neo4j-contrib/neo4j-script-procedures ?

And if you have time, I'd be happy to take a PR that supports executing scripts by name and with params that have been put/registered in a /scripts directory. That should be possible for all registered script engines, or?

vic:
Sure!  I'm closing this issue and will open other for script-procedures.

jexp:
@vic Any updates on this? I think making the scripting procedures more powerful (also support other engines) and secure would be very useful. And we could benefit from your experience in ES.

Please re-open the PR on the https://github.com/neo4j-contrib/neo4j-script-procedures repository.

PR #215, opened 2016-11-21
added apoc.date.systemTimezone() to retrieve the system timezone prog…

ikwattro:
…ramatically


ikwattro:
For example, : 

```
RETURN apoc.date.formatTimezone(timestamp(), "ms", null, apoc.date.systemTimezone())
```

This would also https://github.com/neo-technology-field/knowledge-base/blob/be965a8a65d284760c3c0e2e8ca62547ee9d3cbb/articles/how-do-i-display-date-and-time-of-when-neo4j-was-started-and-other-metrics.adoc to return the correct value

jexp:
Looks good to me, can you also add it to 3.0 ?

jexp:
Oh and please add it to overview.adoc

And perhaps add the usage-example from the KB to the docs?

ikwattro:
```
Looks good to me, can you also add it to 3.0 ?
```

Sure. 

---

```
Oh and please add it to overview.adoc

And perhaps add the usage-example from the KB to the docs?
```

Ok for overview. 

This is a private KB item, so maybe not.

PR #222, opened 2016-12-01
Update guide.html

Dcanzano:
add details that apoc.warmup.run() does not read properties into the cache


jexp:
Please don't update the generated html files (which will be overriden) but the .adoc source files.

Issue #289, opened 2017-02-20
Feature Request: date conversion and add functions


InverseFalcon:
Most of APOC's date/time functions deal with parsing a timestamp (at some time resolution) from a string, and formatting to go the other direction.

APOC could use some additional date functions for converting a timestamp between different resolutions, and adding/subtracting some amount of time at a specific resolution (though keeping the output at original resolution).

Some possible functions:

`apoc.date.convert(<timestamp>, <timestampResolution>, <desired resolution>)`
example usage: 
`return apoc.date.convert(timestamp(), 'ms', 'd') as todayInDayResolution`

`apoc.date.add(<timestamp>, <timestampResolution>, <numberToAdd>, <addResolution>)`
example usage:
`return apoc.date.add(timestamp(), 'ms', -3, 'd') as threeDaysAgoInMs`

jexp:
Thanks a lot @InverseFalcon !

PR #291, opened 2017-02-21
Added apoc.date.convert() and apoc.date.add()

InverseFalcon:
This fulfills the feature request in #289.

### DIFF
--- a/src/main/java/apoc/date/Date.java
+++ b/src/main/java/apoc/date/Date.java
@@ -136,6 +136,19 @@ public class Date {
 		return TimeZone.getDefault().getID();
 	}
 
+	@UserFunction
+	@Description("apoc.date.convert(12345,'ms|s|m|h|d','ms|s|m|h|d') convert a time unit into a different time unit")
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/date/Date.java
```diff
@@ -136,6 +136,19 @@ public String systemTimezone() {
 		return TimeZone.getDefault().getID();
 	}
 
+	@UserFunction
+	@Description("apoc.date.convert(12345,'ms|s|m|h|d','ms|s|m|h|d') convert a time unit into a different time unit")
+	public Long convert(@Name("time") long time, @Name(value = "unit") String unit, @Name(value = "toUnit") String toUnit) {
```
jexp: are there any sensible default values?

jexp:
looks good to me, can you add them to docs/overview.adoc too?

src/main/java/apoc/date/Date.java
InverseFalcon: Maybe 'ms' for the `unit` value, but I'm wondering if in this case a default value would be easier to work with or make usage more error-prone.

jexp: Probably not worth it then.

Issue #287, opened 2017-02-18
Feature Request: Cosine similarity


InverseFalcon:
An example of query calculating cosine similarity:

```
MATCH (p1:Person)-[x:RATED]->(m:Movie)<-[y:RATED]-(p2:Person)
WITH SUM(x.rating * y.rating) AS xyDotProduct,
SQRT(REDUCE(xDot = 0.0, a IN COLLECT(x.rating) | xDot + a^2)) AS xLength,
SQRT(REDUCE(yDot = 0.0, b IN COLLECT(y.rating) | yDot + b^2)) AS yLength,
p1, p2
MERGE (p1)-[s:SIMILARITY]-(p2)
SET s.similarity = xyDotProduct / (xLength * yLength)
```

taken from [Nicole White's graphgist on movie recommendations](https://neo4j.com/graphgist/a7c915c8-a3d6-43b9-8127-1836fecc6e2f).

This proposes a new function `apoc.algo.cosineSimilarity([[rel1A, rel1B], [rel2A, rel2B], ...], weightProp, defaultVal)`

Input is a collection of collected relationship pairs, the property on relationships used for the calculation, and the default value if the weight property is absent.

Output is the calculated similarity.

InverseFalcon:
Changed the relationship input from two collections to a collection of relationship pairs. COLLECT() wouldn't include null relationships, so this alternate type will have to do.

InverseFalcon:
Fulfilled.

PR #293, opened 2017-02-23
Added function apoc.algo.cosineSimilarity()

InverseFalcon:
This takes a list of relationship pairs, a weight property, and a default weight, and returns cosine similarity.

This fulfills feature request #287

### DIFF
--- a/src/main/java/apoc/algo/Similarity.java
+++ b/src/main/java/apoc/algo/Similarity.java
@@ -1,0 +1,51 @@
+package apoc.algo;
+
+import apoc.util.Util;
+import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.graphdb.Relationship;
+import org.neo4j.procedure.Context;
+import org.neo4j.procedure.Description;
+import org.neo4j.procedure.Name;
+import org.neo4j.procedure.UserFunction;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class Similarity {
+    @Context
+    public GraphDatabaseService db;
+
+    @UserFunction
+    @Description("apoc.algo.cosineSimilarity([[rel1A, rel1B], [rel2A, rel2B], …], weightPropertyName, defaultWeight: 0) " +
+            "given a collection of collected pairs of relationships, the weight property of the relationships, " +
+            "and a default weight (if relationship or property is null), calculate cosine similarity")
+    public Double cosineSimilarity(
+            @Name("relPairs") List<List<Relationship>> relPairs,
+            @Name("weightPropertyName") String weightPropertyName,
+            @Name(value = "defaultWeight", defaultValue = "0.0") Double defaultWeight) {
+
+        List<Double> weight1 = new ArrayList<>();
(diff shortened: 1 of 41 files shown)


src/main/java/apoc/algo/Similarity.java
```diff
@@ -0,0 +1,51 @@
...
+    @UserFunction
+    @Description("apoc.algo.cosineSimilarity([[rel1A, rel1B], [rel2A, rel2B], …], weightPropertyName, defaultWeight: 0) " +
+            "given a collection of collected pairs of relationships, the weight property of the relationships, " +
+            "and a default weight (if relationship or property is null), calculate cosine similarity")
+    public Double cosineSimilarity(
+            @Name("relPairs") List<List<Relationship>> relPairs,
+            @Name("weightPropertyName") String weightPropertyName,
+            @Name(value = "defaultWeight", defaultValue = "0.0") Double defaultWeight) {
+
+        List<Double> weight1 = new ArrayList<>();
```
jexp: don't use Double but double primitives (also for default value -> `double defaultValue = defaultWeight.doubleValue();`
can you try to use double arrays, instead of the array-lists (both the lists and arrays should be pre-sized)
I also wonder if we can compute the running totals in one go without filling the lists first? It seems we only need 3 running values: the dotProduct, the xLength and yLength.

and instead of Math.pow(n,2) use n*n

jexp: This would definitely also need a documentation page.

I could imagine one that takes the computed similarities and then if they are above a certain treshhold (e.g. 80%)  creates shortcut relationships between the nodes with the similarity.

The upfront computation & aggregation can be quite expensive, so I wonder if we can add a procedure that gets passed in a cypher statement which would return the nodes and relationships which are then streamed through the computation (keyed on the node-pairs) and returns a stream of node-pairs with a similarity. But that should probably all go into a seperate PR and can use the existing code.

jexp:
Sorry for the delay, I made some comments.

src/main/java/apoc/algo/Similarity.java
InverseFalcon: Sensible suggestions, I made the changes accordingly. Please let me know if you see something I missed.

InverseFalcon:
I'd like to generalize this one, have it take two collection vectors of double values and leave it up to the caller where those are coming from, as they can come from relationships or simply properties of two different nodes. This would require the caller to also handle null properties and relationships, likely with COALESCE() as they assemble the collection vectors before the call.

I'm wondering though if there's any value in keeping the relationship-oriented function as a separate function, as it has means of dealing with null properties and null relationships (from optional matches). Or does that seem unnecessary, once a more general function is available?

jexp:
Can you add docs for this one?

I think the generalized function makes sense, please look into it. I think a list of doubles (or more generally Number) is good enough, you can then use Number.doubleValue() to get the actual values.

Going forward (3.2) we can also look into making this an aggregation function, so we can skip the collect (which potentially creates a huge list).

InverseFalcon:
Retiring. Replaced by #315

PR #296, opened 2017-02-25
Added new functions to apoc.coll - shuffle(), randomItem(), randomItems().

InverseFalcon:
Going to use the 3.1 branch for this the additions can be functions instead of procedures, and so we can allow optional parameters.


jexp:
Thanks a lot, cool stuff.

PR #279, opened 2017-02-13
Added apoc.path procedures: subgraphNodes(), subgraphAll(), spanningTree()

InverseFalcon:
This is meant to fulfill feature requests #273 #274 and #275

Issue #275, opened 2017-02-11
Feature Request: apoc.path.subgraphFull()


InverseFalcon:
This builds off of the proposed [apoc.path.subgraphNodes()](https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/274) request.

Although apoc.path.subgraphNodes() returns nodes for a subgraph, it doesn't return the complete subgraph with all relationships that exist between all subgraph nodes.

The collection of relationships can be obtained by passing the collection of subgraph node ids to apoc.algo.cover().

This requests a new procedure, apoc.path.subgraphFull(), which uses both subgraphNodes() and Cover.cover() (or a similar method which takes the nodes directly instead of the ids), and returns a GraphResult of the subgraph nodes and all relationships between them.

Assiam:
It is exactly what I have done. I suggest to combine the two algorithms.
Thank you for your help

InverseFalcon:
Fulfilled.

Cyoung7:
@Assiam  this method apoc.path.subgraphAll() can get all the nodes and edges within subgraph,but can not get start node and end node in each edge. how can i solve it. wating for your reply, thanks!

InverseFalcon:
@Cyoung7 You can get the start and end nodes of each edge from the relationship. In Cypher, the functions `startNode(rel)` and `endNode(rel)` will give you the start and end nodes of a relationship.

Issue #274, opened 2017-02-11
Feature Request: apoc.path.subgraphNodes()


InverseFalcon:
apoc.path.expandConfig(), with certain configuration values, returns paths whose end nodes form the nodes of a subgraph starting from the given start nodes and adhering to relationship and label filters up to the given maxLevel.

The proposed [apoc.path.spanningTree()](https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/273) procedure already performs the necessary config parameter changes, the only missing step is to remap the PathResult stream to a NodeResult stream of the end nodes of the path.

This requests a new procedure, apoc.path.subgraphNodes(), which wraps the (proposed) apoc.path.spanningTree() call to perform the remapping of the stream values to a NodeResult stream.

InverseFalcon:
Fulfilled.

Issue #273, opened 2017-02-11
Feature Request: apoc.path.spanningTree()


InverseFalcon:
apoc.path.expandConfig(), with certain configuration values, returns paths which form a spanning tree.

This requests a new procedure, apoc.path.spanningTree(), which wraps the expandConfig() call with the following config parameter alternations to return the spanning tree paths:

- "minLevel" removed
- "uniqueness" set to "NODE_GLOBAL"

InverseFalcon:
Fulfilled.

PR #305, opened 2017-03-07
Added new apoc.path procedures: subgraphNodes(), subgraphAll(), and spanningTree()

InverseFalcon:
Also added a static Cover.cover() method for internal usage, and updated path expander documentation.

This replaces PR #279, switching to base on APOC 3.1 branch.

This fulfills feature requests #273 #274 and #275.

### DIFF
--- a/docs/expand.adoc
+++ b/docs/expand.adoc
@@ -1,8 +1,18 @@
 = Expand paths
 
-expand from start node following the given relationships from min to max-level adhering to the label filters
+Expand from start node following the given relationships from min to max-level adhering to the label filters. Several variations exist:
 
-== Usage
+`apoc.path.expand()` expands paths using Cypher's default expansion modes (bfs and 'RELATIONSHIP_PATH' uniqueness)
+
+`apoc.path.expandConfig()` allows more flexible configuration of parameters and expansion modes
+
+`apoc.path.subgraphNodes()` expands to nodes of a subgraph
+
+`apoc.path.subgraphAll()` expands to nodes of a subgraph and also returns all relationships in the subgraph
+
+`apoc.path.spanningTree()` expands to paths collectively forming a spanning tree
+
+== Expand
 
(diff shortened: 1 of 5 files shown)


jexp:
Thanks a lot, looks good!

pducrot:
Thanks for providing these apocs. It's a bit confusing to me that the spanningTree sometimes returns a graph instead of a tree, for instance when you start from several nodes in a directed graph.
MATCH (p:Person) WHERE size((p)<-[:LIKES]-())=0 and size((p)-[:LIKES]->())>0
CALL apoc.path.spanningTree(f, {labelFilter:'+Person', relationshipFilter: 'LIKES'}) 
YIELD path
I suspect one tree to be returned for each start and display to merge all trees. Would it be possible to optionally return a single tree or return multiple trees in which paths would be identified by a tree ID e.g. the starting node?
I haven't tested the 3.2 apoc version yet.

InverseFalcon:
It's a good point, you're only guaranteed a tree when you have a single starting node, otherwise you'll get a union of the trees.

If you only want to view a tree for one starting node at a time, you can either use SKIP and LIMIT on the starting node itself, like `MATCH (p:Person) WHERE size((p)<-[:LIKES]-())=0 and size((p)-[:LIKES]->())>0 WITH p LIMIT 1 ...`, or you can collect the paths per starting node and SKIP and LIMIT at the end:

```
MATCH (p:Person) WHERE size((p)<-[:LIKES]-())=0 and size((p)-[:LIKES]->())>0
CALL apoc.path.spanningTree(f, {labelFilter:'+Person', relationshipFilter: 'LIKES'})
YIELD path
RETURN p, collect(path) SKIP 1 LIMIT 1
```

Keep in mind that a tree is comprised of multiple paths (one for each node in the tree), so collecting the paths or limiting the starting node to 1 is the best way to keep the trees separate.

PR #311, opened 2017-03-10
Generate documentation

fbiville:


fbiville:
@jexp I configured the compiler to generated the global list (SINGLE) and one list per enclosing package (PACKAGE) in `docs/generated/`.

Feel free to have a look.

jexp:
Yay, really cool. 
but we shouldn't check those in if they are generated :)

I have a look later today 

Thank you!!

Michael 

Von meinem iPhone gesendet

fbiville:
@jexp it's just so you can get a preview, I'll remove them in the final version of the PR :-)

PR #319, opened 2017-03-15
Added 'filterStartNode' config parameter to expandConfig()

InverseFalcon:
(and all procedures using the config map). Default is false for versions < 3.2.x.x.


jexp:
Lgtm I would only change the config name to "filterStartNode:true/false"

InverseFalcon:
That does sound better, I'll make the change.

Issue #325, opened 2017-03-18
Feature Request: New label filter operation on path expanders - end node filter ">"


InverseFalcon:
This requests a new label filter operation: ">" - end node filter.

The end node filter is meant to work identically to the termination filter "/", ensuring returned paths are only to nodes with labels in the filter, but it does not prune the path when matched; expansion can continue along the path past the end node, possibly matching on other end nodes further down the path.

NOTE: Updated operator symbol from '%' to '>'

InverseFalcon:
Request fulfilled.

PR #327, opened 2017-03-19
Add end node operator to path expander label filter

InverseFalcon:
The new operator, `>`, works just like the termination operator `/` in that labels in the filter will be used as end nodes in the expansion, except traversal is allowed past these end nodes.

The new `limit` config parameter applies when using this filter.


InverseFalcon:
This fulfills feature request #325

sarmbruster:
thanks @InverseFalcon for your PR.

PR #335, opened 2017-03-20
GephiWeightedRelationships

tomasonjo:
I added an optional weight property parameter, so you can pick which property do you want to use as a weight or set it at null for standard 1.0 weight

### DIFF
--- a/src/main/java/apoc/gephi/Gephi.java
+++ b/src/main/java/apoc/gephi/Gephi.java
@@ -42,31 +42,31 @@ public class Gephi {
     // http://127.0.0.1:8080/workspace0?operation=updateGraph
     // TODO configure property-filters or transfer all properties
     @Procedure
-    @Description("apoc.gephi.add(url-or-key, workspace, data) | streams passed in data to Gephi")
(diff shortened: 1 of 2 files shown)


jexp:
Make the weightProperty (call it weightProperty) optional (i.e. defaultValue="null")
and if none is set also send none.

If a weightProperty is given but none on the node, I'm not sure what the default should generally be for Gephi? 0 or 1 or ... ?

jexp:
Also add a test or two for the new weightProperty

tomasonjo:
I think this looks/works fine, will write a test or two tommorow... thanks for the help

jexp:
Looks good to me, waiting for the tests before I merge.

jexp:
Ping @tomasonjo please add the tests so that we can merge :)

tomasonjo:
Tested it with Gephi running and the results were ok in Gephi. Also Gephi does not handle updating relationship weight, so I created a new label for each test. There seems to be some problem if data is not clean meaning that weight property is randomly a string/number or null within same call.

Issue #199, opened 2016-11-03
Numeric constants for minimum and maximum value


Gudahtt:
It would be helpful to have constants that represent the minimum and maximum possible value for each numeric type (int and float). Something analogous to `Integer.MIN_VALUE`, `Integer.MAX_VALUE`, `Float.MIN_VALUE`, `Float.MAX_VALUE` in Java.

AngeloBusato:
Do you want to expose this constants as UDF?

Gudahtt:
Sure, that works. As long as they're available somehow.

PR #338, opened 2017-03-21
#199 - Numeric constants for minimum and maximum value

AngeloBusato:


jexp:
Looks good.
I think in sake of brevity I would leave off `Value` at the end: `apoc.math.maxLong()` shoudl be good enough.
Could you please add the same for int and byte.

AngeloBusato:
Ok, did it.

jexp:
Thanks a lot.

PR #340, opened 2017-03-22
rest url should have not final question mark if no query fixes #318

### LINKED ISSUE #318: apoc.es.query does not work if arg search is null
When apoc.es.query is called with only a payload but a null search argument, the URL must not have the "?" character at the end.

albertodelazzari:
Refactoring of some methods in ElasticSearch class in order to do unit test for the url creation (according to the parameters: index, type, id, query ...)


jexp:
Looks good. Thanks!

Issue #317, opened 2017-03-14
Improvement: Allow path expander label filters to use multiple operations


InverseFalcon:
The available operations when using path expander label filters are: `+`,`-`, `/`, `>`

Currently, only one such operation is allowed in the label filter. This isn't especially clear, as it's easy to interpret some of the syntax as allowing multiple operations.

This requests a change to the label filter to allow up to all operations to be present in the filter. When an operation is specified as the label filter is parsed, it remains the active operation to apply for all subsequent labels until another operation symbol is detected at the start of a label.

However, this does present a challenge for both evaluation precedence (for multi-labeled nodes), and for general interaction between the operations.

Decisions regarding operator interaction:

1. Whitelists (`+`) - Specifically, when no whitelist labels are provided. This is an issue because `-` and `/` and `>` operations had no concept of whitelists, and never had to filter based on them, or their absence. I propose interpreting lack of whitelist labels (empty whitelist) as always allowed, rather than nothing allowed.

2. Termination and end node operators (`/` and `>`) - These operations have a significant effect on the paths returned. They use `EXCLUDE_AND_CONTINUE` for nodes that don't have the associated labels. The only paths returned when these operations are in play are paths with end nodes of the associated labels. I propose that this behavior is kept (docs have been updated to better describe this behavior), but it will affect the outcome of whitelist operations. 

That is, when a node is allowed by the whitelist (either by the whitelist being empty as in above, or when a node's label is in the whitelist), and not present in the blacklist or the include-and-prune list, the behavior must change based on whether or not the `/` and/or `>` operations are in play on certain labels. If so, then allowed nodes through the whitelist must apply an EXCLUDE_AND_CONTINUE evaluation, otherwise they should apply INCLUDE_AND_CONTINUE.

Proposal regarding operator precedence and multi-labeled nodes:

We're evaluating a node that has labels associated with some combination of the operators, instead of just one. What do we do?

I propose the following precedence, defaulting to respecting exclusion first, and inclusion second:

```
- first
/ second
> third
+ fourth
```

According to this proposed precedence, the first operator whose label list includes the label of the evaluated node will handle it accordingly, and no further evaluation will occur.

InverseFalcon:
Feedback and suggestions welcome.

jexp:
@sarmbruster can you have a look.

InverseFalcon:
Updated to include reference to the newly introduced end-node filter `>` in PR #327

[...]

PR #341, opened 2017-03-26
Allow all labelFilter operations in path expander procedures

InverseFalcon:
This implements proposed order 1 operator precedence from #317, including the adjustment on evaluation for the end node filter which first checks if the a label is in the whitelist before an INCLUDE_AND_CONTINUE, to avoid including paths to further end nodes which inadvertently include a node that violates the whitelist.

See included test cases for examples of operator precedence.

Please don't approve pull request yet, documentation is not yet included.

Documentation will come in a later commit, as I'd like to get some feedback and confirmation on the implementation first.

### DIFF
--- a/docs/expand.adoc
+++ b/docs/expand.adoc
@@ -37,17 +37,25 @@ Syntax: `[<]RELATIONSHIP_TYPE1[>]|[<]RELATIONSHIP_TYPE2[>]|...`
 
 Syntax: `[+-/>]LABEL1|LABEL2|...`
 
-Only one type of operation allowed in the label filter (`+` or `-` or `/` or `>`, never more than one).
+As of APOC 3.1.3.x multiple label filter operations are allowed.
+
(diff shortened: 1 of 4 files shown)


InverseFalcon:
Also looking for advice on documenting which version this feature is available, given that we seem to be supporting a 3.1.2.x branch alongside a 3.1.0.x branch. 

Maybe it will be easier if this feature will only go on the latest 3.1.2.x branch to avoid confusion.

Advice welcome.

jexp:
Yes we will focus on 3.1.3.x going forward, so don't bother about 3.1.0.x.

As I just said on the issue #317 I agree that Option 1 is most suitable.

I think the docs should be succinct but cover the cases by example.

Perhaps with either enough examples a matrix that shows the effects of the parametrization.

InverseFalcon:
Added documentation. This is a complicated feature, so suggestions and refinement are welcome.

Issue #337, opened 2017-03-21
follow redirect for URLs


jexp:
eg. for load.json / load.xml / load.csv

Issue #70, opened 2016-07-22
CALL apoc.load.json path ignores config file setting


jphare:
Great set of stored procs!

I noticed a small issue with CALL apoc.load.json, not sure I'd call it a bug though.  If you are loading JSON locally you have to use a different path convention to the CSV loader if you have a default path set in the config file.  Not a massive issue but using windows 10 I seem to spend a fair amount of time tweaking paths until they work.....

For example with windows 10 the CSV loader format is:

USING PERIODIC COMMIT
LOAD CSV WITH HEADERS FROM "file:///file.csv" AS line

The equivalent in APOC would be:

WITH "file:///C:/default_location/file.json" AS url
CALL apoc.load.json(url) YIELD value

jexp:
Yes it's totally different apis

But good point at aligning them

Von meinem iPhone gesendet

inserpio:
Need to read `dbms.directories.import` in the `neo4j.conf` file and act accordingly

PR #342, opened 2017-03-27
issue #337 follow redirect for URLs

DanielBerton:


src/main/java/apoc/util/Util.java
```diff
@@ -2,10 +2,11 @@
 
 import apoc.ApocConfiguration;
 import apoc.Pools;
-import apoc.export.util.*; // todo
+import apoc.export.util.CountingInputStream;
```
jexp: can we please not change imports if it's not needed?

jexp:
I general it would be super helpful to separate reformatting commits from actual code-change commits.
That makes it much easier to review.

src/main/java/apoc/util/Util.java
```diff
@@ -210,40 +215,71 @@ public static Double toDouble(Object value) {
...
-        if (value instanceof Number) return ((Number)value).longValue();
+        if (value instanceof Number) return ((Number) value).longValue();
         return Long.parseLong(value.toString());
     }
 
     public static URLConnection openUrlConnection(String url, Map<String, Object> headers) throws IOException {
+
         URL src = new URL(url);
-        URLConnection con = src.openConnection();
+        HttpURLConnection con = (HttpURLConnection) src.openConnection();
```
jexp: we cannot assume that it is an HTTPURLConnection
we have to check after the fact for redirect handling

jexp: So we have to have one branch where it returns HttpURLConnection and we can check the redirect. 
And another branch where it is another kind of URLConnection and we cannot check the redirect and just return the connection.

src/main/java/apoc/util/Util.java
```diff
@@ -210,40 +215,71 @@ public static Double toDouble(Object value) {
...
+    public static boolean isRedirect(int code) {
+        return code >= 300 && code < 400;
+    }
+
+    private static void setConnectionParams(HttpURLConnection con, Map<String, Object> headers) throws ProtocolException {
         con.setRequestProperty("User-Agent", "APOC Procedures for Neo4j");
+        con.setInstanceFollowRedirects(false);
         if (headers != null) {
             Object method = headers.get("method");
-            if (method != null && con instanceof HttpURLConnection) {
```
jexp: keep this check

src/main/java/apoc/util/Util.java
```diff
@@ -210,40 +215,71 @@ public static Double toDouble(Object value) {
...
+        setConnectionParams(con, headers);
+        int code = con.getResponseCode();
+        return isRedirect(code) ? openUrlConnection(con.getHeaderField("Location"), headers) : con;
+    }
+
+    public static boolean isRedirect(int code) {
+        return code >= 300 && code < 400;
+    }
+
+    private static void setConnectionParams(HttpURLConnection con, Map<String, Object> headers) throws ProtocolException {
```
jexp: change the method signature back
we want to cater for more url-handlers not fewer :)

src/main/java/apoc/export/util/FileUtils.java
```diff
@@ -4,27 +4,29 @@
...
 
-    public static CountingReader readerFor(String fileName) throws IOException {
+    public static CountingReader readerFor(String fileName) throws IOException, IllegalArgumentException {
         checkReadAllowed(fileName);
-        if (fileName==null) return null;
-        if (fileName.toLowerCase().startsWith("http") || fileName.toLowerCase().startsWith("file:")) {
-            return Util.openInputStream(fileName,null,null).asReader();
-        }
+        if (fileName == null)
+            throw new IllegalArgumentException("The file doesn't exists");
```
inserpio: ok

src/main/java/apoc/export/util/FileUtils.java
```diff
@@ -4,27 +4,29 @@
...
+
+        if (fileName.toLowerCase().startsWith("file:") || fileName.toLowerCase().startsWith("http"))
```
inserpio: we planned to do this for #70, just right after this PR.

src/main/java/apoc/util/Util.java
inserpio: My fault. I always ask my guys to do "reorganize import" before committing, in order to remove warnings for useless imports.

src/main/java/apoc/util/Util.java
inserpio: You mean for example an HTTP could redirect i.e. to FILE o something else like FTP?
In this case you're right: we test the protocol outside the this recursive method.

src/main/java/apoc/util/Util.java
inserpio: Sure, we're trying to separate different protocols in dedicated pieces of code, in order to avoid test like `if (con instanceof HttpUrlConnection)` do X else do Y...
Let us review it.

src/main/java/apoc/util/Util.java
jexp: Only HTTP can redirect, with Location Header, the other protocols can't.
But we still have to handle and support them correctly.

src/main/java/apoc/util/Util.java
jexp: Ah ok, we just have to make sure that the rest of the method still applies to the other connection-types.

src/main/java/apoc/util/Util.java
jexp: At some point we have to have the instanceof check (once) but here, it's just a plain cast.

src/main/java/apoc/util/Util.java
jexp: Then we should have a method that does the additional work for HttpURLConnection which is called from the right place. But the rest should still happen.

src/main/java/apoc/util/Util.java
```diff
@@ -210,40 +215,71 @@ public static Double toDouble(Object value) {
...
+        con.setDoInput(true);
+        con.setConnectTimeout((int) toLong(ApocConfiguration.get("http.timeout.connect", 10_000)));
+        con.setReadTimeout((int) toLong(ApocConfiguration.get("http.timeout.read", 60_000)));
+
         return con;
     }
 
     public static CountingInputStream openInputStream(String url, Map<String, Object> headers, String payload) throws IOException {
-        URLConnection con = openUrlConnection(url, headers);
+        URLConnection con = null;
```
jexp: this whole block should be encapsulated in openUrlConnection

jexp:
Looks good, thank you also for the changes!

Issue #250, opened 2017-01-18
apoc.data.domain not working properly for emails


tomasonjo:
`return apoc.data.domain("foo.bar@foobar.com")` returns foo.bar

seems like the pattern is that when there is a dot on left side of @ function messes up

InverseFalcon:
I can take this one, should be fairly easy, we can just iterate over the split tokens in reverse if we detect an @ sign.

Note though that this may give bad results for URLs with "@" in the query string, so that should probably be avoided by filtering out before the function call, if possible. The most robust solution if needed later on would be splitting this into two separate functions.

PR #344, opened 2017-03-27
Fixed apoc.data.domain() when there are .'s before the @ symbol

InverseFalcon:
Fixes bug #250, but may alter behavior when run on URLs with an @ in the query string which shouldn't be picked up as the domain (if the domain part of an embedded email address, for example, is the last part of the query string, or immediately precedes a split delimiter)

### DIFF
--- a/src/main/java/apoc/data/Extract.java
+++ b/src/main/java/apoc/data/Extract.java
@@ -1,17 +1,10 @@
 package apoc.data;
 
-import org.neo4j.helpers.collection.ArrayIterator;
 import org.neo4j.procedure.Description;
-import apoc.result.StringResult;
 import org.neo4j.procedure.Name;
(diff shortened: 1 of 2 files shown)


jexp:
@InverseFalcon so it should handle URLs with `username:password@host:port` as well?

Cool, thanks a lot.

InverseFalcon:
@jexp Not necessarily. We're not splitting on `:` currently, so the `:port` is part of the domain token, which fails the domain matcher test. We could add it in as a split delimiter, that works.

Also there has to be a "." in the host for the match to work, so `...@localhost:7474` wouldn't pick up "localhost" as the domain even if `:` was a delimiter. We could make an explicit check for "localhost".

jexp:
Ok, parsing emails and URLs is a tricky topic, there are those page-long regexps for it.

Issue #345, opened 2017-03-28
apoc.es - Url encoding on query parameters should be done only on values 


albertodelazzari:
If I perform a rest call to ES with any query parameters (i.e. "?q=name:Neo4j&_source_include=name") all the query string is url-encoded while only "name:Neo4j" and "name" should be encoded.

PR #346, opened 2017-03-28
issue #345 url-encoding only on query values

albertodelazzari:


jexp:
Looks good, thanks a lot.

Issue #179, opened 2016-10-20
Document how to handle large queries from ES using apoc.es.query


Analect:
I've been experimenting with the `apoc.json.load` as well as `apoc.es.query` for loading data to neo4j directly from ES, as part of a large cypher script that gets run from a neo4j-shell.

Since data gets returned as a json document, with the relevant data contained within the nested json at `hits.hits`, I'm using a FOREACH below to access the data.  Alternatively one could also use an UNWIND.

I'm also passing in a size parameter into the ES query payload, since otherwise it was defaulting to just 10.  The approach below is failing for large queries (ie. 40k entities) .. so I need to come up with a smarter way of performing this query in chunks.

```
CALL apoc.es.query("my-server","my-index","my-type",null,{from:0, size:40000, query:{match_all:{}}}) YIELD value as data
WITH data.hits.hits AS result
FOREACH (item IN result | 
MERGE (t:type{id:item._id})
ON CREATE SET t.name=item._source.name);
```

Would anyone be able to suggest an approach where maybe I could have an outer loop in this cypher script to feed the `from` and `size` parameters ... so maybe I would do it in chunks of 1000, setting `from:0 ,size:1000` on the first iteration, `from:1000, size:1000` on the second iteration and so on ... in order to consume all the documents.  I wanted to avoid having to repeat the same script manually.

Or perhaps there is a smarter way of getting this working using an alternative approach?  Such as the ES [scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/search-request-scroll.html), although that would probably necessitate using the `apoc.json.load`, in order to be able to pass `?scroll=1m` as part of the url, since I don't think it can be handled as part of a query payload.

sarmbruster:
I guess you can solve this easily with `apoc.periodic.iterate`:

```
CALL apoc.periodic.commit(
   'CALL apoc.es.query("my-server","my-index","my-type",null,{from:0, size:40000, query:{match_all:{}}}) YIELD value as data RETURN data.hits.hits AS result',
   'UNWIND {result} AS ITEM MERGE (t:type{id:item._id}) ON CREATE SET t.name=item._source.name)',
   {batchSize:1000, parallel:false})
```
(be aware that I've not actually tested the above)

Analect:
@sarmbruster 
Thanks for your response.
[...]

PR #351, opened 2017-03-29
Add documentation for the integration of ES pagination fixes Issue #179

albertodelazzari:
The elasticsearch.adoc is now updated with a section that is focused on pagination with ES.

### DIFF
--- a/src/test/java/apoc/es/ElasticSearchTest.java
+++ b/src/test/java/apoc/es/ElasticSearchTest.java
@@ -1,8 +1,11 @@
 package apoc.es;
 
+import apoc.periodic.Periodic;
 import apoc.util.TestUtil;
+import com.jayway.jsonpath.Configuration;
+import com.jayway.jsonpath.JsonPath;
+import com.jayway.jsonpath.Option;
 import org.junit.AfterClass;
-import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.neo4j.graphdb.GraphDatabaseService;
@@ -11,8 +14,9 @@ import org.neo4j.test.TestGraphDatabaseFactory;
 import java.net.ConnectException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.UUID;
 
-import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.*;
 
 /**
  * @author mh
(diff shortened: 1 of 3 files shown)


src/test/java/apoc/es/ElasticSearchTest.java
```diff
@@ -11,42 +14,272 @@
...
     public static void tearDown() {
         db.shutdown();
     }
 
+    private static void insertDocuments() {
+        TestUtil.ignoreException(() -> {
+            String payload = "{procedurePackage:'es',procedureName:'get',procedureDescription:'performa a GET operation to ElasticSearch'}";
+            String id = UUID.randomUUID().toString();
+
+            TestUtil.testCall(db, "CALL apoc.es.post('" + HOST + "','" + ES_INDEX + "','" + ES_TYPE + "','" + id + "',null," + payload + ") yield value", r -> assertNotNull(r.get("value")));
```
jexp: you should use parameters here, not string concatenation, makes the queries much easier to read.

src/test/java/apoc/es/ElasticSearchTest.java
```diff
@@ -11,42 +14,272 @@
...
+
+            TestUtil.testCall(db, "CALL apoc.es.post('" + HOST + "','" + ES_INDEX + "','" + ES_TYPE + "','" + ES_ID + "',null," + payload + ") yield value", r -> assertNotNull(r.get("value")));
+        }, ConnectException.class);
+    }
+
+    private final Object extractValueFromResponse(Map response, String jsonPath) {
+        Object jsonResponse = response.get("value");
+        assertNotNull(jsonResponse);
+
+        String json = JsonPath.parse(jsonResponse).jsonString();
```
jexp: probably something for Util (later) 
I hoped that there is something in JsonPath that allows you to use the path directly on a Map or List or Object
so you don't have to go map -> string -> parse

src/test/java/apoc/es/ElasticSearchTest.java
jexp: and in the other places too.

jexp: and you can prepare (most of) the params once, statically

jexp:
@albertodelazzari I merge it in now, please fix my comments later.

PR #352, opened 2017-03-29
fixes #70 - CALL apoc.load.json path ignores config file setting

### LINKED ISSUE #70: CALL apoc.load.json path ignores config file setting

AngeloBusato:


src/main/java/apoc/ApocConfiguration.java
```diff
@@ -18,14 +20,24 @@
     private static final Pattern SKIP = Pattern.compile("(ur[il]|pass|cred)",Pattern.CASE_INSENSITIVE);
     private static Map<String, Object> apocConfig = new HashMap<>(10);
     private static Map<String, Object> config = new HashMap<>(32);
+    public static final List<String> PARAM_WHITELIST = new ArrayList<>(2);
```
jexp: use Arrays.asList()

docs/overview.adoc
```diff
@@ -283,6 +284,7 @@ The 3rd value in the `apoc.jdbc.<alias>.url=` effectively defines an alias to be
 
 // tag::xml[]
 Supported protocols are `file`, `http`, `https` with redirect allowed. In case no protocol is passed in the url, procedures try to check whether it is a file.
+Property `dbms.directories.import` is read only if 'apoc.import.file.use_neo4j_config' is enabled.
```
jexp: Not sure what this means? aren't config properties always read-only?

src/main/java/apoc/export/util/FileUtils.java
```diff
@@ -23,6 +25,27 @@ public static CountingReader readerFor(String fileName) throws IOException {
         return new CountingReader(file);
     }
 
+    public static String changeFileUrlIfImportDirectoryAssigned(String url){
+        if(isFile(url) && ApocConfiguration.isEnabled("import.file.use_neo4j_config")) {
+            // check 'dbms.security.allow_csv_import_from_file_urls' and 'dbms.directories.import' parameters
+            boolean allowCsvImportFromFile = Util.toBoolean(ApocConfiguration.list().get("dbms.security.allow_csv_import_from_file_urls"));
```
jexp: Can we use `ApocConfiguration.isEnabled` ? `Util.toBoolean(ApocConfiguration.list().get("dbms.security.allow_csv_import_from_file_urls"));`

Perhaps it's best to transform the two settings into apoc-config 

We could turn the white-list into a map that has as key the neo key and as value the new apoc-config-name and then copy the actual config value with the new name into the config map. 

then we could use: `ApocConfiguration.isEnabled("import.file.allow_read_from_filesystem")`

jexp: Same for `dbms.directories.import` -> `import.file.directory`

src/main/java/apoc/export/util/FileUtils.java
```diff
@@ -23,6 +25,27 @@ public static CountingReader readerFor(String fileName) throws IOException {
...
+            String directory = (String) ApocConfiguration.list().get("dbms.directories.import");
+            if(!allowCsvImportFromFile)
+                throw new RuntimeException("Import from files not enabled, please set dbms.security.allow_csv_import_from_file_urls=true in your neo4j.conf");
```
jexp: perhaps also output the file-name?

src/main/java/apoc/export/util/FileUtils.java
```diff
@@ -23,6 +25,27 @@ public static CountingReader readerFor(String fileName) throws IOException {
         return new CountingReader(file);
     }
 
+    public static String changeFileUrlIfImportDirectoryAssigned(String url){
+        if(isFile(url) && ApocConfiguration.isEnabled("import.file.use_neo4j_config")) {
+            // check 'dbms.security.allow_csv_import_from_file_urls' and 'dbms.directories.import' parameters
+            boolean allowCsvImportFromFile = Util.toBoolean(ApocConfiguration.list().get("dbms.security.allow_csv_import_from_file_urls"));
+            String directory = (String) ApocConfiguration.list().get("dbms.directories.import");
+            if(!allowCsvImportFromFile)
```
jexp: Remove CSV here, we should use in in general terms: `allowImportFromFile`

src/main/java/apoc/export/util/FileUtils.java
```diff
@@ -23,6 +25,27 @@ public static CountingReader readerFor(String fileName) throws IOException {
...
+                throw new RuntimeException("Import from files not enabled, please set dbms.security.allow_csv_import_from_file_urls=true in your neo4j.conf");
+            else if(directory==null || directory.isEmpty()){
+                return url;
+            }else{
+                Pattern pattern = Pattern.compile("^file:/[/]?");
```
jexp: Can't we use new URI(url).getPath() ? and check uri.protocol for file ?
No need for a regexp.
It should also work for regular files, like `/Users/mh/data/test.json` not just urls!

Also check the directory, if it exists, is a directory and is readable.
```
File dir = new File(directory)
dir.exists() && dir.isDirectory() && dir.isReadable()
File newFile = new File(directory, filePath)
return newFile.getAbsolutePath()
// possibly new URI("file", ...,  newFile.getAbsolutePath())
```

docs/overview.adoc
inserpio: reviewing...

src/main/java/apoc/export/util/FileUtils.java
inserpio: > Import file FILE_NAME not enabled, please set dbms.security.allow_csv_import_from_file_urls=true in your neo4j.conf

src/main/java/apoc/export/util/FileUtils.java
```diff
@@ -23,6 +23,21 @@ public static CountingReader readerFor(String fileName) throws IOException {
...
 
+    public static String changeFileUrlIfImportDirectoryConstrained(String url) throws IOException {
+        if (isFile(url) && ApocConfiguration.isEnabled("import.file.use_neo4j_config")) {
+            if (!ApocConfiguration.isEnabled("import.file.allow_read_from_filesystem"))
+                throw new RuntimeException("Import file "+url+" not enabled, please set dbms.security.allow_csv_import_from_file_urls=true in your neo4j.conf");
+            String importDir = ApocConfiguration.get("import.file.directory", null);
+            if (importDir != null && !importDir.isEmpty()) {
+                File dir = new File(importDir);
+                if (dir.exists() && dir.isDirectory() && dir.canRead()) {
+                    return "file:" + new File(importDir, url.substring(url.indexOf("/")+1)).getAbsolutePath();
```
jexp: this is not correct yet.
- if we need to extract the path of the current URL we should use new URI(url).getPath() (after checking that it's an URL) -> try { new URI(url).getPath(); } catch(URLException ue) { return url; }
- if we *need* to construct an URI we should use the new URI("file",...,path) *not* `"file:"+...`
- in this case we don't need to construct an URL
- we should also check if the current url, already starts with the importDir, if so, we should just return it

src/main/java/apoc/export/util/FileUtils.java
```diff
@@ -23,6 +23,21 @@ public static CountingReader readerFor(String fileName) throws IOException {
...
+                }
+            }
+        }
+        return url;
+    }
+
     public static boolean isFile(String fileName) {
         if (fileName==null) return false;
         if (fileName.toLowerCase().startsWith("http")) return false;
```
jexp: we should fix this to work with URLs in general
i.e. `return !isUrl(fileName) || new URI(fileName).getProtocol().equalsIgnoreCase("file");`

src/main/java/apoc/export/util/FileUtils.java
```diff
@@ -23,6 +26,33 @@ public static CountingReader readerFor(String fileName) throws IOException {
...
+                throw new RuntimeException("Import file "+url+" not enabled, please set dbms.security.allow_csv_import_from_file_urls=true in your neo4j.conf");
+            String importDir = ApocConfiguration.get("import.file.directory", null);
+            if (importDir != null && !importDir.isEmpty()) {
+                try {
+                    File dir = new File(importDir);
+                    if (!dir.exists() || !dir.isDirectory() || !dir.canRead()) throw new Exception();
+                    String fileAsUrl = null;
+                    try { fileAsUrl = new URI(url).getPath(); } catch (Exception e) { /*no protocol*/ }
+                    if ("".equals(fileAsUrl)){
+                        String newUrl = url.replace("file://", "file:///");
```
jexp: we should also handle "FILE://" I think at some point.

PR #354, opened 2017-03-30
use parameters so that queries are easier to read fixes #345

### LINKED ISSUE #345: apoc.es - Url encoding on query parameters should be done only on values

albertodelazzari:
use parameters so that queries are easier to read fixes #345

### DIFF
--- a/src/test/java/apoc/es/ElasticSearchTest.java
+++ b/src/test/java/apoc/es/ElasticSearchTest.java
@@ -33,6 +33,16 @@ public class ElasticSearchTest {
     private final static String HOST = "localhost";
 
     protected static GraphDatabaseService db;
+
+    private static Map<String, Object> defaultParams = new HashMap<>();
+
+    static {
+        defaultParams.put("index", ES_INDEX);
+        defaultParams.put("type", ES_TYPE);
+        defaultParams.put("id", ES_ID);
+        defaultParams.put("host", HOST);
+    }
+
     // We need a reference to the class implementing the procedures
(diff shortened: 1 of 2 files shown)


src/test/java/apoc/es/ElasticSearchTest.java
```diff
@@ -33,6 +33,16 @@
     private final static String HOST = "localhost";
 
     protected static GraphDatabaseService db;
+
+    private static Map<String, Object> defaultParams = new HashMap<>();
```
jexp: Util.map()

src/test/java/apoc/es/ElasticSearchTest.java
```diff
@@ -53,25 +63,37 @@ public static void tearDown() {
...
     }
 
+    /**
+     * Default params (host, index, type, id) + payload
+     *
+     * @param payload
+     * @return
+     */
+    private static Map<String, Object> createDefaultProcedureParametersWithPayloadAndId(String payload, String id) {
+        Map<String, Object> params = new HashMap<>(defaultParams);
```
jexp: `Util.merge(defaultParams, Util.map("payload",payload,"id",id))`

albertodelazzari:
Fixes done according to your feedback.
Thanks

jexp:
Thanks a lot

PR #360, opened 2017-04-02
add parameter options with with empty default to index.addAllNodes

sarmbruster:


jexp:
Can't we merge the two procs together? And mark the 2nd as deprecated and just call the other. Remove in 3.2 or 3.3 then

sarmbruster:
the 2nd `apoc.index.addAllNodesExtended` is already marked as deprecated. This PR just added `options` parameter to `apoc.index.addAllNodes` defaulting to empty map. This change makes `addAllNodesExtended` obsolete and a candidate for removal later on.

PR #366, opened 2017-04-04
Util method for transaction termination checking (algo and warmup) fixes #211

### LINKED ISSUE #211: proper termination checks in warmup and graph algorithms

albertodelazzari:
In Utils class is now available a method to check if a transaction is terminated or not.
The following classes were fixed according to this new method:

- Warmup
- Pregel 
- LabelPropagation
- AlgoUtils
- BatchRunnable

All the methods that expect a return value are provided with a default one
We should check all the other packages in order to normalize the transaction checking with only one method

### DIFF
--- a/src/main/java/apoc/algo/LabelPropagation.java
+++ b/src/main/java/apoc/algo/LabelPropagation.java
@@ -1,6 +1,7 @@
 package apoc.algo;
 
 import apoc.Pools;
+import apoc.util.Util;
 import org.neo4j.graphdb.*;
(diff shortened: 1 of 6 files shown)


jexp:
Looks good, there are more places that we have to check:

- Periodic
- Graph Refactorings
- The new renaming methods 
- cypher.runFile
- exports
- load.json, load.jdbc, load.xml

In a sensible place in the **outmost** loop that's iterate a reasonable number of times.

PR #368, opened 2017-04-04
ObjectId can be packed, it will be converted to String fixes #355

### LINKED ISSUE #355: Connect mongo to neo4j using apoc
I'm trying to connect neo4j to mongodb to retrieve some information. When I use this query:
```CALL apoc.mongodb.first('mongodb://localhost:27017','test','airport',{code:'JFK'}) yield value``` 
then I got this error :
```Unpackable value 51dc2f7bb6e4ab4eb769df93 of type org.bson.types.ObjectId``` 

I'm using mongodb v3.2.4 so I added jars v3.2.2

image

Thank you

### PR DESCRIPTION
**Pull request by albertodelazzari**

### CODE REVIEW

File: src/main/java/apoc/mongodb/MongoDBColl.java [hunk 1]
```diff
@@ -39,9 +39,30 @@ public void close() throws IOException {
...
+     */
+    private Map<String, Object> documentToPackableMap(Map<String, Object> document) {
+        /**
+         * A document in MongoDB has a special field "_id" of type ObjectId
+         * This object is not "packable" by Neo4jPacker so it must be converted to a value that Neo4j can deal with
+         *
+         * If the document is not null we simply override the ObjectId instance with its string representation
+         */
+        if (document != null) {
+            document.put("_id", document.get("_id").toString());
```
jexp: what if there is no _id ? then this will yield a NPE

jexp: did you check if there's something in the mongo driver that would provide this?

jexp:
Please add a test

src/main/java/apoc/mongodb/MongoDBColl.java
albertodelazzari: I don't think "_id" can be null in MongoDB but I can do a better NP check.
To be honest I don't remember if there is something in the mongo driver but I can check before updating the PR or if you want I can do this in a separate PR

PR #374, opened 2017-04-06
export now creates 3 different files fixes #270

### LINKED ISSUE #270: have export.cypher split over several files
either one per tx
or at least separate data and schema operations

albertodelazzari:
Given a full path file name like /tmp/myexport.cypher this method will create:

1. /tmp/myexport.schema.cypher
2. /tmp/myexport.nodes.cypher
3. /tmp/myexport.relationships.cypher

We need to consider to normalize the export like the import. We can pass an URI that will be normalized to an export directory.

### DIFF
--- a/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
+++ b/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
@@ -1,6 +1,7 @@
 package apoc.export.cypher;
 
 
+import apoc.export.util.FileUtils;
 import apoc.export.util.FormatUtils;
 import apoc.export.util.Reporter;
 import org.neo4j.cypher.export.SubGraph;
@@ -11,6 +12,7 @@ import org.neo4j.graphdb.Relationship;
 import org.neo4j.graphdb.schema.IndexDefinition;
 import org.neo4j.helpers.collection.Iterables;
 
+import java.io.IOException;
 import java.io.PrintWriter;
 import java.lang.reflect.Array;
 import java.util.*;
@@ -19,17 +21,17 @@ import java.util.*;
  * Idea is to lookup nodes for relationships via a unqiue index
  * either one inherent to the original node, or a artificial one that indexes the original node-id
  * and which is removed after the import.
- *
+ * <p>
  * Outputs indexes and constraints at the beginning as their own transactions
  */
 public class MultiStatementCypherSubGraphExporter {
+    private final static String UNIQUE_ID_LABEL = "UNIQUE IMPORT LABEL";
+    private final static String Q_UNIQUE_ID_LABEL = quote(UNIQUE_ID_LABEL);
+    private final static String UNIQUE_ID_PROP = "UNIQUE IMPORT ID";
     private final SubGraph graph;
     private final Map<String, String> uniqueConstraints;
     Set<String> indexNames = new LinkedHashSet<>();
     Set<String> indexedProperties = new LinkedHashSet<>();
-    private final static String UNIQUE_ID_LABEL = "UNIQUE IMPORT LABEL";
-    private final static String Q_UNIQUE_ID_LABEL = quote(UNIQUE_ID_LABEL);
-    private final static String UNIQUE_ID_PROP = "UNIQUE IMPORT ID";
     private long artificialUniques = 0;
 
     public MultiStatementCypherSubGraphExporter(SubGraph graph) {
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -19,42 +21,90 @@
...
+        return ":`" + id + "`";
+    }
+
+    /**
+     * @param fileName
+     * @param suffix
+     * @return
+     */
+    private String normalizeFileName(final String fileName, String suffix) {
+        // TODO check if this should be follow the same rules of FileUtils.readerFor
```
albertodelazzari: This should be smarter than now.
Accept an URI and follow the same logic of FileUtils.readerFor

src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -19,42 +21,90 @@
...
+     * @param batchSize
+     * @param reporter
+     */
+    public void export(String fileName, int batchSize, Reporter reporter) throws IOException {
         boolean hasNodes = hasData(graph.getNodes());
         if (hasNodes) {
-            begin(out);
-            appendNodes(out, batchSize, reporter);
-            commit(out);
+            PrintWriter nodeWriter = createWriter(fileName, "nodes");
```
albertodelazzari: Default suffix "nodes" for statements that are related to graph nodes

src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -19,42 +21,90 @@
...
+            begin(nodeWriter);
+            appendNodes(nodeWriter, batchSize, reporter);
+            commit(nodeWriter);
+
+            nodeWriter.flush();
         }
-        writeMetaInformation(out);
+
+        PrintWriter schemaWriter = createWriter(fileName, "schema");
```
albertodelazzari: Default suffix "schema" for schema related statement

src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -19,42 +21,90 @@
...
+        writeMetaInformation(schemaWriter);
 
         if (hasData(graph.getRelationships())) {
-            begin(out);
-            appendRelationships(out, batchSize, reporter);
-            commit(out);
+            PrintWriter relationships = createWriter(fileName, "relationships");
```
albertodelazzari: Default suffix "relationships" for schema related statement

jexp:
This was meant to be optional, i.e. configured in the config map { separateFiles: true }

And in that case we should handle:

- when passed a directory, create files like neo4j-export-{uid}.{nodes|relationships|schema}.cypher
- when passed a file-name, add {nodes|relationships|schema}.cypher
- we could also support a config setting like this {fileNames: { schema: "schema.cypher", nodes: "my-nodes.cypher"}}

- the important part is that the cleanup happens at the right time
- i.e. after the import and it is again 1. node / relationships cleanup 2. schema cleanup

src/main/java/apoc/mongodb/MongoDBColl.java
```diff
@@ -39,9 +37,30 @@ public void close() throws IOException {
         mongoClient.close();
     }
 
+    /**
+     * It translates a MongoDB document into a Map where the "_id" field is not an ObjectId
```
jexp: this doesn't belong into this PR it's the mongodb _id conversion

src/test/java/apoc/export/cypher/ExportCypherTest.java
```diff
@@ -16,34 +15,18 @@
...
-            "commit%n" +
-            "begin%n" +
-            "DROP CONSTRAINT ON (node:`UNIQUE IMPORT LABEL`) ASSERT node.`UNIQUE IMPORT ID` IS UNIQUE;%n" +
-            "commit");
+    private static final String EXPECTED_NODES = String.format("begin%nCREATE (:`Foo`:`UNIQUE IMPORT LABEL` {`name`:\"foo\", `UNIQUE IMPORT ID`:0});%nCREATE (:`Bar` {`name`:\"bar\", `age`:42});%nCREATE (:`Bar`:`UNIQUE IMPORT LABEL` {`age`:12, `UNIQUE IMPORT ID`:2});%ncommit");
+
+    private static final String EXPECTED_RELATIONSHIPS = String.format("begin%nMATCH (n1:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`:0}), (n2:`Bar`{`name`:\"bar\"}) CREATE (n1)-[:`KNOWS`]->(n2);%ncommit");
+
+    private static final String EXPECTED_SCHEMA = String.format("begin%nCREATE INDEX ON :`Foo`(`name`);%nCREATE CONSTRAINT ON (node:`Bar`) ASSERT node.`name` IS UNIQUE;%nCREATE CONSTRAINT ON (node:`UNIQUE IMPORT LABEL`) ASSERT node.`UNIQUE IMPORT ID` IS UNIQUE;%ncommit%nschema await%nbegin%nMATCH (n:`UNIQUE IMPORT LABEL`)  WITH n LIMIT 20000 REMOVE n:`UNIQUE IMPORT LABEL` REMOVE n.`UNIQUE IMPORT ID`;%ncommit%nbegin%nDROP CONSTRAINT ON (node:`UNIQUE IMPORT LABEL`) ASSERT node.`UNIQUE IMPORT ID` IS UNIQUE;%ncommit");
```
jexp: this is wrong, the schema commands have to be in the right order in the file(s)

jexp:
we also have to have a separate issue that splits each tx in a separate file so it can be run from sth that can't do transactions

jexp:
@albertodelazzari this PR contains commits from other work, could you please clean it up?

jexp:
@albertodelazzari ping :)

src/test/java/apoc/export/cypher/ExportCypherTest.java
albertodelazzari: Do you mean the "drop" statement should be done at the end of all the other statements?

albertodelazzari:
If you agree I can implement the optional part ("separateFiles" parameter) and update the PR.
Then all the other features could be moved to a new issue.

jexp:
Yes, please implement the separateFiles parameter here.

And please update this PR that only has the commits for this, not an commit + revert for objectId.
You can also do a new local branch and force push to this PR-branch or just create a new PR branch and we close this one

albertodelazzari:
@jexp I added a "cleanup" file when you define separateFiles=true so that there are 4 different files:
- nodes
- relationships
- schema
- cleanup

The default behaviour is exporting all the statements in a single file.
Let me know if it's ok so I can open a new issue with all the other open points

Thanks

jexp:
Ok, looks almost good. The idea was that each of those files can be used by cypher.run.file independently, to resolve the issue that it cannot be a procedure that does schema operations and data operations at the same time.

So we might need to split it up further (i.e. nodes-before, nodes-create, nodes-after, rels-before,rels-create, rels-after). I'll merge it for now and we plan that as a separate issue.

Issue #364, opened 2017-04-03
Bug: path expander 'minLevel' interaction with termination filter


InverseFalcon:
There is an interaction between 'minLevel' and the termination filter that may violate expectations. The more general question I'm asking is "which label filters should be executing when evaluating a node below minLevel depth?

While it makes sense to execute on the whitelist and blacklist, the termination filter's side effect of pruning further traversal down a path still triggers when the node is below minLevel depth.

An example in the movies graph, when 'Clint Eastwood' and 'Gene Hackman' are labeled :Western.

```
MATCH (k:Person {name:'Keanu Reeves'})
CALL apoc.path.expandConfig(k, {relationshipFilter:'ACTED_IN|PRODUCED|DIRECTED', labelFilter:'/Western', uniqueness: 'NODE_GLOBAL', minLevel:3}) yield path 
RETURN path
```

The above query will not return any results, because the termination filter prunes further traversal when 'Gene Hackman' is reached (so Clint Eastwood can never be reached), even though 'Gene Hackman' is never returned as a result, due to `minLevel:3`.

Is this something we should consider correct behavior, or do we want to prevent termination filter pruning until we reach minLevel and are returning results?

Discussion encouraged.

InverseFalcon:
I've come to the conclusion that this is incorrect behavior. The use of the termination filter is to return interested paths up to nodes and pruning further traversal, so it seems to me that pruning traversal before we even reach minLevel is a bug.

InverseFalcon:
Fixed.

PR #379, opened 2017-04-07
Prevent path expander termination filter from filtering below minLevel

InverseFalcon:
This fixes bug #364.


InverseFalcon:
To clarify the first commit message, the terminology changes were only in the ExpandPathTest method names, no terminology changes in the PathExplorer code or the documentation.

PR #389, opened 2017-04-14
Fixes #303 - constraint issue (create before delete) with mergeNodes

### LINKED ISSUE #303: constraint issue (create before delete) with mergeNodes
Failed to invoke procedure `apoc.refactor.mergeNodes`: Caused by: org.neo4j.graphdb.ConstraintViolationException: Node 1250 already exists with label Organization and property "name"=[INVESCO, LTD.]

we need to delete the node first (or at least the key property) before creating a new one

DanielBerton:


src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -299,16 +299,30 @@ public void categorize(
     }
 
     private Node mergeNodes(Node source, Node target, boolean delete) {
-        copyRelationships(source, copyProperties(source, copyLabels(source, target)), delete);
-        if (delete) source.delete();
+        Map<String, Object> propTemp = new HashMap<>();
+        List<String> keys = new ArrayList<>();
```
jexp: why do we need keys?

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -299,16 +299,30 @@ public void categorize(
     }
 
     private Node mergeNodes(Node source, Node target, boolean delete) {
-        copyRelationships(source, copyProperties(source, copyLabels(source, target)), delete);
-        if (delete) source.delete();
+        Map<String, Object> propTemp = new HashMap<>();
```
jexp: can we find a better name? why not just Map<String, Object> properties = source.getAllProperties()

Didn't we say that we just store/keep the existing properties and then set the new ones after deletion of the source node? Not sure why we have to remove the properties individually?

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -299,16 +299,30 @@ public void categorize(
...
+            propTemp.put(prop.getKey(), prop.getValue());
+            keys.add(prop.getKey());
+            source.removeProperty(prop.getKey());
+        }
+        copyRelationships(source, copyLabels(source, target), delete);
+        setProperty(propTemp, keys, target);
         return target;
     }
 
+    private void setProperty(Map<String, Object> propTemp, List<String> keys, Node target) {
```
jexp: isn't that the same as copy properties? only from a map source? Best name it the same and have it return the target node too.

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -299,16 +299,30 @@ public void categorize(
...
+            target.setProperty(keys.get(i), propTemp.get(keys.get(i)));
+        }
+    }
+
     private Node copyRelationships(Node source, Node target, boolean delete) {
         for (Relationship rel : source.getRelationships()) {
             copyRelationship(rel, source, target);
             if (delete) rel.delete();
         }
+        if (delete) source.delete();
```
jexp: will this break any other code that uses this? I think the original contract for this method was that rels are deleted but the nodes are left alone.
so I'd rather move this out of the method back into mergeNodes where it originally was.

src/main/java/apoc/refactor/GraphRefactoring.java
DanielBerton: @jexp so we remane it `copyProperties` with this signature `private Node copyProperties(Map<String, Object> properties, Node target)` right?

src/main/java/apoc/refactor/GraphRefactoring.java
DanielBerton: @jexp ok so we move back the delete into `mergeNodes ` before the call of ` copyProperties`  method

src/main/java/apoc/refactor/GraphRefactoring.java
jexp: yes

src/main/java/apoc/refactor/GraphRefactoring.java
jexp: yes

DanielBerton:
@jexp change requests fixed

jexp:
Looks great, thanks a lot.

Issue #392, opened 2017-04-17
Change request: allow negative time params in apoc.date.format()


InverseFalcon:
Currently, passing negative time values throws an IllegalArgumentException. Given that java.util.Date objects can handle negative time values (representing a time instant before Jan 1, 1970), I'd like to remove this restriction for apoc.date.format().

PR #393, opened 2017-04-17
Allowed negative time parameter to apoc.date.format()

InverseFalcon:
Fulfills change request #392.

If there's a good reason for the restriction that I've overlooked, please reject the PR.


jexp:
Thanks a lot @InverseFalcon

Issue #396, opened 2017-04-19
Feature Request: Implement conditional subquery execution


InverseFalcon:
Should be fairly easy to implement conditional execution of cypher based on apoc.cypher.run(). We'll need a readOnly and write version.

InverseFalcon:
Fulfilled.

PR #398, opened 2017-04-19
Added conditional cypher execution

InverseFalcon:
New procedures apoc.when(), apoc.when.do(), apoc.whenMany(), and apoc.whenMany.do()

This fulfills feature request #396.

Open for suggestions and improvement, please don't merge yet.

Documentation will be updated once final suggestions and improvements are applied.

### DIFF
--- a/src/main/java/apoc/cypher/Cypher.java
+++ b/src/main/java/apoc/cypher/Cypher.java
@@ -365,4 +365,56 @@ public class Cypher {
         Result result = db.execute(cypher, params == null ? Collections.EMPTY_MAP : params);
         return result.stream().map(stringObjectMap -> new MapResult(stringObjectMap));
     }
+
+    @Procedure("apoc.when")
+    @Description("apoc.when(condition, ifQuery, elseQuery:'', params:{}) yield value - based on the conditional, executes read-only ifQuery or elseQuery with the given parameters")
+    public Stream<MapResult> when(@Name("condition") boolean condition, @Name("ifQuery") String ifQuery, @Name(value="elseQuery", defaultValue = "") String elseQuery, @Name(value="params", defaultValue = "") Map<String, Object> params) {
+        if (params == null) params = Collections.emptyMap();
+        String targetQuery = condition ? ifQuery : elseQuery;
+
+        if (targetQuery.isEmpty()) {
+            return Stream.of(new MapResult(Collections.emptyMap()));
+        } else {
+            return db.execute(withParamMapping(targetQuery, params.keySet()), params).stream().map(MapResult::new);
+        }
+    }
+
+    @Procedure(value="apoc.when.do", mode = Mode.WRITE)
+    @Description("apoc.when.do(condition, ifQuery, elseQuery:'', params:{}) yield value - based on the conditional, executes writing ifQuery or elseQuery with the given parameters")
+    public Stream<MapResult> whenDo(@Name("condition") boolean condition, @Name("ifQuery") String ifQuery, @Name(value="elseQuery", defaultValue = "") String elseQuery, @Name(value="params", defaultValue = "") Map<String, Object> params) {
+        return when(condition, ifQuery, elseQuery, params);
+    }
+
+    @Procedure("apoc.whenMany")
+    @Description("apoc.whenMany([[condition, query], [condition, query], ...], elseQuery:'', params:{}) yield value - given conditional / read-only query pairs, executes the query associated with the first conditional evaluating to true (or the else query if none are true) with the given parameters")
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/cypher/Cypher.java
```diff
@@ -365,4 +365,56 @@ private T poll() {
...
+        if (targetQuery.isEmpty()) {
+            return Stream.of(new MapResult(Collections.emptyMap()));
+        } else {
+            return db.execute(withParamMapping(targetQuery, params.keySet()), params).stream().map(MapResult::new);
+        }
+    }
+
+    @Procedure(value="apoc.when.do", mode = Mode.WRITE)
+    @Description("apoc.when.do(condition, ifQuery, elseQuery:'', params:{}) yield value - based on the conditional, executes writing ifQuery or elseQuery with the given parameters")
+    public Stream<MapResult> whenDo(@Name("condition") boolean condition, @Name("ifQuery") String ifQuery, @Name(value="elseQuery", defaultValue = "") String elseQuery, @Name(value="params", defaultValue = "") Map<String, Object> params) {
```
jexp: should do have a void result? so you don't need any yields

src/main/java/apoc/cypher/Cypher.java
```diff
@@ -365,4 +365,56 @@ private T poll() {
...
+        return when(condition, ifQuery, elseQuery, params);
+    }
+
+    @Procedure("apoc.whenMany")
+    @Description("apoc.whenMany([[condition, query], [condition, query], ...], elseQuery:'', params:{}) yield value - given conditional / read-only query pairs, executes the query associated with the first conditional evaluating to true (or the else query if none are true) with the given parameters")
+    public Stream<MapResult> whenMany(@Name("conditionals") List<List<Object>> conditionals, @Name(value="elseQuery", defaultValue = "") String elseQuery, @Name(value="params", defaultValue = "") Map<String, Object> params) {
```
jexp: `whenMulti` or `whens` ?

jexp: Hmm what is a use-case for whenMany ? I just thought that it was a single predicate and multiple if and else queries.

jexp: Something like case ? I wonder if we can find a better way to provide these parameters.

jexp: Shouldn't we just call it apoc.case ? Which takes either this list or functions that wrap condition + query ?
like a function `apoc.pred(condition, query)` 
the other thing we can do is to use defaults, 

i.e. `apoc.case(Boolean pred1, String query1, Boolean pred2=null, String query2=null,....) `
so it looks like varargs from the outside

another function we should add is `apoc.args` which turns a varargs into a list.

`List<Object> apoc.args(Object=null value1, Object=null value2, Object=null value3, Object=null value3, ..., Object=null value10, ...)`

jexp: So we could use `apoc.args(true, "query", false, "query2", ...)` as parameter to apoc.case with your original signature.

InverseFalcon: `case` sounds good. `apoc.args()` sounds like it will be useful, probably not a bad idea to add that in and make this the first usage

src/main/java/apoc/cypher/Cypher.java
InverseFalcon: I'm undecided about that. While it is a write query, there may still be a need to return something, whether it's graph elements that were just created or something else.

src/main/java/apoc/cypher/Cypher.java
InverseFalcon: Though if `apoc.args()` is just going to produce a list, we could just use a list argument instead. Unless I'm missing some other feature or advantage?

src/main/java/apoc/cypher/Cypher.java
jexp: hmm we *could* do that as a separate `when.do.return()` or `when.do.values()`
or we make apoc.when a writeable procedure so you *could* return from it.
But then it's not useable for read-only users anymore.

src/main/java/apoc/cypher/Cypher.java
jexp: Hmm good point, perhaps we then go with the varags in case directly?

I think we can use this approach though in other places too where people expect some kind of varags.

InverseFalcon: I'd prefer a list myself. That lets us keep the current parameter order. Usage of optional parameters must happen at the end of the query, not the start, so the params map would have to go in front, as would the `else` query (unless we tacked on `else` to the end of the optional params, but then that becomes a bit awkward, as that would violate the expected odd-parameters-are-booleans pattern).

InverseFalcon:
Made some changes to the procedure names (apoc.do.when() and apoc.do.case() sounded better than with the reversed order), and changed the expected parameter to a List<Object>.

jexp:
Thanks a lot. This has been long awaited

PR #411, opened 2017-04-25
Adding support for java.sql.Date in JDBC to convert to String

bradnussbaum:


src/test/java/apoc/load/JdbcTest.java
```diff
@@ -37,38 +43,55 @@ public static void tearDown() {
...
+                    put("NAME", "John"); put("HIRE_DATE",hireDate.getTime()); put("EFFECTIVE_FROM_DATE",
+                            effectiveFromDate.getTime());
+                }}, row.get("row")));
     }
+
     @Test
     public void testLoadJdbcSelect() throws Exception {
         testCall(db, "CALL apoc.load.jdbc('jdbc:derby:derbyDB','SELECT * FROM PERSON')",
-                (row) -> assertEquals(singletonMap("NAME", "John"), row.get("row")));
+                (row) -> assertEquals(new HashMap<String,Object>(){{
```
jexp: please use Util.map() in all these places.

src/main/java/apoc/load/Jdbc.java
```diff
@@ -140,6 +141,12 @@ private Object convert(Object value) {
             if (value instanceof UUID || value instanceof BigInteger || value instanceof BigDecimal) {
                 return value.toString();
             }
+            if (value instanceof java.sql.Date) {
```
jexp: don't need this as they all inherit from java.util.Date

PR #412, opened 2017-04-26
Fixes #334 - make sure that manual index operations are read-only

### LINKED ISSUE #334: make sure that manual index operations are read-only
create methods in Util.java that use kernel API to access the legacy indexes in read-only transactions for only query procedures.

so there is no need to call  `db.index().forNodes() or forRelationships()`

especially important for running them in a cluster

DanielBerton:
`apoc.index.forRelationships` and `apoc.index.forNodes` are gets or create so we cannot transform it into read only operations. We can split it into two different procedure, get and create or leave it like this.

Using the method `relationshipLegacyIndexQuery` of `ReadOperation` :
`relationshipLegacyIndexQuery(indexName, query, startingNode, endingNode)`
if we pass `null` on the `endNode` parameter (that we transform in **-1** to ignore the parameter), we have an unexpected behavior. For example if we test `apoc.index.out` (see `testOut)` we have the same result as the incoming (`apoc.index.in`).
We have this problem only if we pass **-1** as parameter, for every other number it works.  If we pass **-1**  on startingNode it works. It seems a bug

### DIFF
--- a/src/main/java/apoc/index/FreeTextSearch.java
+++ b/src/main/java/apoc/index/FreeTextSearch.java
@@ -1,10 +1,11 @@
 package apoc.index;
 
 import apoc.ApocKernelExtensionFactory;
+import apoc.util.Util;
+import org.neo4j.kernel.KernelApi;
 import org.neo4j.procedure.*;
 import apoc.result.WeightedNodeResult;
 import org.apache.lucene.analysis.Analyzer;
-import org.apache.lucene.queryparser.classic.ParseException;
 import org.apache.lucene.search.Sort;
 import org.neo4j.graphdb.Label;
 import org.neo4j.graphdb.Node;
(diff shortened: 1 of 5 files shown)


jexp:
@DanielBerton did you resolve that issue with the wrong relationship vs. node result of the index?

src/main/java/apoc/util/Util.java
```diff
@@ -525,4 +531,20 @@ public static boolean transactionIsTerminated(GraphDatabaseAPI db) {
 
         return false;
     }
+
+    public static List<WeightedNodeResult> toWeightedNodeResultFromLegacyIndex(LegacyIndexHits hits, GraphDatabaseService db){
```
jexp: I don't think these should go into util but stay in the index package, perhaps in an IndexUtils there.
Or even having KernelAPI return a stream of `WeightedNodeResult` or `WeightedRelationshipResult`

DanielBerton:
@jexp PR updated

jexp:
Thanks a lot.

Issue #388, opened 2017-04-13
use export config quotes in CSVWriter


jexp:
if quotes are configured use the configured quote or the default quote-character 

https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.1/src/main/java/apoc/export/csv/CsvFormat.java#L37

PR #417, opened 2017-04-27
issue #388 - use export config quotes in CSVWriter

AngeloBusato:


jexp:
Would be good to add a test @AngeloBusato 

Thanks

Issue #399, opened 2017-04-19
Missing detailed documentation for Graph Refactoring procedures


DanielBerton:
Part of #331

Issue #331, opened 2017-03-20
missing detailed documentation for some procedures


jexp:
Could I get an example for each of these procedures.
 
I am able to do 2 and 3 but still don’t understand when to use 1 and when to use 2. If the team could provide some examples of the patterns and patterfull would be really nice.

```
CALL apoc.create.vNode(['Label'], {key:value,…​})
returns a virtual node
CALL apoc.create.vNodes(['Label'], [{key:value,…​}])
returns virtual nodes
CALL apoc.create.vRelationship(nodeFrom,'KNOWS',{key:value,…​}, nodeTo)
returns a virtual relationship
CALL apoc.create.vPattern({_labels:['LabelA'],key:value},'KNOWS',{key:value,…​}, {_labels:['LabelB'],key:value})
returns a virtual pattern
CALL apoc.create.vPatternFull(['LabelA'],{key:value},'KNOWS',{key:value,…​},['LabelB'],{key:value})
returns a virtual pattern
```
 
 
 
Can someone explain these commands. I am trying to convert the graphs from neo4j into the dataframes for Vertices and edges so that it can be sent to other graphing programs in R.  The first two seem to be the ones I need but not sure.
 
```
apoc.graph.from(data,'name',{properties}) yield graph
creates a virtual graph object for later processing it tries its best to extract the graph information from the data you pass in
apoc.graph.fromData([nodes],[relationships],'name',{properties})
creates a virtual graph object for later processing
apoc.graph.fromPaths(path,'name',{properties})
creates a virtual graph object for later processing
apoc.graph.fromPaths([paths],'name',{properties})
creates a virtual graph object for later processing
apoc.graph.fromDB('name',{properties})
creates a virtual graph object for later processing
apoc.graph.fromCypher('statement',{params},'name',{properties})
creates a virtual graph object for later processing
```
 
Need some good examples for these refactoring procedures
 
```
call apoc.refactor.cloneNodes([node1,node2,…​])
clone nodes with their labels and properties
call apoc.refactor.cloneNodesWithRelationships([node1,node2,…​])
clone nodes with their labels, properties and relationships
call apoc.refactor.mergeNodes([node1,node2])
merge nodes onto first in list
call apoc.refactor.to(rel, endNode)
redirect relationship to use new end-node
call apoc.refactor.from(rel, startNode)
redirect relationship to use new start-node
call apoc.refactor.setType(rel, 'NEW-TYPE')
change relationship-type
[...]

PR #431, opened 2017-05-04
Fixes #176 - Additional documentation on apoc.triggers

### LINKED ISSUE #176: Additional documentation on apoc.triggers
I've been looking at using the apoc trigger functionality. It looks seriously powerful. 

That said I don't have a strong understanding of what statements  are accessible (e.g {createdLabels}, etc ) and what they contain. 

I'd love to see some documentation around this functionality and a few relatively complex examples (e.g update a label on a node, propagate some information to all connected nodes). 

Thank you!!

DanielBerton:
We created some examples and described the available statements, let me know if it's enough or is better add other documentation or examples


jexp:
Cool, some notes.

1. Overview shouldn't include the file, that should be in index.adoc
2. Overview is only for a general listing, which will be replaced by just including the generated documentation csvs.
3. this seems to be mixed with the refactoring examples ? (which are also in overview.adoc, and be moved out). I don't mind if you leave them in a single PR but please move it to a separate adoc file and include that in the right place in index.adoc

Thanks

DanielBerton:
@jexp ok we move the include in the index.adoc, we can put it in a separate section (like before `Appendix: Complete Overview` that contains all the overview.adoc)  or inside an existing one?
The refactoring examples was merged https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/399 we can fix it in this PR or in a new issue.

DanielBerton:
I thought about modifying the refactor example in the issue https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/331 today

DanielBerton:
@jexp PR updated

jexp:
Cool, thanks a lot !

Issue #433, opened 2017-05-05
apoc.trigger with phase:'after' in update actions doesn't work


AngeloBusato:
Related to issue #395 .
On apoc.trigger if we use `phase:'after'` on an update action, procedure doesn't work; it goes in loop because transaction is already close.

jexp:
The transaction is *almost* over, but locks have not been released. So the solution would be to queue it until the tx is really done.

conker84:
we added the `afterAsync` phase in order to manage these use-cases

PR #434, opened 2017-05-05
Fixes #395 - Apoc.trigger.propertiesByKey adding a dynamic label based on newValue vs oldValue does not work

### LINKED ISSUE #395: Apoc.trigger.propertiesByKey adding a dynamic label based on newValue vs oldValue does not work
I have been playing around with apoc.trigger. I created a (a:Test{number:5}) and below query for apoc.trigger, then tried to set the number different, and it does not update labels as it should. Tried out only inner statement with hardcoded values and it works. Using Neo4j 3.1.3 and APOC 3.1.3.6. Nothing in debug.log

```
CALL apoc.trigger.add('test','UNWIND apoc.trigger.propertiesByKey({assignedNodeProperties},"number") AS prop 
WITH prop.node as n,CASE WHEN toFLOAT(prop.oldValue) < toFLOAT(prop.newValue) THEN "increasing" ELSE "decreasing" END AS label 
CALL apoc.create.addLabels( [n], [label]) YIELD node RETURN "none"', {phase:'after'})
```

DanielBerton:
Added suggested changes, we have opened an issue #433 to handle the update on  `phase: after`, when the transaction is already been committed. Now we add some comments in the issue.

### DIFF
--- a/src/main/java/apoc/trigger/Trigger.java
+++ b/src/main/java/apoc/trigger/Trigger.java
@@ -9,6 +9,7 @@ import org.neo4j.graphdb.event.LabelEntry;
 import org.neo4j.graphdb.event.PropertyEntry;
 import org.neo4j.graphdb.event.TransactionData;
 import org.neo4j.graphdb.event.TransactionEventHandler;
+import org.neo4j.helpers.collection.Iterators;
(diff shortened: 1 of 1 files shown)


jexp:
Can you please also add a test ?

Issue #380, opened 2017-04-09
Add config option to cypher.export to use MERGE instead of CREATE


jexp:
Add a config option to export to use MERGE (both for nodes and rels) + `ON CREATE SET x += {key:"value",...}` (except the id-property)

related to: #247

jexp:
`export.cypher(filename, data, { merge: true })`

PR #441, opened 2017-05-14
Improve export output

cleishm:
Generate neo4j-client and cypher-shell compatible output by default.
Use `db.awaitIndex(...)` rather than `schema await`.

### DIFF
--- a/src/main/java/apoc/export/util/ExportFormat.java
+++ b/src/main/java/apoc/export/util/ExportFormat.java
@@ -1,7 +1,5 @@
 package apoc.export.util;
 
-import java.io.PrintWriter;
-
 /**
  * @author AgileLARUS
  *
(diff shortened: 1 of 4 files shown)


src/test/java/apoc/export/cypher/ExportCypherTest.java
```diff
@@ -27,41 +23,130 @@
...
+            return nodes() + schema() + relationships() + cleanup();
+        }
+    }
+
+    private static class CypherOutput implements ExpectedOutputs
+    {
+        @Override
+        public String nodes() {
+            return String.format(":BEGIN%n" +
+                    "CREATE (:`Foo`:`UNIQUE IMPORT LABEL` {`name`:\"foo\", `UNIQUE IMPORT ID`:0});%n" +
```
jexp: I actually wanted to minimize the duplication of the generated cypher strings, that's why they've been defined only once.

src/main/java/apoc/export/util/ExportFormat.java
```diff
@@ -36,7 +37,7 @@ public static final ExportFormat fromString(String format) {
                 }
             }
         }
-        return NEO4J_SHELL;
+        return CYPHER;
```
jexp: If we're changing the default in a .dot release, we have to clearly communicate it, or leave it for 3.2

cleishm: Oh, I thought this format stuff was unreleased. Given it is, then you're right - leave that change for your 3.2 release.

src/test/java/apoc/export/cypher/ExportCypherTest.java
cleishm: Yeah - but I don't like programmatically generating assertion text in test code. It's hard to read what the assertion should be, and you very quickly end up with test code that has the same behavior (and bugs) as the code you're intending to test.

inserpio:
@jexp push with #380. Please note that `db.awaitIndex` is available from 3.1, so we need to remove it when merging into 3.0

PR #446, opened 2017-05-17
Replaced putIfAbsent() with computeIfAbsent() for some map defaults

InverseFalcon:
This should prevent some unnecessary object creation


jexp:
I'm not sure if it makes a big difference as Collections.emptyList() is also just a variable access and it's cheaper that creating a lambda?

PR #452, opened 2017-05-22
fixes #430 - Better error messages when missing drivers

### LINKED ISSUE #430: Better error messages when missing drivers
Probably catch ClassNotFoundExceptions and return/throw a nicer error message, that states
which jars should be put into the plugins directory.

- [ ] JDBC
- [ ] MongoDB
- [ ] Couchbase

Also add tot he docs where to download the right drivers, similar to:
https://neo4j-contrib.github.io/neo4j-apoc-procedures/#_load_jdbc_resources

AngeloBusato:
@jexp for jdbc we have already an error who specifies that driver is missing.
If you want, we can create a map that specifies the drive to be inserted into the plugin folder (Oracle, Derby...). 

missingdriverjdbcerror

### DIFF
--- a/src/main/java/apoc/couchbase/Couchbase.java
+++ b/src/main/java/apoc/couchbase/Couchbase.java
@@ -3,6 +3,7 @@ package apoc.couchbase;
 import java.util.List;
 import java.util.stream.Stream;
 
+import apoc.util.MissingDependencyException;
 import org.neo4j.procedure.Name;
 import org.neo4j.procedure.Procedure;
 
@@ -96,9 +97,10 @@ public class Couchbase {
   @Procedure
   @Description("apoc.couchbase.get(nodes, bucket, documentId) yield id, expiry, cas, mutationToken, content - retrieves a couchbase json document by its unique ID.")
   public Stream<CouchbaseJsonDocument> get(@Name("nodes") List<String> nodes, @Name("bucket") String bucket,
-      @Name("documentId") String documentId) {
+      @Name("documentId") String documentId){
     Stream<CouchbaseJsonDocument> result = null;
-    try (CouchbaseConnection couchbaseConnection = CouchbaseManager.getConnection(nodes, bucket)) {
+
+    try (CouchbaseConnection couchbaseConnection = getCouchbaseConnection(nodes, bucket)) {
       JsonDocument jsonDocument = couchbaseConnection.get(documentId);
       if (jsonDocument != null) {
         result = Stream.of(new CouchbaseJsonDocument(jsonDocument));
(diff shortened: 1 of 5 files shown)


jexp:
Perhaps we can still improve it, ala "Please download and copy the JDBC driver into $NEO4J_HOME/plugins,more details at https://neo4j-contrib.github.io/neo4j-apoc-procedures/#_load_jdbc_resources"

AngeloBusato:
@jexp Done

jexp:
I don't like to add a checked exception. If we can turn the "PluginException" into a "MissingDependencyException" which is a RuntimeException, we can just throw it and it shows up for the user. No need to pollute signatures with a checked exception.

AngeloBusato:
@jexp fixed

src/main/java/apoc/mongodb/MongoDB.java
```diff
@@ -62,38 +66,62 @@ private Coll getColl(@Name("host") String hostOrKey, @Name("db") String db, @Nam
...
 
+    private Coll getMongoColl(String hostOrKey, String db, String collection){
+        Coll coll = null;
+        try {
+            coll = getColl(hostOrKey, db, collection);
+        }
+        catch (NoClassDefFoundError e) {
+            throw new MissingDependencyException("Cannot find the jar into the plugins folder. \n"+
+                    "Please put these jar in the plugins folder :\n\n" +
+                    "bson-3.4.2.jar\n" +
```
jexp: probably leave off the exact versions? and just do -.x.y.z and link to the docs.

src/main/java/apoc/couchbase/Couchbase.java
```diff
@@ -438,4 +440,18 @@
...
     return result;
   }
+
+  private CouchbaseConnection getCouchbaseConnection(List<String> nodes, String bucket){
+    try (CouchbaseConnection couchbaseConnection = CouchbaseManager.getConnection(nodes, bucket)) {
+      return couchbaseConnection;
+    } catch (NoClassDefFoundError e){
+      throw new MissingDependencyException("Cannot find the jar into the plugins folder. \n"+
+              "Please put these jar in the plugins folder : \n\n" +
+              "java-client-2.3.1.jar\n" +
```
jexp: probably leave off the exact versions? and just do -.x.y.z and link to the docs.

AngeloBusato: @jexp fixed

src/main/java/apoc/mongodb/MongoDB.java
AngeloBusato: @jexp fixed

AngeloBusato:
@jexp fixed

PR #454, opened 2017-05-25
fixes #425 - allow apoc.cypher.runFile to take multiple files and run each file in a different tx

### LINKED ISSUE #425: allow apoc.cypher.runFile to take multiple files and run each file in a different tx
the files should be imported in order.

so that we can use the separate files produced by apoc.export.cypher with the directory option.

DanielBerton:


jexp:
@DanielBerton we cannot just change the API

So a separate, new procedure `runFiles` (with s) instead that takes a list makes sense.

Then we also have to make `runStatements` to

1. support schema operations (and test mixed schema + data operations in different files too)
2. always run the file in a new tx
3. make sure that the TOMBSTONE is passed *after* the other operations have finished (I think they are blocking but I am not 100% sure)

jexp:
@DanielBerton ping?

jexp:
We said we wanted to keep runFile

and add a runFiles with a list of files each of which is executed separately.
And a runFilesSchema which uses mode=SCHEMA/DBMS demarcation and schema transactions?

DanielBerton:
@jexp yes I'm preparing a new PR

DanielBerton:
@jexp  fixed with @inserpio support. We have two different apoc for `data` and `schema` files. 
Also we start a new transaction per file, marked with `mode.SCHEMA` the new procedures and introduced `TOMBSTONE`

Issue #453, opened 2017-05-24
Feature Improvement - Add optional parameter on apoc.schema.assert() for whether existing schema is dropped or not


InverseFalcon:
Current behavior in `apoc.schema.assert()` is to drop all indexes and constraints, then add the indexes and constraints that were provided.

According to @jexp, When the procedure was originally written, procedures couldn't do index operations, and it was written to address a customer request to assert the exact schema.

At this point we should have more flexibility in reading and changing the schema from procedures, so it makes sense to add an optional parameter `dropExisting` to control whether or not the entire schema is dropped prior to checking/adding indexes and constraints.

I propose having the default be `dropExisting = false`, as I would guess that most cases of schema assertion will not require dropping previous constraints.

We should also be clear in the documentation, the fact that previous schema was dropped wasn't well documented before, both in the APOC documentation page and the built-in documentation (apoc.help() provided no info on this procedure's behavior).

jexp:
Also rewrite schema.assert to become a regular procedure now.

InverseFalcon:
While I REALLY want to have `dropExisting = false` be default, this may break current usages. I'm going to keep default as `true` for now. Maybe we can flip the default when we reach 3.3.

InverseFalcon:
Fulfilled.

PR #458, opened 2017-06-08
Refactored apoc.schema.assert() into a proper procedure and made schema dropping optional

InverseFalcon:
Added optional argument 'dropExisting', defaulting to true (so we don't break existing usage), giving users the option to retain their current indexes and unique constraints.

This addresses request #453.

We might want to consider flipping the default value for `dropExisting` to false when we hit the next minor version (3.3).


InverseFalcon:
I'll need to work on a 3.2 version, and ensure that it plays nice with node keys.

jexp:
Thanks a lot

Yes we'll have to consider going forward how to best handle node-keys and existence constraints.

PR #459, opened 2017-06-09
Adding apoc.schema.properties.distinctCount(label, key) - quickly ret…

maxdemarzi:
…urns all distinct values and counts for a given key

### DIFF
--- a/src/main/java/apoc/index/SchemaIndex.java
+++ b/src/main/java/apoc/index/SchemaIndex.java
@@ -1,5 +1,6 @@
 package apoc.index;
 
+import apoc.result.MapResult;
 import org.neo4j.procedure.Description;
 import apoc.result.ListResult;
 import apoc.result.NodeResult;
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/index/SchemaIndex.java
```diff
@@ -159,18 +160,37 @@ private IndexReader getIndexReader(String label, String key) throws SchemaRuleNo
...
-        terms = fields.terms("number");
+
+        return new ArrayList<>(values);
+    }
+
+    @Procedure("apoc.schema.properties.distinctCount")
+    @Description("apoc.schema.properties.distinctCount(label, key) - quickly returns all distinct values and counts for a given key")
+    public Stream<MapResult> distinctCount(@Name("label") String label, @Name("key")  String key) throws SchemaRuleNotFoundException, IndexNotFoundKernelException, IOException {
+        Map<String, Object> values = distinctTermsCount(label, key);
+        return Stream.of(new MapResult(values));
```
jexp: wouldn't it be better to have label + key + value + count columns to filter on?
Then you can also do this across labels (just iterate the existing indexes) .i.e provide default values of null for the label and key parameters.

jexp:
@maxdemarzi ping

jexp:
merged manually

PR #462, opened 2017-06-15
Fixes #125 - optional json path / xpath for load.json, load.xml

### LINKED ISSUE #125: optional json path / xpath for load.json, load.xml

DanielBerton:
@jexp as discussed with @albertodelazzari we create a pull request, please take a look at the tests about xPath. There are some comments, we'll remove them


src/main/java/apoc/load/Xml.java
```diff
@@ -27,31 +45,119 @@
     @Context public GraphDatabaseService db;
 
     @Procedure
-    @Description("apoc.load.xml('http://example.com/test.xml', false) YIELD value as doc CREATE (p:Person) SET p.name = doc.name load from XML URL (e.g. web-api) to import XML as single nested map with attributes and _type, _text and _childrenx fields.")
-    public Stream<MapResult> xml(@Name("url") String url, @Name(value = "simple", defaultValue = "false") boolean simpleMode) {
-        return xmlToMapResult(url, simpleMode);
+    @Description("apoc.load.xml('http://example.com/test.xml', 'xPath', false) YIELD value as doc CREATE (p:Person) SET p.name = doc.name load from XML URL (e.g. web-api) to import XML as single nested map with attributes and _type, _text and _childrenx fields.")
+    public Stream<MapResult> xml(@Name("url") String url, @Name(value = "path",defaultValue = "/") String path, @Name(value = "simple", defaultValue = "false") boolean simpleMode) throws Exception {
```
jexp: For efficiency reasons doesn't make it sense to use the previous implementation that doesn't do XPath if there is none given / wanted ?

src/main/java/apoc/load/Xml.java
```diff
@@ -27,31 +45,119 @@
...
+                    result.add(new MapResult(Collections.singletonMap("text", nodeList.item(i).getNodeValue())));
+                }
+            }
+            return result.stream();
         } catch (IOException | XMLStreamException e) {
             throw new RuntimeException("Can't read url " + cleanUrl(url) + " as XML", e);
         }
     }
 
+    private String getXMlfromNode(Node node) throws Exception{
```
jexp: looks expensive

src/main/java/apoc/load/Xml.java
```diff
@@ -27,31 +45,119 @@
...
+                    recursiveFilterByPath(pathType, (List<Map<String, Object>>) c.get("_children"), i);
+                } else {
+                    toBeRemoved.add(c);
+                }
+            });
+            toBeRemoved.forEach(r -> children.remove(r));
+        }
+    }
+
+    @Deprecated
```
jexp: why are these deprecated?

src/test/java/apoc/load/XmlTest.java
```diff
@@ -138,4 +138,81 @@ public void testReturnCollectionElements() {
                 });
     }
 
+    @Test
+    public void testReturnCollectionElementsWithPathTitle() {
+        testResult(db, "call apoc.load.xml('file:src/test/resources/books.xml', '/catalog/book[genre=\"Computer\"]/title') yield value as result\n" +
+                        "WITH result._text as title\n" +
+                        "RETURN  title"
+                , result -> {
+                    assertEquals("+------------------------------------------+\n" +
```
jexp: can we please not use that? we're tying ourselves to the tabular format.
but instead like in other places check the result's columns and then the values.
you can also in the statement order and collect so you only have to check a single collection and can use `testCall`

src/test/java/apoc/load/XmlTest.java
```diff
@@ -138,4 +138,81 @@ public void testReturnCollectionElements() {
...
+                });
+    }
+
+    @Test
+    public void testReturnCollectionElementsWithPathAuthorFromBookId() {
+        testResult(db, "call apoc.load.xml('file:src/test/resources/books.xml', '/catalog/book[@id=\"bk102\"]/author') yield value as result\n" +
+                        "WITH result._text as author\n" +
+                        "RETURN  author"
+                , result -> {
+                    assertEquals("+--------------+\n" +
```
jexp: same here

src/test/java/apoc/load/XmlTest.java
```diff
@@ -138,4 +138,81 @@ public void testReturnCollectionElements() {
...
+    }
+
+    @Test
+    public void testReturnCollectionElementsWithRootPath() {
+        testResult(db, "call apoc.load.xml('file:src/test/resources/books.xml', '/') yield value as catalog\n" +
+                        "UNWIND catalog._children as book\n" +
+                        "WITH book.id as id, [attr IN book._children WHERE attr._type IN ['author','title'] | attr._text] as pairs\n" +
+                        "RETURN id, pairs[0] as author, pairs[1] as title"
+                , result -> {
+                    assertEquals("+-----------------------------------------------------------------------------+\n" +
```
jexp: here you can oder by id and collect into a map and compare with the map

src/test/java/apoc/load/XmlTest.java
```diff
@@ -138,4 +138,81 @@ public void testReturnCollectionElements() {
...
+                });
+    }
+
+    @Test
+    public void testReturnCollectionElementsWithPathText() {
+        testResult(db, "call apoc.load.xml('file:src/test/resources/books.xml', '/catalog/book[genre=\"Computer\"]') yield value as book\n" +
+                        "WITH book.id as id, [attr IN book._children WHERE attr._type IN ['title','price'] | attr._text] as pairs\n" +
+                        "RETURN id, pairs[0] as title, pairs[1] as price"
+                , result -> {
+                    assertEquals("+--------------------------------------------------------------+\n" +
```
jexp: same here

jexp:
Ping

DanielBerton:
@jexp we are working on, we trying to use the XML `Node` instead of `XMLStreamReader` , and we are doing some performance tests to understand if it is better have just one method  or a separate one for  `xPath`.

DanielBerton:
@jexp updated PR, mixedContent solved now respects the document structure

src/test/java/apoc/load/XmlTest.java
```diff
@@ -71,7 +79,9 @@ public void testMixedContent() {
         testCall(db, "CALL apoc.load.xml('file:src/test/resources/mixedcontent.xml')", //  YIELD value RETURN value
                 (row) -> {
                     Object value = row.get("value");
-                    assertEquals("{_type=root, _children=[{_type=text, _children=[text0, {_type=mixed}, text1]}, {_type=text, _text=text as cdata}]}", value.toString());
+                    //assertEquals("{_type=root, _children=[{_type=text, _children=[text0, {_type=mixed}, text1]}, {_type=text, _text=text as cdata}]}", value.toString());
```
albertodelazzari: @sarmbruster we changed the test here because we are using Node objects to create the final JSON representation in order to support both XPath path or not.
With this change we can't guarantee the order of a mixed content as it is in the XML document.
For example in this test before our changes there was `[text0, {_type=mixed}, text1]` while now is `[{_type=mixed}, text0, text1]`. I'm also wondering if `[{_type=mixed}, text0, text1]` is the correct JSON structure or maybe it should be `[{_type=mixed}, {_text=[text0, text1]}]`.

Thanks

PR #465, opened 2017-06-16
Added procedures to fuzzyMatch text and calculate levensthein distance

rweverwijk:


jexp:
Oooh cool, thanks so much Ron !!

src/main/java/apoc/text/Strings.java
```diff
@@ -84,6 +86,27 @@ public boolean compareCleaned(final @Name("text1") String text1, final @Name("te
     }
 
     @UserFunction
+    @Description("apoc.text.getLevenshteinDistance(text1, text2) - compare the given strings with the StringUtils.getLevenshteinDistance(text1, text2) method")
+    public Long getLevenshteinDistance(final @Name("text1") String text1, @Name("text1")final String text2) {
```
jexp: how does it handle it if one or both are null?
perhaps return null then like the rest of cypher

src/main/java/apoc/text/Strings.java
```diff
@@ -84,6 +86,27 @@ public boolean compareCleaned(final @Name("text1") String text1, final @Name("te
...
+        return Long.valueOf(StringUtils.getLevenshteinDistance(text1, text2));
+    }
+
+    @UserFunction
+    @Description("apoc.text.fuzzyMatch(text1, text2) - check if 2 words can be matched in a fuzzy way. Depending on the" +
+                 " length of the String it will allow more characters that needs to be editted to match the second String.")
+    public Boolean fuzzyMatch(final @Name("text1") String text1, @Name("text1")final String text2) {
+        if (text1 == null || text2 == null) {
+            return false;
```
jexp: return rather null to be consistent with the rest of cypher

src/main/java/apoc/text/Strings.java
rweverwijk: fair enough

src/main/java/apoc/text/Strings.java
rweverwijk: fair enough

PR #473, opened 2017-06-23
added apoc.util.fail and apoc.util.assert

ikwattro:
Added the possibility to trigger an exception with `apoc.util.fail` and `apoc.util.assert`.

Note that the latter is quite similar to `apoc.util.validate`


jexp:
Can you then remove the assert and just keep fail?

PR #484, opened 2017-06-29
fixes #332 - add atomic procedures for increment, decrement, etc.

### LINKED ISSUE #332: add atomic procedures for increment, decrement, etc.
that pull a write lock on the node first before reading the property and also handle non-initialized properaties

other ideas:

* add/remove to/from array
* concatenate string
* add/subtract number
* general "compiled-cypher" expression (with optional initializer)

AngeloBusato:


src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,154 @@
...
+	@Context public KernelTransaction tx;
+
+	/**
+	 * increment a property's value
+	 */
+	@Procedure(mode = Mode.WRITE)
+	@Description("apoc.atomic.add(node/relatonship,propertyName,number) Sums the property's value with the 'number' value ")
+	public Stream<AtomicResults> add(@Name("container") Object container, @Name("propertyName") String property, @Name("number") Number number) {
+		PropertyContainer propertyContainer = (PropertyContainer) container;
+		Number oldValue = (Number) propertyContainer.getProperty(property);
```
jexp: this is not atomic, we have to grab a (read)lock first thing, that's what the whole thing is about.

jexp: Same in all the other procedures

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,154 @@
...
+	 * remove a value into an array property value
+	 */
+	@Procedure(mode = Mode.WRITE)
+	@Description("apoc.atomic.remove(node/relatonship,propertyName,position) remove the element at position 'position'")
+	public Stream<AtomicResults> remove(@Name("container") Object container, @Name("propertyName") String property, @Name("position") Long position)
+			throws ClassNotFoundException {
+		PropertyContainer propertyContainer = (PropertyContainer) container;
+		Object oldValue = propertyContainer.getProperty(property);
+		List<?> values = new ArrayList<>();
+		values.addAll(new ArrayBackedList(oldValue));
```
jexp: this is way too complicated, just use arraycopy + Arrays.copyOf to shrink, or just a list.

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,154 @@
...
+		return Stream.of(new AtomicResults(oldValue,newValue));
+	}
+
+	/**
+	 * update the property's value
+	 */
+	@Procedure(mode = Mode.WRITE)
+	@Description("apoc.atomic.update(node/relatonship,propertyName,updateOperation) update a property's value with a cypher operation (ex. \"n.prop1+n.prop2\")")
+	public Stream<AtomicResults> update(@Name("container") Object container, @Name("propertyName") String property, @Name("operation") String operation){
+		PropertyContainer propertyContainer = (PropertyContainer) container;
```
jexp: we have to take a lock again.
not sure why you used iterate here, for a singular update.

AngeloBusato:
@jexp  PR updated

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,213 @@
...
+
+	/**
+	 * increment a property's value
+	 */
+	@Procedure(mode = Mode.WRITE)
+	@Description("apoc.atomic.add(node/relatonship,propertyName,number) Sums the property's value with the 'number' value ")
+	public Stream<AtomicResults> add(@Name("container") Object container, @Name("propertyName") String property, @Name("number") Number number) {
+		checkIsPropertyContainer(container);
+		PropertyContainer propertyContainer = (PropertyContainer) container;
+		checkPropertyExists(propertyContainer,property);
```
jexp: lock has to happen much earlier, here, before the first read

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,213 @@
...
+		Number oldValue = (Number) propertyContainer.getProperty(property);
+		Number newValue;
+		try(Transaction tx = db.beginTx()){
```
jexp: there is already a tx for this procedure, no need for a nested tx

jexp: there is already the KernelTx here.

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,213 @@
...
+	@Description("apoc.atomic.insert(node/relatonship,propertyName,position,value) insert a value into the property's array value at 'position'")
+	public Stream<AtomicResults> insert(@Name("container") Object container, @Name("propertyName") String property, @Name("position") Long position, @Name("value") Object value) throws ClassNotFoundException {
+		checkIsPropertyContainer(container);
+		PropertyContainer propertyContainer = (PropertyContainer) container;
+		checkPropertyExists(propertyContainer,property);
+		Object oldValue = propertyContainer.getProperty(property);
+		Object newValue;
+		try(Transaction tx = db.beginTx()){
+			Lock lock = tx.acquireReadLock(propertyContainer);
+			List<Object> values = insertValueIntoArray(oldValue, position, value);
```
jexp: This is still way to complicated to insert a value into an array.

src/main/java/apoc/atomic/Atomic.java
jexp: in all of the procedures.

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,213 @@
...
+	@Description("apoc.atomic.remove(node/relatonship,propertyName,position) remove the element at position 'position'")
+	public Stream<AtomicResults> remove(@Name("container") Object container, @Name("propertyName") String property, @Name("position") Long position) throws ClassNotFoundException {
+		checkIsPropertyContainer(container);
+		PropertyContainer propertyContainer = (PropertyContainer) container;
+		checkPropertyExists(propertyContainer,property);
+		Object oldValue = propertyContainer.getProperty(property);
+		Object newValue;
+		try(Transaction tx = db.beginTx()){
+			Lock lock = tx.acquireReadLock(propertyContainer);
+			Object[] arrayBackedList = new ArrayBackedList(oldValue).toArray();
```
jexp: Just do a system.arrayCopy to idx -1 if (not first or last index) + Arrays.copyOf
array.getClass().getComponentType()

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,213 @@
...
+	public Stream<AtomicResults> update(@Name("container") Object container, @Name("propertyName") String property, @Name("operation") String operation){
+		checkIsPropertyContainer(container);
+		PropertyContainer propertyContainer = (PropertyContainer) container;
+		checkPropertyExists(propertyContainer,property);
+		Object oldValue = propertyContainer.getProperty(property);
+		try(Transaction tx = db.beginTx()){
+			Lock lock = tx.acquireReadLock(propertyContainer);
+			String cypherStatement = null;
+			if(container instanceof Node)
+				cypherStatement= "start n=node("+ ((Node) propertyContainer).getId() + ") set n."+property+"="+operation;
```
jexp: just pass the propertyContainer instance to db.execute as parameter, e.g. "this" or "n" or "_" and set the value

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,213 @@
...
+			values.add(oldValue);
+		if(position > values.size())
+			values.add(value);
+		else
+			values.add(position.intValue(), value);
+		return values;
+	}
+
+
+	private void checkPropertyExists(PropertyContainer container, String property) {
```
jexp: the "getProperty" already fails if the property doesn't exist.

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,213 @@
...
+		if(!container.hasProperty(property)) throw new RuntimeException("Node/Relationship has not such property:" + property);
+	}
+
+
+	private void checkIsPropertyContainer(Object container) {
+		if(!(container instanceof PropertyContainer))  throw new RuntimeException("You Must pass Node or Relationship");
+	}
+
+	private Periodic newPeriodic() {
```
jexp: no need for this.

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,213 @@
...
+		Periodic periodic = new Periodic();
+		periodic.db = this.db;
+		periodic.log = this.log;
+		periodic.tx = this.tx;
+		return periodic;
+	}
+
+	public class AtomicResults{
+		public Object oldValue;
```
jexp: should also return property name and propertyContainer

src/test/java/apoc/atomic/AtomicTest.java
```diff
@@ -0,0 +1,132 @@
...
+	public void testUpdateNode(){
+		db.execute("CREATE (p:Person {name:'Tom',salary1: 1800, salary2:1500})");
+		Node node = (Node) db.execute("MATCH (n:Person {name:'Tom'}) return n;").next().get("n");
+		testCall(db, "CALL apoc.atomic.update({node},{property},{operation})",map("node",node,"property","salary1","operation","n.salary1 + n.salary2"), (r) -> {});
+		assertEquals(3300L, db.execute("MATCH (n:Person {name:'Tom'}) return n.salary1 as salary;").next().get("salary"));
+	}
+
+	@Test
+	public void testUpdateRel(){
+		db.execute("CREATE (t:Person {name:'Tom'})-[:KNOWS {forYears:5}]->(m:Person {name:'Mary'})");
```
jexp: you can just return the relationship from here
or just test the proc within the procedure.

src/test/java/apoc/atomic/AtomicTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * @author AgileLARUS
+ *
+ * @since 26-06-17
+ */
+public class AtomicTest {
+	private GraphDatabaseService db;
+
+	@Before public void setUp() throws Exception {
```
jexp: needs also concurrent tests, that's what the whole "atomic*" is about :)

DanielBerton:
@jexp  we added two concurrent tests and if we try to access at the same resource, at the same time the lock `tx.acquireReadLock(propertyContainer)` throws a `QueryExecutionException` caused by `DeadlockDetectedException` and each thread wait for the other.
This is the error :

```
Exception in thread "pool-5-thread-1" org.neo4j.graphdb.QueryExecutionException: ForsetiClient[1] can't acquire SharedLock{objectId=2046482643, refCount=2} on NODE(0), because holders of that lock are waiting for ForsetiClient[1].
 Wait list:SharedLock[
Client[1] waits for [0], 
Client[0] waits for [1]]
```

There is something wrong in the code or is the right behavior of the `Lock acquireReadLock( PropertyContainer entity )`?
The test fails even if we try to assert the expected exception `@Test(expected = QueryExecutionException.class)`  or if we try to handle it.

jexp:
@DanielBerton it seems we have to run the atomic operation into a separate tx in an extra thread to be able to retry it in case a deadlock or other race condition happens. Perhaps we should add a default retry parameter, e.g. defaulting to 3 or 5 that retries the operation and only fails after the retries failed

Util.inFuture or Util.inTx should help.

jexp:
@AngeloBusato ping :)

DanielBerton:
@jexp PR updated, we applied the standard behavior for the index position of the arrays,  `index < 0 || index > array.length` is out of range. If you want we can use `-1` to get the last element of the array.

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,250 @@
...
+    /**
+     * increment a property's value
+     */
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.atomic.add(node/relatonship,propertyName,number) Sums the property's value with the 'number' value ")
+    public Stream<AtomicResults> add(@Name("container") Object container, @Name("propertyName") String property, @Name("number") Number number, @Name(value = "times", defaultValue = "5") Long times) {
+        checkIsPropertyContainer(container);
+        PropertyContainer propertyContainer;
+        /*AtomicReference to wrap the object*/
+        final AtomicReference newValue = new AtomicReference();
```
jexp: I think the AtomicReferences are not needed, as those values never leave the method, simple Numbers will do.
The important bit is that it is atomic within neo4j, i.e. taking the write properly lock early on.

src/main/java/apoc/atomic/Atomic.java
```diff
@@ -0,0 +1,250 @@
...
+    @Description("apoc.atomic.update(node/relatonship,propertyName,updateOperation) update a property's value with a cypher operation (ex. \"n.prop1+n.prop2\")")
+    public Stream<AtomicResults> update(@Name("container") Object container, @Name("propertyName") String property, @Name("operation") String operation, @Name(value = "times", defaultValue = "5") Long times) throws InterruptedException {
+        checkIsPropertyContainer(container);
+        PropertyContainer propertyContainer = (PropertyContainer) container;
+        final AtomicReference oldValue = new AtomicReference();
+        final ExecutionContext executionContext = new ExecutionContext(db, propertyContainer, property);
+
+        retry(executionContext, (context) -> {
+            oldValue.set(propertyContainer.getProperty(property));
+            String statement = "UNWIND {container} as n with n set n." + property + "=" + operation + ";";
```
jexp: UNWIND is not needed as the object is not a list
a `WITH {container} as n` should be good enough, but property should be backticked.
Also mention in the docs that the container is `n`.

DanielBerton:
@jexp PR updated

jexp:
Thanks so much !!

PR #488, opened 2017-06-30
Added brand new apoc.sequence and hashing utils from 3.1.

Flegyas:
A sequence has a thread safe _increment_ (not creation).
The problem mentioned here https://github.com/neo4j/neo4j/issues/9599 only appears in test environment. We're not able to reproduce it on different environments (Java embedded and bolt) with the same test code.

### DIFF
--- a/src/main/java/apoc/sequence/Sequence.java
+++ b/src/main/java/apoc/sequence/Sequence.java
@@ -1,0 +1,151 @@
+package apoc.sequence;
+
+import apoc.result.NodeResult;
+import apoc.result.ObjectResult;
+import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.graphdb.Label;
+import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.procedure.*;
+
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
+/**
+ * Created by Valentino Maiorca on 6/29/17.
+ */
+public class Sequence {
+
+    /**
+     * Label for all sequence-node.
+     */
+    public static final Label SEQUENCE_LABEL = Label.label("Apoc_Sequence");
+
+    public static final String PROVIDER_PROPERTY = "provider";
+
+    public static final String NAME_PROPERTY = "name";
+
+    // Simple provider. Increments by 1 the value.
+    public static final SequenceProvider<Long> LONG_SP = new SequenceProvider<Long>() {
+
+        @Override
+        public Long start() {
+            return 0L;
+        }
+
+        @Override
+        public Long next(Long current) {
+            return current + 1;
+        }
+
+        @Override
+        public LinkedList<Long> next(Long current, long length) {
+            return LongStream.range(current + 1, current + length + 1).boxed()
+                    .collect(Collectors.toCollection(LinkedList::new));
+        }
+
+        @Override
+        public String getName() {
+            return "LongCounter";
+        }
+
+    };
+
+    // Map <providerName, sequenceProvider> holding instances of provider for fast access.
+
+    private static final Map<String, SequenceProvider> SEQUENCE_PROVIDER_MAP = new ConcurrentHashMap<>();
+
+    static {
+        SEQUENCE_PROVIDER_MAP.put(LONG_SP.getName(), LONG_SP);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/sequence/Sequence.java
```diff
@@ -0,0 +1,151 @@
...
+            return 0L;
+        }
+
+        @Override
+        public Long next(Long current) {
+            return current + 1;
+        }
+
+        @Override
+        public LinkedList<Long> next(Long current, long length) {
```
jexp: why a linkedlist and not just an long-stream?
also please don't use implementation classes as return values

src/main/java/apoc/sequence/Sequence.java
```diff
@@ -0,0 +1,151 @@
...
+    public static boolean registerProvider(SequenceProvider provider) {
+        return SEQUENCE_PROVIDER_MAP.put(provider.getName(), provider) == null;
+    }
+
+    @Context
+    public GraphDatabaseService db;
+
+    @UserFunction(value = "apoc.sequence.get")
+    @Description("apoc.sequence.get('sequenceName') - returns the node representing that sequence, null if not existing.")
+    public Node get(@Name("sequenceName") String sequenceName) {
```
jexp: 1. I would never expose this, as it should always  be internal, otherwise uses would be able to mess up
2. I would not use nodes for this for the seame reason but graph properties, either a single "apoc.sequence" graph property with a array content, or better `apoc.sequence.<name>` graph properties with long value

jexp: Also the label + property names would pollute the the users namespaces

src/main/java/apoc/sequence/Sequence.java
```diff
@@ -0,0 +1,151 @@
...
+        return db.findNode(SEQUENCE_LABEL, NAME_PROPERTY, sequenceName);
+    }
+
+
+    // TODO: 29/06/17 Can't be completely thread safe. Can't acquire a lock on a node if the node doesn't exist.
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.sequence.create('sequenceName', 'providerName') - create sequence with given provider")
```
jexp: this operation should acquire a global lock
also I'm not sure that I would separate sequence creation and use, I kinda would like if it was implicit

src/main/java/apoc/sequence/Sequence.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.sequence.next('sequenceName') - get&increment sequence value using related provider and length as increments number")
+    @SuppressWarnings("unchecked")
+    public Stream<ObjectResult> next(@Name("sequenceName") String sequenceName, @Name(value = "length", defaultValue = "1") long length) {
+        LinkedList<Object> result = new LinkedList<>();
+
+        try (Transaction tx = db.beginTx()) {
+            Node sequenceNode = db.findNode(SEQUENCE_LABEL, NAME_PROPERTY, sequenceName);
```
jexp: acquire write lock immediately

src/main/java/apoc/sequence/Sequence.java
```diff
@@ -0,0 +1,151 @@
...
+        node.setProperty(PROVIDER_PROPERTY, providerName);
+        node.setProperty(provider.propertyName(), provider.start());
+
+        return Stream.of(new NodeResult(node));
+    }
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.sequence.next('sequenceName') - get&increment sequence value using related provider and length as increments number")
+    @SuppressWarnings("unchecked")
+    public Stream<ObjectResult> next(@Name("sequenceName") String sequenceName, @Name(value = "length", defaultValue = "1") long length) {
```
jexp: in general it would be nice to have the means of acquiring batches of sequence ids (and hold them in memory) and only reach out to the sequence generator when more are needed.

jexp: I would probably limit the id-generation to `long` and not have object-results with boxing etc. What other sequence generators did you think about? The only others that make sense are UUIDs and MongoDB like object-ids which both need no centralized id-generation.

jexp:
@Flegyas thanks a lot for the contribution, I added my thoughts in the comments. Here is the essence:

1. restrict this to long-ids only
2. use graph properties for storage
3. make sequence-generation implict with a separate global lock (similar to the schema locks)

Something that might be interesting to look at is the internal neo4j-id generators, if we could reuse those implementations actually for safe, persistent id-generation

jexp:
@Flegyas ping?

jexp:
Do you plan to work on this in any way?

PR #495, opened 2017-07-03
New schema procedures for indexes and constraints fixes #447

### LINKED ISSUE #447: Feature Request - apoc.schema.existsIndex() and apoc.schema.existsConstraint()
Currently Cypher and APOC don't seem to have any good ways to determine if given indexes and constraints exist (`db.indexes()` can work, but requires string parsing to extract the label/properties).

Some helper functions or procedures would be useful here.

albertodelazzari:
@jexp @InverseFalcon this is the proposal for indexes and constraints procedures.
Property existence constraints are not available for Community Edition so I did some integration tests with an Enterprise instance.

Let me know if it's ok.
Thanks


docs/schema.adoc
```diff
@@ -1,8 +1,17 @@
-[cols="1m,3"]
+[cols="1m,2a"]
 |===
-| CALL apoc.schema.existsIndex(indexLabel, propertyName) yield value | it returns true if an index exists for a label "indexLabel" and a property "propertyName"
-| CALL apoc.schema.existsConstraintForNode(indexLabel, propertyName, constraintType) yield value | it returns true if a constraint exists for a label "indexLabel" and a property "propertyName". The constraint can be a uniqueness constraint or an existence constraint so you need to pass the third parameter to specify what type of constraints you are looking for. Possible values are: UNIQUENESS or NODE_PROPERTY_EXISTENCE.
-| CALL apoc.schema.existsConstraintForRelationship(relationshipType, propertyName) yield value | it returns true if a constraint exists for a relationship type "relationshipType" and a property "propertyName".
+| CALL apoc.schema.nodes() yield name, label, properties, status, type
+| * name of the index/constraint,
```
jexp: perhaps add an example?

jexp: also as this is a "schema.adoc" you don't need to keep the tabular format, just format it so that the page looks good
in terms of examples + list of results columns + general syntax

src/main/java/apoc/result/ConstraintRelationshipInfo.java
```diff
@@ -0,0 +1,101 @@
...
+        status = "";
+        type = lookup.relationshipTypeGetName(propertyConstraint.relationshipType());
+
+        LabelResult labelResult = getNodeLabelsFromRelationshipTypeId(readOperations, propertyConstraint.relationshipType());
+
+        startLabel = labelResult.startLabels;
+        endLabel = labelResult.endLabels;
+    }
+
+    private LabelResult getNodeLabelsFromRelationshipTypeId(ReadOperations readOperations, int relationshipTypeId)
```
jexp: for these operations we don't need the  kernel API or?
as regular schema indexes are available in the java api

src/main/java/apoc/result/IndexConstraintNodeInfo.java
```diff
@@ -0,0 +1,64 @@
...
+     * Constructor using ReadOperations and IndexDescriptor
+     * It is used for wrapping index information
+     *
+     * @param readOperations
+     * @param indexDescriptor
+     * @throws Exception
+     */
+    public IndexConstraintNodeInfo(ReadOperations readOperations, IndexDescriptor indexDescriptor) throws Exception {
+        // We use StatementTokenNameLookUp to get the name
+        name = indexDescriptor.userDescription(new StatementTokenNameLookup(readOperations));
```
jexp: I'd rather have the parameters passed into the constructor of the data class.

src/main/java/apoc/result/IndexConstraintNodeInfo.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public IndexConstraintNodeInfo(ReadOperations readOperations, PropertyConstraint propertyConstraint) throws Exception {
+        // We use StatementTokenNameLookUp to get the names
+        StatementTokenNameLookup lookup = new StatementTokenNameLookup(readOperations);
+
+        name = propertyConstraint.userDescription(lookup);
+        label = lookup.labelGetName(((NodePropertyConstraint) propertyConstraint).label());
+        properties = Collections.singletonList(lookup.propertyKeyGetName(propertyConstraint.propertyKey()));
+        status = "";
+        type = propertyConstraint instanceof UniquenessConstraint ? ConstraintType.UNIQUENESS.name() : ConstraintType.NODE_PROPERTY_EXISTENCE.name();
```
jexp: shouldn't this just constraint.type.name()

src/main/java/apoc/result/ConstraintRelationshipInfo.java
albertodelazzari: @jexp I can change this using IndexDefinition and ConstraintDefinition instead of the kernel classes (descriptors). Honestly I don't know if `IndexDescriptor.userDescription` is equal to `IndexDefinition.toString` but except this little change the real doubt is for the "status" of the index. Is there another way to do the following `readOperations.indexGetState(indexDescriptor)`? I need `ReadOperations` and an `IndexDescriptor` to get the state of an index (or not?)
For the relationships, I chose ReadOperations to find all the start/end node labels. Can I leave this as is or do you prefer anything different?

src/main/java/apoc/result/IndexConstraintNodeInfo.java
albertodelazzari: Yes if I use IndexDefinition is pretty straightforward but with `PropertyContraint` is a little bit tricky

src/main/java/apoc/result/ConstraintRelationshipInfo.java
jexp: there is schema.getIndexState http://neo4j.com/docs/java-reference/current/javadocs/org/neo4j/graphdb/schema/Schema.html#getIndexState-org.neo4j.graphdb.schema.IndexDefinition-

src/main/java/apoc/result/IndexConstraintNodeInfo.java
jexp: there is ConstraintDefinition.getConstraintType()
http://neo4j.com/docs/java-reference/current/javadocs/org/neo4j/graphdb/schema/ConstraintDefinition.html

src/main/java/apoc/result/ConstraintRelationshipInfo.java
albertodelazzari: Ok, completely forgot this method :)

src/main/java/apoc/result/IndexConstraintNodeInfo.java
jexp: Just leave off the startLabels / endLabels fields

jexp:
@albertodelazzari looks good to me, can you also add them to overview.adoc in the appropriate section? (and also mention the user-functions in schema.adoc)

I merge this PR for now as the things mentioned above are just docs fixes.

PR #506, opened 2017-07-10
re-introduce countdown fixes #260

### LINKED ISSUE #260: fix / re-add apoc.periodic.countdown
Figure out when it was removed and then add it back.

albertodelazzari:
"Countdown" function is now annotated as a procedure. I added also a test and a simple example for the documentation.

### DIFF
--- a/src/main/java/apoc/periodic/Periodic.java
+++ b/src/main/java/apoc/periodic/Periodic.java
@@ -152,7 +152,7 @@ public class Periodic {
         return Stream.of(info);
     }
 
-    @Procedure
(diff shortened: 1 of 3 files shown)


src/main/java/apoc/periodic/Periodic.java
```diff
@@ -152,10 +152,11 @@ private long executeNumericResultStatement(@Name("statement") String statement,
         return Stream.of(info);
     }
 
-    // TODO
+    @Procedure
```
jexp: I think it should have mode=WRITE

Issue #329, opened 2017-03-19
apoc.index.* doesn't add or find newly-added nodes in index


igorclark:
Hi all,

I'm trying to use [fulltext indexes](https://neo4j-contrib.github.io/neo4j-apoc-procedures/index30.html#_full_text_search) from APOC, but I'm not having much luck.

I'm running `neo4j 3.1.2` and `apoc-3.1.2.5` on `debian 8.7`.

Creating the index on an existing db works as expected:

```
neo4j-sh (?)$ call apoc.index.addAllNodes( 'search_index', { City: ["name"] } );
+-------------------------------+
| label  | property | nodeCount |
+-------------------------------+
| "City" | "name"   | 1122      |
+-------------------------------+
1 row
866 ms
neo4j-sh (?)$ call apoc.index.search( 'search_index', 'City.name:"La Paz"' ) yield node return node;
+--------------------------------------------------------------------------------------------------------------+
| node                                                                                                         |
+--------------------------------------------------------------------------------------------------------------+
| Node[1593267]{name:"La Paz",url:"/places/bolivia/la-paz",lat:-16.489689,lon:-68.1192936} |
+--------------------------------------------------------------------------------------------------------------+
```

But although I have "autoUpdate" set in neo4j.conf:

```
$ fgrep autoUpdate /etc/neo4j/neo4j.conf
apoc.autoUpdate.enabled=true
```

Adding new items doesn't automatically update the index (I've tried restarting the server after creating the index, but it doesn't make any difference):

```
neo4j-sh (?)$ create (c:City{name:"Made Up City",url:"/places/nowhere/made-up-city"}) return c;
+-----------------------------------------------------------------------+
| c                                                                     |
+-----------------------------------------------------------------------+
| Node[1746889]{url:"/places/nowhere/made-up-city",name:"Made Up City"} |
+-----------------------------------------------------------------------+
1 row
Nodes created: 1
Properties set: 2
Labels added: 1
77 ms
neo4j-sh (?)$ call apoc.index.search( 'search_index', 'City.name:"Made Up"' ) yield node return node;
+------+
| node |
+------+
+------+
0 row
134 ms
neo4j-sh (?)$ start n=node:search_index('name:"Made Up"') return n;
+---+
| n |
+---+
+---+
0 row
46 ms
```

So I try to add the node to the index manually, but whether I have  `autoUpdate` set to `true` or not, adding the node directly with `addNode` doesn't work:

```
[...]

PR #507, opened 2017-07-10
add jmh for performance testing and improvements for background index population

sarmbruster:
I've added JMH library to run micro benchmarks. Note that JMH is not enabled by default, you need to run `./gradlew jmh` for this.

Additionally this PR adds more config options for index population. 

Related to #329.

This PR should be cherry-picked to 3.2 branch as well.

### DIFF
--- a/src/jmh/java/apoc/IndexTrackingGraphDatabaseState.java
+++ b/src/jmh/java/apoc/IndexTrackingGraphDatabaseState.java
@@ -1,16 +1,16 @@
 package apoc;
 
 import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.helpers.collection.MapUtil;
 
-import java.util.Collections;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


src/jmh/java/apoc/IndexTrackingGraphDatabaseState.java
```diff
@@ -0,0 +1,29 @@
...
+import org.neo4j.helpers.collection.MapUtil;
+import org.neo4j.kernel.api.exceptions.KernelException;
+
+import java.util.Map;
+
+public class IndexTrackingGraphDatabaseState extends GraphDatabaseState {
+
+    @Override
+    public Map<String, String> getGraphDatabaseConfig() {
+        return MapUtil.genericMap("apoc.autoIndex.enabled", "true",
```
jexp: try to use Util.map()

PR #508, opened 2017-07-10
adding apoc.merge.node/relationship

sarmbruster:
We already have `apoc.create.node|relationship` to deal with dynamic relationship types or labels. These two new procedures do the very same for `merge`. In the implementation I'm assembling a Cypher `merge` statement.

This PR should be cherry-picked to 3.2 branch as well.

### DIFF
--- a/src/main/java/apoc/merge/Merge.java
+++ b/src/main/java/apoc/merge/Merge.java
@@ -1,0 +1,59 @@
+package apoc.merge;
+
+import apoc.result.*;
+import org.neo4j.graphdb.*;
+import org.neo4j.helpers.collection.Iterators;
+import org.neo4j.procedure.*;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class Merge {
+
+    @Context
+    public GraphDatabaseService db;
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.merge.node(['Label'], {key:value, ...}, {key:value,...}) - merge node with dynamic labels")
+    public Stream<NodeResult> node(@Name("label") List<String> labelNames, @Name("identProps") Map<String, Object> identProps, @Name("props") Map<String, Object> props) {
+        String labels = String.join(":", labelNames);
+
+        Map<String, Object> params = buildParams(identProps, props);
+        String identPropsString = buildIdentPropsString(identProps);
+
+        final String cypher = "MERGE (n:" + labels + "{" + identPropsString + "}) ON CREATE SET n += $props RETURN n";
+        Node node = Iterators.single(db.execute(cypher, params ).columnAs("n"));
+        return Stream.of(new NodeResult(node));
+    }
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.merge.relationship(startNode, relType,  {key:value, ...}, {key:value, ...}, endNode) - merge relationship with dynamic type")
+    public Stream<RelationshipResult> relationship(@Name("startNode") Node startNode, @Name("relationshipType") String relType,
+                                                   @Name("identProps") Map<String, Object> identProps, @Name("props") Map<String, Object> props, @Name("endNode") Node endNode) {
+        Map<String, Object> params = buildParams(identProps, props);
+        String identPropsString = buildIdentPropsString(identProps);
+
+        params.put("startNode", startNode);
+        params.put("endNode", endNode);
+
+        final String cypher = "WITH $startNode as startNode, $endNode as endNode MERGE (startNode)-[r:"+ relType +"{"+identPropsString+"}]->(endNode) ON CREATE SET r+= $props RETURN r";
+        Relationship rel = Iterators.single(db.execute(cypher, params ).columnAs("r"));
+        return Stream.of(new RelationshipResult(rel));
+    }
+
+    private String buildIdentPropsString(Map<String, Object> identProps) {
+        if (identProps==null) {
+            return "";
+        } else {
+            return identProps.keySet().stream().map(s -> "`"+s+"`:$`" + s+"`").collect(Collectors.joining(","));
+        }
+    }
+
+    private Map<String, Object> buildParams(Map<String, Object> identProps, Map<String, Object> props) {
+        Map<String, Object> map = identProps == null ? new HashMap<>() : new HashMap<>(identProps);
+        map.put("props", props == null ? Collections.EMPTY_MAP : props);
+        return map;
+    }
+
+}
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/merge/Merge.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.merge.node(['Label'], {key:value, ...}, {key:value,...}) - merge node with dynamic labels")
+    public Stream<NodeResult> node(@Name("label") List<String> labelNames, @Name("identProps") Map<String, Object> identProps, @Name("props") Map<String, Object> props) {
+        String labels = String.join(":", labelNames);
+
+        Map<String, Object> params = buildParams(identProps, props);
+        String identPropsString = buildIdentPropsString(identProps);
+
+        final String cypher = "MERGE (n:" + labels + "{" + identPropsString + "}) ON CREATE SET n += $props RETURN n";
```
jexp: I don't think empty {} will work when no id-props, isn't that an error condition to have no id-props?

src/main/java/apoc/merge/Merge.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class Merge {
+
+    @Context
+    public GraphDatabaseService db;
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.merge.node(['Label'], {key:value, ...}, {key:value,...}) - merge node with dynamic labels")
+    public Stream<NodeResult> node(@Name("label") List<String> labelNames, @Name("identProps") Map<String, Object> identProps, @Name("props") Map<String, Object> props) {
+        String labels = String.join(":", labelNames);
```
jexp: should we quote / backtick labels too?

src/main/java/apoc/merge/Merge.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.stream.Stream;
+
+public class Merge {
+
+    @Context
+    public GraphDatabaseService db;
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.merge.node(['Label'], {key:value, ...}, {key:value,...}) - merge node with dynamic labels")
+    public Stream<NodeResult> node(@Name("label") List<String> labelNames, @Name("identProps") Map<String, Object> identProps, @Name("props") Map<String, Object> props) {
```
jexp: should we differentiate between on-create props and on-update props?
props should have a default value of an empty map, if that map is empty we can leave off the ON CREATE part

src/main/java/apoc/merge/Merge.java
```diff
@@ -0,0 +1,59 @@
...
+    @Description("apoc.merge.relationship(startNode, relType,  {key:value, ...}, {key:value, ...}, endNode) - merge relationship with dynamic type")
+    public Stream<RelationshipResult> relationship(@Name("startNode") Node startNode, @Name("relationshipType") String relType,
+                                                   @Name("identProps") Map<String, Object> identProps, @Name("props") Map<String, Object> props, @Name("endNode") Node endNode) {
+        Map<String, Object> params = buildParams(identProps, props);
+        String identPropsString = buildIdentPropsString(identProps);
+
+        params.put("startNode", startNode);
+        params.put("endNode", endNode);
+
+        final String cypher = "WITH $startNode as startNode, $endNode as endNode MERGE (startNode)-[r:"+ relType +"{"+identPropsString+"}]->(endNode) ON CREATE SET r+= $props RETURN r";
```
jexp: we should quote the rel-type

src/main/java/apoc/merge/Merge.java
```diff
@@ -0,0 +1,59 @@
...
+
+        final String cypher = "MERGE (n:" + labels + "{" + identPropsString + "}) ON CREATE SET n += $props RETURN n";
+        Node node = Iterators.single(db.execute(cypher, params ).columnAs("n"));
+        return Stream.of(new NodeResult(node));
+    }
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.merge.relationship(startNode, relType,  {key:value, ...}, {key:value, ...}, endNode) - merge relationship with dynamic type")
+    public Stream<RelationshipResult> relationship(@Name("startNode") Node startNode, @Name("relationshipType") String relType,
+                                                   @Name("identProps") Map<String, Object> identProps, @Name("props") Map<String, Object> props, @Name("endNode") Node endNode) {
```
jexp: same for props as above.
I wonder if we should have a variant that allows for multiple start / end-nodes / properties  with UNWIND internally

src/main/java/apoc/merge/Merge.java
```diff
@@ -0,0 +1,59 @@
...
+    private String buildIdentPropsString(Map<String, Object> identProps) {
+        if (identProps==null) {
+            return "";
+        } else {
+            return identProps.keySet().stream().map(s -> "`"+s+"`:$`" + s+"`").collect(Collectors.joining(","));
+        }
+    }
+
+    private Map<String, Object> buildParams(Map<String, Object> identProps, Map<String, Object> props) {
+        Map<String, Object> map = identProps == null ? new HashMap<>() : new HashMap<>(identProps);
```
jexp: there is already an Util.merge(map1,map2)

src/main/java/apoc/merge/Merge.java
jexp: I wonder if we should have a variant that allows for nodes to be merged with UNWIND internally

src/main/java/apoc/merge/Merge.java
sarmbruster: good point, will change it

src/main/java/apoc/merge/Merge.java
sarmbruster: indeed

src/main/java/apoc/merge/Merge.java
sarmbruster: ok

src/main/java/apoc/merge/Merge.java
sarmbruster: ok

src/main/java/apoc/merge/Merge.java
sarmbruster: I think this should be done from outside. Let's first validate it's an issue for our users, if so we can add it later on. Ok?

sarmbruster: Let's leave that off for a feature request. In real life `ON MATCH` is used pretty rarely.

sarmbruster: that PR is already merged, pls create a separate gh issue to track this.

PR #519, opened 2017-07-18
fixes #518 - apoc.meta.schema missing existence, incoming relationships and counts 0 for global relationships

### LINKED ISSUE #518: apoc.meta.schema missing existence, incoming relationships and counts 0 for global relationships
Refers to the issue #384  : 

- counts 0 for global relationships
- existence missing for node properties
- missing incoming relationships

DanielBerton:


src/main/java/apoc/meta/Meta.java
```diff
@@ -234,12 +237,18 @@ public void rel(int typeId, String typeName, long count) {
             }
 
             public void rel(int typeId, String typeName, int labelId, String labelName, long out, long in) {
-                if (out > 0) relStats.put("(:" + labelName + ")-[:" + typeName + "]->()", out);
-                if (in > 0) relStats.put("()-[:" + typeName + "]->(:" + labelName + ")", in);
+                if (out > 0) {
+                    relStats.put("(:" + labelName + ")-[:" + typeName + "]->()", out);
+                    relStatsCount.put(typeName, out);
```
jexp: is relstatscount global? if so shouldn't we add up the different counts per label?
actually there is a global count too of just rel-type in the count store

DanielBerton: Yes it is global, we add it because the global rel count was always 0. Now now we can get total relationship count for each relType `metaStats.relTypesCount.getOrDefault(entityName, 0L)`

jexp: The global relationship-count is available separately in the counts-store.

Here you have to sum it up in my understanding.
`relStatsCount.put(typeName, gelStatsCount.getOrDefault(typeName,0) + out);`

DanielBerton:
@jexp PR updated

Issue #502, opened 2017-07-06
add apoc.text.split function that splits a string using a regexp


jexp:
this should be an easy one.

PR #526, opened 2017-07-25
add apoc.text.split function that splits a string using a regexp (#502)

capgeti:
I've Implemented the apoc.text.split function from the https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/502 issue. Is the null handling as expected or should it return an empty list instead of null?

### DIFF
--- a/docs/text.adoc
+++ b/docs/text.adoc
@@ -2,6 +2,8 @@
 
 Basic utility procedures for string manipulation, comparison, filtering etc.
 
+The `replace`, `split` and `regexpGroups` functions work with regular expressions.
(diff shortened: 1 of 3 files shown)


jexp:
Looks good, I just changed the examples to use actual regexps as parameters.

jexp:
@capgeti can you send in the Neo4j CLA if you haven't yet: https://neo4j.com/developer/cla/

jexp:
@capgeti please don't forget the CLA

capgeti:
@jexp done

PR #529, opened 2017-07-28
Adding apoc.debug.log with user function for logging out messages

bradnussbaum:
Some initial thoughts for debug to think on...


jexp:
Some comments:

- add global apoc config setting for logging
- add a util helper method for checking the config setting and using the passed in `@Context Log`
- add apoc.log.<error,warn,info,debug>(message, [params=[]]) void procedures that also honor the log setting
- add debug logging to relevant procedures, like periodic iterate, load.jdbc and others

bradnussbaum:
@jexp 
1. For settings I think best behavior is to use the global logging unless users have specifically configured an apoc logging level. What is the property for the apoc logging level? I think logging should be configurable at least to the package level to enable silencing higher levels of logging. 
2. I think the scope of this PR should just be introduce the core logging capabilities so we can add logging support to each package i.e. periodic with logging level configuration. This PR will otherwise get very big and may never get merged.

jexp:
There is no log-level config for apoc yet, not sure if the neo4j log level is even configurable.

Agreed on separating the individual logging out.

bradnussbaum:
@jexp Is there a reason this cannot be merged now? We are using this a lot now. Maybe there are other enhancements but I'd rather now have to maintain merging in this separate branch all the time.

jexp:
Oh sorry. I only saw the single commit. So I thought you didn't update it :)
I'll merge it.

bradnussbaum:
@jexp thanks

Issue #152, opened 2016-09-20
Add functions for apoc.convert.to<XXX>List that check and return List<String>, List<Node> etc


dmk255:
This should output Collection<Node> rather than Collection<T>

```
CALL apoc.coll.intersection(Collection<Node>,Collection<Node>) yield value 
```

In the following case Collection<T> is correct (two different collection types, not sure of a good example)

```
CALL apoc.coll.intersection(Collection<Node>,Collection<T>) yield value 
```

Due to this you can't unwind and set labels on the intersection

jexp:
It's a bit tricky as the procedure can only take a List<Object> and return a List<Object>

but what you can do after your unwind is to call apoc.convert.toNode() to turn the Any again into a node. type-wise.

jexp:
We can add some functions for apoc.convert.toXXX that do List<String>, List<Node> etc.?

jexp:
Already in.

PR #531, opened 2017-07-31
Added function apoc.convert.toListOf(value, type) (ref. #152)

conker84:
Hi!
I created the method `apoc.convert.toListOf(value, type)` that converts a collection to the required type, following some examples (taken from the test):
```java
testCall(db, "return apoc.convert.toListOf({a}, 'int') as value", map("a", new Object[]{"1"}),
        		 r -> assertEquals(singletonList(1), r.get("value")));
testCall(db, "CREATE (n) WITH [n] as x RETURN apoc.convert.toListOf(x, 'node') as nodes",
                 r -> {
                	 assertEquals(true, r.get("nodes") instanceof List);
                	 List<Node> nodes = (List<Node>) r.get("nodes");
                	 assertEquals(true, nodes.get(0) instanceof Node);
                 })
testCall(db, "CREATE (n)-[r:KNOWS]->(m) WITH [r] as x RETURN apoc.convert.toListOf(x, 'relationship') AS rels",
                 r -> {
                	 assertEquals(true, r.get("rels") instanceof List);
                	 List<Relationship> rels = (List<Relationship>) r.get("rels");
                	 assertEquals(true, rels.get(0) instanceof Relationship);
                 });
testCall(db, "return apoc.convert.toListOf({a}, 'boolean') as value",
        		 map("a", new Object[]{"false", 0, "no", "", null}),
        		 r -> assertEquals(Arrays.asList(false, false, false, false, false), r.get("value")));
```
Hope it helps.
Andrea

### DIFF
--- a/src/main/java/apoc/convert/Convert.java
+++ b/src/main/java/apoc/convert/Convert.java
@@ -1,6 +1,5 @@
 package apoc.convert;
 
-import java.math.RoundingMode;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -21,7 +20,7 @@ import org.neo4j.procedure.Name;
 import org.neo4j.procedure.UserFunction;
 
 import apoc.coll.SetBackedList;
-import apoc.number.exact.Exact;
+import apoc.meta.Meta.Types;
 import apoc.util.Util;
 
 
@@ -96,38 +95,28 @@ public class Convert {
         	return null;
         }
         Stream<T> stream = null;
-    	if (type.isAssignableFrom(Long.class)) {
-    		final Exact exact = new Exact();
-    		stream = (Stream<T>) convertedList.stream().map(e -> {
-    			if (e == null) {
-    				return null;
-    			}
-    			if (e instanceof Number) {
-    				return ((Number)e).longValue();
-    			}
-    			return exact.toInteger(toString(e), 0L, RoundingMode.HALF_UP.toString());
-    		});
-    	} else if (type.isAssignableFrom(Double.class)) {
-    		final Exact exact = new Exact();
-    		stream = (Stream<T>) convertedList.stream().map(e -> {
-				if (e == null) {
-    				return null;
-    			}
-    			if (e instanceof Number) {
-    				return ((Number)e).doubleValue();
-    			}
-    			return exact.toFloat(toString(e), 0L, RoundingMode.HALF_UP.toString());
-    		});
-    	} else if (type.isAssignableFrom(String.class)) {
+        Types varType = Types.of(type);
+    	switch (varType) {
+    	case INTEGER:
+    		stream = (Stream<T>) convertedList.stream().map(Util::toLong);
+    		break;
+    	case FLOAT:
+    		stream = (Stream<T>) convertedList.stream().map(Util::toDouble);
+    		break;
+    	case STRING:
     		stream = (Stream<T>) convertedList.stream().map(this::toString);
-    	} else if (type.isAssignableFrom(Boolean.class)) {
+    		break;
+    	case BOOLEAN:
     		stream = (Stream<T>) convertedList.stream().map(this::toBoolean);
-    	} else if (type.isAssignableFrom(Node.class)) {
+    		break;
+    	case NODE:
     		stream = (Stream<T>) convertedList.stream().map(this::toNode);
-    	} else if (type.isAssignableFrom(Relationship.class)) {
+    		break;
+    	case RELATIONSHIP:
     		stream = (Stream<T>) convertedList.stream().map(this::toRelationship);
-    	} else {
-    		throw new UnsupportedTypeException("Supported types are: Long, Double, String, Boolean, Node, Relationship");
+    		break;
+		default:
+			throw new RuntimeException("Supported types are: Integer, Float, String, Boolean, Node, Relationship");
     	}
     	return stream.collect(Collectors.toList());
     }
--- a/src/main/java/apoc/convert/Convert.java
+++ b/src/main/java/apoc/convert/Convert.java
@@ -1,22 +1,28 @@
 package apoc.convert;
 
-import org.neo4j.graphdb.PropertyContainer;
-import org.neo4j.procedure.Description;
-import apoc.coll.SetBackedList;
-import apoc.result.*;
-import apoc.util.Util;
+import java.math.RoundingMode;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.PropertyContainer;
 import org.neo4j.graphdb.Relationship;
 import org.neo4j.helpers.collection.Iterables;
 import org.neo4j.helpers.collection.Iterators;
+import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Name;
-import org.neo4j.procedure.Procedure;
 import org.neo4j.procedure.UserFunction;
 
-import java.lang.reflect.Array;
-import java.util.*;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import apoc.coll.SetBackedList;
+import apoc.number.exact.Exact;
+import apoc.util.Util;
 
 
 /**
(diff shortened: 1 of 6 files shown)


jexp:
@conker84 the problem is that you cannot have dynamic generic return types.

I.e. you need to have List<Node> List<Long> etc. as return types.

e.g. in the test something like [x IN list | id(x)] has to work after the toList() operation.

jexp:
@conker84 ping :)

conker84:
@jexp sorry I lost it, so if you agree I'll create 5 specific methods:

1. toIntList
2. toStringList
3. toBooleanList
4. toNodeList
5. toRelationshipList

Ok?

jexp:
@conker84 sounds good to me

src/main/java/apoc/convert/Convert.java
```diff
@@ -91,48 +91,69 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
-    @SuppressWarnings("rawtypes")
 	@UserFunction
-    @Description("apoc.convert.toListOf(value, type) | tries it's best to convert "
-    		+ "the value to a list of supported type, supported types are: "
-    		+ "int, long, double, String, boolean, node, relationship")
-    public List toListOf(@Name("list") Object list, @Name("type") String type) {
-        List objectList = convertToList(list);
+    @Description("apoc.convert.toIntList(value) | tries it's best to convert "
+    		+ "the value to a list of integers")
+    public List<Integer> toIntList(@Name("list") Object list) {
```
jexp: List<Long> as Cypher doesn't know Integers but calls it Integer

and use toInteger function

src/main/java/apoc/convert/Convert.java
```diff
@@ -91,48 +91,69 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+        List<Object> objectList = convertToList(list);
```
jexp: perhaps we can simplify all these functions by having a convertToList(list, Long.class)

that internally does the "list" to null or a stream and then a switch on the type for the element migration
also if the element is already of a type don't convert it again.

if it doesn't exist we should also add the functions toFloat(value) -> Double and toInteger(value) -> Long

```
Double toFloat(Object value) {
if (value==null) return null;
if (value instanceof Number) return ((Number)value).doubleValue();
try {
  return Double.parseDouble(value.toString);
} catch (NFE ) {
   return null;
}
}

```

conker84:
Hi @jexp let me know if it's ok now!
Thanks!

src/main/java/apoc/convert/Convert.java
```diff
@@ -79,12 +88,91 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
-
+    
     @SuppressWarnings("unchecked")
+    private <T> List<T> convertToList(Object list, Class<T> type) {
+        List<Object> convertedList = convertToList(list);
+        if (convertedList == null) {
+        	return null;
+        }
+        Stream<T> stream = null;
+    	if (type.isAssignableFrom(Long.class)) {
```
jexp: should be able to use switch on class? or did java not get that yet ? :)

src/main/java/apoc/convert/Convert.java
```diff
@@ -79,12 +88,91 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+    		final Exact exact = new Exact();
+    		stream = (Stream<T>) convertedList.stream().map(e -> {
+    			if (e == null) {
```
jexp: wrap this whole code-block in toInteger and move it to Util.toInteger

src/main/java/apoc/convert/Convert.java
```diff
@@ -79,12 +88,91 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+    				return null;
+    			}
+    			if (e instanceof Number) {
+    				return ((Number)e).longValue();
+    			}
+    			return exact.toInteger(toString(e), 0L, RoundingMode.HALF_UP.toString());
```
jexp: this is way to expensive. I'd stick with try catch around Long.parseLong(e.toString()) and return null in case of exception.

src/main/java/apoc/convert/Convert.java
```diff
@@ -79,12 +88,91 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+    		});
+    	} else if (type.isAssignableFrom(Double.class)) {
+    		final Exact exact = new Exact();
+    		stream = (Stream<T>) convertedList.stream().map(e -> {
+				if (e == null) {
```
jexp: same here

src/main/java/apoc/convert/Convert.java
```diff
@@ -79,12 +88,91 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+    				return null;
+    			}
+    			if (e instanceof Number) {
+    				return ((Number)e).doubleValue();
+    			}
+    			return exact.toFloat(toString(e), 0L, RoundingMode.HALF_UP.toString());
```
jexp: same here

src/main/java/apoc/convert/Convert.java
```diff
@@ -79,12 +88,91 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+    		});
+    	} else if (type.isAssignableFrom(String.class)) {
+    		stream = (Stream<T>) convertedList.stream().map(this::toString);
```
jexp: use the static methods in Util, makes it easier to inline

src/main/java/apoc/convert/UnsupportedTypeException.java
```diff
@@ -0,0 +1,9 @@
+package apoc.convert;
+
+public class UnsupportedTypeException extends RuntimeException {
```
jexp: don't need that, runtime exception is good enough. the end user will get just string message anyway

src/main/java/apoc/convert/Convert.java
conker84: I found in the Util class the [toLong](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.2/src/main/java/apoc/util/Util.java#L219) method maybe i wrap it in private method in the Covert class, something like this:
```java
    private Long toLong(Object value) {
    	try {
    		return Util.toLong(value);
    	} catch (NullPointerException e) {
    		return null;
    	}
    }
```

Or i can change the method in Util class in something like this:
```java
(-) public static long toLong(Object value) {
(+) public static Long toLong(Object value) {
(+)     if (value == null) return null;        
        if (value instanceof Number) return ((Number)value).longValue();
        return Long.parseLong(value.toString());
    }
```
Or completly change the method as you suggested:

```java
public static Long toLong(Object value) {
   try {
      return Long.parseLong(value.toString());
   } catch (NumberFormatException e) {
      return null;
   }
}
```

What do you think?

src/main/java/apoc/convert/Convert.java
conker84: Se the comment before

src/main/java/apoc/convert/Convert.java
conker84: In this case I'll use the [toDouble](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.2/src/main/java/apoc/util/Util.java#L201) ok?

src/main/java/apoc/convert/Convert.java
conker84: See the comment before

src/main/java/apoc/convert/UnsupportedTypeException.java
conker84: Removed!

src/main/java/apoc/convert/Convert.java
jexp: Can you do a combination of 2 and 3 ?

jexp: Same for toFloat

jexp: Thanks!

jexp:
Looks good to me, thanks so much.
@conker84 did you already sign the Neo4j CLA? If not please do so: http://neo4j.com/developer/cla/

conker84:
It was a pleasure!
I just signed the CLA.

jexp:
Merged manually. Thanks again.
https://github.com/neo4j-contrib/neo4j-apoc-procedures/commit/2f3e124d82157d5f8bbadb621e6238587390693a

PR #533, opened 2017-08-02
Compatible values flag for non-packable values

albertodelazzari:
For all procedures that are related to document retrieval from MongoDB there is a new (boolean) parameter enabling/disabling automatic conversion in order to avoid exception for non-packable values

### DIFF
--- a/docs/overview.adoc
+++ b/docs/overview.adoc
@@ -323,14 +323,16 @@ config contains any of: `{skip:1,limit:5,header:false,sep:'TAB',ignore:['tmp'],a
 
 [cols="3m,2"]
 |===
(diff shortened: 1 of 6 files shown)


albertodelazzari:
@jexp yes, ObjectMapper is costly. I make it static.
I change build.gradle in order to put the dependencies inside the shadowJar and I modify the documentation.

docs/overview.adoc
```diff
@@ -350,7 +350,11 @@ Or you get them locally from your gradle build of apoc.
 
 ----
 gradle copyRuntimeLibs
+<<<<<<< HEAD
+cp lib/mongodb*.jar lib/bson*.jar $NEO4J_HOME/plugins/
+=======
 cp lib/mongodb*.jar lib/bson*.jar lib/jackson-*-2.9.0.jar $NEO4J_HOME/plugins/
```
jexp: jackson should be packaged now

PR #537, opened 2017-08-04
added apoc.load.ldap

kvegter:
when you want to use this on neo4j server you have to put an jldap.jar in the plugins folder. You can get that via this url : https://mvnrepository.com/artifact/com.novell.ldap/jldap/2009-10-07

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ ext {
 }
 
 repositories {
-     mavenLocal()
+//     mavenLocal()
      maven { url "https://m2.neo4j.org/content/repositories/snapshots" }
      mavenCentral()
      maven { url "http://oss.sonatype.org/content/repositories/snapshots/" }
@@ -71,6 +71,8 @@ dependencies {
     compile group: 'commons-codec', name: 'commons-codec', version:'1.9'
     compile 'com.jayway.jsonpath:json-path:2.2.0'
     compileOnly group: 'net.biville.florent', name: 'neo4j-sproc-compiler', version:'1.2'
+
+    compile 'com.novell.ldap:jldap:2009-10-07'
 
     testCompile group: 'junit', name: 'junit', version:'4.12'
     testCompile group: 'org.hamcrest', name: 'hamcrest-library', version:'1.3'
--- a/src/main/java/apoc/load/LDAPResult.java
+++ b/src/main/java/apoc/load/LDAPResult.java
@@ -1,0 +1,11 @@
+package apoc.load;
+
+import java.util.Map;
+
+public class LDAPResult {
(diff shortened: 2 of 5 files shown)


build.gradle
```diff
@@ -72,6 +72,8 @@ dependencies {
     compile 'com.jayway.jsonpath:json-path:2.2.0'
     compileOnly group: 'net.biville.florent', name: 'neo4j-sproc-compiler', version:'1.2'
 
+    compile 'com.novell.ldap:jldap:2009-10-07'
```
neo4j-oss-build: this should be compileOnly + testOnly imho,
otherwise it's packaged into the jar.

src/main/java/apoc/load/LDAPResult.java
```diff
@@ -0,0 +1,11 @@
+package apoc.load;
+
+import java.util.Map;
+
+public class LDAPResult {
```
neo4j-oss-build: can we use MapResult instead?

neo4j-oss-build: Or are there other useful columns that makes sense to add? like data provenance ones?

src/main/java/apoc/load/LoadLdap.java
```diff
@@ -0,0 +1,256 @@
...
+            Map<String, Object> config = new HashMap<>();
+            String[] sConf = ((String) value).split(" ");
+            config.put("ldapHost", sConf[0]);
+            config.put("loginDN", sConf[1]);
+            config.put("loginPW", sConf[2]);
+
+            return config;
+
+        } else {
+            return (Map<String,Object> ) conn;
```
neo4j-oss-build: check if it is really a map?

neo4j-oss-build: could we use a ldap:// URL  instead?

build.gradle
kvegter: I tried it with this and I still needed the ldap.jar in the plugins folder. And yes it is mentioned in the doc

src/main/java/apoc/load/LDAPResult.java
kvegter: I only did this to have the name 'entry' which is more reflecting the ldap name LDAPEntry. But technically we could also use MapResult. Then the examples have to be changed.

src/main/java/apoc/load/LoadLdap.java
kvegter: ldap:// url is not convenient it has also a kind of search syntax in it (so it is more than a connection alone) I only need the connection parameters, search we do via the 'searchMap'

PR #539, opened 2017-08-08
upgrade to gradle 4.1 and gradle plugin updates

sarmbruster:
NB: plugins were also updated to latest release. However I couldn't test nexus plugin.


jexp:
merged manually

PR #552, opened 2017-08-18
Gradle task for automatic version mapping between neo4j and apoc

albertodelazzari:
I introduced a new "versions" task for gradle that does the following steps:

1. retrieve all the tags related to the apoc project
2. for each tag retrieve the raw content of build.gradle
3. extract neo4j version from build.gradle
4. create a JSON containing a list of objects: neo4j version, apoc version, release url and download url
5. write the JSON to version.json

How do we want to handle the "sanity checks" for compatibility? Do we want to do something similar to the ETL project?


jexp:
@albertodelazzari 
Yep some rough sanity check would be good. Perhaps as an extra shell script to be run.

Issue #553, opened 2017-08-20
Cypher timestamp limited to current transaction execution


bradnussbaum:
I recently discovered that Cypher's timestamp function provides the same result for the duration of a transaction. In some cases I need to have System.currentTimeMillis() as the query executes. I have added support for apoc.date.timestamp to serve this purpose.

PR #554, opened 2017-08-20
Adding apoc.date.timestamp

bradnussbaum:


jexp:
What is the difference to the built-in timestamp()  function?

jexp:
Ah ok, read the issue https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/553 just now.

Can you rename it to currentTimestamp() ? so that it's more obvious.

And also add it to the docs for date in overview.adoc and date.adoc. Thanks!

bradnussbaum:
@jexp done and i added an entry to datetime.adoc and overview.adoc.

PR #570, opened 2017-08-26
Fixes #569 - Add Collection Function to update value at an index

### LINKED ISSUE #569: Add Collection Function to update value at an index
This is the required cypher to update a collection index position. There needs to be a simpler way to insert a value into a collection.
`WITH [0,1,2,3,4] AS nums
WITH nums[0..2] AS s1, nums[3..5] AS s3
RETURN s1 + [9] + s3;`

Should be something like....
apoc.coll.set(index, object)

bradnussbaum:


src/main/java/apoc/coll/Coll.java
```diff
@@ -127,6 +127,33 @@ public long indexOf(@Name("coll") List<Object> coll, @Name("value") Object value
     }
 
     @UserFunction
+    @Description("apoc.coll.set(coll, index, value) | set index to value")
```
jexp: should we be more lenient with  these functions

* on coll null -> return null
* on invalid indexes (<0 or > length == null ) ?  -> just return the original coll unchanged
* and then please test for these cases

* can you add a function like `insertList` that inserts a list ?


The only unfortunate drawback is that these functions won't keep the list-type, so one also will have to convert them if a specific type is needed.

jexp:
@bradnussbaum please see my comments.

src/main/java/apoc/coll/Coll.java
bradnussbaum: @jexp yes i should make these more lenient. your suggestions look good. also yes to tests. i'm hoping to get to this soon.

what do you mean these won't keep the list-type? the only type supported by cypher is list type. cypher doesn't support Set so i don't see why these wouldn't always be a List type.

jexp:
@bradnussbaum ping :)

bradnussbaum:
@jexp made some updates see if it looks good

src/main/java/apoc/coll/Coll.java
jexp: Yes but you might pass in a List<Node> but what these functions will return is a List<Object>, so you loose the compound type `Node`

src/main/java/apoc/coll/Coll.java
```diff
@@ -127,6 +127,53 @@ public long indexOf(@Name("coll") List<Object> coll, @Name("value") Object value
...
+    public List<Object> set(@Name("coll") List<Object> coll, @Name("index") long index, @Name("value") Object value) {
+        if (coll == null) return null;
+        if (index < 0 || value == null) return coll;
```
jexp: value == null is not invalid or?

src/main/java/apoc/coll/Coll.java
```diff
@@ -127,6 +127,53 @@ public long indexOf(@Name("coll") List<Object> coll, @Name("value") Object value
...
+        List<Object> list = new ArrayList<>(coll);
+        list.set( (int) index, value );
+        return list;
+    }
+
+    @UserFunction
+    @Description("apoc.coll.insert(coll, index, value) | insert value at index")
+    public List<Object> insert(@Name("coll") List<Object> coll, @Name("index") long index, @Name("value") Object value) {
+        if (coll == null) return null;
+        if (index < 0 || value == null) return coll;
```
jexp: value == null should be allowed

src/main/java/apoc/coll/Coll.java
```diff
@@ -127,6 +127,53 @@ public long indexOf(@Name("coll") List<Object> coll, @Name("value") Object value
...
+        if (index < 0 || values == null) return coll;
+
+        List<Object> list = new ArrayList<>(coll);
+        list.addAll( (int) index, values );
+        return list;
+    }
+
+    @UserFunction
+    @Description("apoc.coll.remove(coll, index, [length=1]) | remove range of values from index to length")
+    public List<Object> remove(@Name("coll") List<Object> coll, @Name("index") long index, @Name(value = "length") long length) {
```
jexp: misses `defaultValue="1"`

src/main/java/apoc/coll/Coll.java
```diff
@@ -127,6 +127,53 @@ public long indexOf(@Name("coll") List<Object> coll, @Name("value") Object value
...
+        if (coll == null) return null;
+        if (index < 0 || length < 0) return coll;
+
+        List<Object> list = new ArrayList<>(coll);
+        for (int i = 0; i < length; i++)
+        {
+            list.remove( (int) index );
```
jexp: we should control that we don't exceed the length of the list? Or is that silently ignored?

jexp:
@bradnussbaum ping :)

jexp:
@bradnussbaum ping :) let me know if you have time, otherwise I try to get it updated and in.

src/main/java/apoc/coll/Coll.java
bradnussbaum: @jexp I can remove this. Usually we see null in collections are bad but it's true that java does allow this so we can as well.

src/main/java/apoc/coll/Coll.java
bradnussbaum: @jexp I can remove this. Usually we see null in collections are bad but it's true that java does allow this so we can as well.

src/main/java/apoc/coll/Coll.java
bradnussbaum: @jexp I will add this.

src/main/java/apoc/coll/Coll.java
bradnussbaum: @jexp Don't you want users to get back raw error? Right now index out of bounds would just surface to the caller. Should we really silence that?

InverseFalcon: Several Cypher operations on lists which exceed the index bounds are allowed, and don't throw exceptions. Seems like we should probably avoid this if possible.

Can we do a `length = Math.min(length, coll.size() - 1)` to prevent the exception?

bradnussbaum: @InverseFalcon yeah i think that's reasonable if that's a standard way of handling out of bounds by user. most these decisions come down to uniform design being the best way. if cypher has a standard already I think we just follow that here

jexp:
ping @bradnussbaum @bennussbaum ?

jexp:
added manually

PR #572, opened 2017-08-27
Xmlimport

kuczera:
NEXT-Edges should be renamed to NEXT-TAG and the new NEXT-Edges should connect only the XmlWord-Nodes.


sarmbruster:
will be dealt with in my forked repo

PR #581, opened 2017-09-05
Fixes #489 - Add support for loading CSVs formatted according the import tool's specification

### LINKED ISSUE #489: helper procedures to use import-tool headers for node/relationship creation
the neo4j-import tool headers already contain information about label, id-property, property-types, skipping etc.

It would be great to load these csv files with load csv and use the header information internally to set labels/rel-types, convert types for properties or skip columns

`apoc.load.csv` should also support that

```
load csv from "url" as line
call apoc.import.create/merge/find/update - Node/Relationship(line, n / r) yield node/relationship
```

* support :LABEL and :TYPE columns
* skip :IGNORE columns
* do type conversion like :INTEGER or :BOOLEAN or arrays
* set (Label) group information as label
* use :ID properties and :START_ID , :END_ID for lookups / merge
* make sure there are constraints/indexes for ID properties (fail if there are none)

From @szarnyasg:

```$ cat comments.csv
:ID(Comment)|id:LONG|creationDate:STRING|locationIP:STRING|browserUsed:STRING|content:STRING|length:INT
274877907725|274877907725|2010-10-17T05:34:03.139+0000|88.80.236.34|Chrome|yes|3

$ bin/neo4j-import --into $DB_DIR --nodes:Message:Comment comment.csv --delimiter '|'
...
```

Is there a simple way to load this from Cypher? the best approach I can think of is to use

```LOAD CSV WITH HEADERS FROM ... AS line FIELDTERMINATOR `|`
```

and instantiate nodes like

```CREATE (:Comment {locationIP: line['locationIP:STRING'], ...})```

szarnyasg:
Fixes #489. Submitting this for review. Not ready to merge.


jexp:
@szarnyasg your branch seems to be messed up
can you do a clean rebase of your changes ? 
and perhaps squash your commits?

Thanks a lot

szarnyasg:
Ok - fixed

szarnyasg:
@jexp I finally got back to this. I revamped the code a bit. I have one fundamental question left: how should we pass the URI of the CSV file to the procedure? Let me elaborate on that one a bit.

The procedure needs to use the filename at least twice:

1. look into it to read the CSV header @ https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/581/files#diff-df1b88452931fbee58e22a3f1d27b3a8R181
2. pass to LOAD CSV, which will use it to open and read the file @ https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/581/files#diff-df1b88452931fbee58e22a3f1d27b3a8R193

The file URI has to be relative to the specified import directory ([related blog post](https://neo4j.com/developer/guide-import-csv/)).

There is a surprising contradiction in this:
- if we pass a relative URI, step #1 will not work - to determine the location of the file, we'd have to know the value of the `dbms.directories.import` directory
- if we pass an absolute URI, step #2 will not work - we'd have to produce an URI that is relative the `dbms.directories.import` directory

So in both cases, the `dbms.directories.import` directory is required. Is there a way for APOC procedures to access that directory?

szarnyasg:
It seems that some Heisenbugs occur during APOC tests - I committed a single whitespace and it the build went from red to green.

szarnyasg:
Hey @jexp, please review it if you get a chance.

jexp:
Hi @szarnyasg will have another look.
Do you think we can simplify the sheer amount of files this PR adds?
Perhpas generate the csvs, e.g. from a large one that's split on a specifc line-marker
Or zip them and extract them on demand?

jexp:
I took a quick look.

Some thoughts.

1. did we discuss back then to forego cypher altogether and create the data with the Java API possibly in parallel?
2. ID type can also be string depending on config option
3. if we use cypher, all statements should use parameters not string interpolation (except where needed for rel-types)
4. as we're in apoc we could also use the apoc calls for creating dynamic nodes and rels but then see 1.
4. add potential cache for node-lookups (see maxdemarzi)
5. the csv files should be generated by the tests, then they are also close together and don't have to be checked in

jexp:
I'm happy to take your work and refactor it quickly next week.

szarnyasg:
Hey @jexp, thanks for the feedback.

> 1. did we discuss back then to forego cypher altogether and create the data with the Java API possibly in parallel?
We discussed this, but I thought it'd be a lot easier to implement one on top of Cypher. It turns out that this solution has dismal performance. I once tested it with a data set of ~10.000 records and it took ~5 mins.

> 2. ID type can also be string depending on config option
Correct, this is missing.

> 3. if we use cypher, all statements should use parameters not string interpolation (except where needed for rel-types)
Good point, [CSV injection](http://georgemauer.net/2017/10/07/csv-injection.html) is actually a thing.

> 6. the csv files should be generated by the tests, then they are also close together and don't have to be checked in
Agreed.

> I'm happy to take your work and refactor it quickly next week.
It would be great if you could refactor the code, esp. with the performance-related aspects.

szarnyasg:
I took another look at this problem.

Historically, my approach was to generate `LOAD CSV` commands. I got that working by the end of last year, but it turned out to be extremely slow, taking ~10 minutes for a small LDBC dataset with ~50k nodes/relationships.

So the next step was to ditch `LOAD CSV`, and instead, use a custom CSV loader and create the entities in the graph manually with `createNode` and `createRelationship`. While doing this, I realized that it no longer make sense to use separate commands for each node/relationship file. Instead, the schema of the procedure is now `apoc.import.csv(nodes, relationships, config`.

Both `nodes` and `relationships` are lists:
* `nodes` contains `{fileName: ..., labels: [...]}` maps,
* `relationships` contains `{fileName: ..., type: ...`} maps.

While this seems a bit cumbersome at first sight, it makes things a lot more easier in the long run: during the first phase of the loading process - loading nodes - we can easily cache the mapping between CSV ids and internal node ids. We can then use these for lookups based on the start id end id of relationships. The resulting code is a *lot* faster, taking only a few seconds for the LDBC dataset that previously took minutes.

As the code is now using APOC's `LoadCsv` class and its mapping, I made a couple of its inner classes public: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/581/commits/4818de88532aa0ece53b9d3b880b66fb897e8de0#diff-f37b66be7dcb534530f538cc87e2ea15

There are a few minor things to be do, e.g. I haven't yet addressed your earlier comment:
> 6. the csv files should be generated by the tests, then they are also close together and don't have to be checked in

szarnyasg:
> 6. the csv files should be generated by the tests, then they are also close together and don't have to be checked in
CSVs are now generated by the tests on-the-fly in the `csv-inputs/` directory and the content of the directory is in gitignore.

szarnyasg:
I addressed one of my earlier comments:

> Also, the whole testing needs to be more thorough, e.g. we should assert on property values of the loaded entities. (Instead of just counting nodes/relationships which only catches the most basic errors.)

I now added tests for each unit tests. I believe the code is good to go for reviews.

jexp:
Cool, thanks so much !! Sorry for the delay.

szarnyasg:
In the meantime, I updated the docs in #844.

I also found an important design decision that's worth discussing. Currently, the whole loading process happens in a single transaction. This implies that the data set size is limited to what a single transaction can handle, i.e. ~20k nodes. However, if we were to split the load process into multiple transactions, then handling failures would become very difficult: if a transaction fails, we would need to roll back previous transactions, otherwise, we have inconsistent data.

I think an acceptable workaround would be to introduce a "force transactions" flag that says "I have a lot of data to load but I know what I am doing and I understand that incorrect CSVs might corrupt my database".

(Of course, this whole problem does not exist for the offline import tool, which can simply crash if it encounters too many errors.)

jexp:
I think we should do the same as we do in other tools, provide config options for batches and parallel execution so the user can choose to make their batch-size large enough to fit all data.

szarnyasg:
I introduced batch transactions in a separate commit. I based it on the `ExportGraphML` class, but I am not very experienced with this part of the APOC API - please let me know if my code needs any fixes.

For parallel execution, we'll need to parallelize the loading processes for the nodes and the relationships separately. Of course, within both phases, individual CSV files can be split into chunks and parallelized.
In any case, it doesn't seem trivial to implement this correctly and I could not find a piece of APOC code to base this on. Any suggestions? Or maybe leave this for a future release along with other improvements?

jexp:
We should use the same batching as we use in periodic.iterate
Export Functions have a different kind of batches.

Let's leave it for a future PR. Thanks so much for your work, I'll merge this and your docs update now.

PR #585, opened 2017-09-08
Fixes #584 - Allow asterisk as node label filter

### LINKED ISSUE #584: allow all-label filter in grouping
Currently, users need to specify explicit labels, e.g. `[Person],[gender]`. However, it might be useful to group all nodes on specific keys. For that, we need to introduce the asterisk in the label filter array.
This also leads to the situation that a node might not have a specific property key and is thus assigned to a `null` group. The resulting virtual node will have the grouping key(s) assigned to `null` in that case.

s1ck:
* also introduces key:null properties for virtual nodes if the property is not present at the group members

### DIFF
--- a/src/main/java/apoc/nodes/Grouping.java
+++ b/src/main/java/apoc/nodes/Grouping.java
@@ -30,51 +30,28 @@ import static java.util.Collections.singletonList;
 public class Grouping {
 
     private static final int BATCHSIZE = 10000;
+
+    private static final String ASTERISK = "*";
+
     @Context
     public GraphDatabaseAPI db;
     @Context
     public Log log;
 
-    static class Key {
-        private final int hash;
-        private final String label;
-        private final Map<String,Object> values;
-
-        public Key(String label, Map<String, Object> values) {
-            this.label = label;
-            this.values = values;
-            hash = 31 * label.hashCode() + values.hashCode();
-        }
-
-        @Override
-        public boolean equals(Object o) {
-            if (this == o) return true;
-            if (o == null || getClass() != o.getClass()) return false;
-
-            Key key = (Key) o;
-            return label.equals(key.label) && values.equals(key.values);
-        }
-
-        @Override
-        public int hashCode() {
-            return hash;
-        }
-    }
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/nodes/Grouping.java
```diff
@@ -83,27 +60,32 @@ public int hashCode() {
             Label label = Label.label(labelName);
             Label[] singleLabel = {label};
 
-            try (ResourceIterator<Node> nodes = db.findNodes(label)) {
+            try (ResourceIterator<Node> nodes = (labelName.equals(ASTERISK)) ? db.getAllNodes().iterator() : db.findNodes(label)) {
```
jexp: Wouldn't it be better to have the full scan on empty-list/null ?

src/main/java/apoc/nodes/Grouping.java
```diff
@@ -232,29 +214,98 @@ public void aggregate(PropertyContainer pc, Map<String, List<String>> aggregatio
...
-                result.add(new Key(label.name(), props));
+    /**
+     * Creates a grouping key for the given node using its label and grouping properties.
+     *
+     * @param node  node
+     * @param label node label
+     * @param keys  property keys
+     * @return grouping key
+     */
+    private NodeKey keyFor(Node node, String label, String[] keys) {
```
jexp: static method on NodeKey

src/main/java/apoc/nodes/Grouping.java
```diff
@@ -232,29 +214,98 @@ public void aggregate(PropertyContainer pc, Map<String, List<String>> aggregatio
...
+    /**
+     * Creates a grouping key for each specified label.
+     *
+     * @param node   node
+     * @param labels node labels
+     * @param keys   property keys
+     * @return grouping keys
+     */
+    private Collection<NodeKey> keysFor(Node node, List<String> labels, String[] keys) {
+        Map<String, Object> props = propertiesFor(node, keys);
```
jexp: static method on node key

PR #615, opened 2017-09-29
Add apoc.cypher.runFirstColumn function

johnymontana:
Executes a Cypher statement and returns first column only,
optionally collected into an array. Used by neo4j-graphql-js


jexp:
Added manually, thanks

Issue #593, opened 2017-09-14
add apoc.convert.toInteger function that converts any object into a whole number


jexp:
* boolean : true=1, false = 0
* string -> parseLong
* Number -> longValue
* other -> parseLong(toString())
* perhaps support hex too
* possibly remove non digit characters
* if we were funny we could even support some basic calculations in strings :)

same for toFloat

dodie:
Hi @jexp ! I'd work on this feature if it ok for you!

zygm0nt:
There's a commit update in connected PR.

albertodelazzari:
Can we close also this issue?

PR #621, opened 2017-10-03
add apoc.convert.toInteger, apoc.convert.toDouble functions - #593

zygm0nt:


src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,85 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+        if (obj == null || obj.equals("")) {
+            return null;
+        }
+
+        Types varType = Types.of(obj);
+        Number result;
+        switch (varType) {
+            case INTEGER:
+            case FLOAT:
+                result = ((Number) obj).intValue();
```
jexp: Cypher only knows long (which it calls Integer), so we should convert it to Long (longValue())

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,85 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
 	public List<Relationship> toRelationshipList(@Name("list") Object list) {
         return convertToList(list, Relationship.class);
 	}
-	
+
+    @UserFunction
+    @Description("apoc.convert.toInteger(value) | tries it's best to convert the value to an integer")
+    public Number toInteger(@Name("object") Object obj) {
```
jexp: Should return Long

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,85 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+            case BOOLEAN:
+                result = ((boolean) obj) ? 1 : 0;
+                break;
+            default:
+                throw new RuntimeException("Supported types are: Integer, Float, String, Boolean, HexString");
+        }
+        return result;
+    }
+
+    private Integer parseIntString(String input) {
```
jexp: Should return long

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,85 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+        if (input.equals("true")) {
+            return 1;
+        } else if (input.equals("false")) {
+            return 0;
+        } else if (input.startsWith("0x")) {
+            return Integer.valueOf(input.substring(2), 16);
```
jexp: Long.valueOf()

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,85 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+        } else {
+            return (int) Float.parseFloat(input);
```
jexp: Shouldn't we have separate parsers for Long.parseLong() and (long)Double.parseDouble() based on existence of decimal point?

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,85 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+    public Double toDouble(@Name("object") Object obj) {
+        if (obj == null || obj.equals("")) {
+            return null;
+        }
+
+        Types varType = Types.of(obj);
+        Double result;
+        switch (varType) {
+            case INTEGER:
+                result = ((Number) obj).doubleValue();
```
jexp: just return without break

src/main/java/apoc/convert/Convert.java
jexp: just return without break

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,85 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+            default:
+                throw new RuntimeException("Supported types are: Integer, Float, String, Boolean, HexString");
+        }
+        return result;
+    }
+
+    private Integer parseIntString(String input) {
+        if (input.equals("true")) {
+            return 1;
+        } else if (input.equals("false")) {
```
jexp: no need for else if you return before.

src/main/java/apoc/convert/Convert.java
zygm0nt: There's a test for that:

'''
testCall(db, "return apoc.convert.toInteger('123') as value", r -> assertEquals(123L, r.get("value")));
        testCall(db, "return apoc.convert.toInteger('123.15') as value", r -> assertEquals(123L, r.get("value")));
'''

and it seems it's not necessary to differentiate based decimal point.

zygm0nt:
@jexp fixed code according to your comments.

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,76 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+        switch (varType) {
+            case INTEGER:
+            case FLOAT:
+                return ((Number) obj).longValue();
+            case STRING:
+                return parseLongString((String)obj);
+            case BOOLEAN:
+                return ((boolean) obj) ? 1L : 0L;
+            default:
+                throw new RuntimeException("Supported types are: Integer, Float, String, Boolean, HexString");
```
jexp: return null

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,76 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+        }
+
+        Types varType = Types.of(obj);
+        switch (varType) {
+            case INTEGER:
+                return ((Number) obj).doubleValue();
+            case STRING:
+                return parseDoubleString((String)obj);
+            case FLOAT:
+                return (Double) obj;
```
jexp: ((Number)).doubleValue()

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,76 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+            case BOOLEAN:
+                return ((boolean) obj) ? 1.0 : 0.0;
+            default:
+                throw new RuntimeException("Supported types are: Integer, Float, String, Boolean, HexString");
```
jexp: return null

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,76 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+        }
+    }
+
+
+    private Double parseDoubleString(String input) {
+        if (input.equals("true")) {
```
jexp: equalsIgnoreCase

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,76 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+            return 1.0;
+        }
+        if (input.equals("false")) {
+            return 0.0;
+        }
+        if (input.startsWith("0x")) {
+            Long i = Long.parseLong(input.substring(2), 16);
+            return Double.longBitsToDouble(i);
+        }
+        return Double.parseDouble(input);
```
jexp: try catch else return null

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,76 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
+
+    private Double parseDoubleString(String input) {
+        if (input.equals("true")) {
+            return 1.0;
+        }
+        if (input.equals("false")) {
+            return 0.0;
+        }
+        if (input.startsWith("0x")) {
+            Long i = Long.parseLong(input.substring(2), 16);
```
jexp: try catch

src/main/java/apoc/convert/Convert.java
```diff
@@ -163,5 +165,76 @@ else if (list.getClass().isArray() && !list.getClass().getComponentType().isPrim
...
-	
+
+    @UserFunction
+    @Description("apoc.convert.toInteger(value) | tries it's best to convert the value to an integer")
+    public Long toInteger(@Name("object") Object obj) {
+        if (obj == null || obj.equals("")) {
+            return null;
+        }
+
+        Types varType = Types.of(obj);
```
jexp: you can also add teh try-catch here, and return null in case of exception

jexp:
ping @zygm0nt

zygm0nt:
I've updated this PR. Please review @jexp

src/main/java/apoc/convert/Convert.java
zygm0nt: This seems pointless - Types.of() implementation returns UNKNOWN type as a fallback. It does not throw any exceptions AFAIK.

jexp:
@zygm0nt there are some conflicts now, would you mind rebasing?

jexp:
@zygm0nt ping?

jexp:
merged manually

Issue #586, opened 2017-09-08
 apoc.coll.max([0.5,1,2.3]) creates a cast exception


gpierrick:
The following creates a casting exception because 1 is considered a long and not a double
`
Failed to invoke function 'apoc.coll.max': Caused by: java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Long
`

```
apoc.coll.max([0.5,1,2.3]) 
```
Example taken from
```
call apoc.help('apoc')
```

InverseFalcon:
Just to note, this is a consequence of using Collections.max().

From the javadocs:

> Furthermore, all elements in the collection must be mutually comparable (that is, e1.compareTo(e2) must not throw a ClassCastException for any elements e1 and e2 in the collection).

jexp:
We could use a custom comparator, that uses Number and toLong / toDouble depending on the number types. Or turn them all into BigDecimals, but that might be a bit overkill.

dodie:
I've submitted a PR to fix this issue and the same problem in ```apoc.coll.min```.
The changed code uses a custom comparator that decides that compares two Objects as numbers.
Because there are many Number implementations, I ended up using BigDecimals:

- The code checks if the parameters are -Infinity, Infinity or NaN, and handles them as the original code would.
- If not, it converts the given Objects to BigDecimals and performs a comparison.

PR #630, opened 2017-10-08
Fix class cast exception apoc.coll.max/min #586

dodie:


src/main/java/apoc/coll/Coll.java
```diff
@@ -74,13 +75,62 @@ public double avg(@Name("numbers") List<Number> list) {
...
+            return 0;
+        }
+        if (isNegativeInfinity(a)) {
+            return -1;
+        }
+        if (isNegativeInfinity(b)) {
+            return 1;
+        }
+
+        return (new BigDecimal((a.toString())).compareTo(new BigDecimal(b.toString())));
```
jexp: I feel this is quite expensive? WDYT? Could we just cast them to Number and compare the double-values?

dodie: I'll create a small performance tests to compare both approaches and check back.
Comparing based on the double value of the numbers might introduce some inaccuracy for larger numbers, but most likely this will be the fastest solution.

Maybe we could use the double-value for floating point numbers, and long-value for non-floating point numbers to avoid inaccuracy and BigDecimals as well.

jexp: Sounds good
What I'm also wondering is if Double.nan and infinity has some builtin comparison logic?

dodie: The Double.compare static method handles it. Tricky part will be to compare doubles to longs and handle these special cases as well.

dodie: I created a small benchmark to compare the different approaches, you can check it out here: https://github.com/dodie/generic-number-comparator-benchmark


| Benchmark                          | Score (ops/s) - Comparing Integers only    | Score (ops/s) - Comparing Integers and Doubles |
| ---------------------------------- | -------------------------------- | ------------------------------------ |
| Original comparator                | 3172804.081                      | X                                    |
| [Long.compare](https://github.com/dodie/generic-number-comparator-benchmark/blob/master/src/main/java/hu/awm/test/comparison/LongValueBasedComparison.java)                       | 3480229.820                      | X                                    |
| [Double.compare](https://github.com/dodie/generic-number-comparator-benchmark/blob/master/src/main/java/hu/awm/test/comparison/DoubleValueBasedComparison.java)                     | 2756652.584                      | 2340389.827                          |
| [Double.compare / Long.compare](https://github.com/dodie/generic-number-comparator-benchmark/blob/master/src/main/java/hu/awm/test/comparison/LongAndDoubleBasedComparison.java)      | 2881121.264                      | 1899611.262                          |
| [BigDecimal.compare + Inf/NaN check](https://github.com/dodie/generic-number-comparator-benchmark/blob/master/src/main/java/hu/awm/test/comparison/BigDecimalBasedComparison.java) |  104337.061                      |   74441.372                          |


I put the original and the Long.compare based versions in the measurement as a baseline.

You were right, the BigDecimal based version is much slower than the other solutions.
The fastest comparator that can handle floating-point numbers is the Double.compare, but promoting longs to doubles might result in loss of precision. From the [JLS](https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.1.2):


>A widening conversion of an int or a long value to float, or of a long value to double, may result in loss of precision - that is, the result may lose some of the least significant bits of the value. In this case, the resulting floating-point value will be a correctly rounded version of the integer value, using IEEE 754 round-to-nearest mode (§4.2.4).


[This](https://github.com/dodie/generic-number-comparator-benchmark/blob/master/src/main/java/hu/awm/test/comparison/LongAndDoubleBasedComparison.java) comparator is not much slower than ```Double.compare```, and it only uses the ```doubleValue``` if both numbers are floating-point values. Otherwise, it compares them as Longs.

I think it might be a good compromise between speed and accuracy.
What do you think, which version should we use?

dodie: One more thing to consider is complexity: using Double.compare is much simpler than the alternatives.

dodie:
I've changed the comparison logic to cast the parameters to Numbers and take their doubleValues. This seems to be the fastest and simplest solution, although it might result in some inaccuracy for really big numbers. Please review it and let me know if it looks ok.

dodie:
Sorry about the failing CI check, I ran the tests on my machine there were no failed tests.

jexp:
Thanks a lot!

jexp:
Thanks a lot for all your effort, to cleanly figure it out.

PR #631, opened 2017-10-10
Instructions for building against local neo4j

fickludd:


sarmbruster:
FYI: there's already a short adoc regarding specfics of apoc build: https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.2/docs/gradle.adoc
Maybe that should be added there? @jexp wdyt?

jexp:
added manually to gradle.adoc

PR #632, opened 2017-10-11
Fixes #610 - JDBC Type Conversation Issues from Oracle

### LINKED ISSUE #610: JDBC Type Conversation Issues from Oracle
1. Apoc is not handling Oracle Timestamps
~~2. Apoc is not handling Oracle Number (converts to String)~~
~~3. Apoc is not handling correct character encoding~~

AngeloBusato:


build.gradle
```diff
@@ -43,10 +43,11 @@ ext {
...
 repositories {
-     maven { url "https://m2.neo4j.org/content/repositories/snapshots" }
-     mavenCentral()
-     maven { url "https://m2.neo4j.org/content/repositories/snapshots" }
-     maven { url "http://oss.sonatype.org/content/repositories/snapshots/" }
+    maven { url "https://m2.neo4j.org/content/repositories/snapshots" }
+    mavenCentral()
+    maven { url "https://m2.neo4j.org/content/repositories/snapshots" }
+    maven { url "http://oss.sonatype.org/content/repositories/snapshots/" }
+    flatDir { dirs 'lib' }
```
jexp: How will that run on other machines (also Travis?)

build.gradle
```diff
@@ -78,6 +79,8 @@ dependencies {
     testCompile group: 'org.neo4j', name: 'neo4j-causal-clustering', version:neo4jVersionEffective, classifier: "tests"
     testCompile group: 'org.neo4j', name: 'neo4j-kernel', version:neo4jVersionEffective, classifier: "tests"
     testCompile group: 'org.neo4j', name: 'neo4j-io', version:neo4jVersionEffective, classifier: "tests"
+    testCompile group: 'com.oracle', name: 'ojdbc7', version: '12.1.0.2'
+    testCompile group: 'postgresql', name: 'postgresql', version: '9.4.1208-jdbc42-atlassian-hosted'
```
jexp: what is this? '9.4.1208-jdbc42-atlassian-hosted'
there is a maven central driver for postgresql?

AngeloBusato: yes there is, sorry,  we change it

build.gradle
AngeloBusato: in the "lib" folder there is the Oracle jar for tests.
we put it there for convenience, otherwise we can remove it but if someone starts the local test with oracle it must have it in his pc.

src/main/java/apoc/load/Jdbc.java
```diff
@@ -169,36 +170,66 @@ public boolean hasNext() {
...
+            try {
+                return new Scanner(is).next();
+            } catch (Exception e) {
+                return "";
+            }
+            finally {
+                try {
+                    is.close();
+                } catch(Exception e) {
+                    throw new RuntimeException(e.getMessage());
```
jexp: don't throw just ignore

src/main/java/apoc/load/Jdbc.java
```diff
@@ -169,36 +170,66 @@ public boolean hasNext() {
...
+                byte[] bytes = (byte[]) value;
+                value = new String(bytes);
             }
+
             return value;
         }
 
+        private Object getValue(InputStream is){
+            try {
+                return new Scanner(is).next();
```
jexp: Scanner is usually used for text only streams. And even then you have to set `\Z` as delimiter for EOF.
Probably better to use one of the included (apache) file-utils to read the full file.

src/main/java/apoc/load/Jdbc.java
```diff
@@ -169,36 +170,66 @@ public boolean hasNext() {
...
             } catch (SQLException e) {
                 throw new RuntimeException("Cannot execute read result-set.", e);
             }
         }
 
-        private Object convert(Object value) {
-            if (value instanceof UUID || value instanceof BigInteger || value instanceof BigDecimal) {
+        private Object convert(int col,int type) throws SQLException {
+            Object value = rs.getObject(col);
+            if(type == Types.TIMESTAMP ||type == Types.DATE){
```
jexp: a switch might be easier :)

src/main/java/apoc/load/Jdbc.java
```diff
@@ -169,36 +170,66 @@ public boolean hasNext() {
...
+                Blob blob = ((Blob) value);
+                return getValue(blob.getBinaryStream());
+            }
+            if (type == Types.CLOB) {
+                Clob clob = ((Clob) value);
+                return getValue(clob.getAsciiStream());
+            }
+            if (type == Types.BINARY) {
+                byte[] bytes = (byte[]) value;
+                value = new String(bytes);
```
jexp: Cypher supports byte[] actually.

src/main/java/apoc/load/Jdbc.java
```diff
@@ -169,36 +170,66 @@ public boolean hasNext() {
...
+                return rs.getTimestamp(col).getTime();
+            }
+            if (type == Types.OTHER || type == Types.BIGINT || type == Types.DECIMAL) {
                 return value.toString();
             }
-            if (value instanceof java.util.Date) {
-                return ((java.util.Date) value).getTime();
+            if (type == Types.BLOB) {
+                Blob blob = ((Blob) value);
+                return getValue(blob.getBinaryStream());
```
jexp: byte-array

jexp:
Looks good only a few small changes.

jexp:
@AngeloBusato ping

src/main/java/apoc/load/Jdbc.java
AngeloBusato: @jexp I would replace the scanner with IOUtils.toByteArray(is).

AngeloBusato:
@jexp  PR updated

PR #635, opened 2017-10-13
3.2 626

dodie:
Change array pattern to be handled as a simple separator character rather than a regular expression.

### DIFF
--- a/src/main/java/apoc/load/LoadCsv.java
+++ b/src/main/java/apoc/load/LoadCsv.java
@@ -90,7 +90,7 @@ public class LoadCsv {
         }
 
         private Object convertArray(String value) {
-            String[] values = value.split(arrayPattern);
+            String[] values = value.split(Pattern.quote(arrayPattern));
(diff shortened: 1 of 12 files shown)


src/main/java/apoc/load/LoadCsv.java
```diff
@@ -90,7 +90,7 @@ public Object convert(String value) {
         }
 
         private Object convertArray(String value) {
-            String[] values = value.split(arrayPattern);
+            String[] values = value.split(Pattern.quote(arrayPattern));
```
jexp: pattern.split(value)

jexp:
Better rebase the 3.2 branch than MERGE

src/main/java/apoc/load/LoadCsv.java
dodie: Ok, I change that.

dodie:
Ok, I create a new PR with clean history.

PR #636, opened 2017-10-13
Enable pipe as CSV arraySep character, fix #626

### LINKED ISSUE #626: Pipe | cannot be used as array separator in apoc.load.csv procedure
In my CSV file, I have a column PumMedIDs with ids separated by '|'
I tried following import statement:
```
CALL apoc.load.csv(
    'file:///file.csv',
    {arraySep:'|',mapping:{PubMedIDs:{array:true,arraySep:'|'}}})
YIELD lineNo, list, map
RETURN lineNo, map.PubMedIDs, limit 20
```
Unfortunately, the pubMedIDs get separated by every character and not by the separator.
By using `\|` as separator I get:
```
Invalid input '|': expected '\', ''', '"', 'b', 'f', 'n', 'r', 't', UTF16 or UTF32 (line 3, column 17 (offset: 109))
"    {arraySep:'\|',mapping:{PubMedIDs:{array:true,arraySep:'\|'}}})"
```
and using `\\|` results in
```
Failed to call procedure `apoc.load.csv(url :: STRING?, config :: MAP?) :: (lineNo :: INTEGER?, list :: LIST? OF ANY?, map :: MAP?)`: Unexpected internal error near index 1
\
```
The split function works just fine with pipes |
So, I just added the PubMedIDs as it is, and splitted the property afterwards.

The CSV I have used can be found here:
[http://ctdbase.org/downloads/#top](URL)

dodie:
Change array pattern to be handled as a simple separator character rather than a regular expression

### DIFF
--- a/src/main/java/apoc/load/LoadCsv.java
+++ b/src/main/java/apoc/load/LoadCsv.java
@@ -73,7 +73,7 @@ public class LoadCsv {
         final boolean array;
         final boolean ignore;
         final char arraySep;
-        private final String arrayPattern;
+        private final Pattern arrayPattern;
 
         public Mapping(String name, Map<String, Object> mapping, char arraySep, boolean ignore) {
             this.name = mapping.getOrDefault("name", name).toString();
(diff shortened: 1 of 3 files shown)


jexp:
Thanks so much !!

PR #637, opened 2017-10-20
608 | add roman <-> arabic conversions

zygm0nt:


src/main/java/apoc/number/ArabicRoman.java
```diff
@@ -0,0 +1,62 @@
...
+        Number number = validateNumberParam(value);
+        if (number == null) return null;
+        return getRoman(number.intValue());
+    }
+
+    private Number validateNumberParam(Object number) {
+        return number instanceof Number ? (Number) number : null;
+    }
+
+    private static int toArabic(String number) {
```
jexp: Wouldn't it make more sense to move all the static methods into the Enum as instance methods where possible ?
And have them be self-aware? I.e. if number.startsWith(this.name()) return ....
etc.

src/main/java/apoc/number/ArabicRoman.java
```diff
@@ -0,0 +1,62 @@
...
+        if (number.startsWith("X")) return 10 + toArabic(number.substring(1));
+        if (number.startsWith("IX")) return 9 + toArabic(number.substring(2));
+        if (number.startsWith("V")) return 5 + toArabic(number.substring(1));
+        if (number.startsWith("IV")) return 4 + toArabic(number.substring(2));
+        if (number.startsWith("I")) return 1 + toArabic(number.substring(1));
+        return 0;
+    }
+
+    public  String getRoman(int number) {
+        String roman[] = {"M","XM","CM","D","XD","CD","C","XC","L","XL","X","IX","V","IV","I"};
```
jexp: Should use the enum and probably be part of the enum.

jexp:
@zygm0nt ping

jexp:
@zygm0nt ping?

jexp:
Added manually, thanks a lot.

Issue #641, opened 2017-10-26
apoc.schema.properties methods do not support `number` Lucene terms


lszumel:
It seems only `string` terms are supported. The methods skip over any indexes for `number` terms.

e.g. in the test dataset you cannot count distinct values of `:Person(age)`

vga91:
Fixed by https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1060

Issue #640, opened 2017-10-26
apoc.schema.properties.distinctCount does not support empty label


lszumel:
The documentation and implementation imply that this method supports empty label and key. In the case of an empty label, all indexes should be searched. However, the actual implementation always tries to load the index using `label` and finds nothing if `label.isEmpty()`.

e.g. the following query will return nothing in the test dataset:
```
CALL apoc.schema.properties.distinctCount("", "bar") YIELD label,key,value,count RETURN * ORDER BY value
```

PR #642, opened 2017-10-26
Support null labels, multiple labels in apoc.schema.properties.distinctCount

### LINKED ISSUE #639: apoc.schema.properties.distinctCount closes transaction after first index examined
This procedure is closing the transaction in two cases: in the `catch` and `finally` blocks. I can't understand why either is necessary--surely the execution engine will close the transaction on exception? 

The close in the `finally` causes a failure if the label has more than one index. After the first index is looked at, the following exception is thrown.
```

org.neo4j.graphdb.QueryExecutionException: Failed to call procedure `apoc.schema.properties.distinctCount(label =  :: STRING?, key =  :: STRING?) :: (label :: STRING?, key :: STRING?, value :: STRING?, count :: INTEGER?)`: The requested operation cannot be performed, because it has to be performed in a transaction. Ensure you are wrapping your operation in the appropriate transaction boilerplate and try again.

	at org.neo4j.kernel.impl.query.QueryExecutionKernelException.asUserException(QueryExecutionKernelException.java:35)
	at org.neo4j.kernel.impl.factory.ClassicCoreSPI.executeQuery(ClassicCoreSPI.java:82)
	at org.neo4j.kernel.impl.factory.GraphDatabaseFacade.execute(GraphDatabaseFacade.java:430)
	at org.neo4j.kernel.impl.factory.GraphDatabaseFacade.execute(GraphDatabaseFacade.java:413)
	at apoc.util.TestUtil.testResult(TestUtil.java:86)
	at apoc.index.SchemaIndexTest.testDistinctCountPropertiesOnFirstIndex(SchemaIndexTest.java:168)
```

e.g. the following query will throw the above exception:
```
CALL apoc.schema.properties.distinctCount({label}, {key}) YIELD label,key,value,count RETURN * ORDER BY value
```

lszumel:
This PR should fix #639 and #640 

I removed the `finally` block that was closing the transaction to fix #639.

For #640, when `labelName.isEmpty()` all indexes are searched (rather than none). This exposed another bug, which is that we would count distinct terms on all indexes for the label, whether or not they contained the key under consideration. So there are two small changes to the logic in `@Procedure("apoc.schema.properties.distinctCount")`

I added test cases to check the behavior with:
1. Labels with more than one index
2. labelName.isEmpty()
3. keyName.isEmpty()
4. both (2) and (3)

In doing so I noticed that only `string` Lucene terms work with these distinct property APOC methods. I'm not sure how to fix that so I just created issue #641 and commented out the test lines that were failing.

### DIFF
--- a/src/main/java/apoc/index/SchemaIndex.java
+++ b/src/main/java/apoc/index/SchemaIndex.java
@@ -203,7 +203,7 @@ public class SchemaIndex {
     @Procedure("apoc.schema.properties.distinctCount")
     @Description("apoc.schema.properties.distinctCount([label], [key]) YIELD label, key, value, count - quickly returns all distinct values and counts for a given key")
     public Stream<PropertyValueCount> distinctCount(@Name(value = "label", defaultValue = "") String labelName, @Name(value = "key", defaultValue = "") String keyName) throws SchemaRuleNotFoundException, IndexNotFoundKernelException, IOException {
-        Iterable<IndexDefinition> labels = (labelName.isEmpty()) ? db.schema().getIndexes(Label.label(labelName)) : db.schema().getIndexes(Label.label(labelName));
+        Iterable<IndexDefinition> labels = (labelName.isEmpty()) ? db.schema().getIndexes() : db.schema().getIndexes(Label.label(labelName));
         return StreamSupport.stream(labels.spliterator(), false).filter(i -> keyName.isEmpty() || isKeyIndexed(i, keyName)).flatMap(
                 index -> {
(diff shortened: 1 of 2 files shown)


lszumel:
@jexp I can't add a reviewer to this issue. Can you review this PR to fix #639 and #640?

jexp:
Great contribution, thanks! Would you mind adding / updating docs on it? (most likely in overview.adoc) or if you want to we can add it to the schema-index.adoc

jexp:
@lszumel oh sorry I forgot, could you please sign the Neo4j CLA if you haven't https://neo4j.com/developer/cla/

lszumel:
@jexp I'll sign the CLA as soon as I get permission from my boss. I opened #650 for the adoc.

PR #649, opened 2017-11-06
Fixes #648 - Export allows parameter mapping

### LINKED ISSUE #648: Export should allow parameters mapping

bradnussbaum:


src/main/java/apoc/export/csv/ExportCSV.java
```diff
@@ -64,7 +66,8 @@ public ExportCSV() {
     @Procedure
     @Description("apoc.export.csv.query(query,file,config) - exports results from the cypher statement as csv to the provided file")
     public Stream<ProgressInfo> query(@Name("query") String query, @Name("file") String fileName, @Name("config") Map<String, Object> config) throws Exception {
-        Result result = db.execute(query);
+        if (config == null) config = Collections.emptyMap();
+        Result result = db.execute(withParamMapping(query, config.keySet()), config);
```
jexp: @bradnussbaum the config is not meant to contain the parameters directly, it's originally meant to have information about batching and file/export. Can you use an "parameters" param in config instead and then also add it to the docs + test. Thanks

bradnussbaum: @jexp Yes I can change to that.

jexp:
Ping @bradnussbaum

jexp:
Already in, thanks for the pointer.

Issue #310, opened 2017-03-10
Feature Request: apoc.coll.combinations()


InverseFalcon:
This requests a function apoc.coll.combinations() which takes a collection, and outputs a list of lists of all possible combinations, with combination sizes from `minK` (default 1) to `maxK` (default -1, meaning the size of the input list).

`apoc.coll.combinations([coll], minK :1, maxK:-1) `

jexp:
What would be a use case?

Von meinem iPhone gesendet

InverseFalcon:
A couple SO questions emerged that could be helped by a function like this. [Here's one](http://stackoverflow.com/questions/42699435/group-nodes-based-on-calculation-of-properties). 

The summary on that one is, given a collection of nodes, find all combinations where the sum of some property in the combination is within a set range. 

Combination functions are bound to be expensive on larger sets of nodes, so there is that to watch out for.

Looking at the original proposal, one k may not be enough, it might be better to have a minK and maxK, allowing more flexibility for the types of combinations to return.

It might also be interesting to consider a predicate parameter to evaluate for each k-set of combinations to filter down results within the function, rather than having to return some gigantic list of results and filter in Cypher after the function return.

szarnyasg:
If I understand correctly, for a set `S` with this would return Cartesian products, e.g. with `minK = 2` and `maxK = 4`, it would return `S×S`, `S×S×S` and `S×S×S×S`.

A closely related use case is to calculate the Cartesian product of k given sets: https://stackoverflow.com/questions/41800990/neo4j-cypher-to-get-combinations/41808490#41808490

szarnyasg:
I guess this is closed by #657.

PR #657, opened 2017-11-12
Added apoc.coll.combinations()

InverseFalcon:
Given a list, returns all combinations of list elements of combination sizes between minSelect and maxSelect inclusive (if omitted, maxSelect defaults to minSelect)

Fulfills request #310


jexp:
Thanks a lot @InverseFalcon

rmspeers:
I feel like I must be missing something -- but can someone point me to how do I get all permutations (e.g., where order matters) instead of the combinations of items in the list?

PR #661, opened 2017-11-16
Adds UDF for extracting date fields from a (UTC) epoch

alexaverbuch:


jexp:
Thanks a lot @alexaverbuch !!

Issue #663, opened 2017-11-16
apoc.export.cypher.all does not support compound indices


fbiville:
I've got the following index in my database: 
`CREATE INDEX ON :HealthProfessional(first_name, last_name)`

When trying to export everything, APOC fails with the following error:

```
Neo.ClientError.Procedure.ProcedureCallFailed
Failed to invoke procedure `apoc.export.cypher.all`: Caused by: java.util.NoSuchElementException: More than one element in java.util.AbstractList$Itr@4d8006ce. First element is 'first_name' and the second element is 'last_name'
```

PR #665, opened 2017-11-16
[#663] Support composite indices in Cypher export

fbiville:
Quick'n'dirty attempt to fix the issue, while listening to the end keynote of Devoxx Morocco :)


jexp:
Thanks so much @fbiville

Issue #205, opened 2016-11-08
apoc.periodic.iterate() seems to fail if using IS NULL, IS NOT NULL, or label checking in the WHERE clause


InverseFalcon:
Attempting to run apoc.periodic.iterate() works just fine when there are comparison operators present in the WHERE like so:

```
CALL apoc.periodic.iterate(
"MATCH (u:User) WHERE u.processed = false RETURN u",
"SET u.processed = true", {batchSize:10000, parallel:true})
```

No errors. The <> operator is also respected.

However, valid predicates using IS NULL or IS NOT NULL or label checks fail.

For example:

```
CALL apoc.periodic.iterate(
"MATCH (u:User) WHERE u.processed IS NULL RETURN u",
"SET u.processed = true", {batchSize:10000, parallel:true})
```

and

```
CALL apoc.periodic.iterate(
"MATCH (u:User) WHERE NOT u:Processed RETURN u",
"SET u:Processed", {batchSize:10000, parallel:true})
```

both fail with the following error:

`Failed to invoke procedure `apoc.periodic.iterate`: Caused by: java.util.IllegalFormatConversionException: d != java.lang.String`

I'm using Neo4j Community 3.0.6 and APOC Procedures 3.0.4.2 on Mac OS X.

jexp:
Can you build the latest version locally and run it; the error message should contain more information.

jexp:
Sorry you missed a bit (that was probably an issue in the docs):
The columns of the driving statement are passed in as parameters.

CALL apoc.periodic.iterate(
"MATCH (u:User) WHERE u.processed IS NULL RETURN u",
"WITH {u} as u SET u.processed = true", {batchSize:10000, parallel:true})
and

CALL apoc.periodic.iterate(
"MATCH (u:User) WHERE NOT u:Processed RETURN u",
"WITH {u} as u SET u:Processed", {batchSize:10000, parallel:true})

PR #666, opened 2017-11-17
apoc.periodic.iterate docs should use parameter

Meatballs1:
Took me a while to get this working until I found:

https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/205#issuecomment-260084621


jexp:
Please remind me, did you already sign the Neo4j CLA?
https://neo4j.com/developer/cla/

Meatballs1:
I have done now. Ooh just noticed the PR number  👿

docs/periodic.adoc
```diff
@@ -48,7 +48,7 @@ So if you were to add an `:Actor` label to several million `:Person` nodes, you
 ----
 CALL apoc.periodic.iterate(
 "MATCH (p:Person) WHERE (p)-[:ACTED_IN]->() RETURN p",
-"SET p:Actor", {batchSize:10000, parallel:true})
+"WITH {p} AS p SET p:Actor", {batchSize:10000, parallel:true})
```
jexp: you don't need that, it's handled automatically.

jexp:
yep cool PR number :)

docs/periodic.adoc
Meatballs1: Is there any differences between 3.0.x and later versions? My query didn't work until I did this and would say that `Variable ``p`` is not defined` etc

Meatballs1: If so I should maybe PR this against the documentation for the older versions but not against later then

Meatballs1: Looks like you automatically do this in https://github.com/neo4j-contrib/neo4j-apoc-procedures/commit/35d83dfa50e547c5b40c64877172a7bd361d7e53#diff-f240d1755aedbb2670af0b9ed9776fc9 (3.2?) and remove the documentation. But this information has been removed from the 3.0.x docs:  https://neo4j-contrib.github.io/neo4j-apoc-procedures/index30.html#_apoc_periodic_iterate

Meatballs1:
Looking at 3.0.x https://github.com/neo4j-contrib/neo4j-apoc-procedures/commit/f3899051732d2738455edd53211e7f258f08a1b9 it seems like the changes are there too? Not sure why it didn't work for me without the `with` statement...

Issue #671, opened 2017-11-22
String similarity - provide Sørensen–Dice coefficient-based user-defined function


fbiville:
This would be called `apoc.text.sorensenDiceScore(String, String)` or something like that

PR #672, opened 2017-11-23
 [#671] Provide Sørensen–Dice similarity function

fbiville:


ikwattro:
Why not leveraging already existing java libraries like https://github.com/tdebatty/java-string-similarity ?

fbiville:
I personally tend to keep my third-party dependencies to a minimum, especially in this case, the implementation is even quite simple (I already had a specific Kotlin implementation around: https://github.com/graph-labs/open-data-with-neo4j/blob/master/similarity-extension/src/main/kotlin/fr/graphlabs/neo4j/StringSimilarityFunction.kt#L32)

I'm happy to pull a third-party implementation if you think re-implementing the algorithm is too risky.

jexp:
@fbiville looks good, can you also add some docs in overview.adoc and text.adoc

fbiville:
Shall I reduce the contribution to 1 UDF with a default language tag to "en-US"?

jexp:
Yes please

fbiville:
Done! (added the documentation for Levenshtein and Fuzzy Match as well)

PR #676, opened 2017-11-28
mongodb: Added paging parameter in the find procedure

stdob:


src/main/java/apoc/mongodb/MongoDB.java
```diff
@@ -70,9 +70,17 @@ private Coll getColl(@Name("host") String hostOrKey, @Name("db") String db, @Nam
...
+    @Description("apoc.mongodb.find(host-or-port,db-or-null,collection-or-null,query-or-null,projection-or-null,sort-or-null,pagination,[compatibleValues=true|false]) yield value - perform a find,project,sort operation on mongodb collection")
+    public Stream<MapResult> find(@Name("host") String hostOrKey, 
+                                  @Name("db") String db, 
+                                  @Name("collection") String collection, 
+                                  @Name("query") Map<String, Object> query, 
+                                  @Name("project") Map<String, Object> project,
+                                  @Name("sort") Map<String, Object> sort, 
+                                  @Name(value = "compatibleValues", defaultValue = "false") boolean compatibleValues,
+                                  @Name(value = "skip", defaultValue = "0") Long skip,
+                                  @Name(value = "limit", defaultValue = "0") Long limit) {
```
jexp: default value should be -1

src/main/java/apoc/mongodb/MongoDBColl.java
```diff
@@ -114,10 +114,12 @@ public long count(Map<String, Object> query) {
     }
 
     @Override
-    public Stream<Map<String, Object>> find(Map<String, Object> query, Map<String, Object> project, Map<String, Object> sort) {
+    public Stream<Map<String, Object>> find(Map<String, Object> query, Map<String, Object> project, Map<String, Object> sort, Long skip, Long limit) {
         FindIterable<Document> documents = query == null ? collection.find() : collection.find(new Document(query));
         if (project != null) documents = documents.projection(new Document(project));
         if (sort != null) documents = documents.sort(new Document(sort));
+        if (skip != null) documents = documents.skip(skip.intValue());
+        if (limit != null) documents = documents.limit(limit.intValue());
```
jexp: the default will be 0 not null

src/main/java/apoc/mongodb/MongoDBColl.java
```diff
@@ -114,10 +114,12 @@ public long count(Map<String, Object> query) {
     }
 
     @Override
-    public Stream<Map<String, Object>> find(Map<String, Object> query, Map<String, Object> project, Map<String, Object> sort) {
+    public Stream<Map<String, Object>> find(Map<String, Object> query, Map<String, Object> project, Map<String, Object> sort, Long skip, Long limit) {
         FindIterable<Document> documents = query == null ? collection.find() : collection.find(new Document(query));
         if (project != null) documents = documents.projection(new Document(project));
         if (sort != null) documents = documents.sort(new Document(sort));
+        if (skip != null) documents = documents.skip(skip.intValue());
```
jexp: default will be 0 noot null

src/main/java/apoc/mongodb/MongoDB.java
stdob: A limit() value of 0 (i.e. .limit(0)) is equivalent to setting no limit.

https://docs.mongodb.com/manual/reference/method/cursor.limit/#cursor.limit

docs/overview.adoc
```diff
@@ -328,10 +328,10 @@ Adding on config the parameter `failOnError:false` (by default `true`), in case
 
 [cols="3m,2"]
 |===
-| CALL apoc.mongodb.get(host-or-port,db-or-null,collection-or-null,query-or-null,[compatibleValues=true|false]) yield value | perform a find operation on mongodb collection
+| CALL apoc.mongodb.get(host-or-port,db-or-null,collection-or-null,query-or-null,[compatibleValues=true\|false]) yield value | perform a find operation on mongodb collection
```
jexp: @stdob would it make sense to add it here too?

src/main/java/apoc/mongodb/MongoDB.java
jexp: Ah ok. :)

docs/overview.adoc
stdob: @jexp Yes, without escaping it broke the layout.

jexp: You can also just use `/` instead :)

jexp: I meant skip/limit to the `get` procedure.

PR #682, opened 2017-11-30
Fixes #681 - Compound index dropping using apoc.schema.assert fails

### LINKED ISSUE #681: Using apoc.schema.assert({},{}) to drop all indexes fails if any of the indexes are compound
Create your index:

`CREATE INDEX ON :Node(A,B)`

Then attempt to remove:

`CALL apoc.schema.assert({}, {})`

You get:

```
ERROR Neo.ClientError.Procedure.ProcedureCallFailed
Failed to invoke procedure `apoc.schema.assert`: Caused by:
java.util.NoSuchElementException: More than one element in java.util.AbstractList$Itr@23ccbcb5. 
First element is 'A' and the second element is 'B'
```

cskardon:
So as not to change the output, I create multiple `AssertSchemaResult` instances per key in a compound index. This should get into 3.2 and 3.3 branches.

### DIFF
--- a/src/main/java/apoc/result/AssertSchemaResult.java
+++ b/src/main/java/apoc/result/AssertSchemaResult.java
@@ -1,13 +1,29 @@
 package apoc.result;
 
+import java.util.Arrays;
+import java.util.List;
+
 public class AssertSchemaResult {
     public final String label;
     public final String key;
+    public final List<String> keys;
     public boolean unique = false;
     public String action = "KEPT";
 
-    public AssertSchemaResult(String label, String key) {
+    public AssertSchemaResult(String label, List<String> keys) {
         this.label = label;
+        if(keys.size() > 1) {
+            this.keys = keys;
+            this.key = null;
+        } else {
+            this.key = keys.get(0);
+            this.keys = null;
+        }
+    }
+
+    public AssertSchemaResult(String label, String key){
+        this.label = label;
+        this.keys = null;
         this.key = key;
     }
 
--- a/src/main/java/apoc/result/AssertSchemaResult.java
+++ b/src/main/java/apoc/result/AssertSchemaResult.java
@@ -1,5 +1,6 @@
 package apoc.result;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
@@ -12,18 +13,18 @@ public class AssertSchemaResult {
 
     public AssertSchemaResult(String label, List<String> keys) {
         this.label = label;
-        if(keys.size() > 1) {
+        if (keys.size() == 1) {
+            this.key = keys.get(0);
+            this.keys = keys;
+        } else {
             this.keys = keys;
             this.key = null;
-        } else {
-            this.key = keys.get(0);
-            this.keys = null;
         }
     }
 
     public AssertSchemaResult(String label, String key){
         this.label = label;
-        this.keys = null;
+        this.keys = Arrays.asList(key);
         this.key = key;
     }
 
--- a/src/main/java/apoc/schema/Schemas.java
+++ b/src/main/java/apoc/schema/Schemas.java
@@ -31,7 +31,7 @@ public class Schemas {
 
     @Procedure(value = "apoc.schema.assert", mode = Mode.SCHEMA)
     @Description("apoc.schema.assert({indexLabel:[indexKeys], ...}, {constraintLabel:[constraintKeys], ...}, dropExisting : true) yield label, key, unique, action - drops all other existing indexes and constraints when `dropExisting` is `true` (default is `true`), and asserts that at the end of the operation the given indexes and unique constraints are there, each label:key pair is considered one constraint/label")
-    public Stream<AssertSchemaResult> schemaAssert(@Name("indexes") Map<String, List<String>> indexes, @Name("constraints") Map<String, List<String>> constraints, @Name(value = "dropExisting", defaultValue = "true") boolean dropExisting) throws ExecutionException, InterruptedException {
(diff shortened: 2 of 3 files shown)


jexp:
Shouldn't we then also extend the result to support a keys column with a list of strings?
And drop the whole compound index as one?

jexp:
I would rather create a single AssertSchemaResult as that's more correct for node-keys.
As you can add new columns, like `keys` this should be fine.

cskardon:
OK this version deletes the index and returns one response, I've also added the ability to create a compound index via `apoc.schema.assert({Foo:[['bar', 'baa']]}, {})`

src/main/java/apoc/result/AssertSchemaResult.java
```diff
@@ -1,13 +1,29 @@
...
     public final String label;
     public final String key;
+    public final List<String> keys;
     public boolean unique = false;
     public String action = "KEPT";
 
-    public AssertSchemaResult(String label, String key) {
+    public AssertSchemaResult(String label, List<String> keys) {
+        this.label = label;
+        if(keys.size() > 1) {
```
jexp: what if size is 0?

src/main/java/apoc/result/AssertSchemaResult.java
```diff
@@ -1,13 +1,29 @@
...
+            this.keys = keys;
+            this.key = null;
+        } else {
+            this.key = keys.get(0);
+            this.keys = null;
```
jexp: single element list

src/main/java/apoc/result/AssertSchemaResult.java
```diff
@@ -1,13 +1,29 @@
...
+        }
+    }
+
+    public AssertSchemaResult(String label, String key){
         this.label = label;
+        this.keys = null;
```
jexp: single element list

src/main/java/apoc/schema/Schemas.java
```diff
@@ -31,7 +31,7 @@
 
     @Procedure(value = "apoc.schema.assert", mode = Mode.SCHEMA)
     @Description("apoc.schema.assert({indexLabel:[indexKeys], ...}, {constraintLabel:[constraintKeys], ...}, dropExisting : true) yield label, key, unique, action - drops all other existing indexes and constraints when `dropExisting` is `true` (default is `true`), and asserts that at the end of the operation the given indexes and unique constraints are there, each label:key pair is considered one constraint/label")
```
jexp: update description

src/main/java/apoc/schema/Schemas.java
```diff
@@ -99,49 +99,100 @@ public Boolean constraintExistsOnRelationship(@Name("type") String type, @Name("
...
-                schema.indexFor(label(index.getKey())).on(key).create();
-                result.add(new AssertSchemaResult(index.getKey(), key).created());
-            }
+    private AssertSchemaResult createSinglePropertyIndex(Schema schema, String lbl, String key) {
+        schema.indexFor(label(lbl)).on(key).create();
+        return new AssertSchemaResult(lbl, key).created();
+    }
+
+    private AssertSchemaResult createCompoundIndex(String label, List<String> keys) {
+        db.execute(String.format("CREATE INDEX ON :%s (%s)", label, String.join(",", keys)));
```
jexp: use backticks for label and keys

src/main/java/apoc/schema/Schemas.java
```diff
@@ -99,49 +99,100 @@ public Boolean constraintExistsOnRelationship(@Name("type") String type, @Name("
...
+        if (keys1 == null && keys2 == null)
+            return true;
+
+        if (keys1 == null || keys2 == null)
+            return false;
+
+        if (keys1.size() != keys2.size())
+            return false;
+
+        for (String item : keys1) {
```
jexp: what about duplicates in list 2?
better to compare both sets?

src/main/java/apoc/result/AssertSchemaResult.java
cskardon: To do that would change the `yield` which people might be relying on - and to yield multiple keys in a field called 'key' would not give any indication of the actual content

src/main/java/apoc/schema/Schemas.java
cskardon: Duplicates would be caught by the `size()` call - if the 2 arrays are of equal length with list2 having duplicates then the `for` loop would fail as an item in `keys1` wouldn't be in `keys2` - or do you mean you want to allow duplicates?

jexp: list a:
a,b,c,a
list b
a,c,c,b

But probably doesn't matter.

src/main/java/apoc/result/AssertSchemaResult.java
jexp: why? just put a single element list in keys for the single key case.
So that new users can always use `keys` if they want to

src/main/java/apoc/schema/Schemas.java
cskardon: Ahh I see duplicates in both, ok, and how would you compare both sets?

src/main/java/apoc/result/AssertSchemaResult.java
cskardon: Ahh so you didn't mean remove the `key` property - gotcha, makes sense

src/main/java/apoc/schema/Schemas.java
cskardon: Don't worry - got it!

Issue #664, opened 2017-11-16
apoc.data.domain fails to recognize WS TLD


moxious:
```RETURN apoc.data.domain('http://P.O.Ws')```

produces null.

WS is the TLD for Samoa.  https://en.wikipedia.org/wiki/.ws

moxious:
I thought this was going to be easy to fix, but it isn't really because of the assumptions in the test suite.  You can use java.net.URL to parse a URL, and another class to parse an email address.  The implementation right now relies on a flawed regex that only sort of works.

On the other hand, the "obvious" way of solving this won't work, because in the tests, APOC is expected to extract a host from <foo@bar.com> and a host from www.foo.bar/baz.  The former isn't technically a valid email address (although would typically be seen that way with a sender app) and the latter isn't a valid URL since it's missing a scheme.

Easiest would be to require that email addresses and URLs be valid, but I'm opting not to submit a patch because if I did it this way, I might break someone else's assumption to work with messier data.   Since APOC already provides regex matching groups though, this could be cleaned up to expect valid email addresses / URLs, with users still retaining the option to do the "messy" way with regex.

Alternatively, you could blow this out into a different set of functions, like `apoc.data.url.host()` and `apoc.data.url.email()` which IMHO would be safer because the cypher user has to explicitly say what kind of string they expect the input to be.

jexp:
You're probably right with 2 separate functions.
I wonder if there's something in apache.commons.lang that does the proper recognition?

ncordon:
Can we close this @jexp ?

PR #687, opened 2017-12-05
Email and URL extraction functions

moxious:
This PR is intended to address https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/664

Additionally it adds new functions mapping to standard Java API provided functions which are easy and useful to expose.

All of the original test cases that were used as part of the original Extract class have been preserved and new test cases have been added to cover the new functions added.

### DIFF
--- a/src/main/java/apoc/data/url/ExtractURL.java
+++ b/src/main/java/apoc/data/url/ExtractURL.java
@@ -1,0 +1,92 @@
+package apoc.data.url;
+
+import java.net.URL;
+import java.net.MalformedURLException;
+
+import org.neo4j.procedure.Description;
+import org.neo4j.procedure.Name;
+import org.neo4j.procedure.UserFunction;
+
+/**
+ * This class is pretty simple.  It just constructs a java.net.URL instance
+ * from the user's input to do validation/parsing, and delegates the actual
+ * functionality to that class.  As such, the behavior of these functions
+ * matches that class, which is nice.
+ */
+public class ExtractURL {
+    /**
+     * In the context of a user function, we want to ignore parse errors.
+     */
+    protected static URL uriOrNull(String value) {
+        if (value == null) return null;
+        try {
+            URL u = new URL(value);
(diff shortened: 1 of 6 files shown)


src/main/java/apoc/data/url/ExtractURL.java
```diff
@@ -0,0 +1,92 @@
...
+ * matches that class, which is nice.
+ */
+public class ExtractURL {
+    /**
+     * In the context of a user function, we want to ignore parse errors.
+     */
+    protected static URL uriOrNull(String value) {
+        if (value == null) return null;
+        try {
+            URL u = new URL(value);
```
jexp: create an URI instead?

moxious: I tried URI first which is why the name here is leftover, but there are some odd tradeoffs with java.net.URI.   For example, in the case of the string "google.com", with a URI you get host="google.com" and path="google.com" which will surprise people, but is correct to the library since that URI is under-specified.   URIs also give you the ability to include URNs in addition to URLs, but I'm unsure how many people understand that since URNs are pretty uncommon these days.    This could be made more generic in order to use URIs rather than URLs though.

jexp:
Any chance to do this without adding the extra library?
We only need InternetAddress or? Can we just copy the code for that?

Sad that javax mail is not in the JDK.

moxious:
@jexp it would be possible to borrow an implementation from somewhere else, but just the parsing of an email address is [fairly substantial](http://grepcode.com/file/repository.springsource.com/javax.mail/com.springsource.javax.mail/1.4.0/javax/mail/internet/InternetAddress.java#591).   The point of using the external library was to do it correctly according to the RFC.  The previous implementation used a regular expression, and there's only so far you can go with that as far as correctness goes.   So if you prefer we can maybe lift an open source implementation of this class into APOC to avoid the dependency.

jexp:
merged manually

PR #690, opened 2017-12-06
Enable parsing of lists in GraphML Import

yourpalal:
First of all, thanks for this awesome project!

This commit adds functionality to import lists, as exported by the GraphML exporter.

### DIFF
--- a/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
+++ b/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
@@ -13,6 +13,8 @@ import javax.xml.stream.events.Attribute;
 import javax.xml.stream.events.StartElement;
 import javax.xml.stream.events.XMLEvent;
 import java.io.Reader;
+import java.lang.reflect.Array;
+import java.util.function.Function;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -64,8 +66,7 @@ public class XmlGraphMLReader {
             }
 
             Object parseList(String value) {
-                List parsedList = JsonUtil.parse(value, null, List.class);
-                return parsedList.toArray(new Boolean[parsedList.size()]);
+                return Type.parseList(value, Boolean.class, (i) -> (Boolean)i);
             }
         }, INT() {
             Object parse(String value) {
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/export/graphml/XmlGraphMLReader.java
```diff
@@ -60,30 +62,82 @@ public XmlGraphMLReader reporter(Reporter reporter) {
...
             Object parse(String value) {
                 return Long.parseLong(value);
             }
+
+            Object parseList(String value) {
+                List parsedList = JsonUtil.parse(value, null, List.class);
+                Long[] result = new Long[parsedList.size()];
+                for (int i = 0; i < parsedList.size(); i++) {
+                    Object item = parsedList.get(i);
+                    if (item instanceof Long) {
```
jexp: why always longs? shouldn't we look into the datatypes of the parsed list?
also ((Number)item).longValue()

jexp: also in the other places

src/main/java/apoc/export/graphml/XmlGraphMLReader.java
```diff
@@ -60,30 +62,82 @@ public XmlGraphMLReader reporter(Reporter reporter) {
             Object parse(String value) {
                 return Boolean.valueOf(value);
             }
+
+            Object parseList(String value) {
+                List parsedList = JsonUtil.parse(value, null, List.class);
```
jexp: we could generify the functions
there is a method in Java lang class to create an array of the type.
and then only call.
`return parseList(value, Boolean.class) `

jexp:
would you mind signing our CLA:  thanks so much
https://neo4j.com/developer/cla/

src/main/java/apoc/export/graphml/XmlGraphMLReader.java
yourpalal: Longs because the list was defined as a list of longs. I'm not sure if the input format makes a difference to how neo4j stores them, or if it can handle different types of numbers in the same array, so I can switch it to whatever you think is best.

yourpalal: I tried converting to a Number[] for the Long/Int/Double/Float methods and it looks like that's not supported:

`java.lang.IllegalArgumentException: java.lang.Number[] is not a supported property value type`

So we'll have to have some kind of type conversion to a specific number type. Although I think it could be possible with reflection to figure out a type based on the input values, forcing them into the type that was declared in the graphml is simpler and probably

yourpalal:
just having my boss review the CLA, but I expect it won't be a problem

yourpalal:
Accepted the CLA and sent it in.

jexp:
Thanks so much!

PR #698, opened 2017-12-11
fixes #527 - Add s3 protocol handler

### LINKED ISSUE #527: Add s3 protocol handler
Eg using this
https://github.com/jenshadlich/S3-URLStreamHandler/blob/master/README.md

AngeloBusato:


build.gradle
```diff
@@ -70,6 +70,7 @@ dependencies {
     compileOnly group: 'net.biville.florent', name: 'neo4j-sproc-compiler', version:'1.2'
 
     compile 'com.novell.ldap:jldap:2009-10-07'
+    compile group: 'com.amazonaws', name: 'aws-java-sdk-s3', version: '1.11.241'
```
jexp: should this be compileOnly ? so that the user would have to provide the jar?

src/main/java/apoc/load/Xml.java
```diff
@@ -61,8 +63,14 @@
             DocumentBuilder documentBuilder = documentBuilderFactory.newDocumentBuilder();
 
             FileUtils.checkReadAllowed(url);
-            URLConnection urlConnection = new URL(url).openConnection();
-            Document doc = documentBuilder.parse(urlConnection.getInputStream());
+            Document doc;
+            if(url.toLowerCase().startsWith("s3")){
```
jexp: don't do any of this here, encapsulate it in file-utils.

src/main/java/apoc/util/Util.java
```diff
@@ -302,6 +307,7 @@ private static String handleRedirect(URLConnection con, String url) throws IOExc
     }
 
     public static CountingInputStream openInputStream(String url, Map<String, Object> headers, String payload) throws IOException {
+        if(url.startsWith("s3")) return openS3InputStream(url);
```
jexp: can we use a proper URL parsing + getProtocol?

src/main/java/apoc/util/s3util/S3Params.java
```diff
@@ -0,0 +1,45 @@
+package apoc.util.s3util;
+
+public class S3Params {
+
+    private final String accessKey;
+    private final String secretKey;
```
jexp: good question, we should have an option to store this in a separate config, either in neo4j.conf or in a more secure place.

src/main/java/apoc/util/s3util/S3ParamsExtractor.java
```diff
@@ -0,0 +1,71 @@
+package apoc.util.s3util;
+
+import com.amazonaws.regions.Regions;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.Objects;
+
+public class S3ParamsExtractor {
+    public static S3Params extract(URL url) throws IOException, IllegalArgumentException {
```
jexp: why is this not a method in S3Params?

src/main/java/apoc/util/s3util/S3ParamsExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+        if (!"s3".equals(url.getProtocol())) {
+            throw new IllegalArgumentException("Unsupported protocol '" + url.getProtocol() + "'");
+        }
+
+        //aws credentials
+        String accessKey = null;
+        String secretKey = null;
+
+        if (url.getUserInfo() != null) {
+            String[] credentials = url.getUserInfo().split("[:]");
```
jexp: you dont' need the square brackets

src/main/java/apoc/util/s3util/S3ParamsExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+            accessKey = credentials[0];
+            secretKey = credentials[1];
+        } else {
+            if (url.getQuery() != null) {
+                String[] queryStrings = url.getQuery().split("&");
+                if (queryStrings.length == 2) {
```
jexp: don't we already have a function for this somewhere? I'm pretty sure.

jexp: If not a proper version should go into Util or FileUtils

src/main/java/apoc/util/s3util/S3ParamsExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+                                secretKey = valueDecoded;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        // bucket
+        String bucket = url.getHost().substring(0, url.getHost().indexOf("."));
```
jexp: you can store the host and the index of "." once.

src/main/java/apoc/util/s3util/S3ParamsExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+        // key
+        String key = url.getPath().substring(1);
+
+        // endpoint
+        String endpoint = url.getHost().substring(url.getHost().indexOf(".") + 1);
+
+        String region = null;
+
+        for (Regions r: Regions.values()){
+            if(endpoint.contains(r.getName())){
```
jexp: what about case, we should probably use tolower on both. And probably split the endpoint on "." ? and access the right segment.

src/main/java/apoc/util/s3util/S3URLConnection.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final String PROP_S3_HANDLER_PROTOCOL = "s3.handler.protocol";
+    public static final String PROP_S3_HANDLER_SIGNER_OVERRIDE = "s3.handler.signerOverride";
+
+    public S3URLConnection(URL url) {
+        super(url);
+    }
+
+    @Override
+    public void connect() throws IOException {
+        // not implemented yet
```
jexp: ???

src/main/java/apoc/util/s3util/S3URLConnection.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public static ClientConfiguration buildClientConfig() {
+        final String userAgent = System.getProperty(PROP_S3_HANDLER_USER_AGENT, null);
```
jexp: are these documented?

src/main/java/apoc/util/s3util/S3ParamsExtractor.java
jexp: is the format documented?

src/test/java/apoc/load/LoadJsonTest.java
```diff
@@ -108,4 +109,34 @@
                     assertFalse(row.hasNext());
                 });
     }
+
+    @Test public void testLoadJsonS3() throws Exception {
+
+        String accessKey = "Q3AM3UQ867SPQQA43P2F";
```
jexp: Can we unify the minio setup in one test helper class?

AngeloBusato:
Ok @jexp we'll have a look

AngeloBusato:
@jexp  The increased size of apoc jar with aws-sdk-s3 library included is about 4 MBytes.

jexp:
Please have a look at the comments. And we would add the s3 library as an external dependency. Does it has any transitive dependencies?

AngeloBusato:
@jexp PR updated.

with `aws-java-sdk-s3-1.11.250.jar` we have to import these jars too: 

* aws-java-sdk-core-1.11.250.jar 
* httpclient-4.4.8.jar 
* httpcore-4.5.4.jar 
* joda-time-2.9.9.jar

jexp:
Weird:

- joda-time is in the JDK these days (since 1.7)
- httpclient/httpcore would be perhaps ok as a default dependency (not sure how big they are)
- currently I avoided pulling them in and used http-url-connection instead

AngeloBusato:
@jexp The increased size of apoc jar with httpclient/httpcore  libraries included is about 1 MBytes.

jexp:
thanks a lot, merged manually.

PR #719, opened 2018-01-14
Provide HDFS Read/Write Support 

conker84:
This PR add the support to HDFS so you can interact with this filesystem in this way:
Read:
```
CALL apoc.load.csv('hdfs://localhost:12345/user/a.santurbano/test.csv',{results:['map','list','stringMap','strings']})
```
Write:
```
CALL apoc.export.csv.all('hdfs://localhost:12345/user/a.santurbano/all.csv',null)
```
The required external dependencies are (the complete size is more or less 15 mb):
* commons-cli
* hadoop-auth
* hadoop-client
* hadoop-common
* hadoop-hdfs
* htrace-core-3.1.0-incubating
* protobuf-java
* guava-11.0.2

Hope it helps!


jexp:
Cool, thanks a lot

I would love to make those dependencies optional, except for commons-cli perhaps, so that a user would need to provide them.

conker84:
Sorry @jexp I think I used a wrong word, the listed dependencies are required to make HDFS work, apoc package still works without these if you don't need the HDFS support.
If you want I can add commons-cli as compiled dependency!

inserpio:
I would do testCompile as mongo and couchbase

conker84:
@inserpio do you mean about commons-cli?

inserpio:
Commons-cli should be 3Mb so no problem with embedding it in the uber jar. The other ones looks to be (1) bigger and (2) not very reusable/generic

conker84:
Hi guys,
I updated the build.gradle file in order to import commons-cli, moreover I just realized that there is another dependency that I forgotten, **guava-11.0.2**, what do you want to do with this library?

jexp:
Merged Manually.

jexp:
Thanks a lot, there were a lot of clashes with the s3 support so I had to refactor it quite a lot.

btw. most [recent guava versions](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.google.guava%22%20AND%20a%3A%22guava%22) are in the 23.x range

conker84:
@jexp just let me know if I can give you an hand in the refactoring process. I'm glad to help!

Yes I specified the guava version used by hadoop library because they are quite old.

conker84:
@jexp just let me know if I can give you an hand in the refactoring process. I'm glad to help!

Yes I specified the guava version used by the hadoop library because it's quite old.

inserpio:
That's why I keep thinking about having different class loaders: soon or later (especially these integration apocs) dependencies will crash between each other or with the neo4j ones :-/

Issue #723, opened 2018-01-24
Add apoc.text.charAt()


InverseFalcon:
Given a string and an index, gets the representation of the character at that index

InverseFalcon:
Manually merged in, request fulfilled.

PR #724, opened 2018-01-24
Added apoc.text.charAt() and apoc.text.hexCharAt()

InverseFalcon:
Fulfills request #723

### DIFF
--- a/src/main/java/apoc/text/Strings.java
+++ b/src/main/java/apoc/text/Strings.java
@@ -372,4 +372,14 @@ public class Strings {
         byte[] decoded = Base64.getDecoder().decode(text.getBytes());
         return new String(decoded);
     }
+
+    @UserFunction
+    @Description("apoc.text.charAt(text, index) - the hex value string of the character at the given index")
+    public String charAt(@Name("text") String text, @Name("index") Long index) {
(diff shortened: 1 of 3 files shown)


src/main/java/apoc/text/Strings.java
```diff
@@ -372,4 +372,14 @@ public String base64Decode(@Name("text") String text) {
         byte[] decoded = Base64.getDecoder().decode(text.getBytes());
         return new String(decoded);
     }
+
+    @UserFunction
+    @Description("apoc.text.charAt(text, index) - the hex value string of the character at the given index")
+    public String charAt(@Name("text") String text, @Name("index") Long index) {
```
jexp: shouldn't this return the long/int value?
and then a separate function do the toHex ?

InverseFalcon: Could do that. Ran into some issues with `Character.getNumericValue()` not returning the correct values for certain characters, I'll check for alternatives.

InverseFalcon:
For hexCharAt, does the empty string return make sense for invalid inputs, or would null work better?

jexp:
Merged manually and added apoc.text.hexValue

PR #726, opened 2018-01-25
adding apoc.import.xml

sarmbruster:
This PR adds `apoc.import.xml` to directly create nodes/rels for a xml document.

Please add this to 3.3 branch as well when merging.


jexp:
merged manually

Issue #729, opened 2018-01-27
Feature Request: delete all nodes with labels


wadael:
I had a need to remove all the newly and badly imported nodes after an import.
Removing all the data was out of question as I already had imported a huge quantity of data from another dataset.
One option is to do several "MATCH (n:LabelA) DETACH DELETE n", one for each label

The other is to have a procedure that takes a list of procedure names as strings and iterate on it to execute the several corresponding cypher queries.
Like 

     CALL wadael.remove.nodes.withlabel(["Activite","Categorie","Effectif","Entreprise","Etablissement","Localisation","NatureEI","NatureJuridique","Region","Responsable"])

which is quicker than x times a MATCH DETACH DELETE query

[LabelProcedure.java.txt](https://github.com/neo4j-contrib/neo4j-apoc-procedures/files/1670616/LabelProcedure.java.txt)

If you guys think it is worthy, what package should it be in, under which name 
I propose apoc.nodes.removeByLabel   (singular because only one label is needed for a node to be deleted)

jexp:
Sounds good to me, just need to batch internally for size otherwise you get into tx-size issues.

You should also handle null both as list and value
And add backticks around the label in the cypher statement.

jexp:
Oh and you would need to sign the Neo4j CLA https://neo4j.com/developer/cla/

wadael:
Can you explain "just need to batch internally for size otherwise you get into tx-size issues." ?

InverseFalcon:
@wadael When altering the graph, the graph changes have to be kept in memory for the transaction. When a large number of these build up for a transaction, you risk heap issues, including exceeding the heap size. We deal with this by batching these larger kinds of queries (see the APOC periodic procedures).

So in this case, we would want to batch the deletions. Unfortunately it's not as simple as JUST that, since a DETACH DELETE operation also deletes all relationships of the nodes being deleted. So it's not enough just to batch on the nodes being deleted, we also need to control the relationship deletion, as high numbers of relationships per node can skew the deletions being committed in the transaction (and supernodes in that set can choke the transaction).

wadael:
Hi Andrew, thanks for the info.

If I sum up. My usage of apoc periodic, which I infered from Michael answer
is right.
He also advised a limit per transaction.
You advise to deal with the relations first.
So, would this solution be ok ?

Loop for périodic delete of relations from node with label1 till 0 lines
are changed
Then remove nodes with label1 nodes

[...]

PR #730, opened 2018-01-30
Added possibility to remove nodes per labels

wadael:
apoc.remove.nodes.withlabel("label1","label2",....) will remove the nodes that one at least one of those labels.

Under the hood, it does N times
CALL apoc.periodic.commit('MATCH (n:`label`) DETACH DELETE n;',null)

### DIFF
--- a/src/main/java/apoc/nodes/Nodes.java
+++ b/src/main/java/apoc/nodes/Nodes.java
@@ -305,18 +305,22 @@ public class Nodes {
     }
 
     @Procedure(mode = Mode.WRITE)
-    @Description("apoc.nodes.removewithlabels(['Person','Movie'],1000) | Will remove all the nodes having one of the given labels, via DETACH DELETE using apoc.periodic.iterate. Same limit used for each label. Default is 100. ")
-    public Stream<MapResult> removewithlabels(@Name("labels") List<String> labels, @Name(value = "limit",defaultValue="100")  Long limit) {
+    @Description("apoc.nodes.removebylabels(['Person','Movie'],1000) | Will remove all the nodes having one of the given labels, via DETACH DELETE using apoc.periodic.iterate. Same limit used for each label. Default is 100. ")
+    public Stream<MapResult> removebylabels(@Name("labels") List<String> labels, @Name(value = "limit",defaultValue="100") long limit) {
         if (limit <= 0)
             throw new RuntimeException("limit must be a positive integer");
 
-        Map<String,Object> results = new HashMap<String, Object>();
-        HashMap<String, Object> parameters = new HashMap<>();
+        Map<String,Object> results = map();
 
         for( String label: labels) {
-            if(label.trim().length()>0) {
-                String query = String.format("call apoc.periodic.iterate( 'MATCH (n:`%s`) RETURN n',  'DETACH DELETE n', {batchsize:%d,iterateList:true, parallel:true}  )", label, limit);
-                db.execute(query, parameters);
+            if(!label.isEmpty()) {
+                String query = String.format("CALL apoc.periodic.iterate('MATCH (n:`%s`) RETURN n',  'DETACH DELETE n', {batchsize:%d,iterateList:true, parallel:true}  )", label, limit);
+                Result result = db.execute(query);
+
+                while(result.hasNext()) {
+                    Map<String, Object> mso = result.next();
+                    results.put(label, mso.get("total") );
+                }
             }
         }
         MapResult mapResult = new MapResult(results);
--- a/src/main/java/apoc/nodes/Nodes.java
+++ b/src/main/java/apoc/nodes/Nodes.java
@@ -17,6 +17,7 @@ import org.neo4j.kernel.impl.api.RelationshipVisitor;
 import org.neo4j.kernel.impl.api.store.RelationshipIterator;
 import org.neo4j.logging.Log;
 import org.neo4j.procedure.*;
+import org.neo4j.procedure.Description;
 import org.neo4j.storageengine.api.Token;
 
 import java.util.*;
@@ -24,6 +25,7 @@ import java.util.stream.Stream;
 
 import static apoc.path.RelationshipTypeAndDirections.parse;
 import static apoc.util.Util.map;
+import apoc.result.MapResult;
 
 public class Nodes {
 
--- a/src/main/java/apoc/nodes/Nodes.java
+++ b/src/main/java/apoc/nodes/Nodes.java
@@ -1,6 +1,9 @@
 package apoc.nodes;
 
-import apoc.result.*;
+import apoc.*;
+import apoc.result.LongResult;
+import apoc.result.NodeResult;
+import apoc.result.RelationshipResult;
 import apoc.util.Util;
 import org.neo4j.collection.primitive.PrimitiveIntIterator;
 import org.neo4j.graphdb.*;
--- a/src/main/java/apoc/nodes/Nodes.java
+++ b/src/main/java/apoc/nodes/Nodes.java
@@ -303,6 +303,16 @@ public class Nodes {
         }
     }
 
+    @Procedure(value = "apoc.remove.nodes.withlabel", mode = Mode.WRITE)
+    @Description("Will remove all the nodes having one of the given labels, via DETACH DELETE")
+    public void removeNodesWithLabels(@Name("labels") List<String> labels) {
+        for (String label : labels) {
+            if (label!=null && label.trim().length()>0)
+                db.execute("CALL apoc.periodic.commit('MATCH (n:`" + label + "`) DETACH DELETE n;',null)");
+        }
+    }
+
+
     private boolean isDense(ReadOperations ops, Node n) {
         try {
             return ops.nodeIsDense(n.getId());
--- a/src/main/java/apoc/nodes/Nodes.java
+++ b/src/main/java/apoc/nodes/Nodes.java
@@ -331,6 +331,7 @@ public class Nodes {
     }
 
 
+
     private boolean isDense(ReadOperations ops, Node n) {
         try {
             return ops.nodeIsDense(n.getId());
--- a/src/main/java/apoc/nodes/Nodes.java
+++ b/src/main/java/apoc/nodes/Nodes.java
@@ -1,14 +1,9 @@
 package apoc.nodes;
 
-import apoc.result.LongResult;
-import apoc.result.NodeResult;
-import apoc.result.RelationshipResult;
+import apoc.result.*;
 import apoc.util.Util;
 import org.neo4j.collection.primitive.PrimitiveIntIterator;
-import org.neo4j.graphdb.Direction;
-import org.neo4j.graphdb.GraphDatabaseService;
-import org.neo4j.graphdb.Node;
-import org.neo4j.graphdb.RelationshipType;
+import org.neo4j.graphdb.*;
 import org.neo4j.helpers.collection.Iterables;
 import org.neo4j.helpers.collection.Pair;
 import org.neo4j.kernel.api.KernelTransaction;
@@ -17,6 +12,7 @@ import org.neo4j.kernel.api.Statement;
 import org.neo4j.kernel.api.exceptions.EntityNotFoundException;
 import org.neo4j.kernel.impl.api.RelationshipVisitor;
 import org.neo4j.kernel.impl.api.store.RelationshipIterator;
+import org.neo4j.logging.Log;
 import org.neo4j.procedure.*;
 import org.neo4j.storageengine.api.Token;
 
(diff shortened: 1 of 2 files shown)


wadael:
Relates to https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/729

wadael:
Works on my machine :)
Errors are in BoltTest, not impacted by my PR

jexp:
Can you change it to withLabels
and the usage of periodic commit is wrong you have to return a count of deleted entities and use a WITH {limit} parameter within the query
`MATCH (n:'+label+') WITH n LIMIT {limit} DETACH DELETE n RETURN count(*)`

wadael:
I was surprised that using WITH n LIMIT {limit}, not forgetting to put limit with a value in the passed params did not work. Query executed. Nodes were not deleted. I'll try to add a testcase later.

wadael:
Other tests failing although not impacted by this PR ... grrr

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -1,14 +1,9 @@
 package apoc.nodes;
 
-import apoc.result.LongResult;
```
jexp: please don't change the imports that makes merges more complicated

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +299,39 @@ public boolean isDense(@Name("node") Node node) {
         }
     }
 
+    @Procedure(value = "apoc.remove.nodes.withlabels", mode = Mode.WRITE)
```
jexp: don't use remove as namespace, before nodes, but instead just call this `apoc.nodes.remove`

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +299,39 @@ public boolean isDense(@Name("node") Node node) {
...
+    @Description("apoc.remove.nodes.withlabels(['Person','Movie'],-1) | Will remove all the nodes having one of the given labels, via DETACH DELETE. Use a negative limit for no limit in deletion. Same limit used for each label.")
+    public Stream<MapResult> removeNodesWithLabels(@Name("labels") List<String> labels, @Name("limit") Long limit) {
```
jexp: Limit should be a `long`primitive and have a defaultValue entry of "-1"

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +299,39 @@ public boolean isDense(@Name("node") Node node) {
...
+        StringBuilder query = new StringBuilder();
+
+        Map<String,Object> results = new HashMap<String, Object>();
+        HashMap<String, Object> parameters = new HashMap<>();
+
+        for( String label: labels) {
+            if(label != null && !label.trim().equals("")) {
```
jexp: !label.trim("").isEmpty()

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +299,39 @@ public boolean isDense(@Name("node") Node node) {
...
+                query.setLength(0);
+                query.append("MATCH (n:");
+                query.append(label);
+                query.append(")");
+                if (limit > 0) {
+                    query.append(" WITH n LIMIT ");
+                    query.append(limit);
```
jexp: limit should be a parameter

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +299,39 @@ public boolean isDense(@Name("node") Node node) {
...
+                }
+                query.append(" DETACH DELETE n RETURN");
+                query.append(" \"");
```
jexp: you don't need to return things, you can just use it from the query result statistics.

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +299,39 @@ public boolean isDense(@Name("node") Node node) {
...
+        HashMap<String, Object> parameters = new HashMap<>();
+
+        for( String label: labels) {
+            if(label != null && !label.trim().equals("")) {
+                query.setLength(0);
+                query.append("MATCH (n:");
+                query.append(label);
+                query.append(")");
+                if (limit > 0) {
+                    query.append(" WITH n LIMIT ");
```
jexp: I think stringbuilder is overkill here, just use string.format
and make sure to use backticks arround the label

src/test/java/apoc/nodes/NodesTest.java
```diff
@@ -175,4 +174,74 @@ public void testDegreeDirectionOnly() {
...
+                    r -> {
+                    });
+        } catch (Exception e) {
+            caught = true;
+        }
+        Assert.assertTrue(caught);
+    }
+
+    @Test
+    public void testRemoveNodesWithLabels_OkParamsWithLimit() {
```
jexp: also test label with spaces

src/test/java/apoc/nodes/NodesTest.java
```diff
@@ -175,4 +174,74 @@ public void testDegreeDirectionOnly() {
...
+                params,
+                r -> {
+                    Map m = r.next();
+                    Map map = (Map) m.get("value");
+                    assertNotNull(map);
+                    assertEquals(3,map.size());
+                });
+    }
+
+    @Test
```
jexp: just add an expected exception to `@Test`

src/test/java/apoc/nodes/NodesTest.java
```diff
@@ -175,4 +174,74 @@ public void testDegreeDirectionOnly() {
...
+    public void testRemoveNodesWithLabels_OkParamsWithLimit() {
+        String[] labels = {"aa", "bb", "cc"};
+        Map params = new HashMap();
+        params.put("labels", labels);
+        params.put("limit", 50);
+        TestUtil.testResult(db,
+                "CALL apoc.remove.nodes.withlabels",
+                params,
+                r -> {
+                    Map m = r.next();
```
jexp: use generics

src/main/java/apoc/nodes/Nodes.java
jexp: ping

jexp:
@wadael do you want to continue on this one?

src/main/java/apoc/nodes/Nodes.java
wadael: As this is thought as a command to help with messed "load csv"s , I'd rather return the count of deleted nodes per label.

wadael:
Hi Michael,
The conflict is the new code here.
Took time to push that Im so lame at git, had to reclone

jexp:
There is a new conflict. Sorry. If you get it in soon today it will be part of the relase..

wadael:
idk, my tests are after the pre existing ones. Why is there a conflict ...

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +304,26 @@ public boolean isDense(@Name("node") Node node) {
...
+    public Stream<MapResult> removewithlabels(@Name("labels") List<String> labels, @Name(value = "limit",defaultValue="100")  Long limit) {
+        if (limit <= 0)
+            throw new RuntimeException("limit must be a positive integer");
+
+        Map<String,Object> results = new HashMap<String, Object>();
+        HashMap<String, Object> parameters = new HashMap<>();
+
+        for( String label: labels) {
+            if(label.trim().length()>0) {
+                String query = String.format("call apoc.periodic.iterate( 'MATCH (n:`%s`) RETURN n',  'DETACH DELETE n', {batchsize:%d,iterateList:true, parallel:true}  )", label, limit);
```
jexp: parallel true doesn't work here
best to prefix the statement with `cypher runtime=slotted MATCH (...)`
quote the labels

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -1,14 +1,12 @@
 package apoc.nodes;
 
+import apoc.*;
```
jexp: don't modify the imports

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +304,26 @@ public boolean isDense(@Name("node") Node node) {
         }
     }
 
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.nodes.removewithlabels(['Person','Movie'],1000) | Will remove all the nodes having one of the given labels, via DETACH DELETE using apoc.periodic.iterate. Same limit used for each label. Default is 100. ")
```
jexp: please change the name to removeByLabel

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +304,26 @@ public boolean isDense(@Name("node") Node node) {
...
+    public Stream<MapResult> removewithlabels(@Name("labels") List<String> labels, @Name(value = "limit",defaultValue="100")  Long limit) {
```
jexp: make limit a `long` no boxing

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +304,26 @@ public boolean isDense(@Name("node") Node node) {
...
+        if (limit <= 0)
+            throw new RuntimeException("limit must be a positive integer");
+
+        Map<String,Object> results = new HashMap<String, Object>();
+        HashMap<String, Object> parameters = new HashMap<>();
+
+        for( String label: labels) {
+            if(label.trim().length()>0) {
```
jexp: isEmpty()

src/main/java/apoc/nodes/Nodes.java
jexp: I'm not 100% sure apoc.periodic.iterate works in the same tx. please also test it in a server environment with a lot of highly connected labels (1M nodes 2M rels)

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +304,26 @@ public boolean isDense(@Name("node") Node node) {
...
+                String query = String.format("call apoc.periodic.iterate( 'MATCH (n:`%s`) RETURN n',  'DETACH DELETE n', {batchsize:%d,iterateList:true, parallel:true}  )", label, limit);
+                db.execute(query, parameters);
```
jexp: you can either use the call without parameters or use them :)

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +304,26 @@ public boolean isDense(@Name("node") Node node) {
...
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.nodes.removewithlabels(['Person','Movie'],1000) | Will remove all the nodes having one of the given labels, via DETACH DELETE using apoc.periodic.iterate. Same limit used for each label. Default is 100. ")
+    public Stream<MapResult> removewithlabels(@Name("labels") List<String> labels, @Name(value = "limit",defaultValue="100")  Long limit) {
+        if (limit <= 0)
+            throw new RuntimeException("limit must be a positive integer");
+
+        Map<String,Object> results = new HashMap<String, Object>();
+        HashMap<String, Object> parameters = new HashMap<>();
+
+        for( String label: labels) {
```
jexp: stream to result one row per label

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -303,6 +304,26 @@ public boolean isDense(@Name("node") Node node) {
...
+            if(label.trim().length()>0) {
+                String query = String.format("call apoc.periodic.iterate( 'MATCH (n:`%s`) RETURN n',  'DETACH DELETE n', {batchsize:%d,iterateList:true, parallel:true}  )", label, limit);
+                db.execute(query, parameters);
+            }
+        }
+        MapResult mapResult = new MapResult(results);
```
jexp: you are not using the results, you should use the results of the iterate call as result for your method

inline

src/main/java/apoc/nodes/Nodes.java
jexp: `labels.stream().filter(empty).map(query)flatMap(execute).map(result)`

src/test/java/apoc/nodes/NodesTest.java
```diff
@@ -27,7 +26,8 @@
     @Before
     public void setUp() throws Exception {
         db = TestUtil.apocGraphDatabaseBuilder().newGraphDatabase();
-        TestUtil.registerProcedure(db,Nodes.class);
+        TestUtil.registerProcedure(db,Nodes.class, Create.class);
+        TestUtil.registerProcedure(db, Periodic.class);
```
jexp: this is varargs, single line

src/test/java/apoc/nodes/NodesTest.java
```diff
@@ -175,4 +175,184 @@ public void testDegreeDirectionOnly() {
         });
 
     }
+
+    @Test
+    public void testRemoveNodesEmptyLabels() {
+        String[] labels = {};
+        Map params = new HashMap();
```
jexp: Utils.map()

src/test/java/apoc/nodes/NodesTest.java
```diff
@@ -175,4 +175,184 @@ public void testDegreeDirectionOnly() {
...
+        params.put("labels", labels);
+        params.put("limit", 50);
+        TestUtil.testResult(db,
+                "CALL apoc.nodes.removewithlabels",
```
jexp: don't use this call syntax, it's deprecated

src/test/java/apoc/nodes/NodesTest.java
```diff
@@ -175,4 +175,184 @@ public void testDegreeDirectionOnly() {
...
+    public void testRemoveNodesWithLabels_AnEmptyLabel() {
+        String[] labels = {"aa", "bb" ,"","cc"};
+        Map params = new HashMap();
+        params.put("labels", labels);
+        params.put("limit", 50);
+        TestUtil.testResult(db,
+                "CALL apoc.nodes.removewithlabels",
+                params,
+                r -> {
+                    Map m = r.next();
```
jexp: generics

src/test/java/apoc/nodes/NodesTest.java
```diff
@@ -175,4 +175,184 @@ public void testDegreeDirectionOnly() {
...
+                    Map map = (Map) m.get("value");
+                    assertNotNull(map);
```
jexp: really?!

jexp: please do some proper checking of the results, now that you have some

src/main/java/apoc/nodes/Nodes.java
wadael: I heard slotted is Entreprise only in your video. Unsure how it would behave in community.

wadael: org.neo4j.graphdb.QueryExecutionException: Failed to invoke procedure  if I use runtime=slotted

As seen in debug:
community single [/home/jerome/OpenSource/wadael/neo4j-apoc-procedures/target/test-data/impermanent-db]

wadael:
I think I cannot solve the not-a conflict

mneedham:
@wadael closing this for now. Let us know if you'd like us to revisit

Issue #691, opened 2017-12-08
add support for repeating patterns to expand


jexp:
I know that the traversal-fw has been used for that where you pass an array of rel-types (and/or) labels and then fit the path-index to index of that array modulo array length

like: `"<FROM,TO>" with an similar syntax for labels, 

or just allow cypher pattern syntax in general

`(:Airport)<-[:FROM]-(:Flight)-[:TO]->(:Airport)`

InverseFalcon:
Implemented by #739

PR #739, opened 2018-02-12
Added sequences to path expander procedures

InverseFalcon:
Added `labelSequence` to path expander config parameters, allowing label filters to define a repeating sequence of nodes to expand.

Further commits also include `relationshipSequence`, and `sequence` (when both are needed)
Addresses #691.

### DIFF
--- a/docs/expand.adoc
+++ b/docs/expand.adoc
@@ -35,30 +35,43 @@ Syntax: `[<]RELATIONSHIP_TYPE1[>]|[<]RELATIONSHIP_TYPE2[>]|...`
 
 ==== Label Filter
 
-Syntax: `[+-/>]LABEL1|LABEL2|...`
+Syntax: `[+-/>]LABEL1|LABEL2|*|...`
+
+
+[opts=header,cols="m,a"]
+|===
+| input | result
+| -Foe | blacklist filter - No node in the path will have a label in the blacklist.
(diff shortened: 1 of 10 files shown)


InverseFalcon:
Not quite ready for merge, I want to change a behavior with end node / termination node filtering for steps in the label sequence. I think I'd rather have the filters apply specifically to the labels matched, rather than at that current step in the sequence as a whole.

InverseFalcon:
@jexp Okay, done here, ready for final review and merge.

Noted in the documentation some differences between the labelFilter and the labelSequence and sequence. 

Personally I prefer the newer behavior, so with 3.4 I'll be updating the labelFilter to match (we can also reuse the same evaluator between the label filter and label sequences, which will be a nice step as well).

InverseFalcon:
I'm actually wondering now if it makes more sense to just keep `labelFilter` and `relationshipFilter` and just allow sequences in both of these. If there are entries separated by `,` it's a sequence, otherwise it's a filter for all node labels / relationships accordingly. 

The only conflict I can see (apart from some differences in behavior between the label filter and label sequences) is that label sequences can currently coexist with the label filter, so using both has to be possible. Maybe we don't actually need that.

Thoughts?

InverseFalcon:
Made a decision. We'll reuse existing `labelFilter` and `relationshipFilter` for sequences, and standardize on a single syntax and behavior for label filters. 

Changed behaviors should be minimal, but are highlighted in the documentation.

jexp:
Cool thanks!

PR #748, opened 2018-02-28
Fixes #747: date.format(null) returns null, also add ISO8601 convenience format

### LINKED ISSUE #747: apoc.data.format(null) should return null.

jasperblues:
- date.format(null) returns null, instead of throw
- add ISO8601 convenience format function, ISO8601 is commonly used, eg in Neo4j OGM, API endpoints, etc.


jasperblues:
Fixes #747

PR #753, opened 2018-03-05
Add missing apoc.coll.combinations() to 3.3

krisgeus:
apoc.coll.combinations was missing in the code of version 3.3 while being in the docs.
Code was added in version 3.2 but only partially merged into 3.3

This code is based on the original from contributor InverseFalcon


InverseFalcon:
Looks good to me. Merging.

PR #760, opened 2018-03-12
Add apoc.diff user function

binarycoded:
The new procedure `apoc.diff.nodes` returns an overview of all node property differences of two nodes given by their nodeIds. The new functionality is documented in a new `*.adoc`-file (with usage and example Cypher) and covered by a unit test.

### DIFF
--- a/src/main/java/apoc/diff/Diff.java
+++ b/src/main/java/apoc/diff/Diff.java
@@ -1,0 +1,65 @@
+package apoc.diff;
+
+import apoc.Description;
+import apoc.util.MapUtil;
+import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.graphdb.Node;
+import org.neo4j.procedure.*;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Benjamin Clauss
+ * @since 15.06.2018
+ */
+public class Diff {
+
+    @Context
+    public GraphDatabaseService db;
+
+    @UserFunction()
+    @Description("apoc.diff.nodes([leftNode],[rightNode]) returns a detailed diff of both nodes")
+    public Map<String, Object> nodes(@Name("leftNode") Node leftNode, @Name("rightNode") Node rightNode) {
+        Map<String, Object> allLeftProperties = leftNode.getAllProperties();
+        Map<String, Object> allRightProperties = rightNode.getAllProperties();
+
+        Map<String, Object> result = new HashMap<>();
+        result.put("leftOnly", getPropertiesOnlyLeft(allLeftProperties, allRightProperties));
+        result.put("rightOnly", getPropertiesOnlyLeft(allRightProperties, allLeftProperties));
+        result.put("inCommon", getPropertiesInCommon(allLeftProperties, allRightProperties));
+        result.put("different", getPropertiesDiffering(allLeftProperties, allRightProperties));
+
+        return result;
+    }
+
+    private Map<String, Object> getPropertiesOnlyLeft(Map<String, Object> left, Map<String, Object> right) {
+        Map<String, Object> leftOnly = new HashMap<>();
+        leftOnly.putAll(left);
+        leftOnly.keySet().removeAll(right.keySet());
+        return leftOnly;
+    }
+
+    private Map<String, Object> getPropertiesInCommon(Map<String, Object> left, Map<String, Object> right) {
+        Map<String, Object> inCommon = new HashMap<>(left);
+        inCommon.entrySet().retainAll(right.entrySet());
+        return inCommon;
+    }
+
+    private Map<String, Map<String, Object>> getPropertiesDiffering(Map<String, Object> left, Map<String, Object> right) {
+        Map<String, Map<String, Object>> different = new HashMap<>();
+        Map<String, Object> keyPairs = new HashMap<>();
+        keyPairs.putAll(left);
+        keyPairs.keySet().retainAll(right.keySet());
+
+        for (Map.Entry<String, Object> entry : keyPairs.entrySet()) {
+            if (!left.get(entry.getKey()).equals(right.get(entry.getKey()))) {
+                Map<String, Object> pairs = new HashMap<>();
+                pairs.put("left", left.get(entry.getKey()));
+                pairs.put("right", right.get(entry.getKey()));
+                different.put(entry.getKey(), pairs);
+            }
+        }
+        return different;
+    }
+}
--- a/docs/diff.adoc
+++ b/docs/diff.adoc
@@ -1,0 +1,33 @@
+= Diff
+
+Diff is a user function to return a detailed diff between two nodes.
+
+*apoc.diff.nodes([leftNode],[rightNode])*
+
+Example
+[source,cypher]
+CREATE
+    (n:Person{name:'Steve',age:34, eyes:'blue'}),
+    (m:Person{name:'Jake',hair:'brown',age:34})
+WITH n,m
+return apoc.diff.nodes(n,m)
+
+Resulting JSON body:
+[source,json]
+{
+  "leftOnly": {
+    "eyes": "blue"
+  },
+  "inCommon": {
+    "age": 34
+  },
+  "different": {
+    "name": {
+      "left": "Steve",
+      "right": "Jake"
+    }
+  },
+  "rightOnly": {
+    "hair": "brown"
+  }
+}
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -120,6 +120,8 @@ include::number.adoc[leveloffset=2]
 
 include::exact.adoc[leveloffset=2]
 
+include::diff.adoc[leveloffset=2]
+
 == Graph Algorithms
 
 include::algo.adoc[leveloffset=2]
--- a/src/test/java/apoc/diff/DiffTest.java
+++ b/src/test/java/apoc/diff/DiffTest.java
@@ -1,0 +1,113 @@
+package apoc.diff;
+
+import apoc.util.TestUtil;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.test.TestGraphDatabaseFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Benjamin Clauss
+ * @since 15.06.2018
+ */
+public class DiffTest {
+
+    private static Node node1;
+    private static Node node2;
+    private static Node node3;
+
+    private static GraphDatabaseService db;
+
+    @BeforeClass
+    public static void setup() throws Exception {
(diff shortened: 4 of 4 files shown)


jexp:
Hi Benjamin,

thanks a lot for your contribution.

1. Please sign our CLA if you haven't yet:
2. I think apoc.diff.nodes should be also available for maps in general and relationships
3. I'm a bit torn about it being a procedure vs. function, perhaps we could add a boolean function that returns true / false 
4. do we really need the guava dependency? I mean after all it is just 2 x entrySet.removeAll for disjuction, and 1x entrySet.retainAll  for intersection
5. please change the test to run with embedded not spinning up a neo4j server otherwise it slows down the testing too much

Thanks
Michael

jexp:
Sorry fogot the CLA: https://neo4j.com/developer/cla/

binarycoded:
Hi Michael,

thanks for your feedback!

Short question on point 3: Do you think we should add both, a procedure that returns detailed diff information and a function that only returns a boolean or just the function?

Thanks
Benjamin

jexp:
Yes both would be good.

the equality would be the same as `properties(n) = properties(m) AND labels(n) = labels(m)` or?

We could also add sub.functions that provide the difference, intersection, disjunctions of maps

Michael

binarycoded:
@jexp I just rebased to the last commit with the Travis fix - it's all green now! :+1:

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+ * @author Benjamin Clauss
+ * @since 29.03.2018
+ */
+public class Diff {
+
+    @Context
+    public GraphDatabaseService db;
+
+    @UserFunction
+    @Description("apoc.diff.maps([leftMap],[rightMap])" )
```
jexp: please add an explanation to all descriptions ?

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+    public boolean maps(@Name("leftMap") Map<String, Object> leftMap, @Name("rightMap") Map<String, Object> rightMap) {
+        return !leftMap.equals(rightMap);
+    }
+
+    @UserFunction
+    @Description("apoc.diff.properties([leftNodeId],[rightNodeId])")
+    public boolean properties(@Name("leftNodeId") Long leftNodeId, @Name("rightNodeId") Long rightNodeId) {
```
jexp: support nodes and not node-ids, if someone has node-ids they can use cypher to look them up.
as you want to use this from cypher you most often have nodes, and not node ids.

src/main/java/apoc/diff/Diff.java
jexp: it's not really a diff, it's just an equals, which is the same as  `WHERE leftMap = rightMap` in cypher, or?
I thought you'd have a function that outputs the differences between both, which then should go to `apoc.map.*`

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
+package apoc.diff;
```
jexp: perhaps rather call it apoc.equals.*

a diff is more detailed difference computation and output

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+                && (leftRelationship.getType().name().equals(rightRelationship.getType().name()))
+                && (leftRelationship.getAllProperties().equals(rightRelationship.getAllProperties()))) {
+            return false;
+        }
+        return true;
+    }
+
+    @Procedure(mode = Mode.READ)
+    @Description("apoc.diff.nodes([leftNodeId],[rightNodeId])")
+    public Stream<DiffResult> nodes(@Name("leftNodeId") Long leftNodeId, @Name("rightNodeId") Long rightNodeId) {
```
jexp: you could have this both as function and procedure, the function then would return a map with the fields of the DiffResult being keys.

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+        result.inCommon = getPropertiesInCommon(leftNode, rightNode);
+        result.differing = getPropertiesDiffering(leftNode, rightNode);
+
+        return Stream.of(result);
+    }
+
+    private Map<String, Object> getPropertiesLeftOnly(Node leftNode, Node rightNode) {
+        Map<String, Object> leftOnly = new HashMap<>();
+        leftOnly.putAll(leftNode.getAllProperties());
+        leftOnly.putAll(rightNode.getAllProperties());
```
jexp: don't need it
also you should access the node's properties only once per node, it's an expensive operation
so all these private methods should take Maps instead.

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+        leftOnly.keySet().removeAll(rightNode.getAllProperties().keySet());
+        return leftOnly;
+    }
+
+    private Map<String, Object> getPropertiesRightOnly(Node leftNode, Node rightNode) {
```
jexp: it's the same as `getPropertiesLeftOnly` just reversed parameters. no need for two methods :)

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+        rightOnly.putAll(rightNode.getAllProperties());
+        rightOnly.keySet().removeAll(leftNode.getAllProperties().keySet());
+        return rightOnly;
+    }
+
+    private Map<String, Object> getPropertiesInCommon(Node leftNode, Node rightNode) {
+        Map<String, Object> inCommon = new HashMap<>();
+        inCommon.putAll(leftNode.getAllProperties());
+        inCommon.keySet().retainAll(rightNode.getAllProperties().keySet());
+        inCommon.entrySet().retainAll(rightNode.getAllProperties().entrySet());
```
jexp: not needed.

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+        Map<String, Object> rightOnly = new HashMap<>();
+        rightOnly.putAll(leftNode.getAllProperties());
+        rightOnly.putAll(rightNode.getAllProperties());
+        rightOnly.keySet().removeAll(leftNode.getAllProperties().keySet());
+        return rightOnly;
+    }
+
+    private Map<String, Object> getPropertiesInCommon(Node leftNode, Node rightNode) {
+        Map<String, Object> inCommon = new HashMap<>();
+        inCommon.putAll(leftNode.getAllProperties());
```
jexp: you can just pass it to the constructor

src/main/java/apoc/result/DiffResult.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * @author Benjamin Clauss
+ * @since 29.03.2018
+ */
+public class DiffResult {
+
+    public Map<String, Object> leftOnly;
+    public Map<String, Object> rightOnly;
+    public Map<String, Object> inCommon;
+    public Map<String, Map<String, Object>> differing;
```
jexp: different

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+        inCommon.keySet().retainAll(rightNode.getAllProperties().keySet());
+        inCommon.entrySet().retainAll(rightNode.getAllProperties().entrySet());
+        return inCommon;
+    }
+
+    private Map<String, Map<String, Object>> getPropertiesDiffering(Node leftNode, Node rightNode) {
```
jexp: again use maps as params

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+        Map<String, Map<String, Object>> differing = new HashMap<>();
+        Map<String, Object> keyPairs = new HashMap<>();
+        keyPairs.putAll(leftNode.getAllProperties());
+        keyPairs.keySet().retainAll(rightNode.getAllProperties().keySet());
+
+        Iterator it = keyPairs.entrySet().iterator();
```
jexp: use foreach

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+        while (it.hasNext()) {
+            Map.Entry entry = (Map.Entry)it.next();
+            if (!leftNode.getAllProperties().get(entry.getKey()).equals(rightNode.getAllProperties().get(entry.getKey()))) {
```
jexp: you already have it in keyPairs

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+                Map<String, Object> pairs = new HashMap<>();
```
jexp: Probably eaiser to use a 2 element array or list

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+                pairs.put("leftValue", leftNode.getAllProperties().get(entry.getKey()));
+                pairs.put("rightValue", rightNode.getAllProperties().get(entry.getKey()));
+                differing.put(entry.getKey().toString(), pairs);
```
jexp: keys are always strings

src/main/java/apoc/diff/Diff.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private Map<String, Map<String, Object>> getPropertiesDiffering(Node leftNode, Node rightNode) {
+        Map<String, Map<String, Object>> differing = new HashMap<>();
+        Map<String, Object> keyPairs = new HashMap<>();
+        keyPairs.putAll(leftNode.getAllProperties());
+        keyPairs.keySet().retainAll(rightNode.getAllProperties().keySet());
+
+        Iterator it = keyPairs.entrySet().iterator();
+        while (it.hasNext()) {
+            Map.Entry entry = (Map.Entry)it.next();
```
jexp: missing generics, they come with foreach

src/main/java/apoc/diff/Diff.java
jexp: btw. Java equality is not the same as Cypher equality

e.g. in Cypher all these are equal `(int)1 == long)1 == (byte)1 == (float)1.0 == (double)1.0`

jexp:
@binarycoded ping :)

binarycoded:
@jexp I'm still here! :) Will have a look at your comments asap.

jexp:
ping again :)

jexp:
I'd really love to merge this in :) Please have a look at my feedback. Thank you!!

binarycoded:
Hey @jexp, I hope I find some time in the next few days ...

binarycoded:
Hi @jexp 
I tried to reduce this PR to the minimum. I was able to just create a user function for my diff - no procedure. All other additional functionality should be moved to a new PR I think. Would be great if we could finally close this one. ;)
Best regards

src/main/java/apoc/diff/Diff.java
binarycoded: reworked

jexp:
Looks great, thanks so much.

Issue #768, opened 2018-03-21
apoc.cypher.runTimeboxed may terminate other cypher statements


anti-peng:
I'm running neo4j-enterprise-3.2.3 on Ubuntu 14.04, leave `dbms.transaction.timeout` and `dbms.rest.transaction.idle_timeout` as default, with APOC APOC 3.2.3.5 installed.
One of  the query statement is requested through HTTP REST API with a high request rate.
```
CALL apoc.cypher.runTimeboxed("cypher", params,  1000 * 20)
YIELD value RETURN value
```
Neo4j log keeps output `WARN [o.n.k.i.p.Procedures] marking KernelTransaction[statementLocks == null] for termination`.

The cypher statement works fine, terminal a long time cost query in time, but it sometimes cause other cypher query statement a termination with error
```
{
  "code": "Neo.DatabaseError.Statement.ExecutionFailed",
  "message": "The transaction has been terminated. Retry your operation in a new transaction, and you should see a successful result. Explicitly terminated by the user. ",
[...]

PR #774, opened 2018-03-28
fixes #720: apoc.cypher.runTimeboxed uses separate thread

### LINKED ISSUE #720: apoc.cypher.runTimeboxed() not terminating transaction in Neo4j 3.3.1, 3.3.2
Specific to 3.3.1 and 3.3.2, procedure is working fine in 2.2.9 and 3.3.0.

In 3.3.1 the procedure is not terminating long-running queries as designed.

To replicate, on a new 3.3.1 or 3.3.2 instance, create the movies graph from `:play movies`.

Execute the following and observe the hang:

```
call apoc.cypher.runTimeboxed("match (n)-[*]-(m)
return n, m", {}, 1000) yield value
return value
```

You may need to use :queries to terminate the hanging query.

sarmbruster:


sarmbruster:
this PR should be cherry-picked to 3.2 as well.
Will fix #768 as well.

src/main/java/apoc/cypher/Timeboxed.java
```diff
@@ -24,26 +24,85 @@
...
-
     @Context
     public GraphDatabaseService db;
-    @Context
-    public KernelTransaction tx;
 
     @Context
     public Log log;
 
+    private static Map<String,Object> POISON = Collections.singletonMap("__magic", "POISON");
```
jexp: final?

src/main/java/apoc/cypher/Timeboxed.java
```diff
@@ -24,26 +24,85 @@
...
+        });
+
+        //
         Pools.SCHEDULED.schedule(() -> {
-            String txString = tx == null ? "<null>" : tx.toString();
-            log.warn("marking " + txString + " for termination");
-            tx.markForTermination(Status.Transaction.Terminated);
+            Transaction tx = txAtomic.get();
+            if (tx==null) {
+                log.error("oops, tx is null, maybe other thread already finished?");
```
jexp: is that an error?

src/main/java/apoc/cypher/Timeboxed.java
```diff
@@ -24,26 +24,85 @@
...
+
+        // run query to be timeboxed in a separate thread to enable proper tx termination
+        // if we'd run this in current thread, a tx.terminate would kill the transaction the procedure call uses itself.
+        Pools.DEFAULT.submit(() -> {
+            try (Transaction tx = db.beginTx()) {
+                txAtomic.set(tx);
+                Result result = db.execute(cypher, params == null ? Collections.EMPTY_MAP : params);
+                while (result.hasNext()) {
+                    final Map<String, Object> map = result.next();
+                    queue.put(map);
```
jexp: what happens if the blocking queue is not emptied?

src/main/java/apoc/cypher/Timeboxed.java
sarmbruster: sure :-)

src/main/java/apoc/cypher/Timeboxed.java
sarmbruster: this could potentially happen if you set the timeout value so small that the other transaction has not yet been started. I consider it an error. Will amend the message to something more meaningful

src/main/java/apoc/cypher/Timeboxed.java
sarmbruster: good find. It should not happen that the queue is not emptied. I'll switch from `put` to `offer` and throw a RuntimeException if the queue doesn't accept.

Issue #738, opened 2018-02-09
apoc.index.search return too little result


vstruchkov:
I work with russian string and use apoc.search to increase search speed instead of contain function.
I make two different query
call apoc.index.search('NameIndex','Лен*') yield node, weight match (p:Person)-[:Name]->(node) return id(p) as id, node.value as value, id(node) as nid, weight
And receive one string  (Result1)
╒═════╤══════════════╤═════╤════════╕
│"id" │"value" │"nid"│"weight"│
╞═════╪══════════════╪═════╪════════╡
│11619│"Лена Амбьёрн"│11514│1 │
└─────┴──────────────┴─────┴────────┘
After that I add one russian letter 'и' to search string
call apoc.index.search('NameIndex','Лени*') yield node, weight match (p:Person)-[:Name]->(node) return id(p) as id, node.value as value, id(node) as nid, weight
╒═══════╤══════════════╤═══════╤════════╕
│"id" │"value" │"nid" │"weight"│
╞═══════╪══════════════╪═══════╪════════╡
│1086330│"Лениа Кунени"│1086359│1 │
└───────┴──────────────┴───────┴────────┘
And receive other string (Result2)

Why Result2 is not included in Result1? Pattern 'Лен*' corresponds to all query results

jexp:
Can you try to use this analyiser when creating the index:

http://lucene.apache.org/core/4_2_0/analyzers-common/org/apache/lucene/analysis/ru/RussianAnalyzer.html

you can pass this kind of config to the createIndex procedure:

https://neo4j.com/docs/java-reference/current/indexing/#indexing-create-advanced

vstruchkov:
Michael, 
Using lucene with supporting multilanguage is very complicated.
I found several identical solutions. Look at this page http://proiot.ru/blog/posts/2012/10/27/lucene-iazykovye-analizatory/ .Skip russian description and look at the code
This example show who to support russian and english languages simultaneously

jexp:
Would you mind helping us add documentation for it as a pull request?

PR #786, opened 2018-04-11
Update fulltext.adoc

vstruchkov:
I think that realisation of this feature fix this issue https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/738

### DIFF
--- a/docs/fulltext.adoc
+++ b/docs/fulltext.adoc
@@ -146,7 +146,7 @@ All https://neo4j.com/docs/java-reference/3.1/#indexing-create-advanced[standard
 | name | value | description
 | type | fulltext/exact | type of the index
 | to_lower_case | false/true | if terms should be converted to lower case before indexing
-| analyzer | classname | classname of lucene analyzer to be used for this index
+| analyzers | classnames | list of classname of lucene analyzer to be used for this index
 | similarity | classname | classname for lucene similarity to be used for this index
 | autoUpdate | true/false | if this index should be tracked for graph updates
 |===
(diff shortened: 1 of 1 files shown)


docs/fulltext.adoc
```diff
@@ -146,7 +146,7 @@ All https://neo4j.com/docs/java-reference/3.1/#indexing-create-advanced[standard
 | name | value | description
 | type | fulltext/exact | type of the index
 | to_lower_case | false/true | if terms should be converted to lower case before indexing
-| analyzer | classname | classname of lucene analyzer to be used for this index
+| analyzers | classnames | list of classname of lucene analyzer to be used for this index
```
jexp: why the change in the config name?

docs/fulltext.adoc
```diff
@@ -252,3 +252,29 @@ Again we're search for "Johnny", expecting now two of them:
 CALL apoc.index.search("people", "Johnny") YIELD node, weight
 RETURN node.name, weight
 ----
+
+=== Fulltext Index and multilingual string
+By default standart configuration of fultext index use standart analyzer for extract token from text.
+For support fultext search on multilangual content we need to set additional configuration to parameter `<option>` in function `apoc.index.addAllNodes`.
+
+Make sure `apoc.multilingual.enabled=true` is set.
```
jexp: where is the defined in the code?

docs/fulltext.adoc
```diff
@@ -252,3 +252,29 @@ Again we're search for "Johnny", expecting now two of them:
...
+
+Bellow there is example of creating fulltext index with multilingual support
+[source,cypher]
+----
+CALL apoc.index.addAllNodes('locations',{
+  Person:  ["address"],
+  Address: ["address"]}, {analyzer:["English","Russian","Thai"]})
```
jexp: uses analyzer here without an `s`?

docs/fulltext.adoc
vstruchkov: It's mistake. The right string is "anayzer" but you must pay attention that analizer is array now

docs/fulltext.adoc
vstruchkov: I write "apoc.multilingual.enabled=true" if you decide to make multilingual index optional

docs/fulltext.adoc
vstruchkov: See my comment for first questions.

vstruchkov: In this code I found error. After parameter "analizer" I had to write correct class names for analyzer

docs/fulltext.adoc
jexp: But where is the code that uses it?

vstruchkov: There is no code that use it.

docs/fulltext.adoc
jexp: Could you update your PR then with the fixes?

jexp:
Hey @vstruchkov as Neo4j 3.5.x added fulltext procedures to the product we plan to deprecated and remove the ones in APOC so I guess we might not need to add this anymore. What do you think?

vstruchkov:
Cool. 
I test it in new version

PR #793, opened 2018-04-25
Add apoc.cypher.runFirstColumnWrites procedure

johnymontana:
Similar to apoc.cypher.runFirstColumn function, but used for writes (since user-defined functions are read-only). This is used in neo4j-graphql-js for handling mutations.

### DIFF
--- a/src/main/java/apoc/cypher/Cypher.java
+++ b/src/main/java/apoc/cypher/Cypher.java
@@ -2,12 +2,15 @@ package apoc.cypher;
 
 import apoc.Pools;
 import apoc.result.MapResult;
+import apoc.result.NodeResult;
 import apoc.util.FileUtils;
 import apoc.util.QueueBasedSpliterator;
 import apoc.util.Util;
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.QueryStatistics;
 import org.neo4j.graphdb.Result;
+import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.ResourceIterator;
 import org.neo4j.helpers.collection.Iterables;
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/cypher/Cypher.java
```diff
@@ -434,4 +437,15 @@ private static Collection asCollection(Object value) {
...
     }
+
+    @Procedure(name = "apoc.cypher.runFirstColumnWrites", mode = Mode.WRITE)
+    @Description("apoc.cypher.runFirstColumnWrites(statement, params) - executes write statement with given parameters, yields first column only (nodes)")
+    public Stream<NodeResult> runFirstColumnWrites(@Name("cypher") String statement, @Name("params") Map<String, Object> params) {
+        if (params == null) params = Collections.emptyMap();
+        try (Result result = db.execute(withParamMapping(statement, params.keySet()), params)) {
+            String firstColumn = result.columns().get(0);
+            ResourceIterator<Node> resultColumn = result.columnAs(firstColumn);
+            return resultColumn.stream().map(NodeResult::new);
```
jexp: what happens if the query doesn't return anything or no nodes?
How should it fail/finish then?
With no results, with a single null result?
Would it make more sense to just have an ObjectResult and have the consumer decide?
Alternatively we could add an optional param to apoc.cypher.run ? that says which column names or indexes to return? We could also use map-deconstruction on 

```
CALL apoc.cypher.run(query,params) yield row
WITH apoc.map.values(row,[index/name])[0]
```

I kinda  want to avoid a too specific procedure.

Btw. please note that this can't be inlined like we do with functions within pattern comprehensions.

PR #796, opened 2018-04-27
Added whitelistNodes and blacklistNodes

InverseFalcon:
This finishes up node-specific filtering for path expanders.


InverseFalcon:
The failed checks don't look related to my changes.

src/test/java/apoc/path/ExpandPathTest.java
```diff
@@ -325,96 +325,4 @@ public void testCompoundLabelWorksInBlacklist() {
 					assertEquals("Gene Hackman", node.getProperty("name"));
 				});
 	}
-
```
jexp: should we have equivalent tests for the new behavior?

src/main/java/apoc/path/PathExplorer.java
```diff
@@ -243,116 +261,49 @@ public static Traverser traverse(TraversalDescription traversalDescription,
...
-			Node node = path.endNode();
-			boolean belowMinLevel = depth < minLevel;
-
-			// if start node shouldn't be filtered, exclude/include based on if using termination/endnode filter or not
-			// minLevel evaluator will separately enforce exclusion if we're below minLevel
-			if (depth == 0 && (!filterStartNode || !beginSequenceAtStart)) {
-				return whitelistAllowedEvaluation;
+			if (!whitelistNodes.isEmpty()) {
+				// ensure endNodes and terminatorNodes are whitelisted
+				whitelistNodes.addAll(endNodes);
```
jexp: is this a Set?

src/main/java/apoc/path/PathExplorer.java
```diff
@@ -243,116 +261,49 @@ public static Traverser traverse(TraversalDescription traversalDescription,
...
+			if (!blacklistNodes.isEmpty()) {
+				td = td.evaluator(NodeEvaluators.blacklistNodeEvaluator(blacklistNodes));
 			}
 
-			// if true for one matcher, need to set true for all matchers
-			if (endNodesOnly) {
-				for (LabelMatcherGroup group : sequenceMatchers) {
-					group.setEndNodesOnly(endNodesOnly);
-				}
+			Evaluator endAndTerminatorNodeEvaluator = NodeEvaluators.endAndTerminatorNodeEvaluator(endNodes, terminatorNodes);
```
jexp: I think you can add multiple, individual evaluators, but then you are bound by the order and prioritization of the traversal-framework.

jexp:
Looks good, I'll merge. But if possible please add a few tests if the new functionality is not yet covered enough.

InverseFalcon:
Just noting that the commits/functionality here didn't make it into any releases yet as of August 1 2018

jexp:
@InverseFalcon can we see to add this to the 3.4 / 3.5 branches too, sorry that I missed it in cherry picking back then?

InverseFalcon:
@marccelani 

whitelistNodes and blacklistNodes are present and working in 3.5.0.4, so they did get it, I'll check for what other versions contain the additions.

As for tests, there's a new test file I added for the node filters here:
https://github.com/InverseFalcon/neo4j-apoc-procedures/blob/683fec97bb4098ce04a2b40f9e77b1d97467dd0c/src/test/java/apoc/path/NodeFilterTest.java

If you have an example case that demonstrates a breakage with terminatorNodes, can you create a new issue for that for us to investigate?

InverseFalcon:
I believe the release notes are accurate, I can see proper behavior for whitelistNodes and blacklistNodes in 3.3.0.4 and 3.4.0.2, when they were first introduced.

@marccelani If you were using a version after these, please create a new issue and let us know the version of Neo4j and APOC you were using, and a sample for recreating the behavior you noted.

PR #800, opened 2018-05-02
Fixes #797 - sortMulti now orders by ASC by default

### LINKED ISSUE #797: Inconsistent ordering in apoc.coll.sort*
There is an inconsistency in the default way the `apoc.coll.sort` methods sort, bearing in mind that the default for Cypher is to order in Ascending, I would expect _all_ the sort methods to also sort in that way by default.

`apoc.coll.sort` orders in ASC 
`apoc.coll.sortMaps` orders in ASC
`apoc.coll.sortMulti` orders in **DESC**
`apoc.coll.sortNodes` orders in ASC

I think they should all be ASC, as the default is.

Fix wise - it's pretty simple, the problem is - it would be a **breaking** change, as anyone who is currently using `sortMulti` and expecting it to work in it's current way would have problems.

If it's something that _should_ be fixed, let me know and I'll get on it, but I won't work on it unless it's decided it's a 'goer'. :)

cskardon:
Now matches the other sort methods, and the default for Cypher.

### DIFF
--- a/src/main/java/apoc/coll/Coll.java
+++ b/src/main/java/apoc/coll/Coll.java
@@ -701,7 +701,7 @@ public class Coll {
     }
 
     @UserFunction("apoc.coll.sortMulti")
-    @Description("apoc.coll.sortMulti(coll, ['^name','age'],[limit],[skip]) - sort list of maps by several sort fields (ascending with ^ prefix) and optionally applies limit and skip")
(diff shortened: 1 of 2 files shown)


jexp:
Oh sorry, now I see what you meant. It was intentional actually here.

As 99% of the cases are descending sort.
And the `^` is meant to be an up-arrow to indicate "ascending".

cskardon:
OK, but isn't it a bit odd to have only one of the sort methods default that way, and the rest (including Cypher) default the other way?

PR #813, opened 2018-05-23
Allow passing parameters to `apoc.cypher.runFile`

thobe:
Parameters are passed by adding a `parameters: {...}` entry to the config parameter of the `runFile` procedure.


jexp:
Thanks a lot !!

PR #831, opened 2018-06-15
Remove unused imports

binarycoded:
I tried to cleanup the imports of different classes. There were so much unused imports that looked _ugly_.


binarycoded:
Seems no one is interested in that! :)

Issue #835, opened 2018-06-20
missing unit in apoc.date.field()


ArGeBre:
The function apoc.date.fields() returns a map of fields, including weekdays:
` {
    'weekdays': 5,
    'years': 2015,
    'seconds': 5,
    'zoneid': 'EET',
    'minutes': 4,
    'hours': 3,
    'months': 1,
    'days': 2
  }`

In contrast, the function apoc.date.field() is missing the capability to return the field weekdays. This would be very simple to add to apoc.date.Date class. See attached diff file:

[Date_diff.txt](https://github.com/neo4j-contrib/neo4j-apoc-procedures/files/2119795/Date_diff.txt)

jexp:
Can you please sign the CLA and send a PR? Thanks :)

PR #838, opened 2018-06-21
Added unit "weekday" to apoc.date.field()

ArGeBre:
This should fix issue #835. I have not been able to fully test it, because my tool chain is maven, not gradle based. But it's pretty straight forward. Also this ist my first pull request,


ArGeBre:
I already did - with my email to cla@neotechnology.com dated 21.06.2018. Do you want me to send it again?

PR #855, opened 2018-07-12
Fixes #852 - Added optional 4th parameter to procedure 'apoc.periodic.repeat'. (Branch 3.2)

### LINKED ISSUE #852: Make new apoc.periodic.repeat2 that support parameters
There is a need for a repeat procedure that can support parameters.

This should be defined as apoc.periodic.repeat2 to maintain backwards compatibility with the existing apoc.periodic.repeat. The signature for repeat2 should accept a parameters map.

alexiudice:
Added optional 4th parameter to procedure 'apoc.periodic.repeat' that allows for configuration to be passed in.
    
Added test for apoc.periodic.repeat that makes sure parameters passed in work as expected. Named testRepeatParams(). This PR is for branch 3.2.

### DIFF
--- a/src/main/java/apoc/periodic/Periodic.java
+++ b/src/main/java/apoc/periodic/Periodic.java
@@ -157,9 +157,10 @@ public class Periodic {
     }
 
     @Procedure(mode = Mode.WRITE)
-    @Description("apoc.periodic.repeat('name',statement,repeat-rate-in-seconds) submit a repeatedly-called background statement")
-    public Stream<JobInfo> repeat(@Name("name") String name, @Name("statement") String statement, @Name("rate") long rate) {
-        JobInfo info = schedule(name, () -> Iterators.count(db.execute(statement)),0,rate);
+    @Description("apoc.periodic.repeat('name',statement,repeat-rate-in-seconds, config) submit a repeatedly-called background statement. Fourth parameter 'config' is optional.")
+    public Stream<JobInfo> repeat(@Name("name") String name, @Name("statement") String statement, @Name("rate") long rate, @Name(value = "config", defaultValue = "{}") Map<String,Object> config ) {
+        Map<String,Object> params = (Map)config.getOrDefault("params", Collections.emptyMap());
+        JobInfo info = schedule(name, () -> Iterators.count(db.execute(statement, params)),0,rate);
         return Stream.of(info);
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/periodic/Periodic.java
```diff
@@ -163,6 +163,13 @@ private long executeNumericResultStatement(@Name("statement") String statement,
         return Stream.of(info);
     }
 
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.periodic.repeat2('name',statement,repeat-rate-in-seconds, config) submit a repeatedly-called background statement with configuration")
```
jexp: you don't need a second procedure

just add config to the regular one and default it to an empty map

src/main/java/apoc/periodic/Periodic.java
```diff
@@ -163,6 +163,13 @@ private long executeNumericResultStatement(@Name("statement") String statement,
...
+    public Stream<JobInfo> repeat2(@Name("name") String name, @Name("statement") String statement, @Name("rate") long rate, @Name("config") Map<String,Object> config ) {
+        JobInfo info = schedule(name, () -> Iterators.count(db.execute(statement, config)),0,rate);
```
jexp: don't call execute with the config directly, either rename it to `params`
or add a `params` entry into `config`

I would want the 2nd.

bradnussbaum: @jexp Okay so preference is pass in config.params?

src/main/java/apoc/periodic/Periodic.java
bradnussbaum: @jexp Okay will do.

src/main/java/apoc/periodic/Periodic.java
alexiudice: Ok, so similar to how apoc.periodic.iterate does it.

src/main/java/apoc/periodic/Periodic.java
alexiudice: If we only use one procedure wouldn't it result in a breaking change?

alexiudice: @jexp If we did that wouldn't it cause a breaking change? We would be going from 3 parameters to 4 parameters, even if we set the default value to an empty map.

PR #860, opened 2018-07-13
fixes #699 - implement functions from pgsql

### LINKED ISSUE #699: implement functions from pgsql
- [ ] apoc.label.exists(element, name) | boolean | returns true if the node or rel has the given label.
- [ ] apoc.coll.different(val1, val2, .., valn) | boolean | returns true if the values are all different, a function typically used for specifying isomorphic constraints (see Subgraph Isomorphism).
- [ ] apoc.node.degree[.out/in](node, [rel-type], [direction]) | exact numeric | returns the number of incoming neighbors.

AngeloBusato:


src/main/java/apoc/coll/Coll.java
```diff
@@ -768,4 +768,16 @@ public long occurrences(@Name("coll") List<Object> coll, @Name("item") Object it
 
         return combinations;
     }
+
+    @UserFunction
+    @Description("apoc.coll.different(values) - returns true if values are different")
+    public boolean different(@Name("values") List<Object> values) {
+        for(int i=0; i<values.size()-1;i++){
```
jexp: always use foreach
this one is just `return new HashSet(values).size() == values.size()`

src/main/java/apoc/label/Label.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class Label {
+
+    @Context public GraphDatabaseService db;
+
+    @UserFunction("apoc.label.exists")
+    @Description("apoc.label.exists(element, label) - returns true or false related to label existance")
+    public boolean exists(@Name("node") Object element, @Name("label") String label) {
+
+        return element instanceof Node?((Node) element).hasLabel(() -> label) :
```
jexp: while the lambda approach is cool, I'm not sure if it's more efficient htan just using `Label.label(label)`

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -347,6 +347,23 @@ public long degree(@Name("node") Node node, @Name(value = "types",defaultValue =
         return degree;
     }
 
+    @UserFunction("apoc.node.degree.in")
```
jexp: just use `node.getDegree(type, direction)`
please no manual counting.

src/main/java/apoc/nodes/Nodes.java
```diff
@@ -386,4 +403,31 @@ private int getDegreeSafe(Node node, RelationshipType relType, Direction directi
         return node.getDegree(relType, direction);
     }
 
-}
+    private long getTotalNeighbors(Node node, String types, Direction direction) {
```
jexp: remove this

jexp:
Thanks a lot!

PR #865, opened 2018-07-18
Fixes #864 - Adds text similarity/distance methods and double metaphone text encoder.

### LINKED ISSUE #864: Text Similarity/Distance Functions and Double Metaphone Text Encoder
There is a need for the following similarity/distance functions:
- Levenshtein similarity/distance
- Hamming distance
- Jaro-Winkler distance

and the need for a Double Metaphone text encoder.

The Levenshtein, Hamming, and Jaro-Winkler similarity/distance methods should be added to ‘Similarity.java’ in the ‘algo’ package. This will also require adding apache commons-text to build.gradle for imports used. 

The Double Metaphone text encoder should be added to ‘Phonetic.java’ in the ‘text’ package.

alexiudice:
Adds text similarity/distance methods and double metaphone text encoding.

Added Levenshtein Distance and Similarity code and test.

Added Hamming Distance code and test.

Added Jaro-Winkler Distance code and test.

Added Double Metaphone text encoding and test.

Added Apache commons-text dependency.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -127,6 +127,7 @@ dependencies {
     compile group: 'com.github.javafaker', name: 'javafaker', version: '0.10'
 
     compile group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
+    compile group: 'org.apache.commons', name: 'commons-text', version: '1.2'
     jmh group: 'org.neo4j', name: 'neo4j-lucene-index', version: neo4jVersionEffective
     jmh group: 'org.neo4j', name: 'neo4j-kernel', version: neo4jVersionEffective, classifier: "tests"
 
--- a/docs/overview.adoc
+++ b/docs/overview.adoc
@@ -879,7 +879,11 @@ Example: `'FRIEND|MENTORS>|<REPORTS_TO'` will match to :FRIEND relationships in 
 
 [cols="1m,5"]
 |===
-| apoc.text.distance(text1, text2) | compare the given strings with the StringUtils.distance(text1, text2) method
+| apoc.text.distance(text1, text2) | compare the given strings with the Levenshtein distance algorithm
+| apoc.text.levenshteinDistance(text1, text2) | compare the given strings with the Levenshtein distance algorithm
+| apoc.text.levenshteinSimilarity(text1, text2) | calculate the similarity (a value within 0 and 1) between two texts based on Levenshtein distance.
+| apoc.text.hammingDistance(text1, text2) | compare the given strings with the Hamming distance algorithm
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -127,6 +127,7 @@ dependencies {
     compile group: 'com.github.javafaker', name: 'javafaker', version: '0.10'
 
     compile group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
+    compile group: 'org.apache.commons', name: 'commons-text', version: '1.2'
```
jexp: how big is this dependency?

nammmm: Just checked on my machine, `commons-text-1.2.jar` is 136,544 bytes (139 KB on disk)

alexiudice:
I’ve updated the PR with the following changes:
   1. I've moved all of the added functions in `apoc.algo` to `apoc.text` and the tests from `SimilarityTest.java` to `StringsTest.java`. 

   2.  I've changed `apoc.text.distance` function a bit.  The  `apoc.text.levenshteinDistance` function has a more descriptive name so I kept it in. It now it just calls `apoc.text.distance`. I edited the `apoc.text.distance` so it does not use the deprecated method `StringUtils.getLevenshteinDistance`

3. All three distance functions were instantiating a new object every call which can be expensive in a loop. To remedy this I've replaced these with calls to three static final objects added in the Strings.java class so they are instantiated only one time.

jexp:
Great, thanks so much. 

For the reuse, it is only an issue if the object creation is really expensive (think jackson ObjectMapper) can you make sure the static instances are thread-safe? Otherwise please revert it back if it's not expensive to create the instances.

nammmm:
@jexp Hi Michael, I am actually doing some clean up for the code at the moment, so please don't merge it yet.

nammmm:
I think they are all thread safe as they are all `static final` fields. In addition, I just finished cleaning up the code (code style, description and parameter names).

jexp:
Can you please add it to the docs, at least to overview.adoc ?
Thanks so much.

alexiudice:
@jexp They should be thread safe as the Distance objects themselves are immutable after being instantiated. If you want to be really certain I can add a synchronized block when the `apply` methods of the Distance objects are being called, which I'm pretty sure will guarantee thread safety.

Other than that this PR should be ready for merge.

alexiudice:
I'm pretty sure the added methods are thread safe since the Distance objects are immutable once instantiated and the `apply` methods parameters are `final`. We could add a synchronized block around the `apply` method calls if we really want to make sure they are thread safe.

nammmm:
I have added the functions to overview.adoc. Let me know if there is anything else I need to do.

alexiudice:
Please do not merge yet! Testing some last things.

PR #871, opened 2018-07-27
fixes #829 - Add null string check and optional error message for apoc.spatial.geocode* functions

### LINKED ISSUE #829: Add null string check and optional error message for apoc.spatial.geocode* functons
A null string sent to apoc.spatial.geocode will give the error.  

`CALL apoc.spatial.geocodeOnce(NULL) YIELD location RETURN *)
Failed to invoke procedure apoc.spatial.geocode: Caused by: java.lang.NullPointerException`

This is possible when the spatial procedures are called with a property that doesn't exist.  Suggestions to address and avoid having to do a lot of pre-validating of data:

1. check to see if the input string is null and return error or zero length string, and do not call the geospatial provider
2. provide a parameter that would be an optional error message string if parameter is null
3. provide a parameter that can be the default geospatial lookup string if the input string is null

AngeloBusato:


jexp:
Instead of an exception and potentially rolling back the whole transaction, I'd rather have an error field in the result, so that the user can filter that out instead.

jexp:
needs rebase too

AngeloBusato:
@jexp PR updated

PR #874, opened 2018-07-31
fixes #679 - add an apoc.export.json procedure

### LINKED ISSUE #679: add an apoc.export.json procedure
That writes graph data into a json/graph-json file.

albertodelazzari:
Added also documentation with examples and tests of course


docs/exportJson.adoc
```diff
@@ -0,0 +1,208 @@
...
+[source,cypher]
+----
+CALL apoc.export.json.query("MATCH (u:User) WHERE u.age > {age} return u","/tmp/query.json",{params:{age:10}})
+----
+
+result:
+
+[source,json]
+----
+[ {
```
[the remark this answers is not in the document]
albertodelazzari: Do you mean a format like that?

```
[	
  	{
		"object_type": "node",
		"id": 1,
		"labels": ["Person"],
		"properties": {
			"key1": "value1",
			"keyN": "valueN"
		},
		"relationships": [{
			"object_type": "relationship",
			"direction": "INCOMING",
			"properties": {
				"key1": "value1",
				"keyN": "valueN"
			},
			"type": "KNOWS"
		}]
	},
	{
		"object_type": "relationship",
		"direction": "OUTGOING",
		"properties": {
			"key1": "value1",
			"keyN": "valueN"
		},
		"type": "KNOWS",
		"start_node": {
			"object_type": "node",
			"id": 1,
			"labels": ["Person"],
			"properties": {
				"key1": "value1",
				"keyN": "valueN"
			}
		},
		"end_node": {
			"object_type": "node",
			"id": 2,
			"labels": ["Person"],
			"properties": {
				"key1": "value1",
				"keyN": "valueN"
			}
		}
	}
]
```

For each object, we are serializing we also take into account the first level and serialize the related objects as well. For example, if we are serializing a node, we also serialize the node's relationships while if we are serializing a relationship we also serialize the relationship's start and end nodes.

src/main/java/apoc/export/json/serialize/NodeSerializer.java
```diff
@@ -0,0 +1,43 @@
...
+
+        jsonGenerator.writeArrayFieldStart("labels");
+        for (Label arg: node.getLabels()) {
+            jsonGenerator.writeString(arg.toString());
+        }
+        jsonGenerator.writeEndArray();
+        serializerProperty(jsonGenerator, node);
+        jsonGenerator.writeArrayFieldStart("relationships");
+        for (Relationship rel: node.getRelationships()) {
+            jsonGenerator.writeString(rel.toString());
```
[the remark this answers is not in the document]
albertodelazzari: Another json-graph format could be [JSON Graph Format](http://jsongraphformat.info)

docs/exportJson.adoc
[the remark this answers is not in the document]
jexp: I would only put in the id's for the start/end-node
and for the nodes I wouldn't put the relationships

and the format should have one json object per entity similar to twitter streaming or the yelp dataset, not one big object
so you can stream it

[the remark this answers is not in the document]
albertodelazzari: So you would have a streaming API from graph entities (or subgraphs) to JSON objects and then different "writers" that accept this stream of JSON objects and serialize them to different formats.
The default writer simply creates a file with an array of these JSON objects.

src/test/java/apoc/export/json/ExportJsonTest.java
```diff
@@ -0,0 +1,260 @@
...
+import java.io.File;
+import java.util.Map;
+import java.util.Scanner;
+
+import static apoc.util.MapUtil.map;
+import static org.junit.Assert.assertEquals;
+
+public class ExportJsonTest {
+
+    private static final String EXPECTED_QUERY_NODES = String.format("[ {\n" +
```
jexp: you can read this from a file.

src/test/java/apoc/export/json/ExportJsonTest.java
```diff
@@ -0,0 +1,260 @@
...
+            "  \"id\" : 0,\n" +
+            "  \"labels\" : [ \"User\" ],\n" +
+            "  \"properties\" : {\n" +
+            "    \"name\" : \"foo\",\n" +
+            "    \"age\" : \"42\",\n" +
+            "    \"male\" : \"true\",\n" +
+            "    \"kids\" : [ \"a\", \"b\", \"c\" ]\n" +
+            "  },\n" +
+            "  \"relationships\" : [ \"(0)-[KNOWS,0]->(1)\" ]\n" +
```
jexp: I don't want to see this format. anywhere, it's opaque. Please *do not* use `relationship.toString()`

src/main/java/apoc/export/json/serialize/NodeSerializer.java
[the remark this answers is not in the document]
jexp: `rel.toString()` is completely useless

src/main/java/apoc/export/json/JsonFormatSerializer.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        @Override
+        public void serializerProperty(JsonGenerator jsonGenerator, PropertyContainer propertyContainer) throws IOException {
+            jsonGenerator.writeObjectFieldStart("properties");
+            propertyContainer.getAllProperties().forEach((key, value)-> {
+                try {
+                    if(value.getClass().isArray()){
+                        jsonGenerator.writeArrayFieldStart(key);
+                        for(String property : (String[]) value){
```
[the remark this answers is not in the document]
jexp: You cannot assume that it's always a string array, please add tests for all datatypes, see: https://neo4j.com/docs/developer-manual/current/cypher/syntax/values/

src/main/java/apoc/export/json/JsonFormatSerializer.java
```diff
@@ -0,0 +1,74 @@
...
+                            jsonGenerator.writeString(property);
+                        }
+                        jsonGenerator.writeEndArray();
+                    } else {
+                        jsonGenerator.writeStringField(key, value.toString());
```
jexp: I don't think this is correct as JSON natively supports: String, Boolean, Numbers
Only datetime and spatial has to be formatted accordingly
for spatial we could use geojson (afaik there is an impl in neo4j-server Tx endpoint for serializing spatial types)
for time values

please try out to return differnt types via the tx-http-endpoint

```
curl -i http://localhost:7474/db/data/transaction/commit/ -H accept:application/json -H content-type:application/json -u neo4j:test -d'{"statements":{"statement":"return datetime()"}}'
curl -i http://localhost:7474/db/data/transaction/commit/ -H accept:application/json -H content-type:application/json -u neo4j:test -d'{"statements":[{"statement":"return datetime()"}]}'
curl -i http://localhost:7474/db/data/transaction/commit/ -H accept:application/json -H content-type:application/json -u neo4j:test -d'{"statements":[{"statement":"return date()"}]}'
curl -i http://localhost:7474/db/data/transaction/commit/ -H accept:application/json -H content-type:application/json -u neo4j:test -d'{"statements":[{"statement":"return time()"}]}'

curl -i http://localhost:7474/db/data/transaction/commit/ -H accept:application/json -H content-type:application/json -u neo4j:test -d'{"statements":[{"statement":"return point({latitude:53.2,longitude:18})"}]}'

curl -i http://localhost:7474/db/data/transaction/commit/ -H accept:application/json -H content-type:application/json -u neo4j:test -d'{"statements":[{"statement":"return localdatetime()"}]}' 
```

enabled with an _optional_ config parameter we could allow to store the type information for each field
ala

"data":...
"data_types": {"name":"string"} ... etc.

src/test/resources/exportJson/all.json
```diff
@@ -0,0 +1,4 @@
+{"object_type":"node","id":0,"labels":["User"],"properties":{"name":"foo","age":"42","male":"true","kids":["a","b","c"]}}
```
jexp: can we inline all these files? i.e. generate them from literal maps/collections + jackson?

jexp:
Ping?

AngeloBusato:
@jexp PR Updated

src/main/java/apoc/export/json/JsonFormatSerializer.java
conker84: Hi @jexp , for this one I create a separate issue ok?

jexp: yes, thanks.

src/main/java/apoc/export/json/JsonFormatSerializer.java
```diff
@@ -0,0 +1,86 @@
...
+            jsonGenerator.writeStringField("label", type);
+            serializeProperties(jsonGenerator, properties);
+            jsonGenerator.writeNumberField("start", startNodeId);
+            jsonGenerator.writeNumberField("end", endNodeId);
+            jsonGenerator.writeEndObject();
+        }
+
+        @Override
+        public void serializeProperties(JsonGenerator jsonGenerator, Map<String, Object> properties) throws IOException {
+            jsonGenerator.writeObjectFieldStart("data");
```
jexp: only if there are actually properties

src/main/java/apoc/export/json/JsonFormatSerializer.java
```diff
@@ -0,0 +1,86 @@
...
+
+    DEFAULT() {
+        @Override
+        public void writeNode(JsonGenerator jsonGenerator, long id, Iterable<Label> labels, Map<String,Object> properties) throws IOException {
+
+            jsonGenerator.writeStartObject();
+            jsonGenerator.writeStringField("type", "node");
+            jsonGenerator.writeNumberField("id", id);
+            jsonGenerator.writeArrayFieldStart("labels");
+            for (Label label : labels) { jsonGenerator.writeString(label.toString()); }
```
jexp: only if there are actually labels

docs/exportJson.adoc
```diff
@@ -0,0 +1,101 @@
...
+----
+
+result:
+
+[source,json]
+----
+{"type":"node","id":0,"labels":["User"],"data":{"name":"foo","age":"42","male":"true","kids":["a","b","c"]}}
+{"type":"node","id":1,"labels":["User"],"data":{"name":"bar","age":"42"}}
+{"type":"node","id":2,"labels":["User"],"data":{"age":"12"}}
+{"type":"relationship","id":0,"label":"KNOWS","properties":{},"start":0,"end":1}
```
jexp: no empty properties, btw. that should be "data" :)

docs/exportJson.adoc
```diff
@@ -0,0 +1,101 @@
...
```
jexp: no empty properties

docs/exportJson.adoc
```diff
@@ -0,0 +1,101 @@
...
+----
+
+result:
+
+[source,json]
+----
+{"type":"node","id":0,"labels":["User"],"data":{"name":"foo","age":"42","male":"true","kids":["a","b","c"]}}
+{"type":"node","id":1,"labels":["User"],"data":{"name":"bar","age":"42"}}
+{"type":"node","id":2,"labels":["User"],"data":{"age":"12"}}
+{"type":"relationship","id":0,"label":"KNOWS","data":{},"start":0,"end":1}
```
jexp: no empty data

docs/exportJson.adoc
```diff
@@ -0,0 +1,101 @@
...
+[source,cypher]
+----
+CALL apoc.export.json.query("MATCH (u:User) WHERE u.age > {age} return u","/tmp/query.json",{params:{age:10}})
+----
+
+result:
+
+[source,json]
+----
+{"data":{"u.age":42,"u.name":"foo","u.male":true,"u.kids":["a","b","c"],"labels(u)":["User"]}}
```
jexp: this is still wrong

docs/exportJson.adoc
```diff
@@ -0,0 +1,101 @@
...
+// tag::export.json[]
+`YIELD file, source, format, nodes, relationships, properties, time, rows`
+[cols="1m,5"]
+|===
+| apoc.export.json.query(query,file,config) | exports results from the Cypher statement as Json to the provided file
+| apoc.export.json.all(file,config) | exports whole database as Json to the provided file
+| apoc.export.json.data(nodes,rels,file,config) | exports given nodes and relationships as Json to the provided file
+| apoc.export.json.graph(graph,file,config) | exports given graph object as Json to the provided file
+|===
+// end::export.json[]
```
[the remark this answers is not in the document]
jexp: ping

src/test/java/apoc/export/json/ExportJsonTest.java
```diff
@@ -0,0 +1,252 @@
...
+        assertEquals(output.getAbsolutePath(), r.get("file"));
+        assertEquals("json", r.get("format"));
+        assertEquals(true, ((long) r.get("time")) >= 0);
+    }
+
+    private static void populateJsonResultsMap(){
+
+        jsonResults = new HashMap<>();
+
+        jsonResults.put("all", "{\"type\":\"node\",\"id\":0,\"labels\":[\"User\"],\"data\":{\"name\":\"foo\",\"age\":42,\"male\":true,\"kids\":[\"a\",\"b\",\"c\"]}}\n" +
```
jexp: I actually meant to populate them from regular maps and lists to avoid all these escaped quotes

docs/exportJson.adoc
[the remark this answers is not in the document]
AngeloBusato: added

AngeloBusato:
@jexp PR is updated and there is the fix of issue #919 too.

docs/exportJson.adoc
```diff
@@ -0,0 +1,109 @@
...
+[source,cypher]
+----
+CALL apoc.export.json.query("MATCH (u:User) WHERE age > {age} return u","/tmp/query.json",{params:{age:10}})
+----
+
+result:
+
+[source,json]
+----
+{"data":{"age":42,"name":"foo","male":true,"kids":["a","b","c"],"labels":["User"]}}
```
jexp: this is not corrct, it should be a json object with one field per column of the query

docs/exportJson.adoc
```diff
@@ -0,0 +1,109 @@
...
+|===
+
+
+== Examples
+
+For the examples below we create the following dataset:
+
+[source,cypher]
+----
+CREATE (f:User {name:'foo',age:42,male:true,kids:['a','b','c']})-[:KNOWS]->(b:User {name:'bar',age:42}),(c:User {age:12})
```
jexp: add a property {since: 1993}  to the relationship

docs/exportJson.adoc
jexp: also the structure of the data is wrong, it should be the same for graph entities as the usual export format

src/test/java/apoc/export/json/ExportJsonTest.java
```diff
@@ -0,0 +1,213 @@
...
+                    assertResults(output, r, "database");
+                }
+        );
+        assertEquals(JsonResults.expectedAllGraphData(), new Scanner(output).useDelimiter("\\Z").next());
+    }
+
+    @Test
+    public void testExportPointMapDatetimeJson() throws Exception {
+        String filename = "mapPointDatetime.json";
+        File output = new File(directory, filename);
```
jexp: add a test where the nested lists/ maps contains nodes/rels/paths

src/test/java/apoc/export/util/JsonResults.java
```diff
@@ -0,0 +1,273 @@
...
+                put("data_types",datatypes);
+            }
+        }};
+
+        return map;
+    }
+
+    public static Map<String, Object> nodeAge(boolean datatypes) throws JsonProcessingException {
+
+        Map<String, Object> map = new LinkedHashMap<String, Object>() {{
```
jexp: use Util.map()

jexp:
@conker84 did we align the format with neo4j-streams?
For id we use string there.

Also we need fix that "writeNodeProperties" thing.

conker84:
@jexp sorry after we talked about the management of composite keys I thought that we should put that on hold, we will update it.
Regarding the property **writeNodeProperties** we created it because the Path representation is just a detailed version of Relationship representation where the nodes are exported with their properties, following a [Gist](https://gist.github.com/conker84/29b1e72bdb492b359178446770670841) with an example.
If you want we can separate the code for path e relationship.

src/main/java/apoc/meta/Meta.java
```diff
@@ -61,6 +62,8 @@ public static Types of(Class<?> type) {
             if (Relationship.class.isAssignableFrom(type)) return RELATIONSHIP;
             if (Path.class.isAssignableFrom(type)) return PATH;
             if (Iterable.class.isAssignableFrom(type)) return LIST;
+            if (TemporalAccessor.class.isAssignableFrom(type)) return TEMPORAL;
```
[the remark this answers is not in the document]
jexp: Yes, please.

src/main/java/apoc/export/util/ExportConfig.java
```diff
@@ -110,4 +110,8 @@ private ExportFormat format(Object format) {
     public boolean streamStatements() {
         return streamStatements;
     }
+
+    public boolean writeNodeProperties() {
```
jexp: we should probably extract this once into field,
as it's used in the hot loop of the json generator

jexp:
Thanks a lot everyone @AngeloBusato  @conker84  @albertodelazzari

PR #886, opened 2018-08-14
Fix issue885. Special characters are not allowed in jdbc input url.

oaicstef:
The solution is not clean enough, because I'm using try catch to detect the use case. Once I know better why exactly we're using the object URL, the solution will be clearer.


jexp:
So it fails during URI parsing for the auth info? 

That's really weird, I thought that JDBC URLs always have to be correct URI's.

oaicstef:
I think that the JDBC connection string is not always a valid URL, infact with most providers it can contain user and password and the password can contain special characters like "$#!%@¨.
Furthermore the jdbc driver will work even if the connection string contains special characters.

With my solution I have 2 failing tests but I don´t see how they are related with the changes I added :-(

src/main/java/apoc/load/Jdbc.java
```diff
@@ -43,13 +44,18 @@
     public Log log;
 
     private static Connection getConnection(String jdbcUrl) throws Exception {
-        URI uri = new URI(jdbcUrl.substring("jdbc:".length()));
-        String userInfo = uri.getUserInfo();
-        if (userInfo != null) {
-            String[] user = userInfo.split(":");
-            String cleanUrl = jdbcUrl.substring(0,jdbcUrl.indexOf("://")+3)+jdbcUrl.substring(jdbcUrl.indexOf("@")+1);
-            return DriverManager.getConnection(cleanUrl, user[0], user[1]);
+        if (!jdbcUrl.contains("#")) {
```
jexp: I don't understand this check.

jexp: User Info can also be in `scheme://user:password@host`

oaicstef: Firstly, I'm sorry not to give you an answer sooner.

You are right, that check is not so understandable and it doesn't make a lot of sense, I pushed an old version of the correction. The issue happen when the jdbc connection string contains some special character as tested in the unit test Jdbc.testLoadJdbcWithSpecialChar .

In the last commit I just catch the URISyntaxException and in that case I skip the code that extract the user info from the URL, but I pass it directly to the driver. I know that neither this is a good solution because I'm thinking only now that the url can contain several kind of mistakes which are not related to this problem.  A more clean solution is needed I suppose.

oaicstef: I think that the method getConnection shouldn't use the URI class to parser the jdbc url, because the jdbc driver accept characters that are not allowed in a normal URI. The Jdbc driver itself is parsing the parameter passed in order to validate it depending on the database used.

oaicstef: Hi again, 
my code is making failing the CI but I didn't change anything else that could have broken 34 unit tests. Could you help me? The point is that we need to solve this bug in order to go on with my Company project. For now we're using our compiled jar, but in order to go to production, we need an official release.

Thank you in advance.
Stefano

conker84:
Hi @oaicstef and @jexp, the problem is that all after the first `#` is  managed by the [URI](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html) class as a _fragment_ so the test provided by `testLoadJdbcWithSpecialChar` is a fake positive because `@!6@;` is treated as fragment, "it works" because there is no authetication configuration on Derby DB
I think we can just remove the URI part and delegate all the Jdbc url parse to the `DriverManager.getConnection`, it's its job.
What do you think?

jexp:
@conker84 but then we have no means of passing credentials with the URL if the db driver doesn't allow it in a "custom" syntax.

So far we could use the URI getAuthInfo to extract username and password and pass it via the JDBC API.

conker84:
@jexp the problem is that if we have a URI like `scheme://user:pass#word@host`:
- `URI#getUserInfo` is null
- `URI#getAuthority` return `user:pass`

A more worse case is if we have an uri like `scheme://user:pass#w#ord@host`, this throws an `IllegalArgumentException`. Check this [java fiddle](https://repl.it/@conker84/Test-URI)

Maybe a better approach could be remove the URI part and provide for the cases where the db driver doesn't allowing pass credentials in the url a config map `{username: <user>, password: <pass>}` and manage it with `DriverManager #getConnection(String url, String user, String password)`.

jexp:
Yes, probably, your last suggestion.
Doesn't really make me happy though.

oaicstef:
> @jexp the problem is that if we have a URI like `scheme://user:pass#word@host`:
> 
> * `URI#getUserInfo` is null
> * `URI#getAuthority` return `user:pass`
> 
> A more worse case is if we have an uri like `scheme://user:pass#w#ord@host`, this throws an `IllegalArgumentException`. Check this [java fiddle](https://repl.it/@conker84/Test-URI)
> 
> Maybe a better approach could be remove the URI part and provide for the cases where the db driver doesn't allowing pass credentials in the url a config map `{username: <user>, password: <pass>}` and manage it with `DriverManager #getConnection(String url, String user, String password)`.

Thank you very much for your effort. I see that the issue is already been solved. I suppose I´ll see the change applied in the next release?

conker84:
Yes and thanks for your feedback!

PR #898, opened 2018-08-28
Support for both CB Server 4.x and 5.0 fixes #645

### LINKED ISSUE #645: couchbase connectivity
Hi

I need some guidance on how to use apoc.couchbase with couchbase 4.6 and 5.

5 has deprecated bucket passwords in favor of user passwords
4.6 has bucket passwords using sasl auth

Please provide an example of proper connection syntax for each version (eg with bucket passwords for 4.6 and without bucket passwords for 5.0)

https://forums.couchbase.com/t/running-n1ql-via-rest-api-with-bucket-password/12084/2

I'd also like some guidance on proper versions of java-client, core-io and rx-java for both 4.6 and 5.0

thanks for your help!

Michael

albertodelazzari:
- Added more tests to check different authentications between CB 4.x (bucket based) and 5.x (user based)
- Added "host or key" management (as MongoDB load procedures)
- Couchbase Java client updated to 2.5.9
- Bucket password is handled correctly
- check CB server version in order to handle different authentication properly
- Documentation updated


src/main/java/apoc/couchbase/Couchbase.java
```diff
@@ -71,389 +71,338 @@
...
+                                             @Name("documentId") String documentId) {
+        Stream<CouchbaseJsonDocument> result = null;
 
-    try (CouchbaseConnection couchbaseConnection = getCouchbaseConnection(nodes, bucket)) {
-      JsonDocument jsonDocument = couchbaseConnection.get(documentId);
-      if (jsonDocument != null) {
-        result = Stream.of(new CouchbaseJsonDocument(jsonDocument));
-      }
+        try (CouchbaseConnection couchbaseConnection = getCouchbaseConnection(hostOrKey, bucket)) {
+            JsonDocument jsonDocument = couchbaseConnection.get(documentId);
```
jexp: I wonder if couchbase has an API that returns a stream/optional or reactive type that we could just `map`.

jexp:
@graphadvantage could you have a look if this is what you need?

src/main/java/apoc/couchbase/Tuple2.java
```diff
@@ -0,0 +1,24 @@
+package apoc.couchbase;
+
+/**
+ * Created by alberto.delazzari on 23/08/2018.
+ */
+public final class Tuple2<T1, T2> {
```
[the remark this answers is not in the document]
albertodelazzari: Ok, I'll change with the internal one. Yes, it misses comparability. I did the simplest implementation just to have only a simple return type without using a map.

src/main/java/apoc/couchbase/Couchbase.java
albertodelazzari: Honestly, I have to take a look because I don't remember.

albertodelazzari:
Currently, we don't have a dedicated page for Couchbase. I just added the missing parts according to the changes but probably it's better to move all the section to a dedicated adoc file

PR #901, opened 2018-08-30
fixes #591 - Handle zip/tgz/tar.gz (compressed files) with !syntax to identify single files

### LINKED ISSUE #591: Handle zip/tgz/tar.gz (compressed files) with !syntax to identify single files
apoc.load.json("file://foo.zip!foo/bar.json")

AngeloBusato:


src/main/java/apoc/util/Util.java
```diff
@@ -281,32 +282,68 @@ private static String handleRedirect(URLConnection con, String url) throws IOExc
...
+
         if ("gzip".equals(encoding) || urlAddress.endsWith(".gz")) {
-            stream = new GZIPInputStream(stream);
+             return new GZIPInputStream(stream);
         }
         if ("deflate".equals(encoding)) {
-            stream = new DeflaterInputStream(stream);
+            return new DeflaterInputStream(stream);
         }
-        return new CountingInputStream(stream, sc.getLength());
```
jexp: as you remove CountingInputStream here, did you fix that in all places that call: `getInputStream` ?

src/main/java/apoc/util/Util.java
```diff
@@ -281,32 +282,68 @@ private static String handleRedirect(URLConnection con, String url) throws IOExc
 
     public static CountingInputStream openInputStream(String urlAddress, Map<String, Object> headers, String payload) throws IOException {
         StreamConnection sc;
+        InputStream stream;
+        if (urlAddress.contains(".zip") || urlAddress.contains(".tar") || urlAddress.contains(".tgz")) {
```
jexp: shouldn't we do a `contains("!")` check first here
or at least check `zipFileName` for null before passing it into `getFileStreamIntoCompressedFile`

src/main/java/apoc/util/Util.java
```diff
@@ -281,32 +282,68 @@ private static String handleRedirect(URLConnection con, String url) throws IOExc
...
-            stream = new DeflaterInputStream(stream);
+            return new DeflaterInputStream(stream);
         }
-        return new CountingInputStream(stream, sc.getLength());
+
+        return stream;
+    }
+
+    private static InputStream getFileStreamIntoCompressedFile(InputStream is, String fileName) throws IOException {
+        try (ZipInputStream zip = new ZipInputStream(is)) {
```
jexp: there is no null check on `fileName` !!!

jexp:
Thanks for adding all the tests.

src/main/java/apoc/util/Util.java
```diff
@@ -281,32 +282,68 @@ private static String handleRedirect(URLConnection con, String url) throws IOExc
...
+            sc = getStreamConnection(urlAddress,headers,payload);
+            stream = getFileStreamIntoCompressedFile(sc.getInputStream(), zipFileName);
+
+            return new CountingInputStream(stream, sc.getLength());
+        }
+
+        sc = getStreamConnection(urlAddress,headers,payload);
+        stream = getInputStream(sc,urlAddress);
+
+        return new CountingInputStream(stream, sc.getLength());
```
AngeloBusato: @jexp is here the CountingInputStream return.

src/main/java/apoc/util/Util.java
AngeloBusato: this is old, I done a small refactoring, i put I comment where is the new return countingInputStream, is inside the openInputStream method as before

AngeloBusato:
@jexp PR updated

src/main/java/apoc/util/Util.java
```diff
@@ -281,32 +282,72 @@ private static String handleRedirect(URLConnection con, String url) throws IOExc
 
     public static CountingInputStream openInputStream(String urlAddress, Map<String, Object> headers, String payload) throws IOException {
         StreamConnection sc;
+        InputStream stream;
+        if (urlAddress.contains("!")) {
+            String[] tokens = urlAddress.split("!");
+            urlAddress = tokens[0];
+            String zipFileName;
+            
+            if(tokens.length > 1) {
```
jexp: ==2

src/main/java/apoc/util/Util.java
```diff
@@ -281,32 +282,72 @@ private static String handleRedirect(URLConnection con, String url) throws IOExc
 
     public static CountingInputStream openInputStream(String urlAddress, Map<String, Object> headers, String payload) throws IOException {
         StreamConnection sc;
+        InputStream stream;
+        if (urlAddress.contains("!")) {
```
jexp: this whole branch should go into a separate, properly named method

AngeloBusato:
@jexp PR Updated

jexp:
Thanks a lot.

PR #903, opened 2018-08-30
Inform users what's wrong when triggers are not enabled

moxious:
At present, if apoc.trigger.enabled is not set, or set to false, and a user calls any of the APOC trigger procedures, they get a stacktrace and a NullPointerException.

This PR just sets a flag and checks prior to executing any trigger related function whether triggers are enabled.  If they are not, a RuntimeException is thrown that gives a meaningful message the user can act on.

Submitting this PR because I've run into it twice recently, it confused me both times.  Internal neo4j slack #apoc channel had two other users who got caught by same confusion, raising questions.

### DIFF
--- a/src/main/java/apoc/trigger/Trigger.java
+++ b/src/main/java/apoc/trigger/Trigger.java
@@ -56,23 +56,6 @@ public class Trigger {
 
     @Context public GraphDatabaseService db;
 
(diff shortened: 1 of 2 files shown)


jexp:
Cool, thanks so much.

PR #922, opened 2018-09-19
fixes #921: Updated apoc.cypher.doIt with config map and retry feature.

### LINKED ISSUE #921: Retry on failure for `apoc.cypher.doIt`
It would be useful for the apoc command `apoc.cypher.doIt` to be able to retry a query on transaction failure. This would be part of an optional configuration map added to the procedure.

A call could would look like
`CALL apoc.cypher.doIt("RETURN 1", {}, {retries: 3})`

This would try to run the query `RETURN 1` at most three times, and would return the result after the first successful transaction commit. 

Note the middle map `{}` is the "params" map and this issue looks to add a configuration map.

alexiudice:
Added optional config map to `apoc.cypher.doIt`. Currently only takes in a `retries` option that specifies the number of times the cypher query should be retried.

### DIFF
--- a/src/main/java/apoc/cypher/Cypher.java
+++ b/src/main/java/apoc/cypher/Cypher.java
@@ -380,8 +380,15 @@ public class Cypher {
 
     @Procedure(mode = WRITE)
     @Description("apoc.cypher.doIt(fragment, params) yield value - executes writing fragment with the given parameters")
-    public Stream<MapResult> doIt(@Name("cypher") String statement, @Name("params") Map<String, Object> params) {
+    public Stream<MapResult> doIt(@Name("cypher") String statement, @Name("params") Map<String, Object> params, @Name(value = "config",defaultValue = "{}") Map<String,Object> config) {
         if (params == null) params = Collections.emptyMap();
+
+        if (config.containsKey( "retries" ))
+        {
+            List<String> retryStatusCodes = (List<String>) config.getOrDefault( "retryStatusCodes", Collections.emptyList() );
+            return Util.runWithRetry( db, log, statement, (Long) config.get( "retries" ),retryStatusCodes, params);
+        }
+
         return db.execute(withParamMapping(statement, params.keySet()), params).stream().map(MapResult::new);
     }
 
--- a/src/main/java/apoc/util/Util.java
+++ b/src/main/java/apoc/util/Util.java
@@ -4,6 +4,8 @@ import apoc.ApocConfiguration;
 import apoc.Pools;
 import apoc.export.util.CountingInputStream;
 import apoc.path.RelationshipTypeAndDirections;
+import apoc.result.MapResult;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.neo4j.collection.primitive.PrimitiveLongIterator;
@@ -12,6 +14,7 @@ import org.neo4j.graphdb.Node;
 import org.neo4j.helpers.collection.Iterables;
 import org.neo4j.helpers.collection.Pair;
 import org.neo4j.internal.kernel.api.security.SecurityContext;
+import org.neo4j.kernel.api.exceptions.Status;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
 import org.neo4j.logging.Log;
 import org.neo4j.procedure.TerminationGuard;
@@ -23,6 +26,7 @@ import java.net.*;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.LockSupport;
 import java.util.function.Function;
 import java.util.stream.*;
 import java.util.zip.DeflaterInputStream;
@@ -30,6 +34,8 @@ import java.util.zip.GZIPInputStream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 
+import static apoc.cypher.Cypher.POOL;
+import static apoc.cypher.Cypher.withParamMapping;
 import static java.lang.String.format;
 
 /**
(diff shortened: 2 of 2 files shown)


src/main/java/apoc/cypher/Cypher.java
```diff
@@ -436,4 +445,39 @@ private static Collection asCollection(Object value) {
     public Stream<MapResult> doWhenCase(@Name("conditionals") List<Object> conditionals, @Name(value="elseQuery", defaultValue = "") String elseQuery, @Name(value="params", defaultValue = "") Map<String, Object> params) {
         return whenCase(conditionals, elseQuery, params);
     }
+
```
jexp: shouldn't this go somewhere else?
I think I have similar code already for periodic.iterate?

Also I'm not sure that 

1. you get an error from the pooled execution as it usually just kills the thread in the execution, most likely at Future.get and then you get an InvocationTargetException where you have unpack the nested exception
1. it is a transient error? Not many errors make sense to retry, mostly deadlocks only or if some pre-codition is not met yet.

alexiudice: Yes the code was taken from periodic.iterate and was repurposed for `apoc.cypher` calls. The periodic.iterate method I believe you are referring to is`iterateAndExecuteBatchedInSeparateThread`. It has parameters, such as batching and parallel processing which do not make sense in this context. This PR looks to allow a configurable retry. 

I can move this `runWithRetry` method to a Utils class if you think it would be better there.

1. Also I didn’t think the exception itself mattered so I didn't think it needed to be unpacked. It just matters that there is an exception and it is caught so a new transaction can be retried. 

2. Though there may be multiple uses, this PR has one specific use I was thinking of. This is useful for generating unique values in a trigger and then retrying if the value is already in use. 
For example, suppose we have a trigger set up called “addUniqueValueToCreatedNodes” that generates a `uuid` that must be unique (has a uniqueness CONSTRAINT on it) and adds it to created nodes. If a node is added to the graph the trigger will start and generate the `uuid` and try to add it.  If the generated `uuid` is already in use, the trigger will fail and the node will not be created. The ability to run the generation of `uuid` and being able to SET it on the node in `apoc.cypher.doIt` will allow for multiple retries in the trigger itself, so the node will be created. 

Here is the example in code:

`CALL apoc.trigger.add("addUniqueValueToCreatedNodes", "UNWIND {createdNodes} AS n CALL apoc.cypher.doIt(\"WITH {node} AS node SET node.uuid = 'uuid:' + apoc.create.uuid()\", {node: n}, {retries: 5}) YIELD value RETURN value", {phase: "after”})`

This would fail after 5 retries, making it much more likely the node would be created compared to the code below where there is no retry and a conflicting `uuid` would fail to create the node.

`CALL apoc.trigger.add("addUniqueValueToCreatedNodes", "UNWIND {createdNodes} AS n WITH n SET n.uuid = 'uuid:' + apoc.create.uuid()", {phase: "before”})`

src/main/java/apoc/cypher/Cypher.java
```diff
@@ -436,4 +445,39 @@ private static Collection asCollection(Object value) {
...
+                    {
+                        return null;
+                    }
+                    return consumer.apply( params );
+                } );
+
+                // Query has not finished yet, wait.
+                while ( !resultFuture.isDone() )
+                {
+                    LockSupport.parkNanos( 1000 );
```
alexiudice: Would a `Util.sleep` suffice here rather than `LockSupport.parkNanos`?

alexiudice:
I changed how the transaction was being executed, now the `runWithRetry` method is able to get the type of Neo4j error that occurs. 

For exceptions of type `QueryExecutionException` (the standard type of exception thrown by the graphdb)  I went through the different StatusCodes and created a whitelist for which failed transactions can be retried by default. Like you mentioned many exceptions do not make sense to retry. We can add or remove from the list, it was just a first pass through.
 ```
General.UnknownError,
Network.CommunicationError,
Procedure.ProcedureTimedOut,
Schema.ConstraintValidationFailed,
Schema.SchemaModifiedConcurrently,
Statement.ConstraintVerificationFailed,
Transaction.ConstraintsChanged,
Transaction.DeadlockDetected,
Transaction.InstanceStateChanged,
Transaction.LockAcquisitionTimeout,
Transaction.LockSessionExpired,
Transaction.Outdated,
Transaction.TransactionCommitFailed,
Transaction.TransactionStartFailed,
Transaction.TransactionTimedOut```
(Documentation can be found here: https://neo4j.com/docs/developer-manual/current/reference/status-codes/)

Something like a “SyntaxError” will not retried and the StatusCode of the exception is pushed up to the user.
I also made an optional parameter to pass in StatusCodes, so if you wanted to retry on a SyntaxError (or any other errors not in the whitelist) you could.

Lastly I moved all the methods into the Util class, it seemed like the best place for these methods.

I'm thinking the last thing this PR needs is to add to the documentation and maybe writing some tests. I've tested this method out manually in several contexts and it works. I can create some tests once I know the PR is not going to change too much.

@jexp Anything else that I should change? Did I address your concerns?

src/main/java/apoc/util/Util.java
```diff
@@ -686,4 +691,99 @@ public static boolean classExists(String className) {
...
+                    return result;
+                }
+                catch ( Exception e )
+                {
+                    throw e;
+                }
+            } );
+            Thread threadForTransaction = new Thread( resultFuture );
+
+            try
```
jexp: can you use the thread pool and not start manual threads?

src/main/java/apoc/util/Util.java
```diff
@@ -686,4 +691,99 @@ public static boolean classExists(String className) {
             return null;
         }
     }
+
+    public static Stream<MapResult> runWithRetry( GraphDatabaseService db, Log log, String statement, long retries, List<String> additionalRetryCodes,
+            Map<String,Object> params )
+    {
```
jexp: please use the regular apoc code format

src/main/java/apoc/util/Util.java
```diff
@@ -686,4 +691,99 @@ public static boolean classExists(String className) {
...
+                Transaction.TransactionCommitFailed,
+                Transaction.TransactionStartFailed,
+                Transaction.TransactionTimedOut
+        );
+    }
+
+    private static boolean isRetriableException( String statusCode, List<String> customRetriable )
+    {
+
+        if ( RetriableExceptions.QueryExecutionExceptionStatusCodeList.stream().map( sc -> sc.code().serialize().equals( statusCode ) ).reduce( ( x, acc ) -> acc || x ).get()
```
jexp: probably a loop + if with early return is easier here ?
there is also any/none() etc on streams

src/main/java/apoc/util/Util.java
```diff
@@ -686,4 +691,99 @@ public static boolean classExists(String className) {
...
+            transactionFailureMessage = "Failed after " + timesTried + " of " + retries + " retries.";
+        }
+        while ( timesTried <= retries );
+
+        throw new TransactionFailureException( transactionFailureMessage );
+    }
+
+    private interface RetriableExceptions extends Status
+    {
+        List<Status> QueryExecutionExceptionStatusCodeList = Arrays.asList(
```
jexp: why is this in an interface?
did you check if status has equals perhaps a set with contains is good enough?

src/main/java/apoc/util/Util.java
```diff
@@ -686,4 +691,99 @@ public static boolean classExists(String className) {
...
+    {
+        String transactionFailureMessage = "";
+        Long timesTried = 0L;
+        do
+        {
+
+            FutureTask<Result> resultFuture = new FutureTask<>( () -> {
+                try ( Transaction tx = db.beginTx() )
+                {
+                    Result result = ((Function<Map<String,Object>,Result>) ( Map<String,Object> p ) -> db.execute( withParamMapping( statement, p.keySet() ),
```
jexp: what is wrong with just calling it?

src/main/java/apoc/util/Util.java
```diff
@@ -686,4 +691,99 @@ public static boolean classExists(String className) {
...
+                            p )).apply( params );
+                    tx.success();
+                    return result;
+                }
+                catch ( Exception e )
+                {
+                    throw e;
```
jexp: this doesn't make sense?

src/main/java/apoc/util/Util.java
```diff
@@ -686,4 +691,99 @@ public static boolean classExists(String className) {
...
+
+            try
+            {
+                threadForTransaction.start();
+                threadForTransaction.join();
+
+                // Query has not finished yet, wait.
+                while ( !resultFuture.isDone() )
+                {
+                    LockSupport.parkNanos( 1000 );
```
jexp: 1ms is a very long time, I think we should rather stream and apply batches and then handled the failed ones on the side, e.g. by putting them into a queue.

src/main/java/apoc/util/Util.java
jexp: we should utilize Code.classification for transient errrors
https://github.com/neo4j/neo4j/blob/3.5/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java#L80-L91

src/main/java/apoc/util/Util.java
alexiudice: Where could I find the apoc code format scheme?

jexp: Regular Java Style guide (e.g. IntelliJ / Eclipse defaults)

jexp:
@alexiudice ping, let me know if/when I should review again.

src/main/java/apoc/cypher/Cypher.java
jexp: Ok, makes sense.

We meant to add an uuid adding feature to apoc anyway though :)

Btw. it's not recommended to uses prefixes like `uuid:` as that affects the ability for string compression as it leaves the character set of hex-code+separator.

jexp:
@alexiudice ping

alexiudice:
Sorry I've been swamped. I haven't forgotten about this!

src/main/java/apoc/util/Util.java
alexiudice: @jexp It is an interface since the inner enums of Status are `protected`. And I did check the Status interface, it does not have the exceptions grouped in a nice way for what we would want to retry on. We want to have control over which exceptions can cause a retry.

What were you thinking/meaning by utilizing `code.classification` for transient errors?

src/main/java/apoc/util/Util.java
alexiudice: I'll fix this up.

src/main/java/apoc/util/Util.java
alexiudice: I was thinking that we only want the result once the query has finished. By putting it in a FutureTask I was thinking we would only return the result once the entire query was done. I will remove the FutureTask and just run it.

alexiudice: I was changing it back and I realized why I had it as a FutureTask rather than just calling it. 

I did this because if the Tx fails, it marks every transaction created in the thread for `rollback`. So even if the retry is successful the transaction is always rolled back. 

This is why I put it in a FutureTask object and then ran it in a separate thread each time.

src/main/java/apoc/util/Util.java
```diff
@@ -700,4 +706,99 @@ public static boolean classExists(String className) {
...
+
+            try
+            {
+                POOL.submit( resultFuture );
+
+                // Query has not finished yet, wait.
+                while ( !resultFuture.isDone() )
+                {
+                    LockSupport.parkNanos( 100 );
+                }
```
alexiudice: Does this wait time work? Or do we want to scrap the waiting entirely and try to do something by streaming the result? I'm just not sure how I'd go about doing that.

alexiudice:
@jexp ping.

I've updated the PR, though I was unsure of some of the changes you wanted me to make. Some of the comments/questions I left are not showing up on the Files Changed, not sure if you can seem them or you need me to retype them.

src/main/java/apoc/util/Util.java
alexiudice: I'll update to the regular style guide as my final step. Want to get all the code completed before changing the aesthetics.

PR #932, opened 2018-09-30
Adding Neighbors stored procedures.

maxdemarzi:


src/main/java/apoc/neighbors/Neighbors.java
```diff
@@ -0,0 +1,240 @@
...
+    @Description("apoc.neighbors(node, rel-direction-pattern, distance) - returns distinct nodes of the given relationships in the pattern up to a certain distance, can use '>' or '<' for all outgoing or incoming relationships")
+    public Stream<NodeResult> neighbors(@Name("node") Node node, @Name(value = "types", defaultValue = "") String types, @Name(value="distance", defaultValue = "1") Long distance) {
+        if (distance < 1) return Stream.empty();
+        if (types==null || types.isEmpty()) return Stream.empty();
+
+        // Initialize bitmaps for iteration
+        RoaringBitmap seen = new RoaringBitmap();
+        RoaringBitmap nextA = new RoaringBitmap();
+        RoaringBitmap nextB = new RoaringBitmap();
+        int nodeId = (int) node.getId();
```
sarmbruster: I think that's a unsafe operation in case we're working on a graph with more than 2bn nodes - which even is possible when using standard store format. Shouldn't we fall back on either aborting the operation with an exception or somehow deal gracefully with larger node ids?

maxdemarzi: There is NeighborsLarge for those rare folks.

maxdemarzi: Do we want to just drop the regular and rename the .large so it won't be an issue?

sarmbruster: Maybe leave both as they are. But the int version should throw an execption in case you're using node ids > 2bn. This way we have both, efficency of the int variant and scale of the long variant. just my 2ct

jexp:
Thanks a lot @maxdemarzi

PR #938, opened 2018-10-02
Fixes #937: Fix trigger configuration parameter.

### LINKED ISSUE #937: Issue regarding apoc.trigger configuration parameter.
There is an issue regarding apoc.trigger configuration parameter. The configuration parameter only stores the sub-map `params` rather than the entire `config` passed in. 

As we can see in the `apoc.trigger.add` procedure:
```
Map<String,Object> params = (Map)config.getOrDefault("params", Collections.emptyMap());
```
is stored in the TriggerInfo object (and the underlying Trigger object represented by a Map<String,Object>) .

What should be stored in those objects is instead the full `config` map.

Requires a small fix in some coding and some refactoring.

alexiudice:
Fixes #937: Fix trigger configuration parameter. 

Makes changes so the whole trigger config map is useable rather than just the sub-map "params".

### DIFF
--- a/src/main/java/apoc/trigger/Trigger.java
+++ b/src/main/java/apoc/trigger/Trigger.java
@@ -31,7 +31,7 @@ public class Trigger {
         public String name;
         public String query;
         public Map<String,Object> selector;
-        public Map<String, Object> params;
+        public Map<String, Object> config;
         public boolean installed;
         public boolean paused;
 
@@ -43,12 +43,12 @@ public class Trigger {
             this.paused = paused;
         }
 
-        public TriggerInfo( String name, String query, Map<String,Object> selector, Map<String,Object> params, boolean installed, boolean paused )
+        public TriggerInfo( String name, String query, Map<String,Object> selector, Map<String,Object> config, boolean installed, boolean paused )
         {
             this.name = name;
             this.query = query;
             this.selector = selector;
-            this.params = params;
+            this.config = config;
             this.installed = installed;
             this.paused = paused;
         }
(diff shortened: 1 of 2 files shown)


jexp:
Can you explain more about this change?
And add tests to confirm it fixes what it intents to?

alexiudice:
Sure I can explain more about this change. 

When I originally wrote the use of params in Triggers I wrote it too restricted. I wrote it so the Triggers only save the params map itself, not the entire configuration map. I meant to have an entire configuration map, with params _inside_ as a submap.

This change is to make sure the whole config map gets saved, not just the params.
The config map is useful for further updates to how Triggers work internally, making it possible to pass in configurations for triggers.

alexiudice:
I've added two tests, one that tests the params work as expected and one that tests the config map works as expected.

jexp:
Btw, while you're on it, can you change the `kernelTransaction` to `statement`?
It was a stupid accidental search and replace issue :(

Also, we need to make the changes backward compatible at least for a few releases.

So if folks stored params we have to take that into account and merge it with a stored config (and remove params when storing the map again.

alexiudice:
I can work on making those changes for backwards compatibility.

Also would it make more sense to update `kernalTransaction` to `statement` in a separate issue/PR, since the change is also in Trigger.java?

jexp:
Yes on the separate PR for the renaming. I already fixed some of those too so yu might need to rebase.

I added a procedure to just remove the stored property as a last resort.

But in general we should be lenient if we  change structures so the users that already use the feature are not hit with an incompatible change.

bradnussbaum:
@jexp I think in this case b/c we just added this and it's under development I'm not expecting anyone is using and we can just make this change. It will keep the code clean.

alexiudice:
Closing this because it is outdated.

PR #939, opened 2018-10-03
fixes #890 - apoc.schema.assert are not deleted all indexes

### LINKED ISSUE #890: apoc.schema.assert are not deleted all indexes
When i run `CALL apoc.schema.assert({}, {})` not all indexes are deleted. 

The indexes with more propertys are not deleted.

I have manual deleted that index `DROP CONSTRAINT ON (n:Coordinates) ASSERT (n.Latitude, n.Longitude) IS NODE KEY`

AngeloBusato:


jexp:
Shouldn't there be some tests covering the new cases?

AngeloBusato:
@jexp yes but Node Key constraint requires Neo4j Enterprise Edition, we don't have an Enterprise Edition on tests

jexp:
Thank you

PR #946, opened 2018-10-08
fixes #820 - Support Temporal types in apoc.load.jdbc

### LINKED ISSUE #820: Support Temporal types in apoc.load.jdbc
Currently all `Date`s  are converted to milliseconds since epoch when loading data through `apoc.load.jdbc`.  Now that we're supporting native dates, it would be nice to have the option to convert these into temporal types.

Here is the conversion line in question:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/cba3e556409f5c31a6ac9c4eec4e53705a7bea72/src/main/java/apoc/load/Jdbc.java#L227

Nice to haves:
- Additional map of configuration options
- Choose between Date/Time and Local DateTime/Time
- Pass a timezone for DateTime and Time

AngeloBusato:


src/main/java/apoc/load/Jdbc.java
```diff
@@ -221,11 +226,23 @@ public boolean hasNext() {
...
 
         private Object convert(Object value) {
+
             if (value instanceof UUID || value instanceof BigInteger || value instanceof BigDecimal) {
                 return value.toString();
             }
             if (value instanceof java.util.Date) {
-                return ((java.util.Date) value).getTime();
+                if(!config.isEmpty()){
+                    LoadJdbcConfig loadJdbcConfig = new LoadJdbcConfig(config);
```
jexp: Shouldn't this be handled upfront? The JDBC config created etc? And instead of the many ifs an enum or lambda would be better suited to be passed in as "DateTransformer" or so.

docs/loadjdbc.adoc
```diff
@@ -457,4 +457,53 @@ You can call the procedure without param:
...
+
+Config param is optional, the default value is an empty map.
+
+[cols="3m,2"]
+|===
+|localDateTime| value: true/false i.e "2007-12-03T10:15:30"
+|localDate| value: true/false i.e "2007-12-03"
+|zoneDateTime| value: true/false i.e "2007-12-03T10:15:30+01:00 Europe/Paris"
+|offSetTime| value: true/false i.e "2007-12-03T10:15:30+01:00"
+|timezone| default value: 'UTC'
```
jexp: Can we have an example for timezone and the different types?

src/main/java/apoc/load/util/LoadJdbcConfig.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private ZoneId zoneId;
+    private boolean localDateTime;
+    private boolean localDate;
+    private boolean zoneDateTime;
+    private boolean offSetTime;
+
+    public LoadJdbcConfig(Map<String,Object> config) {
+        config = config != null ? config : Collections.emptyMap();
+        this.zoneId = (ZoneId) config.getOrDefault("timezone",ZoneId.of("UTC"));
```
jexp: what if more than one of them is true? Should we validate that?

I think a different format is better: `{temporal: "localdatetime", timezone: 'UTC'}`

with the names from the cypher temporal functions

src/main/java/apoc/load/Jdbc.java
```diff
@@ -221,11 +226,23 @@ public boolean hasNext() {
         }
 
         private Object convert(Object value) {
+
             if (value instanceof UUID || value instanceof BigInteger || value instanceof BigDecimal) {
                 return value.toString();
             }
             if (value instanceof java.util.Date) {
```
jexp: shouldn't this also take the format of the SQL datatype into account?
java.sql.Timestamp  vs. java.sql.Date vs. ... ?

conker84: Hi @jexp the java.sql date types all inherits from `java.util.Date`

docs/loadjdbc.adoc
conker84: Done

conker84: Done

src/main/java/apoc/load/Jdbc.java
```diff
@@ -224,8 +263,8 @@ private Object convert(Object value) {
             if (value instanceof UUID || value instanceof BigInteger || value instanceof BigDecimal) {
                 return value.toString();
             }
-            if (value instanceof java.util.Date) {
-                return ((java.util.Date) value).getTime();
+            if (value instanceof Date) {
```
jexp: if we have mixed types (date, timestamp, etc) in the relational data we want to convert them correctly.

So dates should stay dates, and timestamps should become datetime etc.
java.sql.Types -> resultsetmeta
java.sql.Date -> Date
java.sql.Timestamp -> datetime

I'm not sure if the SQL dates have timezones by default.

if there is a TZ we can do localdate/time

conker84: @jexp pr updated

docs/loadjdbc.adoc
```diff
@@ -457,4 +457,80 @@ You can call the procedure without param:
...
 ----
 CALL apoc.load.jdbcUpdate('jdbc:mysql:....','INSERT INTO RECOMMENDATIONS values(user.id, reco.id, score)') YIELD row;
+----
+
+
+=== Load JDBC format date
+
+[source, cypher]
+----
+CALL apoc.load.jdbc('key or url','table or kernelTransaction', config) YIELD row
```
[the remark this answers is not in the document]
conker84: Done

src/main/java/apoc/load/Jdbc.java
jexp: What about the other temporal types? like Types.DATE ?

conker84: It goes into the switch's `default`, all of sql temporal types inherit from `java.util.Date` so we group by it:
* [java.sql.Date](https://docs.oracle.com/javase/8/docs/api/java/sql/Date.html)
* [java.sql.Time](https://docs.oracle.com/javase/8/docs/api/java/sql/Time.html)
* [java.sql.Timestamp](https://docs.oracle.com/javase/8/docs/api/java/sql/Timestamp.html)

An then we use the JDBCType in order to make the conversion

jexp: So Types.DATE goes into the `default:` branch? which means it stays an `java.sql.Date` and is not converted?

PR #949, opened 2018-10-10
Handle separator and quote chars properly, fixes #948

### LINKED ISSUE #948: Passing a delimiter to `apoc.import.csv` results in "String cannot be cast" error
I am experiencing the same bug as reported on SO: https://stackoverflow.com/questions/52046831/neo4j-apoc-apoc-import-csv-error

```
CALL apoc.import.csv(
  [{fileName: 'file:/<path>/persons.csv', labels: ['Person']}],
  [{fileName: 'file:/<path>/knows.csv', type: 'KNOWS'}],
  {delimiter: '|', arrayDelimiter: ',', stringIds: false}
)
```
Results in 
```
apoc "java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Character"
```

This does not happen when `delimiter` is not set.

szarnyasg:
Fixes #948. Also fix handling quotation characters, which was made possible by PR #879

### DIFF
--- a/src/main/java/apoc/export/csv/CsvEntityLoader.java
+++ b/src/main/java/apoc/export/csv/CsvEntityLoader.java
@@ -79,7 +79,7 @@ public class CsvEntityLoader {
                 )
         );
 
-        final CSVReader csv = new CSVReader(reader, clc.getDelimiter());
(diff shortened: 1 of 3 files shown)


jexp:
Thank you

PR #953, opened 2018-10-16
fixes #568 - similar to the TTL listener, add an uuid listener that automatically set's uids for nodes

### LINKED ISSUE #568: similar to the TX listener, add an uuid listener that automatically set's uids for nodes

AngeloBusato:


src/main/java/apoc/uuid/UUIDLifeCycle.java
```diff
@@ -0,0 +1,82 @@
...
+        } catch (Exception e) {
+            log.error("UUID: Error creating index", e);
+        }
+    }
+
+    public void setUUID(List<String> labels) {
+        if (!Util.isWriteableInstance(db)) return;
+        labels.forEach(label -> {
+            try (Transaction tx = db.beginTx()) {
+                try (ResourceIterator<Node> result = db.execute(String.format("MATCH (n:%s) WHERE not exists(n.uuid) RETURN n", label)).columnAs("n")) {
```
alexiudice: I would recommend putting backticks around that `%s` so this works for labels that have spaces or special character.

src/main/java/apoc/uuid/UUIDLifeCycle.java
```diff
@@ -0,0 +1,82 @@
...
+        String label = ApocConfiguration.get("uuid.labels", StringUtils.EMPTY);
+        List<String> labels = Stream.of(label.split(",")).map(String::trim).collect(Collectors.toList());
+
+        uuidIndexJobHandle = scheduler.schedule(UUID_GROUP, () -> createUUIDIndex(labels), (int)(uuidSchedule*0.8), TimeUnit.SECONDS);
+        uuidJobHandle = scheduler.scheduleRecurring(UUID_GROUP, () -> setUUID(labels), uuidSchedule, uuidSchedule, TimeUnit.SECONDS);
+    }
+
+    private void createUUIDIndex(List<String> labels) {
+        try {
+            labels.forEach(label -> db.execute(String.format("CREATE CONSTRAINT ON (%s:%s) ASSERT %s.uuid IS UNIQUE", label.toLowerCase(),
```
alexiudice: Same issue here with labels that have spaces/special characters. Fix is adding backticks. See comment below.

jexp:
Sorry, I wasn't clear when putting in this issue back then.

I actually didn't mean TTL but tx listener.

Issue #706, opened 2017-12-27
apoc.load.json not restricted to dbms.directories.import


Dcanzano:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/70  indicates that one can for example run

call apoc.load.json("file:////home/neo4j/neo4j-enterprise-3.3.0/bin/myfile.json") yield value return value; 

and this will properly read said file.   this is probably not a good idea for the file://// could represent any file on the filesystem.   Similar to LOAD CSV we should probably restrict apoc.load.json to the value reported by 

conf/neo4j.conf parameter of  dbms.directories.import

conker84:
closed via 3558084

Issue #605, opened 2017-09-20
Missing tests for import file from folder (csv, json and xml)


DanielBerton:
Add tests for  `apoc.load.csv`, `apoc.load.json`  and `apoc.load.xml`
Set a constraint in a specific folder, and test the relative path.
Configuration example for tests: 
```
.setConfig("apoc.import.file.enabled","true")
.setConfig("apoc.import.file.use_neo4j_config", "true")
.setConfig("dbms.directories.import", "build/resources/test")
.setConfig("dbms.security.allow_csv_import_from_file_urls","true")
```
Test the relative path `CALL apoc.load.json("file://map.json")` and the exception if we try to use a full path.

conker84:
closed via 3558084

PR #966, opened 2018-10-29
relative path issue on load csv,json,xml  - fixes #841 #605 #706

### LINKED ISSUE #841: apoc.load.csv requires uri to be absolute path
Can we change apoc.load.csv to mimic LOAD CSV where it will use the import directory by default unless the #dbms.directories.import=import in the configuration file is commented out?

AngeloBusato:


src/main/java/apoc/ApocConfiguration.java
```diff
@@ -28,6 +30,10 @@
     public static void initialize(GraphDatabaseAPI db) {
         Static.clear();
         Map<String, String> params = db.getDependencyResolver().resolveDependency(Config.class).getRaw();
+        Map<String, String> configValues = db.getDependencyResolver().resolveDependency(Config.class)
```
jexp: What's the difference?

src/main/java/apoc/ApocConfiguration.java
```diff
@@ -28,6 +30,10 @@
...
+                .getConfigValues().entrySet().stream()
+                .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue().valueAsString().orElse(StringUtils.EMPTY)));
```
jexp: We definitely don't need StringUtils.EMPTY as a dependency here.

src/main/java/apoc/util/FileUtils.java
```diff
@@ -42,68 +42,71 @@ public static CountingInputStream inputStreamFor(String fileName) throws IOExcep
...
-                    try { fileAsUrl = new URI(url).getPath(); } catch (Exception e) { /*no protocol*/ }
-                    if ("".equals(fileAsUrl)){
-                        String newUrl = url.replace("file://", "file:///");
-                        fileAsUrl = new URI(newUrl).getPath();
-                        if ("".equals(fileAsUrl)) fileAsUrl=newUrl.replace("file://","");
-                    }
-                    File file = url.startsWith(importDir) ? new File(fileAsUrl) : new File(dir.getAbsolutePath(), fileAsUrl);
-                    if (!file.exists() || !file.isFile() || !file.canRead()) throw new Exception();
-                    return new URL("file","",file.getAbsolutePath()).toString();
+                    String relativeFilePath = !uri.getPath().isEmpty() ? uri.getPath() : uri.getHost();
```
jexp: where did all that code go?

src/test/java/apoc/config/ConfigTest.java
```diff
@@ -21,64 +16,23 @@
 public class ConfigTest {
 
     private GraphDatabaseService db;
-    private File testFile;
 
     @Before
     public void setUp() throws Exception {
         db = new TestGraphDatabaseFactory().newImpermanentDatabaseBuilder()
```
jexp: where did all these tests go?

src/test/java/apoc/load/relative/LoadCsvRelativePathTest.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static apoc.util.MapUtil.map;
+import static apoc.util.TestUtil.testResult;
+import static java.util.Arrays.asList;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+public class LoadCsvRelativePathTest {
```
jexp: to just test the correct loading we don't need a full copy of each test file.

a single simple test if the file is found and readable (e.g. by reading all rows is good enough)
same goes for JSON/XML

src/main/java/apoc/ApocConfiguration.java
conker84: We merged `params` and `configValues` because all non-neo4j configuration such as apoc.* is not included into configValues; we don't get just a singular configuration but the whole one, some other parameter can be useful for some other feature; but if you want we could get them one by one by the need

src/main/java/apoc/util/FileUtils.java
conker84: As we moved to `URI#create method this allowed us to simplify the code base

src/test/java/apoc/config/ConfigTest.java
conker84: We moved all the tests into the class `FileUtilsTest`

src/main/java/apoc/ApocConfiguration.java
jexp: As there might be some security critical things, I'd prefer to only extract the ones we need.

src/main/java/apoc/ApocConfiguration.java
```diff
@@ -27,7 +29,12 @@
 
     public static void initialize(GraphDatabaseAPI db) {
         Static.clear();
-        Map<String, String> params = db.getDependencyResolver().resolveDependency(Config.class).getRaw();
+        Config neo4jConfig = db.getDependencyResolver().resolveDependency(Config.class);
```
jexp: this looks horribly complicated with little benefit.

Let's only grab the things we really need from the main config and that have the correct shape.

PR #967, opened 2018-10-31
fixes #789 - APOC Intersection JSON Number parsing issue

### LINKED ISSUE #789: APOC Intersection issue
Case2 below should work without explicitly casting between Integers or Longs.  Case4 shows a current work-around for this.

/* case1: this works */
WITH [1,2] as set1, [2,3] as set2
WITH apoc.coll.intersection(set1, set2) as intersection
RETURN intersection

/* case2: this doesn't work */
WITH apoc.convert.fromJsonMap('{"numbers":[1,2]}') as set1, [2,3] as set2
WITH apoc.coll.intersection(set1.numbers, set2) as intersection
RETURN intersection

/* case3: this works */
WITH apoc.convert.fromJsonMap('{"numbers":[1,2]}') as set1, [2,3] as set2
RETURN filter(number IN set1.numbers WHERE number IN set2) as intersection

/* case4: works - work around for case2 */
WITH apoc.convert.fromJsonMap('{"numbers":[1,2]}') as set1, [2,3] as set2
WITH apoc.coll.intersection([x in set1.numbers | tointeger(x)], set2) as intersection
RETURN intersection

AngeloBusato:


src/test/java/apoc/spatial/GeocodeTest.java
```diff
@@ -151,7 +151,7 @@ private void testGeocodeAddress(Map map, String provider) {
                 }
                 testCallCount(db, "CALL apoc.spatial.geocode({url},0)",
                         map("url", map.get("address").toString()),
-                        (int) ((Map) map.get("count")).get(provider));
+                        Integer.parseInt(((Map) map.get("count")).get(provider).toString()));
```
jexp: use Number.intValue()

src/test/java/apoc/schema/SchemasTest.java
```diff
@@ -268,7 +268,7 @@ public void testIndexes() {
             assertEquals("NO FAILURE", r.get("failure"));
             assertEquals(new Double(100), r.get("populationProgress"));
             assertEquals(new Double(1), r.get("valuesSelectivity"));
-            assertEquals("Index( GENERAL, bar )", r.get("userDescription"));
+            assertEquals("Index( GENERAL, :Foo(bar) )", r.get("userDescription"));
```
jexp: this seems to have changed between Neo4j versions.

src/main/java/apoc/util/JsonUtil.java
```diff
@@ -73,10 +77,13 @@ public void close() throws IOException {
     public static <T> T parse(String json, String path, Class<T> type) {
         if (json==null || json.isEmpty()) return null;
         try {
-            if (path == null || path.isEmpty()) return OBJECT_MAPPER.readValue(json, type);
-            return JsonPath.parse(json,JSON_PATH_CONFIG).read(path,type);
+            if (path == null || path.isEmpty()) {
+                return OBJECT_MAPPER.readValue(json, type);
+            }
+            String stringValue = OBJECT_MAPPER.writeValueAsString(JsonPath.parse(json,JSON_PATH_CONFIG).read(path, type));
```
jexp: Can we please not do this, and find a way to avoid creating new huge string objects in memory?

src/test/java/apoc/spatial/GeocodeTest.java
conker84: We restored the previous version the value of `((Map) map.get("count")).get(provider))` is aready and integer.

src/main/java/apoc/util/JsonUtil.java
conker84: Done we now inject Jackson into JsonPath

jexp: Great!

jexp:
@conker84 something interesting about the "intersection" in general is that there might be lists of mixed numbers. like doubles and longs and then folks might expect cypher semantics of `1.0 == 1`.

But we should cover that in a separate issue I think.
https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/972

jexp:
Thanks a lot.

PR #971, opened 2018-11-05
fixes #957 - apoc.export.cypher.all doesn't work with temporal types

### LINKED ISSUE #957: apoc.export.cypher.all doesn't work with temporal types
The cypher generated by `apoc.export.cypher.all` for temporal values is not valid cypher

For example when I have a node with a time property the cypher generated by `apoc.export.cypher.all` for that node looks like:
```
CREATE (:`TimeEntry`:`UNIQUE IMPORT LABEL` {`Time`:15:55Z, `UNIQUE IMPORT ID`:99500});
```
which will fail if executed with:
```
Variable `15:55Z` not defined (line 2, column 57 (offset: 57))
```

in this case I believe the correct cypher would be:
```
`Time`:time('15:55Z')
```
although there may be other ways to create a time (I am not an expert!)

AngeloBusato:


src/main/java/apoc/export/cypher/formatter/CypherTemporalFormat.java
```diff
@@ -0,0 +1,50 @@
+package apoc.export.cypher.formatter;
+
+import org.neo4j.values.storable.DurationValue;
+
+import java.time.*;
+
+public enum CypherTemporalFormat {
```
jexp: why do we roll our own? 

isn't that handled by cypher values?
and for bolt via org.neo4j.driver.v1.Values ?

src/test/java/apoc/export/cypher/ExportCypherResults.java
```diff
@@ -0,0 +1,168 @@
+package apoc.export.cypher;
```
jexp: why do we pull it out to a separate file? is it reused in some other place?
Otherwise makes it harder to keep test + expected results together?

jexp: I get that the file gets quite long, we can push these also to the end of the file if that makes it more readable.

PR #984, opened 2018-11-12
fixes #816 - requesting ability to bury url for apoc.load.json into neo4j.conf

### LINKED ISSUE #816: requesting ability to bury url for apoc.load.json into neo4j.conf
apoc.load.jdbc allows one to bury the url into an alias and in the neofj.conf via

apoc.jdbc.<alias>.url= ..... ..... ......

requesting the same functionality with apoc.load.json such that i would like to define in my neo4j.conf 

apoc.json.<alias>.url = .... .... .... ....

and then be able to call

call apoc.load.json(<alias>, .........) yield value

AngeloBusato:


jexp:
We also use it for mongo and elastic at least, if you want to unify this functionality.

src/main/java/apoc/util/Util.java
```diff
@@ -528,7 +528,7 @@ public static String cleanUrl(String url) {
             if (source.getRef() != null) file += "#"+source.getRef();
             return new URL(source.getProtocol(),source.getHost(),source.getPort(),file).toString();
         } catch (MalformedURLException mfu) {
-            return "invalid URL";
+            return String.format("invalid URL (%s)", url);
```
jexp: are we sure this doesn't expose credentials?

conker84: This method is (at this moment) used in these classes:
- LoadCsv
- LoadXls
- Xml
- JsonUtil
So we have no DB connector that could expose credentials, the only credentials that could be exposed are the http embedded which are deprecated. [Here](https://medium.com/@lmakarov/say-goodbye-to-urls-with-embedded-credentials-b051f6c7b6a3) an article that explains all.
If you think we must prevent this case we'll revert it.

AngeloBusato:
@jexp PR rebased

Issue #951, opened 2018-10-12
apoc.import.csv creates superflous nodes when relationship references don't exist


ArthurClemens:
Test case, using the example from "Loading nodes and relationships":

`/tmp/persons.csv`:
```
:ID,name:STRING,speaks:STRING[]
1,John,"en,fr"
2,Jane,"en,de"
```

`/tmp/knows.csv`:
```
:START_ID,:END_ID,since:INT
1,2,2016
10,11,2014
11,12,2013
```
(note the 2 extra lines)

Command:
```
CALL apoc.import.csv(
  [{fileName: '/tmp/persons.csv', labels: ['Person']}],
  [{fileName: '/tmp/knows.csv', type: 'KNOWS'}],
  {stringIds: false}
)
```

The result creates 6 nodes - John and Jane are tripled:
```
╒═══════════════════════════════════════════════╕
│"n"                                            │
╞═══════════════════════════════════════════════╡
│{"name":"John","speaks":["en,fr"],"__csv_id":1}│
├───────────────────────────────────────────────┤
│{"name":"Jane","speaks":["en,de"],"__csv_id":2}│
├───────────────────────────────────────────────┤
│{"name":"John","speaks":["en,fr"],"__csv_id":1}│
├───────────────────────────────────────────────┤
│{"name":"Jane","speaks":["en,de"],"__csv_id":2}│
├───────────────────────────────────────────────┤
│{"name":"John","speaks":["en,fr"],"__csv_id":1}│
├───────────────────────────────────────────────┤
│{"name":"Jane","speaks":["en,de"],"__csv_id":2}│
└───────────────────────────────────────────────┘
```

This is the full response:
```
[
  {
    "keys": [
      "n"
    ],
    "length": 1,
    "_fields": [
      {
        "identity": {
          "low": 393,
          "high": 0
        },
        "labels": [
          "Person"
        ],
        "properties": {
          "name": "John",
          "speaks": [
            "en,fr"
          ],
          "__csv_id": {
            "low": 1,
            "high": 0
          }
        }
      }
    ],
    "_fieldLookup": {
      "n": 0
    }
  },
  {
    "keys": [
      "n"
    ],
    "length": 1,
    "_fields": [
      {
        "identity": {
          "low": 394,
          "high": 0
        },
        "labels": [
          "Person"
        ],
        "properties": {
          "name": "Jane",
          "speaks": [
            "en,de"
          ],
          "__csv_id": {
[...]

PR #993, opened 2018-11-16
Expect failure for 'testRelationshipEndNodesNotExist'. Fixes #992

### LINKED ISSUE #992: Test failure:  ImportCsvTest.testNoDuplicationsCreated
@szarnyasg 

It's probably faster if you have a look. I guess it doens't find the node it's looking for.

This test fails with `Caused by: java.lang.IllegalStateException: Node for id space __CSV_DEFAULT_IDSPACE and id 10 not found`

szarnyasg:
Fixes #992 

Expect test `testRelationshipEndNodesNotExist` to fail.

- This test was added in #976 as a followup to #951. It is expected to fail, but my previous implementation was sloppy and it only `@Ignore`d the test.
- Imports were reorganized by IntelliJ.

### DIFF
--- a/src/test/java/apoc/export/csv/ImportCsvTest.java
+++ b/src/test/java/apoc/export/csv/ImportCsvTest.java
@@ -1,8 +1,12 @@
 package apoc.export.csv;
 
 import apoc.util.TestUtil;
-import org.junit.*;
+import org.junit.After;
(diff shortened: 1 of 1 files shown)


jexp:
Fixed manually, thanks!

PR #997, opened 2018-11-21
fixes #994 - Provide a configuration parameter map with user/pass for JDBC

### LINKED ISSUE #994: Provide a configuration parameter map with user/pass for JDBC
Connect to a DB via JDBC if the url contains one or more `#` in the user e/o password

Cannot connect to a DB via JDBC if the url contains one or more `#` in the user e/o password

### Steps (Mandatory)

  1. Just lunch this test case in `JdbcTest`

```java
    @Test
    public void testLoadJdbcWithSpecialChar() {
        String connStr = "jdbc:derby:derbyDB;user=apoc;password=foo#bar#";
        db.execute("CALL apoc.load.jdbc('" + connStr + "', 'PERSON')").next();
    }
```

Related to #885 and #886

AngeloBusato:
Fixes #994 

Provide a configuration parameter map with user/pass for JDBC

  - added a new config parameter _credentials_ in which user can write user and password 
`{credentials:{user:"",password:""}}`

### DIFF
--- a/src/main/java/apoc/load/Jdbc.java
+++ b/src/main/java/apoc/load/Jdbc.java
@@ -39,15 +39,19 @@ public class Jdbc {
     @Context
     public Log log;
 
-    private static Connection getConnection(String jdbcUrl) throws Exception {
-        URI uri = new URI(jdbcUrl.substring("jdbc:".length()));
-        String userInfo = uri.getUserInfo();
-        if (userInfo != null) {
-            String[] user = userInfo.split(":");
-            String cleanUrl = jdbcUrl.substring(0,jdbcUrl.indexOf("://")+3)+jdbcUrl.substring(jdbcUrl.indexOf("@")+1);
-            return DriverManager.getConnection(cleanUrl, user[0], user[1]);
+    private static Connection getConnection(String jdbcUrl, LoadJdbcConfig config) throws Exception {
+        if(config.hasCredentials()) {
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/load/Jdbc.java
```diff
@@ -39,13 +39,9 @@
...
     public Log log;
 
-    private static Connection getConnection(String jdbcUrl) throws Exception {
-        URI uri = new URI(jdbcUrl.substring("jdbc:".length()));
-        String userInfo = uri.getUserInfo();
-        if (userInfo != null) {
-            String[] user = userInfo.split(":");
-            String cleanUrl = jdbcUrl.substring(0,jdbcUrl.indexOf("://")+3)+jdbcUrl.substring(jdbcUrl.indexOf("@")+1);
-            return DriverManager.getConnection(cleanUrl, user[0], user[1]);
+    private static Connection getConnection(String jdbcUrl, LoadJdbcConfig config) throws Exception {
```
jexp: We cannot change it directly. We have to keep the current way for at least one version and deprecated it.
So if there is no credentials passed in config we have to keep the existing approach for the time being.

src/main/java/apoc/load/util/LoadJdbcConfig.java
```diff
@@ -13,6 +15,8 @@
 
     private ZoneId zoneId = null;
 
+    private Map<String,String> credentials;
```
jexp: I would make this explicit into username and password.
And add a method "hasCredentials"

PR #1002, opened 2018-11-30
fixes #991 - Add ability to provide label to schema functions for filtered results

### LINKED ISSUE #991: Add ability to provide label to schema functions for filtered results
Currently when using e.g. apoc.schema.nodes() If i want to find out information about a specific label I have to post filter the results e.g. to find out constraints on 'Email' i would do:

CALL apoc.schema.nodes() yield label, properties
with * where label="Email" return properties

It would be great if we could provide the label directly to the procedure e.g.

CALL apoc.schema.nodes("Email") yield properties

AngeloBusato:
Fixes #991

Add ability to provide label to schema functions for filtered results

  - Now it can be possible to find all the nodes contraints/Indexes by label e.g (CALL apoc.schema.nodes(labelName) yield properties)

### DIFF
--- a/src/main/java/apoc/schema/SchemaConfig.java
+++ b/src/main/java/apoc/schema/SchemaConfig.java
@@ -1,0 +1,46 @@
+package apoc.schema;
+
+import java.util.*;
+
+/**
+ * @author ab-larus
+ * @since 17.12.18
+ */
+public class SchemaConfig {
+
+    private Set<String> labels;
+    private Set<String> excludeLabels;
+    private Set<String> relationships;
+    private Set<String> excludeRelationships;
+
+    public Set<String> getLabels() {
+        return labels;
+    }
+
+    public Set<String> getExcludeLabels() {
+        return excludeLabels;
+    }
+
+    public Set<String> getRelationships() {
+        return relationships;
+    }
+
+    public Set<String> getExcludeRelationships() {
+        return excludeRelationships;
+    }
+
+    public SchemaConfig(Map<String,Object> config) {
+        config = config != null ? config : Collections.emptyMap();
+        this.labels = new HashSet<>((Collection<String>)config.getOrDefault("labels", Collections.EMPTY_SET));
+        this.excludeLabels = new HashSet<>((Collection<String>) config.getOrDefault("excludeLabels", Collections.EMPTY_SET));
+        validateParameters(this.labels, this.excludeLabels, "labels");
+        this.relationships = new HashSet<>((Collection<String>)config.getOrDefault("relationships", Collections.EMPTY_SET));
+        this.excludeRelationships = new HashSet<>((Collection<String>)config.getOrDefault("excludeRelationships", Collections.EMPTY_SET));
+        validateParameters(this.labels, this.excludeLabels, "relationships");
+    }
+
+    private void validateParameters(Set<String> include, Set<String> exclude, String parametrType){
+        if(!include.isEmpty() && !exclude.isEmpty())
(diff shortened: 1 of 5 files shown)


jexp:
Can we change this to a consistent, optional config map that works for all schema/meta procedures.

My suggestion is to take it from apoc.meta.subGraph().

`{ labels:[], relationships:[], exclude:[], sample:number}`

Anything I miss? @sarmbruster @kvegter ?

conker84:
Maybeis it better to split the `exclude` property?
`{ labels:[], relationships:[], excludeLabels:[], excludeRelationships:[], sample:number}`

jexp:
Yes we could, makes it more explicit.
In general you shouldn't have an overlap between the two.

We could also work with a `"-"` prefix for exclusion, not sure which one is nicer for the user.

jexp:
Perahaps also add a include/excluding of property-names.

AngeloBusato:
@jexp PR Updated

AngeloBusato:
@jexp PR updated

src/main/java/apoc/schema/SchemaConfig.java
```diff
@@ -0,0 +1,48 @@
...
+        return relationships;
+    }
+
+    public Set<String> getExcludeRelationships() {
+        return excludeRelationships;
+    }
+
+    public SchemaConfig(Map<String,Object> config) {
+        config = config != null ? config : Collections.emptyMap();
+        this.labels = (Set<String>) config.getOrDefault("labels", Collections.EMPTY_LIST);
```
jexp: emptySet()

src/main/java/apoc/schema/Schemas.java
```diff
@@ -44,15 +44,15 @@
     }
 
     @Procedure(value = "apoc.schema.nodes", mode = Mode.SCHEMA)
-    @Description("CALL apoc.schema.nodes() yield name, label, properties, status, type")
-    public Stream<IndexConstraintNodeInfo> nodes() throws IndexNotFoundKernelException {
-        return indexesAndConstraintsForNode();
+    @Description("CALL apoc.schema.nodes(config) yield name, label, properties, status, type")
```
jexp: config optional in descripton

jexp: also for the other procs

src/main/java/apoc/schema/Schemas.java
```diff
@@ -296,12 +335,68 @@ private Boolean constraintsExistsForRelationship(String type, List<String> prope
...
+                        containsLabelName(excludes, relationship.getEndNode())) {
+                    relationshipIterator.remove();
+                }
+            }
+
+            return gr;
+        });
+    }
+
+    private boolean containsLabelName(Set<String> excludes, Node node) {
```
jexp: should probably go to util class

src/main/java/apoc/schema/Schemas.java
```diff
@@ -269,15 +269,54 @@ private Boolean constraintsExistsForRelationship(String type, List<String> prope
...
         try ( Statement ignore = tx.acquireStatement() ) {
             TokenRead tokenRead = tx.tokenRead();
             TokenNameLookup tokens = new SilentTokenNameLookup(tokenRead);
 
             SchemaRead schemaRead = tx.schemaRead();
-            Iterable<IndexReference> indexesIterator = () -> schemaRead.indexesGetAll();
+            Iterable<IndexReference> indexesIterator;
+            Iterable<ConstraintDescriptor> constraintsIterator;
+
+            if (!includeLabels.isEmpty()) {
```
jexp: flip condition

AngeloBusato:
@jexp PR Updated

Issue #466, opened 2017-06-19
APOC GraphML attributes not loadable by Gephi


phelps-sg:
I landed on [Issue 258 (closed)](https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/258)  having encountered the same problem as the OP.

jexp said something was fixed in 3.1.0.4, but I am unable to import GraphML attributes when using APOC 3.1.3.7 and importing it into Gephi 0.9.1.

Qingmu:
Same problem in `neo4j-cmmunity-3.2.2` with `apoc-3.2.0.4.jar`, and `gephi` version 0.9.1. Besides, the replacement of `apoc` with the version 3.1.0.4  stop the running of `neo4j-cmmunity-3.2.2`.

jexp:
You have to use a 3.2.x version of APOC for the 3.2.x versions of Neo4j.

We can look into transmitting other attributes to gephi.

It currently transmits a property that fully or partially contains one of: "name", "title", "label" as caption.
And a configurable "weigth" property.

Which other properties would you want to see in Gephi?

jexp:
@phelps-sg how does this fail: `import GraphML attributes when using `

Do you have an example / error message / screenshot , e.g. for the movies graph?

Qingmu:
@jexp It would be better if apoc can get all the properties automatically instead of  fixed values as "name" and “title”. or how can I change the config files in apoc to add the properties I want to show in gephi ?

`apoc-3.2.0.4` can generate all the properties, e.g. `phonenumber`, into `graphml`, but it fails to add headers, like
`<key id="labels" for="node" attr.name="phonenumber" attr.type="string"/>`

jexp:
What confuses me is that this test, which stores two different attributes also writing the correct key-entries then works.

https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.2/src/test/java/apoc/export/graphml/ExportGraphMLTest.java#L93

jexp:
I'll check again.

Qingmu:
Thanks a lot !
And sorry for my poor coding with java or I can contribute to apoc myself.

conker84:
Related to #478

PR #1008, opened 2018-12-06
fixes #478 - Export graphml missing labels of Nodes and Edges

### LINKED ISSUE #478: Export graphml missing labels of Nodes and Edges
hello, i encountered an issue when i try to export a graphML document for Graphi.
i am using **apoc.export.graphml.query** command as follows ( refrence from (#295) )

`call apoc.export.graphml.query('MATCH p=()-[r]->() RETURN p limit 1000','../../../Ken/GraphML_output/AR.graphml',{useTypes:true})
`
the option **useTypes:true** is worthy for detecting Attributes of Nodes & Edges in Gephi.
However, i found the original Labels of Nodes & Edges in Neo4j is missing !
the meaning is i could not find the Labels of Nodes & Edges in Gephi which are exist in Neo4j.

To solve this problem, i try to Add following Tags to the GraphML document.
```
< key id="label" for="node" attr.name="label" attr.type="string"/ >
< key id="label" for="edge" attr.name="label" attr.type="string"/ >
```
and replace all nodes tag from
`<node id="n150" labels=":CORP"><data key="**labels**">:CORP</data></node>`
into 
`<node id="n150" label=":CORP"><data key="**label**">:CORP</data></node>`

After adjusting, i could see the Labels of Nodes & Edges in Gephi successfully.
node ex.
id             **Label**
n157	**:CORP**	.

edge ex.
Source     Target       Type         id     **Label**
n292	n188	directed	e0	**Order**		

However, all above works are done by myself,
I wonder to know is there any option parameter can i used in  **apoc.export.graphml.query** ??
Or, how could i solve these problem ?? 

Thanks for your help~

AngeloBusato:
fixes #478 #466

- #478 Export graphml missing labels of Nodes and Edges
- #466 APOC GraphML attributes not loadable by Gephi

  - All attributes/key are exported and generated
  - By deafult key are generated, for types use `useTypes:true`

### DIFF
--- a/docs/overview.adoc
+++ b/docs/overview.adoc
@@ -278,7 +278,27 @@ GraphML is used by other tools, like Gephi and CytoScape to read graph data.
 | apoc.export.graphml.graph(graph,file,config) | exports given graph object as graphml to the provided file
 | apoc.export.graphml.query(query,file,config) | exports nodes and relationships from the Cypher statement as graphml to the provided file
 |===
+
+All `Point` or `Temporal` data types are exported formatted as a String
+
+e.g:
+[cols="1m,2"]
+|===
+|Point 3d | {"crs":"wgs-84-3d","latitude":56.7,"longitude":12.78,"height":100.0}
+|Point 2d | {"crs":"wgs-84-3d","latitude":56.7,"longitude":12.78,"height":null}
+|Date | 2018-10-10
+|LocalDateTime | 2018-10-10T00:00
+|===
 // end::export.graphml[]
+
+.configuration options
+[options=header]
+|===
+| param | default | description
+| format | | In export to Graphml script define the export format. Possible value is: "gephi"
+| caption | | It's an array of string (i.e. ['name','title']) that define an ordered set of properties eligible as value for the `Label` value, if no match is found the there is a fallback to the node label, if the node label is missing the then the ID is used
+| useTypes | false | Write the attribute type information to the graphml output
+|===
 
 ==== Note:
 
--- a/src/main/java/apoc/export/graphml/XmlGraphMLWriter.java
+++ b/src/main/java/apoc/export/graphml/XmlGraphMLWriter.java
@@ -1,9 +1,6 @@
 package apoc.export.graphml;
 
-import apoc.export.util.ExportConfig;
-import apoc.export.util.FormatUtils;
-import apoc.export.util.MetaInformation;
-import apoc.export.util.Reporter;
+import apoc.export.util.*;
 import org.neo4j.cypher.export.SubGraph;
 import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.PropertyContainer;
@@ -12,13 +9,11 @@ import org.neo4j.graphdb.Relationship;
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
-import java.io.IOException;
 import java.io.Writer;
 import java.util.HashMap;
 import java.util.Map;
 
-import static apoc.export.util.MetaInformation.getLabelsString;
-import static apoc.export.util.MetaInformation.updateKeyTypes;
+import static apoc.export.util.MetaInformation.*;
 
 /**
  * @author mh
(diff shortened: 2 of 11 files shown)


docs/overview.adoc
```diff
@@ -265,6 +265,17 @@ GraphML is used by other tools, like Gephi and CytoScape to read graph data.
 | apoc.export.graphml.graph(graph,file,config) | exports given graph object as graphml to the provided file
 | apoc.export.graphml.query(query,file,config) | exports nodes and relationships from the Cypher statement as graphml to the provided file
 |===
+
+All `Point` or `Temporal` data types are exported formatted as a String
+
+e.g:
+[cols="1m,2"]
+|===
+|Point 3d | {"crs":"wgs-84-3d","latitude":56.7,"longitude":12.78,"height":100.0}
```
jexp: good question, can gephi / other graphml tools work with that? 
can we quickly check if there is any precedent about geo-coordinates in graphml?

src/main/java/apoc/export/graphml/XmlGraphMLWriter.java
```diff
@@ -97,7 +103,7 @@ private void writeLabels(XMLStreamWriter writer, Node node) throws IOException,
     private void writeLabelsAsData(XMLStreamWriter writer, Node node) throws IOException, XMLStreamException {
         String labelsString = getLabelsString(node);
         if (labelsString.isEmpty()) return;
-        writeData(writer, "labels", labelsString);
+        writeData(writer, "label", labelsString);
```
jexp: this changes the format (backwards compat) (perhaps better to write both, or add a config option for the old format)

are we sure that what Gephi calls label is the same as our labels?
I thought it was rather something like a caption.

docs/overview.adoc
conker84: We checked if there was a spatial support in new versions of GraphML, but it is not. So we decided to render it as a json String. We tested it with *Gephi* and *yEd* and is managed as simple string column.

src/main/java/apoc/export/graphml/XmlGraphMLWriter.java
conker84: This change is mandatory in order to correctly display labels in Gephi, maybe we can add, as you suggested, a property like `{exportFormat:'Gephi'}`, what do you think?

jexp: perhaps just "format"

jexp: as of our conversation, "label" in gephi is rather a caption and what we export here seems to be "TYPE", see apoc.export.gephi procedure implementation.

jexp:
Also needs rebase

AngeloBusato:
examples of results of nodes and edges exported with `useTypes:true, format:'gephi', caption:['name']`, and imported in Gephi

edges 1
nodes 1

AngeloBusato:
@jexp PR Updated

src/main/java/apoc/export/graphml/XmlGraphMLWriter.java
```diff
@@ -107,6 +114,7 @@ private int writeRelationship(XMLStreamWriter writer, Relationship rel) throws I
         writer.writeAttribute("target", id(rel.getEndNode()));
         writer.writeAttribute("label", rel.getType().name());
         writeData(writer, "label", rel.getType().name());
+        writeData(writer, "TYPE", rel.getType().name());
```
jexp: please only do this for format==Gephi

src/main/java/apoc/export/util/ExportConfig.java
```diff
@@ -124,6 +127,12 @@ private ExportFormat format(Object format) {
         return format != null && format instanceof String ? ExportFormat.fromString((String)format) : ExportFormat.NEO4J_SHELL;
     }
 
+    private static String[] convertCaption(Object value) {
+        if (value == null) return null;
+        List<String> strings = (List<String>) value;
```
jexp: please check before if this is a list otherwise we could treat it as a single string or fail with an error

src/main/java/apoc/export/util/ExportConfig.java
```diff
@@ -56,16 +58,17 @@ public boolean useTypes() {
...
+
+    public CypherFormat getCypherFormat() { return cypherFormat; }
 
     public ExportConfig(Map<String,Object> config) {
         config = config != null ? config : Collections.emptyMap();
         this.silent = toBoolean(config.getOrDefault("silent",false));
         this.batchSize = ((Number)config.getOrDefault("batchSize", DEFAULT_BATCH_SIZE)).intValue();
         this.delim = delim(config.getOrDefault("d", String.valueOf(DEFAULT_DELIM)).toString());
         this.useTypes = toBoolean(config.get("useTypes"));
+        this.caption = convertCaption(config.getOrDefault("caption", null));
```
jexp: perhaps as default: [name, title, label, id]

src/main/java/apoc/export/util/MetaInformation.java
```diff
@@ -46,19 +51,28 @@ public static void updateKeyTypes(Map<String, Class> keyTypes, PropertyContainer
...
+            default:
+                return isAllowed ? name : "string"; // We manage all other data types as strings
+        }
     }
 
-    public static String getLabelsString(Node node) {
-        if (!node.getLabels().iterator().hasNext()) return "";
-        String delimiter = ":";
-        return delimiter + FormatUtils.joinLabels(node, delimiter);
+    public static String getLabelsString(ExportConfig config, Node node) {
```
jexp: this does not belong together labels != caption

src/main/java/apoc/gephi/GephiFormatUtils.java
```diff
@@ -0,0 +1,18 @@
+package apoc.gephi;
+
+import apoc.export.util.FormatUtils;
+import org.neo4j.graphdb.Node;
+
+public class GephiFormatUtils {
+
+    public static String captionGraphml(Node n, String[] captions) {
+        for (String caption : captions) {
+            if (n.hasProperty(caption)) return n.getProperty(caption).toString();
```
jexp: first do one loop with the exact names 
2nd loop with lowercase
3rd loop with contains or endsWith

src/main/java/apoc/gephi/GephiFormatUtils.java
```diff
@@ -0,0 +1,18 @@
...
+            for (String key : n.getPropertyKeys()) {
+                if (key.toLowerCase().equalsIgnoreCase(caption)) return n.getProperty(key).toString();
+            }
+        }
+        String delimiter = ":";
+        return n.getLabels().iterator().hasNext() ? delimiter + FormatUtils.joinLabels(n, delimiter) : String.valueOf(n.getId());
```
jexp: not the labels but the first property

caption (instance)  != label(category or group)

src/test/java/apoc/load/LoadCsvTest.java
```diff
@@ -34,7 +34,7 @@
     }
 
     @Test public void testLoadCsv() throws Exception {
-        URL url = ClassLoader.getSystemResource("test.csv");
+        URL url = Thread.currentThread().getContextClassLoader().getResource("test.csv");
```
jexp: movie in a method in TestUtil

jexp:
ping @AngeloBusato

AngeloBusato:
@jexp PR Updated

src/main/java/apoc/gephi/GephiFormatUtils.java
```diff
@@ -0,0 +1,40 @@
...
+            }
+        }
+        return result == null ? String.valueOf(n.getId()) : result; // if the node has no property return the ID
+    }
+
+    public static String filterCaption(Node n, Set<String> captions, BiPredicate<String, String> predicate) {
+        for (String caption : captions) {
+            for (String key : n.getPropertyKeys()) {
+                if (predicate.test(key, caption))
+                    return n.getProperty(key.toLowerCase()).toString();
```
jexp: how would this work? if they key is camelCase and we then do getProperty with `key.toLowerCase()` 
and *why*?

AngeloBusato:
@jexp PR Updated

PR #1022, opened 2018-12-14
fixes #775 - apoc.export.graphml.all requires apoc.import.file.enabled needs to be enabled rather than apoc.export.file.enabled.

### LINKED ISSUE #775: apoc.export.graphml.all requires apoc.import.file.enabled needs to be enabled rather than apoc.export.file.enabled.
apoc.export.graphml.all requires apoc.import.file.enabled needs to be enabled rather than apoc.export.file.enabled.

When using `apoc.export.graphml.all` I get the following error if `apoc.import.file.enabled` is not set to true. 

```
neo4j> call apoc.export.graphml.all("/Users/rohankharwar/test-graphml",{});
Failed to invoke procedure `apoc.export.graphml.all`: Caused by: java.lang.RuntimeException: Import from files not enabled, please set apoc.import.file.enabled=true in your neo4j.conf
```

Since we are exporting and writing to the file system ideally `apoc.export.file.enabled` should be required and set to true and `apoc.import.file.enabled` should not be required.

Thanks,
Rohan

AngeloBusato:
fixes #775

apoc.export.graphml.all requires apoc.import.file.enabled needs to be enabled rather than apoc.export.file.enabled.

  - Import required only the `apoc.import.file.enabled`
  - Export required only the `apoc.export.file.enabled`

### DIFF
--- a/src/main/java/apoc/export/graphml/ExportGraphML.java
+++ b/src/main/java/apoc/export/graphml/ExportGraphML.java
@@ -1,10 +1,10 @@
 package apoc.export.graphml;
 
 import apoc.export.util.ExportConfig;
-import apoc.util.FileUtils;
 import apoc.export.util.NodesAndRelsSubGraph;
 import apoc.export.util.ProgressReporter;
 import apoc.result.ProgressInfo;
+import apoc.util.FileUtils;
 import apoc.util.Util;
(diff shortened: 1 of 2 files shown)


jexp:
Thanks a lot

PR #1029, opened 2018-12-19
fixes #548 - apoc.import.graphml ignores Node properties with empty String

### LINKED ISSUE #548: apoc.import.graphml ignores Node properties with empty String
Following scenerio:

1. `CREATE (n:Test {name : "", limit : 3}) RETURN n` 
resulted in a new Node:  
```json
{
  "name": "",
  "limit": 3
}
```
2. exporting graph to graphml `call apoc.export.graphml.all('/tmp/example.graphml',{storeNodeIds:true, readLabels:true, useTypes:true, defaultRelationshipType:'RELATED'})`
resulted in 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
<key id="name" for="node" attr.name="name" attr.type="string"/>
<key id="limit" for="node" attr.name="limit" attr.type="long"/>
<graph id="G" edgedefault="directed">
<node id="n3733" labels=":Test"><data key="labels">:Test</data><data key="name"></data><data key="limit">3</data></node>
</graph>
</graphml>
```
3. Delete all nodes 
4. Reimport node via apoc.graph.import
`call apoc.import.graphml("/tmp/example.graphml",{batchSize: 10000, readLabels: true, storeNodeIds: false, defaultRelationshipType:'RELATED'})`
5. `MATCH (n:Test) RETURN n`
resulted in 
```json
{
  "limit": 2
}
```

###Conclusion: 
The problem is that all properties, where the value is an empty string are ignored and not reimported.  
Maybe this would make sense for Null values, but not for empty strings. So ideally an additional flag is added in the config, like `includeEmptyProperties: true`.  

Kind Regards,
Stefan

AngeloBusato:
fixes #548

apoc.import.graphml ignores Node properties with empty String

  - Now properties with space are imported

### DIFF
--- a/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
+++ b/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
@@ -9,6 +9,7 @@ import org.neo4j.graphdb.*;
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLEventReader;
 import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLStreamConstants;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.events.Attribute;
 import javax.xml.stream.events.StartElement;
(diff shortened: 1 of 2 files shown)


AngeloBusato:
@jexp Rebased

PR #1033, opened 2018-12-28
Add missing cases in #1005

### LINKED ISSUE #1005: apoc.export.csv.query ignores the settings 'quotes = false'
## Expected Behavior
Related to #811 
Calling
`CREATE (n:test_quote {name: "test", type: "type"}) RETURN n;`

`CALL apoc.export.csv.query('MATCH (n) RETURN  n.name, n.type', '/import/test.csv', {batchSize:100, d: ';', quotes: false, format: 'plain'})`

The Fields in the file should not have any quote, i.e. 

```
n.name;n.type
test;type
```

## Actual Behavior 

The file looks like:
```
"n.name";"n.type"
"test";"type"
```

Running the following script, reproduce the issue

```
#!/bin/bash
set -e
docker_name="neo4j:3.5"
apoc_version="3.5.0.1"

NEO4J_DIR=$HOME/neo4j/neo4test/

echo "=========== Running neo4j =============="
echo "  * create directories for the data"
mkdir -p ${NEO4J_DIR}/data
mkdir -p ${NEO4J_DIR}/import
mkdir -p ${NEO4J_DIR}/plugins
mkdir -p ${NEO4J_DIR}/logs

echo "  * Downloading plugins"
if [ ! -e "${NEO4J_DIR}/plugins/apoc-${apoc_version}-all.jar"  ]; then
    pushd ${NEO4J_DIR}/plugins
    wget https://github.com/neo4j-contrib/neo4j-apoc-procedures/releases/download/"${apoc_version}"/apoc-"${apoc_version}"-all.jar
    popd
fi

echo "  * Running container in background"
docker_id=$(docker run \
    --publish=7474:7474 \
    --publish=7687:7687 \
    --volume=${NEO4J_DIR}/data:/data \
    --volume=${NEO4J_DIR}/import:/import \
    --volume=${NEO4J_DIR}/plugins:/plugins \
    --volume=${NEO4J_DIR}/logs:/logs \
    --env=NEO4J_dbms_memory_pagecache_size=2G \
    --env=NEO4J_dbms_memory_heap_maxSize=2G \
    --env NEO4J_apoc_export_file_enabled=true \
    --env NEO4J_apoc_import_file_enabled=true \
    --user="$(id -u):$(id -g)" \
    --env=NEO4J_AUTH=none \
    -d ${docker_name})

i=0
 while (true); do
    i=$((i+1))
    started_=$(docker logs "${docker_id}" | grep "INFO  Started." || true)
    if [ ! -z "${started_}" ]; then
       echo
       echo "Server started ..."
       sleep 4
       break
    fi  
    echo -ne "."
    sleep 2
    if [ "${i}" -gt 10 ]; then
       break
    fi  
 done
 rm ${NEO4J_DIR}/import/test.csv

curl -X POST -H 'Content-type: application/json' http://localhost:7474/db/data/transaction/commit -d '{"statements": [{"statement": "CREATE (n:test_quote {name: \"test\", type: \"type\"}) RETURN n;"}]}' && echo

curl -X POST -H 'Content-type: application/json' http://localhost:7474/db/data/transaction/commit -d '{"statements": [{"statement": "CALL apoc.export.csv.query(\"MATCH (n) RETURN  n.name, n.type\", \"/import/test.csv\", {batchSize:100, d: \";\", quotes: false, format: \"plain\"})"}]}' && echo

echo "EXPORTED FILE: "
cat ${NEO4J_DIR}/import/test.csv

```

### Simple Dataset

`CREATE (n:test_quote {name: "test", type: "type"}) RETURN n;`

`CALL apoc.export.csv.query('MATCH (n) RETURN  n.name, n.type', '/import/test.csv', {batchSize:100, d: ';', quotes: false, format: 'plain'})`

### Versions

**Neo4j**
```
Version: | 3.5.0
Edition: | community
```
**Apoc**

```
apoc.version(): "3.5.0.1"
```
**OS**
```
uname -a
Linux nbbell-debian 4.9.0-8-amd64 #1 SMP Debian 4.9.130-2 (2018-10-27) x86_64 GNU/Linux
```

StefanoBelloni:
Complement of the Fixes #1005

- Fixes in #1005 did not applied for export all graph, quotes rules should be now valid for all types of csv export
- counter of the exported relationships has been fixed

### DIFF
--- a/src/main/java/apoc/export/csv/CsvFormat.java
+++ b/src/main/java/apoc/export/csv/CsvFormat.java
@@ -34,6 +34,7 @@ import static apoc.export.util.MetaInformation.getLabelsString;
  */
 public class CsvFormat implements Format {
     private final GraphDatabaseService db;
+    private boolean quoteIfNeeded = true;
 
     public CsvFormat(GraphDatabaseService db) {
         this.db = db;
@@ -47,7 +48,7 @@ public class CsvFormat implements Format {
     @Override
     public ProgressInfo dump(SubGraph graph, Writer writer, Reporter reporter, ExportConfig config) throws Exception {
         try (Transaction tx = db.beginTx()) {
-            CSVWriter out = config.isQuotes() ? new CSVWriter(writer,config.getDelimChar(), ExportConfig.QUOTECHAR) : new CSVWriter(writer,config.getDelimChar(), '\0', '\0' );
+            CSVWriter out = getCsvWriter(writer, config);
             writeAll(graph, reporter, config, out);
             tx.success();
             reporter.done();
@@ -56,9 +57,38 @@ public class CsvFormat implements Format {
         }
     }
 
+    private CSVWriter getCsvWriter(Writer writer, ExportConfig config)
+    {
+        CSVWriter out;
+        switch (config.isQuotes()) {
+            case "none":
+                out = new CSVWriter(writer,
+                                    config.getDelimChar(),
+                                    '\0',
+                                    '\0');
+                quoteIfNeeded = false;
+                break;
+            case "ifNeeded":
+                out = new CSVWriter(writer,
+                                    config.getDelimChar(),
+                                    '\0',
+                                    '\0');
+                quoteIfNeeded = true;
+                break;
+            case "always":
+            default:
+                out = new CSVWriter(writer,
+                                    config.getDelimChar(),
+                                    ExportConfig.QUOTECHAR);
+                quoteIfNeeded = false;
+                break;
+        }
+        return out;
+    }
+
     public ProgressInfo dump(Result result, Writer writer, Reporter reporter, ExportConfig config) throws Exception {
         try (Transaction tx = db.beginTx()) {
-            CSVWriter out = config.isQuotes() ? new CSVWriter(writer,config.getDelimChar(), ExportConfig.QUOTECHAR) : new CSVWriter(writer,config.getDelimChar(), '\0', '\0');
+            CSVWriter out = getCsvWriter(writer, config);
 
             String[] header = writeResultHeader(result, out);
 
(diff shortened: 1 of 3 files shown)


src/main/java/apoc/export/csv/CsvFormat.java
```diff
@@ -96,10 +98,8 @@ public ProgressInfo dump(Result result, Writer writer, Reporter reporter, Export
             result.accept((row) -> {
                 for (int col = 0; col < header.length; col++) {
                     Object value = row.get(header[col]);
-                    data[col]= FormatUtils.toString(value);
```
jexp: do we really need manual quoting here? I thought CsvWriter does that for us?

StefanoBelloni: you're right. It should now be better.

src/main/java/apoc/export/csv/CsvFormat.java
```diff
@@ -34,7 +34,7 @@
  */
 public class CsvFormat implements Format {
     private final GraphDatabaseService db;
-    private boolean quoteIfNeeded = true;
+    private boolean quoteIfNeeded = false;
```
jexp: why was the the default changed?

src/test/java/apoc/export/csv/ExportCsvTest.java
```diff
@@ -134,6 +179,17 @@ public void testExportAllCsvHDFS() throws Exception {
...
+        TestUtil.testCall(db, "CALL apoc.graph.fromDB('test',{}) yield graph " +
+                        "CALL apoc.export.csv.graph(graph, {file},{quotes: 'none'}) " +
+                        "YIELD nodes, relationships, properties, file, source,format, time " +
+                        "RETURN *", map("file", output.getAbsolutePath()),
+                (r) -> assertResults(output, r, "graph"));
+        assertEquals(EXPECTED_NONE_QUOTES, new Scanner(output).useDelimiter("\\Z").next());
+    }
+
+    @Test
+    public void testExportGraphCsvWithouQuotes() throws Exception {
```
jexp: typo

src/main/java/apoc/export/csv/CsvFormat.java
StefanoBelloni: This was an extra flag to distinguish between _no-quotes_ and _quotes only if needed_. It's value is overwritten in the method `getCsvWriter`, de default is never used.
In the current implementation, however, the name is now misleading: there's only need to distinguish between _apply quotes to all_ and **not** _apply quotes to all_, so I changed the name to `applyQuotesToAll`

StefanoBelloni:
I keep having problem running the first time all tests in **apoc_tests** with the case `Periodictest.testTerminateCommit`. When I run however the test a second time, all tests are O.K.

jexp:
Thanks so much!!

PR #1035, opened 2019-01-04
Fix link to issue page.

michael-simons:
The relative link doesn't work on the user guide nor the readme on GH itself.


jexp:
thanks a lot!

PR #1046, opened 2019-01-14
3.4 filter duplicate nodes

### LINKED ISSUE #1038: apoc.import.csv does not have the functionality to ignore duplicate nodes
Using `neo4j-admin import` one can use the parameter `--ignore-duplicate-nodes` and it will force the bad node to be ignored during the import process. 
I cannot achieve functionality similar to `--ignore-duplicate-nodes` from the config of `apoc.import.csv()` . Is it possible to have an alternative to this functionality from apoc.

szarnyasg:
Fixes #1038

Adapt the [`--ignore-duplicate-nodes` option](https://neo4j.com/docs/operations-manual/current/tools/import/options/#import-tool-option-ignore-duplicate-nodes) from the offline import tool.

## Proposed Changes

- Add `ignoreDuplicateNodes` option to the `CsvLoaderConfig` class
- When detecting a duplicate node id, based on the value of `ignoreDuplicateNodes`, the import either crashes (throws an `IllegalStateException`) or ignores the duplicate.
- Added relevant tests.

### DIFF
--- a/src/main/java/apoc/export/csv/CsvEntityLoader.java
+++ b/src/main/java/apoc/export/csv/CsvEntityLoader.java
@@ -92,10 +92,22 @@ public class CsvEntityLoader {
                         loadCsvCompatibleHeader, line, lineNo, false, mapping, Collections.emptyList(), results
                 );
 
-                // create node and add id to the mapping
+                final String nodeCsvId = result.map.get(idAttribute.get()).toString();
+
+                // if 'ignore duplicate nodes' is false, there is an id field and the mapping already has the current id,
+                // we either fail the loading process or skip it depending on the 'ignore duplicate nodes' setting
+                if (idField.isPresent() && idspaceIdMapping.containsKey(nodeCsvId)) {
+                    if (clc.getIgnoreDuplicateNodes()) {
+                        continue;
+                    } else {
+                        throw new IllegalStateException("Duplicate node with id " + nodeCsvId + " found/");
+                    }
+                }
+
+                // create node and add its id to the mapping
                 final Node node = db.createNode();
                 if (idField.isPresent()) {
-                    idspaceIdMapping.put(result.map.get(idAttribute.get()).toString(), node.getId());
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/export/csv/CsvEntityLoader.java
```diff
@@ -92,10 +92,22 @@ public void loadNodes(final String fileName, final List<String> labels, final Gr
...
-                // create node and add id to the mapping
+                final String nodeCsvId = result.map.get(idAttribute.get()).toString();
+
+                // if 'ignore duplicate nodes' is false, there is an id field and the mapping already has the current id,
+                // we either fail the loading process or skip it depending on the 'ignore duplicate nodes' setting
+                if (idField.isPresent() && idspaceIdMapping.containsKey(nodeCsvId)) {
+                    if (clc.getIgnoreDuplicateNodes()) {
+                        continue;
+                    } else {
+                        throw new IllegalStateException("Duplicate node with id " + nodeCsvId + " found/");
```
jexp: probably emit the whole line with line-number ot make it easier to find?

MerkinMa:
Before calling apoc.import.csv, if __csv_id already exists in the DB, can it be ignoredDuplicateNodes? Try printing idspaceIdMapping.containsKey(nodeCsvId) and find that it returns false, but the same __csv_id is existed in DB before.

szarnyasg:
@MerkinMa good question. `apoc.import.csv` is designed to be used for loading a set of nodes and edges together (in a single APOC call) as it builds the id mappings when loading the nodes and keeps this in memory for loading the edges.

To determine whether an id (e.g. the `__csv_id` property) already exists in the database, we would have to repeatedly query the database for this id. This would be slow if this property is not indexed so first we'd have to build an index for it.

In any case, we'd need to make some design decisions before implementing this. Can you share your scenario for using it (e.g. multiple APOC calls for loading from different sets of CSVs)? Also, can you open a separate issue?

MerkinMa:
@szarnyasg, added new issue #1048

PR #1050, opened 2019-01-23
fixes #1027 - Graph Refactoring Procedures Documentation

### LINKED ISSUE #1027: Graph Refactoring Procedures Documentation
[Documentation listed here ](https://neo4j-contrib.github.io/neo4j-apoc-procedures/#_graph_refactoring_procedures) states that one of the operation types is named "**override**". However, after looking at the source code, the real key is **overwrite**. The documentation needs to be updated.

The **override** key behaves as if using the **discard** key because it's not a recognized value. The real value is **overwrite**.

AngeloBusato:
Fixes #1027 

Graph Refactoring Procedures Documentation

  - Update documentation with right parameter value


jexp:
Can we allow both in the code?

AngeloBusato:
@jexp PR Updated

PR #1052, opened 2019-01-24
fixes #1044 - Drop Duplicate Neighbors

### LINKED ISSUE #1044: Drop Duplicate Neighbors
It'd be cool if we had `apoc.coll.dropDuplicateNeighbors` that would remove consecutive values that are the same. Here's how to do it in Cypher:

```
WITH [1,2,3,4,4,5,6,6,7] AS items
RETURN [i in range(0, size(items)-1) WHERE i=0 OR items[i] <> items[i-1] | items[i] ] AS items
```

Here I use it for numeric values, but in my actual problem I wanted to use it to remove duplicate consecutive nodes in a path

AngeloBusato:
Fixes #1044 

Drop Duplicate Neighbors

  - new procedure which drop duplicate neighbors in a list of elements (node, relationships, numbers, strings ..)

### DIFF
--- a/src/main/java/apoc/coll/Coll.java
+++ b/src/main/java/apoc/coll/Coll.java
@@ -828,4 +828,22 @@ public class Coll {
 		if (values == null) return false;
         return new HashSet(values).size() == values.size();
     }
+
+
+
+    @UserFunction
+    @Description("apoc.coll.dropDuplicateNeighbors(list) - remove duplicate consecutive objects in a list")
+    public List<Object> dropDuplicateNeighbors(@Name("list") List<Object> list){
+        List<Object> newList = new ArrayList<>();
+
+        Object last = null;
+        for (Object element : list) {
+            if (!element.equals(last)) {
+                newList.add(element);
+                last = element;
+            }
+        }
+
+        return newList;
+    }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/apoc/coll/Coll.java
```diff
@@ -828,4 +828,42 @@ public boolean different(@Name("values") List<Object> values) {
...
+                    newList.add(element);
+                } else {
+                    hasSameType = false;
+                    break;
+                }
+            }
+        }
+
+        if (!hasSameType) {
+            throw new IllegalArgumentException("List elements must be the same type");
```
jexp: I don't think we need this.

src/main/java/apoc/coll/Coll.java
```diff
@@ -828,4 +828,42 @@ public boolean different(@Name("values") List<Object> values) {
...
+    @Procedure
+    @Description("apoc.coll.dropDuplicateNeighbors(list) - remove duplicate consecutive objects in a list")
+    public Stream<ListResult> dropDuplicateNeighbors(@Name("list") List<Object> list){
+        boolean hasSameType = true;
+        List<Object> newList = new ArrayList<>();
+        newList.add(list.get(0));
+        Class first = list.get(0).getClass();
+
+        for (int i = 1; i < list.size(); i++) {
+            Object element = list.get(i);
```
jexp: Cypher lists are not good for index access.
Just use foreach, you don't use the index anywhere.
And it's fine to have a "last" element.

AngeloBusato:
@jexp PR Updated

jexp:
Please turn it into a function.

jexp:
@mneedham as it will lose the type of the list do you think, we need specific ones for nodes/numbers/strings ?

AngeloBusato:
@jexp PR Updated

PR #1053, opened 2019-01-24
fixes #899 - make apoc.meta.* sample at batches

### LINKED ISSUE #899: make apoc.meta.* sample at batches
i.e. if there are 100000 user nodes, sample each 1000th node (jump over the iterateator from findNodes()

AngeloBusato:
Fixes #899 

make apoc.meta.* sample at batches

  - add `sample` config on apoc.meta procedures for sample the dataset

### DIFF
--- a/docs/overview.adoc
+++ b/docs/overview.adoc
@@ -150,8 +150,8 @@ Returns a virtual graph that represents the labels and relationship-types availa
 | CALL apoc.meta.graphSample() | examines the database statistics to build the meta graph, very fast, might report extra relationships
 | CALL apoc.meta.graph | examines the database statistics to create the meta-graph, post filters extra relationships by sampling
 | CALL apoc.meta.subGraph({labels:[labels],rels:[rel-types],excludes:[label,rel-type,...]}) | examines a sample sub graph to create the meta-graph
-| CALL apoc.meta.data({sample:<number>}) | examines a subset of the graph to provide a tabular meta information, with sample config the dataset is sampled by the provided number
-| CALL apoc.meta.schema({sample:<number>}) | examines a subset of the graph to provide a map-like meta information, with sample config the dataset is sampled by the provided number
+| CALL apoc.meta.data([{sample:1000,maxRels:100}]) | examines a subset of the graph to provide a tabular meta information, with sample config the dataset is sampled by the provided number  (default 1000 nodes per label, -1 for all)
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/meta/Meta.java
```diff
@@ -443,12 +446,14 @@ private void collectStats(Collection<String> labelNames, Collection<String> relT
                 }
             }
             try (ResourceIterator<Node> nodes = db.findNodes(label)) {
-                int count = 0;
-                while (nodes.hasNext() && count++ < SAMPLE) {
+                int count = 1;
+                while (nodes.hasNext()) {
                     Node node = nodes.next();
+                    if(count++ % config.getSample() == 0) {
```
jexp: I don't think this is correct yet. As each node label has different counts we have to normalize the sample rate for each label. 

E.g. if you want 1000 samples
and you have 

- 100 Node A you want all
- 1000 Node B you want all
- 10000 Node C you want each 10th
- 1M Node D you want each 1000th
etc.

AngeloBusato:
@jexp PR Updated

docs/overview.adoc
```diff
@@ -150,8 +150,8 @@ Returns a virtual graph that represents the labels and relationship-types availa
 | CALL apoc.meta.graphSample() | examines the database statistics to build the meta graph, very fast, might report extra relationships
 | CALL apoc.meta.graph | examines the database statistics to create the meta-graph, post filters extra relationships by sampling
 | CALL apoc.meta.subGraph({labels:[labels],rels:[rel-types],excludes:[label,rel-type,...]}) | examines a sample sub graph to create the meta-graph
-| CALL apoc.meta.data | examines a subset of the graph to provide a tabular meta information
-| CALL apoc.meta.schema | examines a subset of the graph to provide a map-like meta information
+| CALL apoc.meta.data({sample:{label1:<number of each elements>, label2:<number of each elements>, ...}}) | examines a subset of the graph to provide a tabular meta information, with sample config the dataset is sampling by label
```
jexp: No we don't want to provide that per label. We just want to provide: "sample at least XXX" elements.
E.g. `sample: 1000`.

jexp: We can grab the total counts from the count store and compute the partial skip numbers.

jexp: ping @AngeloBusato

AngeloBusato:
@jexp pr updated

PR #1058, opened 2019-01-28
fixes #1049 - apoc.export.cypher.all fails when a compound constraint exists

### LINKED ISSUE #1049: apoc.export.cypher.all fails when a compound constraint exists
`CALL apoc.export.cypher.all('/data/seed.cypher',{format:'cypher-shell'});` works when compound constraint exist.

```
CALL apoc.export.cypher.all('/data/seed.cypher',{format:'cypher-shell'});
Failed to invoke procedure `apoc.export.cypher.all`: Caused by: java.util.NoSuchElementException: More than one element in java.util.AbstractList$Itr@693ec7a0. First element is 'adTagType' and the second element is 'name'
```

Create a compound constraint on a node. I have the following:

```
create constraint on (t:AdTag) assert (t.adTagType, t.name) is node key;
```

```
MERGE (a:AdTag:test {name: "E", adTagType: "ESRB-test"});
MERGE (a:AdTag:test {name: "Under 18", adTagType: "Age-test"});
MERGE (a:AdTag:test {name: "13-17",  adTagType: "Age-test"});
MERGE (a:AdTag:test {name: "18-24", adTagType: "Age-test"});
MERGE (a:AdTag:test {name: "25-34", adTagType: "Age-test"});
```

### Steps (Mandatory)

  1. Run `create constraint on (t:AdTag) assert (t.adTagType, t.name) is node key;`
  1. Add the sample data above.
  1. Run `CALL apoc.export.cypher.all('/data/seed.cypher',{format:'cypher-shell'});`

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: OS X 10.13.6
  - Neo4j: Enterprise 3.5.0
  - Neo4j-Apoc: 3.5.0.1

AngeloBusato:
Fixes #1049 

apoc.export.cypher.all fails when a compound constraint exists

  - compound constraind are now managed

### DIFF
--- a/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
+++ b/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
@@ -1,5 +1,6 @@
 package apoc.export.cypher;
 
+import apoc.export.cypher.formatter.CypherFormatterUtils;
 import apoc.export.util.*;
 import apoc.export.cypher.formatter.CypherFormatter;
 import apoc.util.Util;
@@ -13,6 +14,8 @@ import org.neo4j.helpers.collection.Iterables;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 import static apoc.export.cypher.formatter.CypherFormatterUtils.*;
 
@@ -26,7 +29,7 @@ import static apoc.export.cypher.formatter.CypherFormatterUtils.*;
 public class MultiStatementCypherSubGraphExporter {
 
     private final SubGraph            graph;
-    private final Map<String, String> uniqueConstraints = new HashMap<>();
+    private final Map<String, Set<String>> uniqueConstraints = new HashMap<>();
     private Set<String> indexNames        = new LinkedHashSet<>();
     private Set<String> indexedProperties = new LinkedHashSet<>();
     private Long artificialUniques = 0L;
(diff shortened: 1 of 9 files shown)


jexp:
For compound keys we also have to create MERGE and MATCh statements with multiple keys to assert the concrete uniqueness

So we might turn the query generator into always using multiple properties and then a size=1 is just a special case.

src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -160,7 +160,12 @@ private void exportSchema(PrintWriter out) {
             String label = index.getLabel().name();
             Iterable<String> props = index.getPropertyKeys();
             if (index.isConstraintIndex()) {
-                String cypher = this.cypherFormat.statementForConstraint(label, Iterables.single(props));
+                String cypher;
+                if(Iterables.count(props) > 1) {
+                    cypher = this.cypherFormat.statementForCompoundConstraint(label, props);
```
jexp: can't we handle that inside of the this.cypherFormat ?

jexp:
So mean mean this as a fix to the original exception but not as a proper fix b/c those composite keys are actually ignored and lead to different nodes being collapsed on whichever property comes first in an constraint?

jexp:
@conker84 ^^^

AngeloBusato:
@jexp PR Updated

jexp:
@sarmbruster suggested to move the enterprise tests to a testcontainer, at least for 3.5 we'd need that.

What about my comment above? 
Can we make sure that it doesn't happen then?

I.e. you have  two nodes with a composite key, and you export a relationship.
What is used as "MATCH" key?

conker84:
@jexp  I'll handle the #1079 ok?

conker84:
> you have two nodes with a composite key, and you export a relationship.
> What is used as "MATCH" key?

We'll add a test case for that

AngeloBusato:
@jexp PR Updated

jexp:
Thanks a lot.

PR #1062, opened 2019-01-31
APOC CSV Metrics and File Streaming

moxious:
Introduces ```apoc.file.stream("logs/debug.log") and apoc.metrics.get("metric_name")``` which permit pure cypher access to files on disk, in two specific cases: line-by-line logfiles, and CSV access to internal neo4j metrics that help debug common issues.

`apoc.file.stream` will accept either relative paths (interpreted to be relative to $neo4jHome) or absolute paths on the host.

Reference: https://github.com/moxious/halin/issues/60

  - Permit loading CSV metrics as a special case of apoc.load.csv() -- but where user does not have to know/guess the $NEO4J_HOME path ahead of time.
  - Permit loading of files line-by-line, which can be thought of as a special case of CSV loading, except without a sensible delimiter.

## Other Issues

Need to talk about unit tests.  This is fairly easy to unit test, but since this code does local file access and the impermanent graph database test harness doesn't materialize files on disk, it's not straightforward to test these functions without an integration testing setup.


jexp:
I guess we should do: `apoc.logs.stream("debug")` for safety reasons.

moxious:
How come?  If we do this, then you can't stream non-log files as a utility

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            try {
+                v = Double.parseDouble(entry.map.get("value").toString());
+            } catch(NumberFormatException exc) {
+                return null;
+            }
+
+            return new Neo4jMetricEntry(t, v, metricName);
+        }).filter(p -> p != null);
```
moxious: Talked to Dana Canzano -- this odd little bit is necesary due to a bug in the way that neo4j does CSV metrics.  Apparently after restarts and some other situations, the CSV header will be inserted into the file a second or a third time.   Note above that if the Double.parseDouble fails, we return null, and here they get filtered out. This does come up in practice.

src/main/java/apoc/file/Neo4jFileStream.java
```diff
@@ -0,0 +1,89 @@
...
+                return -1;
+            } else if(this.lineNo > o.lineNo) {
+                return 1;
+            }
+
+            return 0;
+        }
+    }
+
+    @Procedure
```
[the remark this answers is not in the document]
jexp: to limit use to admins !?

src/main/java/apoc/file/Neo4jFileStream.java
```diff
@@ -0,0 +1,89 @@
...
+    @Description( "apoc.file.stream(path, config) - retrieve system or log file contents" )
+    public Stream<FileEntry> stream(
+            @Name("path") String path,
+            @Name(value = "config",defaultValue = "{}") Map<String, Object> config) {
+        String neo4jHome = ApocConfiguration.allConfig.get("unsupported.dbms.directories.neo4j_home");
+        // String logDir = ApocConfiguration.allConfig.get("dbms.directories.logs");
+
+        // Prepend neo4jHome if it's a relative path, and use the user's path otherwise.
+        String fullPath = path.startsWith(File.separator) ? path : (neo4jHome + File.separator + path);
```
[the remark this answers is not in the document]
jexp: new File(parent, child) does the correct separators and avoids injection.

src/main/java/apoc/file/Neo4jFileStream.java
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            Stream<String> stream = Files.lines(Paths.get(fullPath));
+            final AtomicLong lineNumber = new AtomicLong(0);
+            final String p = canonicalPath;
+
+            Stream<FileEntry> entries = stream.map(line -> new FileEntry(lineNumber.getAndIncrement(), line, p));
+
+            // Useful for tailing logfiles.
+            if(config.containsKey("last")) {
+                return entries.sorted(Collections.reverseOrder()).limit(new Double(config.get("last").toString()).longValue());
```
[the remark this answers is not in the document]
jexp: extract the limit to a variable makes it easier to read.

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Arrays;
+import java.util.Map;
+import java.util.stream.Stream;
+
+public class Metrics {
+    /**
+     * DAO for a single line in a metrics/*.csv file.
+     * Value is abstracted to double; in reality some are int, some are float, some are double.
+     */
+    public class Neo4jMetricEntry {
```
jexp: public static class

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+        public long t;
```
jexp: better name: `timestamp`

jexp: all variables better final

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+            this.t = t;
+            this.value = value;
+            this.metric = metric;
+        }
+    }
+
+    public class Neo4jMetric {
+        public String name;
+        public long lastUpdated;
+        public String path;
```
jexp: should we expose the path to the outside world?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+        public String metric;
+
+        public Neo4jMetricEntry(long t, double value, String metric) {
+            this.t = t;
+            this.value = value;
+            this.metric = metric;
+        }
+    }
+
+    public class Neo4jMetric {
```
jexp: public static class

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+        public String name;
```
jexp: public final

src/main/java/apoc/metrics/Metrics.java
jexp: should we encode the resolution? utcEpocTimeMillis ?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+            this.name = name;
+            this.lastUpdated = lastUpdated;
+            this.path = path;
+        }
+    }
+
+    @Procedure
+    @Description("apoc.metrics.list() - get a list of available metrics")
+    public Stream<Neo4jMetric> list() {
+        Object metricsSetting = ApocConfiguration.allConfig.get("dbms.directories.metrics");
```
jexp: whitelist it and get it from the apoc config.

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+
+        File metricsDir = new File(metricsSetting == null ?
+                ApocConfiguration.allConfig.get("unsupported.dbms.directories.neo4j_home") + File.separator + "metrics"
```
jexp: whitelist it and get it from the apoc config.

jexp: `new File(neo4jHome, "metrics")`

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+                : metricsSetting.toString());
+
+        if (!metricsDir.exists() || !metricsDir.isDirectory()) {
```
jexp: canRead() ?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+            return Stream.empty();
+        }
+
+        FileUtils.checkReadAllowed(metricsDir.toURI().toString());
+
+        return Arrays.asList(metricsDir.list())
```
jexp: listFiles() ?

jexp: listFiles can also take a filter so no need to post-filter.

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+                .map(metricFile -> {
+                    String metricName = metricFile.substring(0, metricFile.length() - 4);
+                    File f = new File(metricsDir + File.separator + metricFile);
+
+                    return new Neo4jMetric(metricName, f.lastModified(), f.getAbsolutePath());
+                });
+    }
+
+    @Procedure
+    @Description("apoc.metrics.get(metric_name) - retrieve a system metric")
```
jexp: metricName

jexp: Explain config

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+     * relative paths.
+     */
+    public Stream<Neo4jMetricEntry> get(
+            @Name("metric_name") String metricName,
+            @Name(value = "config",defaultValue = "{}") Map<String, Object> config) {
+        // These config parameters are generally true of Neo4j metrics.
+        config.putIfAbsent("sep", ",");
+        config.putIfAbsent("header", true);
+
+        String neo4jHome = ApocConfiguration.allConfig.get("unsupported.dbms.directories.neo4j_home");
```
jexp: whitelist and get from apoc config

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+        String url = "file://" + neo4jHome + File.separator + "metrics" + File.separator + metricName + ".csv";
```
jexp: use file constructor and then URI

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+
+        FileUtils.checkReadAllowed(url);
+
+        return new LoadCsv().csv(url, config).map(entry -> {
```
jexp: is apoc load csv overkill for this? because it does a lot of mapping and stuff, which might affect performance
compared to a plain CsvReader?

perahps something to consider for later.

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,111 @@
...
+            long t = 0;
+            double v = 0.0;
+
+            try {
+                t = Long.parseLong(entry.map.get("t").toString());
+            } catch (NumberFormatException exc) {
```
jexp: combine the two try-catches

src/main/java/apoc/metrics/Metrics.java
jexp: can we easier just compare the value to `"v"` / `"t"`

src/main/java/apoc/file/Neo4jFileStream.java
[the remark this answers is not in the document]
moxious: based on system doc I pick "DBMS" because its' definition is:

> This procedure will perform system operations - i.e. not against the graph

src/main/java/apoc/file/Neo4jFileStream.java
```diff
@@ -0,0 +1,89 @@
...
+                return 1;
+            }
+
+            return 0;
+        }
+    }
+
+    @Procedure
+    @Description( "apoc.file.stream(path, config) - retrieve system or log file contents" )
+    public Stream<FileEntry> stream(
```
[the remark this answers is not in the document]
moxious: Intended to be wider than logs.

src/main/java/apoc/metrics/Metrics.java
moxious: I agree that timestamp is a better name but I pick t quite on purpose.  This naming is chosen to match the metrics documentation in the system manual.

The column they expose is t, so I expose t.  I would prefer it be named timestamp, but for avoiding confusion I was thinking it's best to follow the naming convention of what the product docs already say.

src/main/java/apoc/metrics/Metrics.java
moxious: I thought it might be useful to some but it's not a hard requirement so safer to leave out, it can always be added later.

src/main/java/apoc/metrics/Metrics.java
moxious: I don't think/hope it's too heavy -- my thought here was that I would need some options in some cases to read values back, and I wanted to reuse all of the config options that come with apoc.load.csv so they are not necessary to re-implement.

src/main/java/apoc/metrics/Metrics.java
jexp: I think as consumers of the procedure might not know or care where the data comes from and I don't think that the single character naming by the kernel team is such a good idea. I'd rather on the side of clear column names.

src/main/java/apoc/metrics/Metrics.java
jexp: Ok, then let's leave it out for now.

src/main/java/apoc/file/Neo4jFileStream.java
```diff
@@ -0,0 +1,89 @@
...
+
+        try {
+            Stream<String> stream = Files.lines(Paths.get(fullPath));
+            final AtomicLong lineNumber = new AtomicLong(0);
+            final String p = canonicalPath;
+
+            Stream<FileEntry> entries = stream.map(line -> new FileEntry(lineNumber.getAndIncrement(), line, p));
+
+            // Useful for tailing logfiles.
+            if(config.containsKey("last")) {
```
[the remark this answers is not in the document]
moxious: I've investigated.  Time range filters, no.  Because the format of logfiles aren't consistent, and because parsing would be required.  I think to make it simple to stream lines of files, but not to go further than this, otherwise a bunch of additional procedures are needed specialized for each file.  Because these files can include stacktraces in them, the parsing gets quite tricky quickly if what you want to do is give a stream of "log entries" rather than a stream of lines.   They're not the same unfortunately.   This is intended as a utility just to provide access to this data but it's left to the user how to access.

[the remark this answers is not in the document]
jexp: Makes sense.

src/main/java/apoc/metrics/Metrics.java
jexp: Oki

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,107 @@
...
+            double v = 0.0;
+
+            try {
+                t = Long.parseLong(entry.map.get("t").toString());
+                v = Double.parseDouble(entry.map.get("value").toString());
+            } catch (NumberFormatException exc) {
+                return null;
+            }
+
+            return new Neo4jMetricEntry(t, v, metricName);
```
jexp: you can move this into the try catch too so it's cleaner

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,107 @@
...
+        config.putIfAbsent("header", true);
+
+        String neo4jHome = ApocConfiguration.get("unsupported.dbms.directories.neo4j_home", "");
+        String url = new File(neo4jHome + File.separator + "metrics", metricName + ".csv")
+                .toURI().toString();
+
+        FileUtils.checkReadAllowed(url);
+
+        return new LoadCsv().csv(url, config).map(entry -> {
+            long t = 0;
```
jexp: move these into the try catch too

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public Stream<Neo4jMetricEntry> get(
+            @Name("metricName") String metricName,
+            @Name(value = "config",defaultValue = "{}") Map<String, Object> config) {
+        // These config parameters are generally true of Neo4j metrics.
+        config.putIfAbsent("sep", ",");
+        config.putIfAbsent("header", true);
+
+        String neo4jHome = ApocConfiguration.get("unsupported.dbms.directories.neo4j_home", "");
+        String url = new File(neo4jHome + File.separator + "metrics", metricName + ".csv")
```
jexp: new File(new File(neo4jHome, "metrics"), metricName + ".csv")

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,107 @@
...
+                .toURI().toString();
+
+        FileUtils.checkReadAllowed(url);
```
jexp: remove to allow fro cloud?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,107 @@
...
+        File metricsDir = (
+                (metricsSetting == null) ?
+                        new File(ApocConfiguration.get("unsupported.dbms.directories.neo4j_home", ""), "metrics") :
+                        new File(metricsSetting.toString()));
+
+        if (!metricsDir.exists() || !metricsDir.isDirectory() || !metricsDir.canRead()) {
+            return Stream.empty();
+        }
+
+        FileUtils.checkReadAllowed(metricsDir.toURI().toString());
```
jexp: Remove for cloud?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,107 @@
...
+        Object metricsSetting = ApocConfiguration.get("dbms.directories.metrics", "");
+
+        File metricsDir = (
+                (metricsSetting == null) ?
+                        new File(ApocConfiguration.get("unsupported.dbms.directories.neo4j_home", ""), "metrics") :
+                        new File(metricsSetting.toString()));
+
+        if (!metricsDir.exists() || !metricsDir.isDirectory() || !metricsDir.canRead()) {
+            return Stream.empty();
+        }
```
jexp: Perhaps you can extract line 46 to 55 into a function that returns a File or null and return Stream.empty() on null
And then reuse the same function in `apoc.metrics.get`

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+
+public class Metrics {
+    /**
+     * DAO for a single line in a metrics/*.csv file.
+     * Value is abstracted to double; in reality some are int, some are float, some are double.
+     */
+    public static class Neo4jMetricEntry {
+        public final long t;
```
jexp: _please_ change it to timestamp :)

src/main/java/apoc/file/Neo4jFileStream.java
```diff
@@ -0,0 +1,83 @@
...
+
+        return canonicalPath.contains(neo4jHome);
+    }
+
+    @Procedure(mode=Mode.DBMS)
+    @Description( "apoc.file.stream(path, { last: n }) - retrieve system or log file contents, optionally return only the last n lines" )
+    public Stream<FileEntry> stream(
+            @Name("path") String path,
+            @Name(value = "config",defaultValue = "{}") Map<String, Object> config) {
+        String neo4jHome = ApocConfiguration.get("unsupported.dbms.directories.neo4j_home", null);
```
jexp: Seeing this repeated feels as if we might want to add a dedicated method to ApocConfiguration instead to retrieve the home.

src/main/java/apoc/file/Neo4jFileStream.java
```diff
@@ -0,0 +1,83 @@
...
+
+        // Prepend neo4jHome if it's a relative path, and use the user's path otherwise.
+        File f = path.startsWith(File.separator) ? new File(path) : new File(neo4jHome, path);
```
jexp: Perhaps even have a method in `FileUtils.directoryInNeo4jHome(dirName)`  -> file or null
`FileUtils.fileInNeo4jHome(fileName)` -> file or null

moxious: I'm going to do a getLogDir and getMetricsDir in FileUtil.  I don't want to do a boolean check to see if something is in that directory because the File API will do this for me I think if I make it easy to get the right directory.

Putting those things in FileMEtrics also allows me to centralize access to ApocConfiguration instead of sprinkling it everywhere.

jexp:
Something I would love to see before the next release @moxious is a documentation page on the new procedures, so the parameters & results are documented and also the usage in Halin (with a pointer) is  explained.

PR #1083, opened 2019-02-04
fix #1071 apoc.periodic.commit should check for `limit`

### LINKED ISSUE #1071: apoc.periodic.commit should check for `limit`
if the statement passed in as an argument does not contain the word `limit`, it should fail with a reasonable error message.

sarmbruster:
This - of course - is a somewhat breaking changes. Usages that do not have a `limit` will get an exception.

### DIFF
--- a/src/main/java/apoc/periodic/Periodic.java
+++ b/src/main/java/apoc/periodic/Periodic.java
@@ -2,6 +2,7 @@ package apoc.periodic;
 
 import apoc.Pools;
 import apoc.util.Util;
+import org.apache.commons.lang.StringUtils;
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.Result;
 import org.neo4j.helpers.collection.Iterators;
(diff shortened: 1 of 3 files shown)


src/main/java/apoc/periodic/Periodic.java
```diff
@@ -49,11 +50,15 @@
 
     @Procedure(mode = Mode.WRITE)
     @Description("apoc.periodic.commit(statement,params) - runs the given statement in separate transactions until it returns 0")
-    public Stream<RundownResult> commit(@Name("statement") String statement, @Name("params") Map<String,Object> parameters) throws ExecutionException, InterruptedException {
+    public Stream<RundownResult> commit(@Name("statement") String statement, @Name(value = "params", defaultValue = "") Map<String,Object> parameters) throws ExecutionException, InterruptedException {
         Map<String,Object> params = parameters == null ? Collections.emptyMap() : parameters;
         long total = 0, executions = 0, updates = 0;
         long start = nanoTime();
 
+        if (!StringUtils.containsIgnoreCase(statement, "limit")) {
```
jexp: perhaps with white-spaces around it?

src/main/java/apoc/periodic/Periodic.java
```diff
@@ -49,11 +50,15 @@
 
     @Procedure(mode = Mode.WRITE)
     @Description("apoc.periodic.commit(statement,params) - runs the given statement in separate transactions until it returns 0")
```
jexp: should we document that it needs a `LIMIT $limit` ?

sarmbruster: Agreed. Will fix this.

src/main/java/apoc/periodic/Periodic.java
sarmbruster: Agreed. Will fix this.

PR #1085, opened 2019-02-05
fixes #1076 - apoc.coll.flatten recursive

### LINKED ISSUE #1076: apoc.coll.flatten -> optiona recursive parameter
also allow mixed types

tomgeudens:
Fixes #1076 

Replaced single level flatten of lists with any-level flatten of lists

- added recursive parameter to apoc.coll.flatten
- apoc.coll.flatten does now allow plain objects at the lowest level


jexp:
Thanks a lot Tom, can you add a few tests for the recursive flattening? And also if the actual list is mixed between scalars and collections?

PR #1087, opened 2019-02-07
Added apoc.refactor.cloneSubgraph() for cloning and reanchoring a subgraph of nodes and relationships between them

### LINKED ISSUE #1075: Request new refactoring procedure: apoc.refactor.cloneSubgraph()
## Feature description
We've had requests via the community on copying parts of a graph, but current approach to cloning nodes and relationships is messy when the intent is to copy out part of a graph and attach it elsewhere, such as attempting to copy a branch of a tree structure and attach it to a different root node. In these cases, what's really being requested is a subgraph consisting of the cloned nodes with the cloned relationships between those nodes, but omitting any relationships outside of those cloned nodes (including to the originals).

This requests a new procedure `apoc.refactor.cloneSubgraph()` which takes in the collection of nodes and relationships to clone (such as those returned by apoc.path.subgraphAll()). The caller could alternately use `apoc.algo.cover()` to get the collection of relationships between the list of nodes, if all are desired.

Signature:

`apoc.refactor.cloneSubgraph([nodes], [relationships], [[nodeToBeReplaced, replacementNode]]) YIELD input, output, error`
The first two parameters are mandatory, the lists of nodes and relationships forming the subgraph to be cloned.
The third parameter is optional, a list of pairs of nodes, the node to be replaced and the replacement node.
The yielded values would be the same as the cloneNodes() and cloneNodesWithRelationships() procedures.

The procedure would first clone the nodes, then using an id mapping from originals to clones, the relationships would be cloned between the cloned nodes only. If no mapping exists (user somehow supplied relationships that don't have both connections in the node list parameter) the relationship will be ignored, with the exception of the behavior below.

If the third parameter (replacement nodes list) is present, then it would modify the relationship creation step. Rather than using the id map to find the cloned nodes to use to clone the each relationship, the supplied map pairs would be used to construct a secondary mapping of to-be-replaced nodes to replacement nodes, and this map would be consulted first. 

## Considered alternatives
I considered using the existing clone procedures instead, adding a new parameter to control the functionality, but I wanted to provide more control over the relationships that were cloned rather than always cloning all relationships. The signatures were also getting a bit long, especially considering the replacement pairs parameter.

Here is an example query showing cloning of a branch of a tree structure and attaching it to a new root:

```
MATCH (a:Root {name:'A'}), (b:Root {name:'B'})
CALL apoc.path.subgraphAll(a, {relationshipFilter:'CHILD>'}) YIELD nodes, relationship
CALL apoc.refactor.cloneSubgraph(tail(nodes), relationships, [[a, b]]) YIELD input, output, error
RETURN input, output, error
```

Note that by using `tail(nodes)` we skip the starting node `a`, so it does not get cloned, but the list of replacement pairings will ensure that `b` will be a standin for `a` when the relationships are cloned.

InverseFalcon:
Fixes / implements feature request #1075 

Introduces apoc.refactor.cloneSubgraph for cloning and reanchoring a subgraph of nodes and relationships between them.

JUnit tests and overview documentation included (refactoring procs).

Requesting backport to 3.4 as well if possible.


docs/overview.adoc
```diff
@@ -708,8 +708,9 @@ CALL apoc.periodic.rock_n_roll('match (p:Person) return id(p) as id_p', 'MATCH (
 
 [cols="1m,5"]
 |===
-| call apoc.refactor.cloneNodes([node1,node2,...],[withRelationships=false],[skipProperties=[]]) |  clone nodes with their labels and properties
-| call apoc.refactor.cloneNodesWithRelationships([node1,node2,...]) | deprecated
+| call apoc.refactor.cloneNodes([node1,node2,...],[withRelationships=false],[skipProperties=[]]) YIELD input, output, error |  clone nodes with their labels and properties
+| call apoc.refactor.cloneNodesWithRelationships([node1,node2,...]) YIELD input, output, error | deprecated
+| call apoc.refactor.cloneSubgraph([node1,node2,...], [rel1,rel2,...], [reanchorNodePairs=[]], [skipProperties=[]]) YIELD input, output, error | clone nodes with their labels and properties (optionally skipping specific properties), and clone the given relationships (will exist between cloned nodes only or be reanchored according to reanchor node pairings)
```
jexp: should it also support a list of paths? that constitute the same subgraph?
Can we move config into a config map, so we're more flexible going forward?

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -114,6 +113,91 @@
...
+     * to the original nodes) and to reanchor any relationships from an old root node (which will not be cloned) to a different existing root node.
+     *
+     */
+    @Procedure(mode = Mode.WRITE)
+    @Deprecated
+    @Description("apoc.refactor.cloneSubgraph([node1,node2,...], [rel1,rel2,...], [reanchorNodePairs = [] ], [skipProperties = [] ]) YIELD input, output, error | " +
+            "clone nodes with their labels and properties (optionally skipping specific properties), and clone the given relationships (will exist between cloned nodes only or be reanchored according to reanchor node pairings)")
+    public Stream<NodeRefactorResult> cloneSubgraph(@Name("nodes") List<Node> nodes,
+                                                    @Name("rels") List<Relationship> rels,
+                                                    @Name(value = "reanchors", defaultValue = "[]") List<List<Node>> reanchors,
```
jexp: move into a config map so we're more flexible going forward.

jexp: don't fully understand the reachoring, as the "new" nodes don't yet exist.
Or do you mean reconstruct the relationships between the old pair with the new node and the old target?
So the old target node would have twice the relationships in the end?

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -114,6 +113,91 @@
...
+                                                    @Name(value = "skipProperties", defaultValue = "[]") List<String> skipProperties) {
+        Map<Node, Node> copyMap = new LinkedHashMap<>(nodes.size());
```
jexp: is the order important? otherwise HashMap is good enough.

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -114,6 +113,91 @@
...
+        Map<Node, Node> reanchorMap = reanchors.isEmpty() ? Collections.emptyMap() : new LinkedHashMap<>(reanchors.size());
+        List<NodeRefactorResult> resultStream = new ArrayList<>();
+        if (nodes == null) return Stream.empty();
```
jexp: move to earlier

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -114,6 +113,91 @@
...
+
+        // populate reanchor map
+        for (List<Node> pairing : reanchors) {
```
jexp: move this into a method that returns just the reanchorMap

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -114,6 +113,91 @@
...
+        // clone nodes and populate copy map
+        for (Node node : nodes) {
+            if (node == null || reanchorMap.containsKey(node)) continue;
+            // nodes being reanchored from will NOT be cloned
+
+            NodeRefactorResult result = new NodeRefactorResult(node.getId());
+            try {
+                Node copy = copyLabels(node, db.createNode());
+
+                Map<String, Object> properties = node.getAllProperties();
```
jexp: don't we have the code for that already? including skipping? if node we should probably extend the copyProperties method with the skip-list.

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -114,6 +113,91 @@
...
+                copy = copyProperties(properties, copy);
+
+                resultStream.add(result.withOther(copy));
+                copyMap.put(node, copy);
+            } catch (Exception e) {
+                resultStream.add(result.withError(e));
+            }
+        }
+
+        // clone relationships, will be between cloned nodes and/or reanchors
```
jexp: how easy is it actually to provide the relationships within the structure?

I originally thought we would clone all rels _within_ the structure automatically?

src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -114,6 +113,91 @@
...
+        for (Relationship rel : rels) {
+            if (rel == null) continue;
+
+            Node oldStart = rel.getStartNode();
+            Node newStart = reanchorMap.get(oldStart);
```
jexp: getOrElse

jexp:
See my comments, please also add an section in the graph refactoring documentation page.

docs/overview.adoc
InverseFalcon: A list of paths may be a bit tricky, we would need to extract and de-duplicate the nodes and relationships from all paths. Easy enough to do in code with Sets, and I can see use cases for this. The problem is this messes a bit with the signature, 2 parameters for nodes and rels vs 1 parameter for paths. Remind me, can we overload procedure methods with the same name but different signatures? I didn't think we could. Maybe I could write a companion proc for this?

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: Config map makes sense, I'll refactor.

For reanchoring, rather than cloning one of the nodes and creation relationships to it, we create relationships to some existing node.

The prototypical use case is cloning a tree (could be a branch, but we'll use tree here) onto a different root node.

If we have (a:Root{name:'A'}) with the tree we want to clone and (b:Root{name:'B'}) where we want to reanchor, then when we have the nodes and rels, we can clone the tree, but reanchor to `b`:

```
...
CALL apoc.refactor.cloneSubgraph(treeNodes, treeRels, [[a, b]]) YIELD input, output, error
...
```
Any relationship in treeRels that went to `a`, rather than cloning `a` and cloning the rels to a's clone, we will not clone `a`, and instead clone the rels but attached to `b` instead.

We've had a couple asks on this exact case via stack overflow and the community site, so it's a valuable use case, but any insight into naming and description would be welcome.

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: Right, when I started implementation I thought I would need to iterate over the map's contents, but that's not happening, so HashMap it is.

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: Right

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: You could use apoc.algo.cover() for that. I wanted to support cases where we could be selective of which relationships get cloned. Other methods to get these relationships, such as via the path expander procs, may be more efficient.

Maybe we could say that if no relationships are provided that we automatically get all the relationships between the cloned nodes?

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: getOrDefault()? We pay the cost of the copyMap.get() access even when not needed, but I suppose that's trivial enough for map access.

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: Got it

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: The doCloneNodes() method has something similar, but it's processing this via a stream, which we can't use since we need to further process the results.

We also add the nodes to a map structure which doesn't look like is needed in the other method.

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: I'll rename these to redirections, that may be less confusing. New signature looks like this:

> apoc.refactor.cloneSubgraph([node1,node2,...], [rel1,rel2,...], {redirectRels:[], skipProperties:[]}) YIELD input, output, error | clone nodes with their labels and properties (optionally skipping any properties in the skipProperties list), and clone the given relationships (will exist between cloned nodes only or be redirected according to redirectRels node pairings)

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: Because of the above, I'm not going to reuse or combine the code. Maybe will look at it for a future refactor if needed.

src/main/java/apoc/refactor/GraphRefactoring.java
jexp: Can you add that to the docs too. (Can also be a later PR9.

src/main/java/apoc/refactor/GraphRefactoring.java
jexp: yep or a config option.

docs/overview.adoc
jexp: I guess we can just have a single List<Object> parameter, which can be paths, relationships or a mix of both.
And we just use a function to extract the distinct nodes and rels from those.

InverseFalcon: I'm just going to make a separate proc for it: apoc.refactor.cloneSubgraphFromPaths()

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: Okay, made `rels` optional, if missing or an empty list we'll internally run `apoc.algo.cover()` to get all the rels between the nodes and use that. Test cases updated too. Will update docs accordingly.

src/main/java/apoc/refactor/GraphRefactoring.java
InverseFalcon: Docs updated

PR #1088, opened 2019-02-07
add functions for node, relationship or graph hash calculation

sarmbruster:
-

### DIFF
--- a/docs/hashing.adoc
+++ b/docs/hashing.adoc
@@ -3,6 +3,8 @@
 The following functions do calculate hashsums over nodes, relationship or the entire graph. It takes into account all
 properties, node labels and relationship types.
 
+NOTE: be aware that the algorithm used for hashing might be changed from one apoc version to another. So you can only compare hashing results of two entities/graphs from the same or from different graph using the *very same* apoc version.
+
 The hashsum of a graph first calculates the hashsums for each node. This hashsum list is ordered and for each node the
 hashsum for all relationships and their endnode are added. This approach provides independence of internal ids.
 
@@ -12,7 +14,7 @@ Optionally you can supply a list of `propertyKeys` that should be ignored on all
 [opts=header,cols="m,a"]
 |===
 | function name | description
-| `apoc.hashing.fingerprintNodeOrRel(node_or_rel, <list_of_props_to_ignore>)` | calculates a md5 hashsum over the given node or relationship.
+| `apoc.hashing.fingerprint(object, <list_of_props_to_ignore>)` | calculates a md5 hashsum over the object. It deals gracefully with ordering (in case of maps), scalars, arrays.
 | `apoc.hashing.graph(<list_of_props_to_ignore>)` | calculates a md5 hashsum over the full graph.
 |===
 
--- a/docs/hashing.adoc
+++ b/docs/hashing.adoc
@@ -1,0 +1,19 @@
+= Hashing
+
+The following functions do calculate hashsums over nodes, relationship or the entire graph. It takes into account all
+properties, node labels and relationship types.
+
+The hashsum of a graph first calculates the hashsums for each node. This hashsum list is ordered and for each node the
+hashsum for all relationships and their endnode are added. This approach provides independence of internal ids.
+
+Optionally you can supply a list of `propertyKeys` that should be ignored on all nodes. This is, e.g. useful if you store
+`created=timestamp()` properties that should be ignored.
+
+[opts=header,cols="m,a"]
+|===
+| function name | description
+| `apoc.hashing.fingerprintNodeOrRel(node_or_rel, <list_of_props_to_ignore>)` | calculates a md5 hashsum over the given node or relationship.
+| `apoc.hashing.graph(<list_of_props_to_ignore>)` | calculates a md5 hashsum over the full graph.
+|===
+
+
--- a/src/main/java/apoc/hashing/Fingerprinting.java
+++ b/src/main/java/apoc/hashing/Fingerprinting.java
@@ -31,14 +31,17 @@ public class Fingerprinting {
         return withMessageDigest(md -> fingerprint(md, thing, excludedPropertyKeys));
     }
 
-    private void fingerprint(MessageDigest md, Object thing, List<String> excludedPropertyKeys) {
+    private void fingerprint(DiagnosingMessageDigestDecorator md, Object thing, List<String> excludedPropertyKeys) {
         if (thing instanceof Node) {
             fingerprintNode(md, (Node) thing, excludedPropertyKeys);
         } else if (thing instanceof Relationship) {
             fingerprintRelationship(md, (Relationship) thing, excludedPropertyKeys);
         } else if (thing instanceof Map) {
             Map<String, Object> map = (Map<String, Object>) thing;
-            map.entrySet().stream().sorted(Comparator.comparing(Map.Entry::getKey, String::compareTo)).forEach(entry -> {
+            map.entrySet().stream()
+                    .filter(e -> !excludedPropertyKeys.contains(e.getKey()))
+                    .sorted(Map.Entry.comparingByKey())
+                    .forEachOrdered(entry -> {
                 md.update(entry.getKey().getBytes());
                 md.update(fingerprint(entry.getValue(), excludedPropertyKeys).getBytes());
             });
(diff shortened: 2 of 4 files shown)


docs/hashing.adoc
```diff
@@ -0,0 +1,19 @@
...
+The hashsum of a graph first calculates the hashsums for each node. This hashsum list is ordered and for each node the
+hashsum for all relationships and their endnode are added. This approach provides independence of internal ids.
+
+Optionally you can supply a list of `propertyKeys` that should be ignored on all nodes. This is, e.g. useful if you store
+`created=timestamp()` properties that should be ignored.
+
+[opts=header,cols="m,a"]
+|===
+| function name | description
+| `apoc.hashing.fingerprintNodeOrRel(node_or_rel, <list_of_props_to_ignore>)` | calculates a md5 hashsum over the given node or relationship.
```
jexp: just use `fingerprint`

would be nice if it also worked with maps

src/main/java/apoc/hashing/Fingerprinting.java
```diff
@@ -0,0 +1,198 @@
...
+    @UserFunction
+    @Description("calculate a checksum (md5) over a node or a relationship. This deals gracefully with array properties. Two identical entities do share the same hash.")
+    public String fingerprintNodeOrRel(@Name("node_relation") Object thing, @Name(value = "propertyExcludes", defaultValue = "") List<String> excludedPropertyKeys) {
+
+        if (thing instanceof Node) {
+            return fingerprintNode((Node) thing, excludedPropertyKeys);
+        } else if (thing instanceof Relationship) {
+                return fingerprintRelationship((Relationship) thing, excludedPropertyKeys);
+        } else {
+            throw new UnsupportedOperationException("cannot fingerprint a " + thing.getClass());
```
jexp: perhaps just return null?

src/main/java/apoc/hashing/Fingerprinting.java
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+    }
+
+    @UserFunction
```
jexp: perhaps this is better as a procedure so it could return more columns on number of nodes, rels, properties observed ?

src/main/java/apoc/hashing/Fingerprinting.java
```diff
@@ -0,0 +1,198 @@
...
+    @Description("calculate a checksum (md5) over a the full graph. Be aware that this function does use in-memomry datastructures depending on the size of your graph.")
+    public String fingerprintGraph(@Name(value = "propertyExcludes", defaultValue = "") List<String> excludedPropertyKeys) {
+
+        return withMessageDigest(messageDigest -> {
+            // step 1: load all nodes, calc their hash and map them to id
+            Map<Long, String> idToNodeHash = db.getAllNodes().stream().collect(Collectors.toMap(
```
jexp: - could we perhaps user the eclipse collections that neo4j users now internally?
- do we need the String hash or would the byte[] be better/more efficient?
- can we perhaps build this up in a streaming manner to save space?

jexp: i.e. recompute hash vs. store it?
if we sort by internal id we still rely on the internal id?
shouldn't we instead sort by hash or such?

src/main/java/apoc/hashing/Fingerprinting.java
```diff
@@ -0,0 +1,198 @@
...
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private String convertValueToString(Object value) {
+        if (value.getClass().isArray()) {
+            return nativeArrayToString(value);
+        } else {
+            return value.toString();
```
jexp: what about nested values / iterables / maps?
And would it be more efficient to actually add the value's bytes to the MD?

src/test/java/apoc/hashing/FingerprintingTest.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+
+public class FingerprintingTest {
+
+    @Rule
+    public Neo4jRule neo4j = new Neo4jRule()
```
jexp: harness is really slow for tests (firing up a server)
please stick to the embedded db tests

src/test/java/apoc/hashing/FingerprintingTest.java
```diff
@@ -0,0 +1,75 @@
...
+            .withFunction(Fingerprinting.class);
+
+    @Test
+    public void fingerprintNodeWithArrayProperty() {
+        GraphDatabaseService db = neo4j.getGraphDatabaseService();
+        db.execute("CREATE (:Person{name:'ABC',emails:['aa@bb.de', 'cc@dd.ee'], integers:[1,2,3], floats:[0.9,1.1]})");
+
+        String value = Iterators.single(db.execute("MATCH (n) RETURN apoc.hashing.fingerprintNodeOrRel(n) AS hash").columnAs("hash"));
```
jexp: you can just create + fingerprint without doing the extra match, makes it more 1:1

src/main/java/apoc/hashing/Fingerprinting.java
```diff
@@ -0,0 +1,198 @@
...
+                return fingerprintRelationship((Relationship) thing, excludedPropertyKeys);
+        } else {
+            throw new UnsupportedOperationException("cannot fingerprint a " + thing.getClass());
+        }
+
+    }
+
+    @UserFunction
+    @Description("calculate a checksum (md5) over a the full graph. Be aware that this function does use in-memomry datastructures depending on the size of your graph.")
+    public String fingerprintGraph(@Name(value = "propertyExcludes", defaultValue = "") List<String> excludedPropertyKeys) {
```
jexp: please use a consistent naming like `excludeProperties` 
`propertyExcludes <> excludedPropertyKeys`

src/main/java/apoc/hashing/Fingerprinting.java
```diff
@@ -0,0 +1,198 @@
...
+                    Map.Entry::getKey,
+                    (o, o2) -> { throw new RuntimeException(); },
+                    () -> new TreeMap<>()
+            ));
+
+            // step 3: iterate nodes in order of their hash (we cannot rely on internal ids)
+            nodeHashToId.entrySet().stream().forEach(entry ->{
+                messageDigest.update(entry.getKey().getBytes());
+
+                Node node = db.getNodeById(entry.getValue());
```
jexp: we can just store the node

src/main/java/apoc/hashing/Fingerprinting.java
jexp: can we just create a hash->node map? and use that? so we don't have to look them up again?

src/main/java/apoc/hashing/Fingerprinting.java
```diff
@@ -0,0 +1,198 @@
...
+            // step 1: load all nodes, calc their hash and map them to id
+            Map<Long, String> idToNodeHash = db.getAllNodes().stream().collect(Collectors.toMap(
+                    Node::getId,
+                    node -> fingerprintNode(node, excludedPropertyKeys),
+                    (aLong, aLong2) -> { throw new RuntimeException(); },
+                    () -> new TreeMap<>()
+            ));
+
+            // step 2: build inverse map
+            Map<String, Long> nodeHashToId = idToNodeHash.entrySet().stream().collect(Collectors.toMap(
```
jexp: what happens if 2 nodes have the same hash? 
will we lose one from the comparison?
should we fail?
should we merge them?

src/main/java/apoc/hashing/Fingerprinting.java
```diff
@@ -0,0 +1,198 @@
...
+            nodeHashToId.entrySet().stream().forEach(entry ->{
+                messageDigest.update(entry.getKey().getBytes());
+
+                Node node = db.getNodeById(entry.getValue());
+                List<EndNodeRelationshipHashTuple> endNodeRelationshipHashTuples = StreamSupport.stream(node.getRelationships(Direction.OUTGOING).spliterator(), false)
+                        .map(relationship -> {
+                            String endNodeHash = idToNodeHash.get(relationship.getEndNodeId());
+                            String relationshipHash = fingerprintRelationship(relationship, excludedPropertyKeys);
+                            return new EndNodeRelationshipHashTuple(endNodeHash, relationshipHash);
+                        }).collect(Collectors.toList());
```
jexp: do we have to materialize the list here in between?
we can just continue with the `.sorted().forEach(..)`

src/main/java/apoc/hashing/Fingerprinting.java
jexp: would be good to parallelize in a future version by batch of nodes

jexp: how do we make sure that the hashes stay consistent? or should we document that we only guarantee the hashes to be the same with the exactly same apoc version??

src/main/java/apoc/hashing/Fingerprinting.java
sarmbruster: I changed the code so it should always have a way to create a hashsum.

src/main/java/apoc/hashing/Fingerprinting.java
sarmbruster: hmm. I'm not sure if these additional fields do make sense. At the end I just want a hashsum to compare a graph with another one.

src/main/java/apoc/hashing/Fingerprinting.java
sarmbruster: nesting should work now. I've refactored the code to work recursively.

src/main/java/apoc/hashing/Fingerprinting.java
sarmbruster: For ease of maintenance I've added a `NOTE` to the docs making clear that hashsums might differ in different apoc version.

jexp:
Do you have time to look at the other bits or shall i merge and we look at those later?

PR #1092, opened 2019-02-11
fixes #1091 - Procedure: apoc.model.jdbc

### LINKED ISSUE #1091: Procedure: apoc.model.jdbc
Create a procedure `apoc.model.jdbc` that extract metadata information via jdbc as Graph.
The procedure must:

- provide a config param to persist the resulting graph on Neo4j
- provide config params in order to filter columns/view/tables

conker84:
Fixes #1091
Create the procedure `apoc.model.jdbc`

  - created the procedure `apoc.model.jdbc`
  - updated documentation

@jexp do you think could be useful add a compose file with the same example of the documentation?


build.gradle
```diff
@@ -135,6 +135,11 @@ dependencies {
     compileOnly group: 'org.ow2.asm', name: 'asm', version: '5.0.2'
     compile group: 'com.github.javafaker', name: 'javafaker', version: '0.10'
 
+    // schemacrawler
+    compile group: 'us.fatehi', name: 'schemacrawler', version: '15.04.01'
```
jexp: I guess we can add it as a regular dependency depending on the size.

jexp:
Looks good to me, thanks so much!

PR #1095, opened 2019-02-12
Adds apoc.metrics.storage(dirName) to allow inspection of free storage

moxious:
Adds apoc.metrics.storage(dirName) to allow inspection of free storage

  - A common operational problem with Neo4j is the disk filling up. When this happens, the database can't commit transactions and make them durable, which results in run-time "freak outs".  :)
  - Existing JMX metrics do not provide a free disk space metric to tell if this is going to happen.
  - Halin would like to be able to know if this is going to happen to warn the user as part of a diagnostic.

I've opted to implement this purely on a setting basis.  Users will have no external knowledge of where the mount points on disk are.  They will however be able to tell that the spot where they've put their TX logs is 99% full, and that would be a potential problem!

Reference: https://github.com/moxious/halin/issues/30


moxious:
As with my last PR, unfortunately I can't provide tests for this because the ephemeral in memory graph database used for unit testing has no valid directory configurations or filesystem presence (I tried).  As part of containerized testing, I can provide later tests.

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -92,6 +95,57 @@ public Neo4jMetric(String name, long lastUpdated) {
...
+     * Note that we divulge the storage directory setting, not the internal path.
+     * It isn't a secret that neo4j has a dbms.directories.logs, but user doesn't need
+     * to know where that is on a physical hard disk.
+     */
+    public static class StorageMetric {
+        public final String setting;
+        public final long freeSpaceBytes;
+        public final long totalSpaceBytes;
+        public final long usableSpaceBytes;
+        public final double pctFree;
```
jexp: perhaps spell it out?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -92,6 +95,57 @@ public Neo4jMetric(String name, long lastUpdated) {
...
+
+        public StorageMetric(String setting, long freeSpaceBytes, long totalSpaceBytes, long usableSpaceBytes) {
+            this.setting = setting;
+            this.freeSpaceBytes = freeSpaceBytes;
+            this.totalSpaceBytes = totalSpaceBytes;
+            this.usableSpaceBytes = usableSpaceBytes;
+            this.pctFree = (double)freeSpaceBytes / (double)totalSpaceBytes;
```
jexp: avoid division by zero?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -92,6 +95,57 @@ public Neo4jMetric(String name, long lastUpdated) {
...
+        }
+    }
+
+    /**
+     * DAO that gets streamed to the user with apoc.metric.storage()
+     * Note that we divulge the storage directory setting, not the internal path.
+     * It isn't a secret that neo4j has a dbms.directories.logs, but user doesn't need
+     * to know where that is on a physical hard disk.
+     */
+    public static class StorageMetric {
```
jexp: Cool that we don't expose paths.

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -125,6 +179,50 @@ public Neo4jMetric(String name, long lastUpdated) {
...
+            "as a directorySetting, you will get back all available directory settings.  For a list of available directory settings, " +
+            "see the Neo4j operations manual reference on configuration settings.   Directory settings are **not** paths, they are " +
+            "a neo4j.conf setting key name")
+    public Stream<StorageMetric> storage(@Name("directorySetting") String directorySetting) {
+        // Permit case-insensitive checks.
+        String input = directorySetting == null ? null : directorySetting.toLowerCase();
+
+        boolean validSetting = (
+            input == null ||
+            Arrays.stream(neo4jDirectoryConfigurationSettingNames).anyMatch(input::equals)
```
jexp: contains?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -125,6 +179,50 @@ public Neo4jMetric(String name, long lastUpdated) {
...
+                        return directorySetting.equals(dirSetting);
+                    }
+
+                    return true;
+                })
+                .map(StoragePair::fromDirectorySetting)
+                .filter(sp -> {
+                    if (sp == null) { return false; }
+
+                    if (!sp.dir.exists() || !sp.dir.isDirectory() || !sp.dir.canRead()) {
```
jexp: can we invert the condition so it's more readable?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -125,6 +179,50 @@ public Neo4jMetric(String name, long lastUpdated) {
...
+        if (!validSetting) {
+            String validOptions = String.join(", ", neo4jDirectoryConfigurationSettingNames);
+            throw new RuntimeException("Invalid directory setting specified.  Valid options are one of: " +
+                    validOptions);
+        }
+
+        return Stream.of(neo4jDirectoryConfigurationSettingNames)
+                .filter(dirSetting -> {
+                    // If user specified a particular one, immediately cut list to just that one.
+                    if (directorySetting != null) {
```
jexp: `directorySetting == null || directorySetting.equals(dirSetting)`

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -204,4 +302,24 @@ private final double doubleFrom(LoadCsv.CSVResult r, String column) {
...
     }
+
+
+    // This is the list of dbms.directories.* valid configuration items for neo4j.
+    // https://neo4j.com/docs/operations-manual/current/reference/configuration-settings/
+    // Usually these reside under the same root but because they're separately configurable, in the worst case
+    // every one is on a different device.
+    //
+    // More likely, they'll be largely similar metrics.
+    public static final String [] neo4jDirectoryConfigurationSettingNames = new String [] {
```
jexp: constants should be NEO4J_DIRECTORY_CONFIGURATION_SETTING_NAMES
use Arrays.asList(...) then you can use contains / stream directly on it.

jexp:
Testing would be interesting with testcontainers :)
And docs entry.

src/main/java/apoc/metrics/Metrics.java
moxious: Don't think so...no.   If the input contains the valid setting, this means you could ask for storage metrics for "dbms.directories.logs@)#&*)&!)#$&!)#&$!#" --- not sure that makes sense, isn't it better to give an error in this case.

jexp: contains as in list.contains(setting) not string contains :)

PR #1102, opened 2019-02-16
fixes #1101: Test Export CSV Enterprise version

### LINKED ISSUE #1101: Test Export CSV Enterprise version
Provide tests in v 3.5 for Export CSV Enterprise features (NODE KEY constraints)

conker84:
Fixes #1101 

Added the `ExportCypherEnterpriseFeaturesTest` for specific enterprise features tests. On Travis we get FileNotFoundException with "Permission Denied" so we test "stream" feature using the result map in orde to assert the statement

  - Added the `ExportCypherEnterpriseFeaturesTest` for specific enterprise features tests

### DIFF
--- a/src/test/java/apoc/export/cypher/ExportCypherEnterpriseFeaturesTest.java
+++ b/src/test/java/apoc/export/cypher/ExportCypherEnterpriseFeaturesTest.java
@@ -1,0 +1,111 @@
+package apoc.export.cypher;
+
+import apoc.util.TestUtil;
+import apoc.util.Util;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.neo4j.driver.v1.AuthTokens;
+import org.neo4j.driver.v1.Driver;
+import org.neo4j.driver.v1.GraphDatabase;
+import org.neo4j.driver.v1.Session;
+import org.testcontainers.containers.Neo4jContainer;
+
+import java.io.File;
+import java.util.Map;
+
+import static apoc.export.cypher.ExportCypherTest.ExportCypherResults.*;
+import static apoc.util.MapUtil.map;
+import static apoc.util.TestContainerUtil.*;
+import static apoc.util.TestUtil.isTravis;
+import static apoc.util.TestUtil.readFileToString;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assume.assumeNotNull;
+
+/**
+ * @author as
+ * @since 13.02.19
+ */
+public class ExportCypherEnterpriseFeaturesTest {
+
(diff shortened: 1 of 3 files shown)


src/test/java/apoc/export/cypher/ExportCypherEnterpriseFeaturesTest.java
```diff
@@ -0,0 +1,136 @@
...
+                (r) -> assertExportStatement(EXPECTED_NEO4J_SHELL_WITH_COMPOUND_CONSTRAINT, output, r));
+    }
+
+    private void assertExportStatement(String expectedStatement, File output, Map<String, Object> result) {
+        assertEquals(expectedStatement, isTravis() ? result.get("cypherStatements") : readFile(output));
+    }
+
+    private static String readFile(File output) {
+        try {
+            return new Scanner(output).useDelimiter("\\Z").next() + String.format("%n");
```
jexp: we should move those into utilities or use it from there.
also in Java8 we should probably rather use something from java.nio.Files https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#lines-java.nio.file.Path-

src/test/java/apoc/export/cypher/ExportCypherEnterpriseFeaturesTest.java
```diff
@@ -0,0 +1,136 @@
...
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private String getFilePath(File output) { // TODO on Travis we get FileNotFoundException with "Permission Denied" so we test the Cypher from the map (investigate why)
```
jexp: should this go into testutils too?

src/test/java/apoc/export/cypher/ExportCypherEnterpriseFeaturesTest.java
```diff
@@ -0,0 +1,136 @@
...
+        return isTravis() ? "null" : PREFIX + output.getPath().replace("target/", "");
+    }
+
+    private File getFile(String fileName) {
```
jexp: what is this achieving?
the method name is not really clear

it actually writes a  new, empty file to the location?

What about file.exits(), canWrite(), isDirectory() etc?

conker84: They are tries that I made when I realized that on Travis I had permissions issue, I forgot to delete them sorry.

conker84:
I cleaned/moved the code a requested, the only exceptions are `getFilePath` and `getImportDirectoryPath` which are test specific.

PR #1118, opened 2019-03-06
Revise Metrics and Log Streaming Approach

moxious:
- Changes metrics to return a timestamp and map of values instead of breaking those values out.  This accomodates a range of different schemas in different metric files across Neo4j versions.  In particular, there were a lot of metric schema changes between 3.4 and 3.5 which makes supporting the individual file format of particular metrics not a terribly feasible option

### DIFF
--- a/src/main/java/apoc/log/Neo4jLogStream.java
+++ b/src/main/java/apoc/log/Neo4jLogStream.java
@@ -4,6 +4,7 @@ import apoc.util.FileUtils;
 import org.neo4j.procedure.*;
 import org.neo4j.logging.Log;
 
+import java.nio.file.NoSuchFileException;
 import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
@@ -13,6 +14,10 @@ import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.stream.Stream;
 
+/**
+ * @author moxious
+ * @since 27.02.19
+ */
 public class Neo4jLogStream {
     @Context
     public Log log;
@@ -73,7 +78,11 @@ public class Neo4jLogStream {
             }
 
             return entries;
-        } catch(IOException exc) {
+        } catch(NoSuchFileException nsf) {
+            // This special case we want to throw a custom message and not let this error propagate, because the
+            // trace exposes the full path we were checking.
+            throw new RuntimeException("No log file exists by that name");
+        } catch (IOException exc) {
             throw new RuntimeException(exc);
         }
     }
--- a/src/main/java/apoc/log/Neo4jLogStream.java
+++ b/src/main/java/apoc/log/Neo4jLogStream.java
@@ -1,6 +1,5 @@
 package apoc.log;
 
-import apoc.ApocConfiguration;
 import apoc.util.FileUtils;
 import org.neo4j.procedure.*;
 import org.neo4j.logging.Log;
@@ -51,9 +50,8 @@ public class Neo4jLogStream {
         File f = new File(logDir, logName);
 
         try {
-            // This is just here as an added safety check; this proc is limited to the logs directory but it would
-            // still be possible to get cute and create symlinks from the logs directory elsewhere.
-            if (!FileUtils.canonicalPathInNeo4jHome(f)) {
+            String canonicalPath = f.getCanonicalPath();
+            if (!canonicalPath.contains(logDir.getAbsolutePath())) {
                 throw new RuntimeException("The path you are trying to access has a canonical path outside of the logs " +
                         "directory, and this procedure is only permitted to access files in the log directory.  This may " +
                         "occur if the path in question is a symlink or other link.");
--- a/src/main/java/apoc/log/Neo4jLogStream.java
+++ b/src/main/java/apoc/log/Neo4jLogStream.java
@@ -56,7 +56,7 @@ public class Neo4jLogStream {
 
         try {
             String canonicalPath = f.getCanonicalPath();
-            if (!canonicalPath.contains(logDir.getAbsolutePath())) {
+            if (!canonicalPath.startsWith(logDir.getAbsolutePath())) {
                 throw new RuntimeException("The path you are trying to access has a canonical path outside of the logs " +
                         "directory, and this procedure is only permitted to access files in the log directory.  This may " +
                         "occur if the path in question is a symlink or other link.");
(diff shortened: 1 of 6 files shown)


src/main/java/apoc/meta/Meta.java
```diff
@@ -101,6 +101,9 @@ public static Types of(Class<?> type) {
         }
 
         public static Types from(String typeName) {
+            if (typeName == null) {
```
moxious: In current APOC if user specifies a type mapping of null, they get an NPE.

The column doesn't even have to exist in the input. 

Reproduction: 
```
CALL apoc.load.csv("https://raw.githubusercontent.com/moxious/cypher-surface/master/surface-3.4.0.csv", { mapping: { whateverColumn: { type: null }}});
```

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -147,9 +74,35 @@ public static StorageMetric fromStoragePair(StoragePair storagePair) {
...
+    /**
+     * DAO for a single line in a metrics/*.csv file.
+     * Value is abstracted to double; in reality some are int, some are float, some are double.
+     */
+    public static class GenericMetric {
+        public final long timestamp;
+        public final String metric;
+        public final Map<String,Object> map;
+
+        public GenericMetric(String metric, long t, Map<String,Object> map) {
```
moxious: This DAO is about all that is generic to all metrics.

The "value" field is not generic.  It's in some but not in others.

Keeping in mind -- timestamp is UNIX time, not necessarily in UTC, but in the timezone the database is configured to use (which is usually UTC).  I'm leaving it as its published by the database in order to not deal with issues like fetching the configured TZ prior to getting the metric, which would allow me to convert to a timestamp, but would only then create the secondary issue of what TZ to represent the output in (is database timezone or UTC better?  Hmm.....)

src/main/java/apoc/log/Neo4jLogStream.java
```diff
@@ -51,9 +55,8 @@ public int compareTo(FileEntry o) {
         File f = new File(logDir, logName);
 
         try {
-            // This is just here as an added safety check; this proc is limited to the logs directory but it would
-            // still be possible to get cute and create symlinks from the logs directory elsewhere.
-            if (!FileUtils.canonicalPathInNeo4jHome(f)) {
+            String canonicalPath = f.getCanonicalPath();
+            if (!canonicalPath.contains(logDir.getAbsolutePath())) {
```
jexp: perhaps better "startsWith"?

src/main/java/apoc/metrics/Metrics.java
jexp: probably a choice for the user on the UI?

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -176,8 +129,44 @@ public static StorageMetric fromStoragePair(StoragePair storagePair) {
...
+                // case to avoid checking entire map every time.  If the value of the "t" field
+                // is null, it's also a repeated header or bad row.  We're specifying type mappings
+                // from the CSV string t -> long.  So if the actual value is "t" this will turn into
+                // a null long.
+                Object t = map.get("t");
+                if (map.containsKey("t") && ("t".equals(t.toString()) || t == null)) {
+                    return false;
+                } else {
+                    for(Object value : map.values()) {
+                        if (value != null && value instanceof Number) {
```
jexp: instanceof already does the null check.

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -176,8 +129,44 @@ public static StorageMetric fromStoragePair(StoragePair storagePair) {
...
+                Map<String,Object> map = o.map;
+
+                // Most commonly CSV has a timestamp "t" field, if its value = "t"
+                // Then it's a repeated header.  This is just a shortcut for the most common
+                // case to avoid checking entire map every time.  If the value of the "t" field
+                // is null, it's also a repeated header or bad row.  We're specifying type mappings
+                // from the CSV string t -> long.  So if the actual value is "t" this will turn into
+                // a null long.
+                Object t = map.get("t");
+                if (map.containsKey("t") && ("t".equals(t.toString()) || t == null)) {
```
jexp: I think an `if ("t".equals(map.get("t")))` should work fine.

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -176,8 +129,44 @@ public static StorageMetric fromStoragePair(StoragePair storagePair) {
...
+                // Final case: no number data.  Likely all null values, as headers failed type mapping to
+                // numbers.
+                return false;
+            }
+        };
+
         String url = new File(metricsDir, metricName + ".csv").toURI().toString();
-        return new LoadCsv().csv(url, config);
+        return new LoadCsv().csv(url, config)
+                .filter(csvResult -> duplicatedHeaderRows.test(csvResult))
```
jexp: I guess you can just do this::duplicatedHeaderRows

moxious:
Don't merge just yet.

Tested on 3.5 within desktop, and 3.5 cluster running on GCP derived from debian packages (different directory structures).   Needs test on 3.4, which necessitates pulling the commits to another branch.

Here's a useful testing query: 

List metrics and pull data from all of them to ensure nothing's going wrong
```
call apoc.metrics.list() 
yield name 
with name 
call apoc.metrics.get(name) 
yield timestamp, metric, map 
return * 
order by timestamp desc limit 1000;
```

(That query takes a lot of memory and a lot of in-memory filtering -- so it's a bit expensive and may take a bit of time)

Pull a log:
```
call apoc.log.stream("debug.log");
```

Try but fail to pull a log:
```
call apoc.log.stream('nonexistant');
```

Yields: `Neo.ClientError.Procedure.ProcedureCallFailed: Failed to invoke procedure apoc.log.stream: Caused by: java.lang.RuntimeException: No log file exists by that name`

STorage metrics:

```
call apoc.metrics.storage(null);
```

Just one metric:

```
call apoc.metrics.storage("dbms.directories.data")
```

(Returns only one row)

jexp:
PR looks good.

moxious:
3.4 testing:

```
git cherry-pick 5b960d3a43966116f24a867d524b3eac9168f51f 6a099dfb1cd2b3ed2ca663a00f6560cf6bb866d0 c52a1453ead8d37e85decaa2cee95dc636b30dec 996d65d2c6f1f39e3a1b2ded47724bea827844ec
```

Recompile and deploy.  Tested all same queries & approaches and LGTM!

PR #1124, opened 2019-03-20
Remove broken tests from build

szarnyasg:
Removed two broken tests that made the CI fail (non-deterministically?). The idea is to raise an issue for both and then merge this PR.


szarnyasg:
Fixed #1126.

PR #1128, opened 2019-03-21
fix #1055 adding includeFailedBatches for apoc.periodic.iterate

### LINKED ISSUE #1055: Apoc.periodic.iterate - get list of failed batches
## Background
In my case around 1% of the batches of apoc.periodic.iterate with {parallel:true, concurrency:16} fail most likely because 2 threads are trying to create relationship to/from the same node at the same time. Retries do not seem to help. When running with parallel:false I get 0 failed batches, but it takes too long.

apoc.periodic.iterate to rerurn a collection of order numbers of failed batches.

Ideally on top of that an additional supporting procedure with inputs (cypher:STRING, batchSize:INT, returnBatchesAt:LIST) and returns the results of the "cypher" query at the given batches. The user then could rerun the failed batches with periodic.iterate with different settings to fix the failed batches (e.g. with parallel:false), But also would be ok to leave users the possibility to write own custom functions to post-process)

## Considered alternatives
Fix retries to correctly deal with the failed batches where the required nodes were released, add possibility to switch to parallel:false after certain retry.

Failed batches could be rerun or at least nodes/relationships in the failed batches could be flaged to a user.

sarmbruster:
adding a new optional parameter to apoc.periodic.iterate.

If `includeFailedBatches:true` is set, the data of all failed batches is included in the result being passed back to the caller. 

Since it defaults to false I don't expect any compatibility issues.

### DIFF
--- a/src/main/java/apoc/periodic/Periodic.java
+++ b/src/main/java/apoc/periodic/Periodic.java
@@ -349,7 +349,7 @@ public class Periodic {
                     } catch (Exception e) {
                         failedOps.addAndGet(batchsize);
                         if (failedParams >= 0) {
-                            failedParamsMap.put(Long.toString(finalBatches), batch.subList(0, Math.min(failedParams+1, batch.size())));
+                            failedParamsMap.put(Long.toString(finalBatches), new ArrayList<Map<String,Object>>>(batch.subList(0, Math.min(failedParams+1, batch.size()))));
                         }
                         recordError(operationErrors, e);
                     }
@@ -369,7 +369,7 @@ public class Periodic {
                             } catch (Exception e) {
                                 failedOps.incrementAndGet();
                                 if (failedParams >= 0) {
-                                    failedParamsMap.put(Long.toString(finalBatches), batch.subList(0, Math.min(failedParams+1, batch.size())));
+                                    failedParamsMap.put(Long.toString(finalBatches), new ArrayList<Map<String,Object>>>(batch.subList(0, Math.min(failedParams+1, batch.size()))));
                                 }
                                 recordError(operationErrors, e);
                             }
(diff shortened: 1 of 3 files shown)


jexp:
Should we also record which batch failed? So instead of having all failed ones in a huge list we could have a CHM with batch-id -> list of params?

jexp:
Perhaps we can turn includeFailedBatches into a number which when > 0 limits the number of entries per batch. So that folks get an idea of what failed without returning millions of records? And when = 0 it only returns the failed batch ids and when -1 then all.

sarmbruster:
like that idea. I'll update the PR.

src/main/java/apoc/periodic/Periodic.java
```diff
@@ -346,11 +348,15 @@ public Pattern regNoCaseMultiLine(String pattern) {
                         retried.addAndGet(retry(consumer,params,0,retries));
                     } catch (Exception e) {
                         failedOps.addAndGet(batchsize);
+                        if (failedParams >= 0) {
+                            failedParamsMap.put(Long.toString(finalBatches), batch.subList(0, Math.min(failedParams+1, batch.size())));
```
jexp: Remember subList keeps a pointer to the original list. So it might be better to use new ArrayList(sublist)

src/main/java/apoc/periodic/Periodic.java
```diff
@@ -362,6 +368,9 @@ public Pattern regNoCaseMultiLine(String pattern) {
                                 retried.addAndGet(retry(consumer,params,0,retries));
                             } catch (Exception e) {
                                 failedOps.incrementAndGet();
+                                if (failedParams >= 0) {
+                                    failedParamsMap.put(Long.toString(finalBatches), batch.subList(0, Math.min(failedParams+1, batch.size())));
```
jexp: same here.

jexp:
Thanks a lot Stefan, really good functionality.

PR #1138, opened 2019-03-28
Move deprecation warning to correct headline

### LINKED ISSUE #1137: Deprecation warning in documentation has wrong scope
According to the documentation, the whole "Text and Lookup Indices" are deprecated: https://neo4j-contrib.github.io/neo4j-apoc-procedures/#indexes

But the deprecation text references fulltext indices only.

This is misleading and should be clarified by moving the deprecation warning to the appropriate headline.

LarsKumbier:
Fixes #1137 

Moves deprecation warning to correct headline to clarify, which namespace specifically will be deprecated and removed in the future.

properly scope and move the deprecation warning.

### DIFF
--- a/docs/asciidoc/_indexes.adoc
+++ b/docs/asciidoc/_indexes.adoc
@@ -1,9 +1,5 @@
 [[indexes]]
-== Deprecated: Text and Lookup Indexes
-
-[WARNING]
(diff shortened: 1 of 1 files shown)


jexp:
Thanks so much!!

PR #1143, opened 2019-03-28
fixes #1140: Prevent unsafe redirect

### LINKED ISSUE #1140: Prevent unsafe redirect
95% of HTTPS servers vulnerable to trivial MITM attacks | Netcraft
Only 1 in 20 HTTPS servers correctly implements HTTP Strict Transport Security, a widely-supported security feature that prevents visitors making unencrypted

In case of redirect with protocol change we must return an exception with the url provided by the `Location` header

It follows any redirection so we are sensitive to this kind of attacks:
https://news.netcraft.com/archives/2016/03/17/95-of-https-servers-vulnerable-to-trivial-mitm-attacks.html

`CALL apoc.load.csv("<URL_THAT_REDIRECTS_TO_FS>")`

### Versions

Any

conker84:
Fixes #1140 

In case of the HTTP response code is 3XX and the `Location` header provides a different protocol we throw an Exception with the value of the `Location` in the message

  - changed the `Util#isRedirect` method in order to implement that logic
  - provided a simple test for `CALL apoc.load.csv({url})` in order to highlight the new behaviour
  - added the `UtilIT` class that tests the method `Util#openInputStream` for two call, one that has as redirect location `http://www.google.com` and the other one has `file:/etc/passwd`

### DIFF
--- a/src/main/java/apoc/util/Util.java
+++ b/src/main/java/apoc/util/Util.java
@@ -264,8 +264,18 @@ public class Util {
         return con;
     }
 
-    public static boolean isRedirect(int code) {
-        return code >= 300 && code < 400;
+    public static boolean isRedirect(HttpURLConnection con) throws IOException {
+        int code = con.getResponseCode();
+        boolean isRedirectCode = code >= 300 && code < 400;
+        if (isRedirectCode) {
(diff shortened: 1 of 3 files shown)


jexp:
Thank you Andrea!

PR #1144, opened 2019-03-29
Aquire write lock for both Nodes and Relationships.

michael-simons:


jexp:
Thanks so much for spotting this.

PR #1155, opened 2019-04-08
fixes #1141: add config option to "garble" log messages

### LINKED ISSUE #1141: add config option to "garble" log messages
apoc.user.log=none/safe/<nothing>

add an option to disable apoc logging altogether `none`

for `safe` do:

if set then replace all `.` and whitespace (space and tab) with underscore
and lowercase all characters

so that log messages from apoc cannot trigger operations errors

global variable for log messages per second (over a window of 1 minute)
ignore logging if it exceeds 1 every 10 seconds

conker84:
Fixes #1141

add config option to "garble" log messages

* Added three properties:
  * `apoc.user.log.type` | enum[none, safe, raw] (default value `safe`) | type of logging:
    * `node`: disable the procedures
    * `safe`: replace all `.` and whitespace (space and tab) with underscore and lowercase all characters
    * `raw`: left the messages as-is
  * `apoc.user.log.window.ops` | int (default value `10`) | num of logs permitted in a time-window, exceeded this quota every log message will be skip
  * `apoc.user.log.window.time` | long (default value `10000`) | the length (in milliseconds) of the time-window (default 10 seconds)
* Added documentation
* Added tests

### DIFF
--- a/src/main/java/apoc/util/SimpleRateLimiter.java
+++ b/src/main/java/apoc/util/SimpleRateLimiter.java
@@ -1,0 +1,27 @@
+package apoc.util;
+
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class SimpleRateLimiter {
+
+    private final AtomicInteger countDownLatch = new AtomicInteger(0);
+    private final AtomicLong lastUpdate = new AtomicLong(0);
+
+    private final long timeWindow;
+    private final int operationPerWindow;
+
+    public SimpleRateLimiter(long timeWindow, int operationPerWindow) {
+        this.timeWindow = timeWindow;
+        this.operationPerWindow = operationPerWindow;
+    }
+
+    public boolean canExecute() {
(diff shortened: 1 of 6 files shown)


src/main/java/apoc/util/SimpleRateLimiter.java
```diff
@@ -0,0 +1,27 @@
...
+
+    private final long timeWindow;
+    private final int operationPerWindow;
+
+    public SimpleRateLimiter(long timeWindow, int operationPerWindow) {
+        this.timeWindow = timeWindow;
+        this.operationPerWindow = operationPerWindow;
+    }
+
+    public boolean canExecute() {
```
jexp: If I'm not mistaken these are fixed time windows, so that means at the boundary of two windows it can write the full amount on either side?

I was thinking about something like this:
rate = 1/s

on each update we increment a counter
and we also decrement the counter by (now-lastUpdate)*rate
and allow if the counter > 0

which is more of a sliding window.

jexp:
Merging for now, we will add a different rate-limiter later

Issue #1153, opened 2019-04-04
support Excel (xlsx) as format for `apoc.export....`


sarmbruster:
use different sheets for each label and relationship type.

PR #1162, opened 2019-04-11
#1153 export to excel

sarmbruster:
This PR addes couple of procedure for exporting to an Excel file using the POI library.

Note: docs are still missing. Working on this.

### DIFF
--- a/src/main/java/apoc/export/xls/ExportXls.java
+++ b/src/main/java/apoc/export/xls/ExportXls.java
@@ -76,14 +76,15 @@ public class ExportXls {
 
     private Stream<ProgressInfo> exportXls(@Name("file") String fileName, String source, Object data, Map<String,Object> configMap) throws Exception {
         checkWriteAllowed();
-        try (Transaction tx = db.beginTx()) {
+        try (Transaction tx = db.beginTx();
+             OutputStream out = getOutputStream(fileName, null);
+             SXSSFWorkbook wb = new SXSSFWorkbook(100)) {
+
             XlsExportConfig config = new XlsExportConfig(configMap);
             ProgressInfo progressInfo = new ProgressInfo(fileName, source, "xls");
             progressInfo.batchSize = config.getBatchSize();
             ProgressReporter reporter = new ProgressReporter(null, null, progressInfo);
 
-            SXSSFWorkbook wb = new SXSSFWorkbook(100);
-            OutputStream out = getOutputStream(fileName, null);
             Map<Class, CellStyle> styles = buildCellStyles(config, wb);
 
             if (data instanceof SubGraph) {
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/export/xls/ExportXls.java
```diff
@@ -0,0 +1,275 @@
...
+    private Stream<ProgressInfo> exportXls(@Name("file") String fileName, String source, Object data, Map<String,Object> configMap) throws Exception {
+        checkWriteAllowed();
+        try (Transaction tx = db.beginTx()) {
+            XlsExportConfig config = new XlsExportConfig(configMap);
+            ProgressInfo progressInfo = new ProgressInfo(fileName, source, "xls");
+            progressInfo.batchSize = config.getBatchSize();
+            ProgressReporter reporter = new ProgressReporter(null, null, progressInfo);
+
+            SXSSFWorkbook wb = new SXSSFWorkbook(100);
+            OutputStream out = getOutputStream(fileName, null);
```
jexp: resource handling?

src/main/java/apoc/util/FileUtils.java
```diff
@@ -124,20 +124,23 @@ public static boolean isFile(String fileName) {
     }
 
     public static PrintWriter getPrintWriter(String fileName, Writer out) throws IOException {
```
jexp: what's the reason behind these changes?

src/main/java/apoc/util/FileUtils.java
```diff
@@ -124,20 +124,23 @@ public static boolean isFile(String fileName) {
...
+        OutputStream outputStream;
         if (isHdfs(fileName)) {
             try {
-                writer = new OutputStreamWriter(HDFSUtils.writeFile(fileName));
+                outputStream = HDFSUtils.writeFile(fileName);
             } catch (Exception e) {
                 throw new RuntimeException(e);
             }
         } else {
-            writer = fileName.equals("-") ? out : new BufferedWriter(new FileWriter(fileName));
```
jexp: we shouldn't remove the buffering

src/main/java/apoc/util/FileUtils.java
sarmbruster: I couldn't get excel export workin correctly when acquiring a writer and converting it to a outputstream. With an output stream it works, so I changed this.

src/main/java/apoc/export/xls/ExportXls.java
sarmbruster: good point, will amend

jexp:
Looks good thanks a lot.

To be done in a subsequent PR:

- documentation
- handling of complex(nested) types via export.json converter

PR #1169, opened 2019-04-15
fixes #1072: Make sure custom procedures are also recreated on other cluster members

### LINKED ISSUE #1072: Make sure custom procedures are also recreated on other cluster members
at runtime.

conker84:
Fixes #1072

Make sure custom procedures are also recreated on other cluster members

  - added "lastUpdate" to the procedure/function structure
  - added timer that checks every `x` (default 60000) milliseconds if the graph property got updated and register the updated procedure/function version
  - added integration test in Causal Cluster environment (3 core + 1 read replica)
  - updated the documentation

### DIFF
--- a/src/main/java/apoc/custom/CypherProcedures.java
+++ b/src/main/java/apoc/custom/CypherProcedures.java
@@ -1,5 +1,6 @@
 package apoc.custom;
 
+import apoc.ApocConfiguration;
 import apoc.util.JsonUtil;
 import apoc.util.Util;
 import org.neo4j.collection.PrefetchingRawIterator;
@@ -28,6 +29,7 @@ import org.neo4j.values.AnyValue;
 import org.neo4j.values.virtual.MapValue;
 
 import java.util.*;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -327,10 +329,12 @@ public class CypherProcedures {
         public List<List<String>>inputs;
         public Object outputs;
         public Boolean forceSingle;
+        public long lastUpdate;
 
         public CustomProcedureInfo(String type, String name, String description, String mode,
                                    String statement, List<List<String>> inputs, Object outputs,
-                                   Boolean forceSingle){
+                                   Boolean forceSingle,
+                                   long lastUpdate){
             this.type = type;
             this.name = name;
             this.description = description;
@@ -339,6 +343,7 @@ public class CypherProcedures {
             this.inputs = inputs;
             this.forceSingle = forceSingle;
             this.mode = mode;
+            this.lastUpdate = lastUpdate;
         }
     }
 
@@ -347,6 +352,7 @@ public class CypherProcedures {
         private GraphProperties properties;
         private final GraphDatabaseAPI api;
         private final Log log;
+        private Timer timer;
 
         public CustomProcedureStorage(GraphDatabaseAPI api, Log log) {
             this.api = api;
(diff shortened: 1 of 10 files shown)


jexp:
I think we misunderstood each other.

- we don't need a timer
- just a graphproperty with a long value that's update with the update of the main graph-property
- in memory storage of our own last update AtomicLong
- every "refresh-seconds" after our last update when there is an execution, check the time-graph-property
- if it's newer than our in-memory one update the functions

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -356,38 +381,48 @@ public CustomProcedureStorage(GraphDatabaseAPI api, Log log) {
...
-            stored.get(FUNCTIONS).forEach((name, data) -> {
-                registry.registerFunction(name, (String) data.get("statement"), (String) data.get("output"),
-                        (List<List<String>>) data.get("inputs"), (Boolean)data.get("forceSingle"), (String) data.get("description"));
-            });
+            restoreFunctions(registry, stored);
+            restoreProcedures(registry, stored);
+        }
+
+        public static void restoreProcedures(CustomStatementRegistry registry, Map<String, Map<String, Map<String, Object>>> stored) {
             stored.get(PROCEDURES).forEach((name, data) -> {
```
jexp: shouldn't this set the last update in memory?

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -356,38 +381,48 @@ public CustomProcedureStorage(GraphDatabaseAPI api, Log log) {
...
                 registry.registerProcedure(name, (String) data.get("statement"), (String) data.get("mode"),
                         (List<List<String>>) data.get("outputs"), (List<List<String>>) data.get("inputs"), (String) data.get("description"));
             });
         }
 
+        public static void restoreFunctions(CustomStatementRegistry registry, Map<String, Map<String, Map<String, Object>>> stored) {
+            stored.get(FUNCTIONS).forEach((name, data) -> {
```
jexp: see above

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -53,6 +55,9 @@
     @Context
     public Log log;
 
+    private static final AtomicLong PROCEDURES_LAST_UPDATE = new AtomicLong(0);
```
jexp: also add a timer for last check

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -126,8 +133,16 @@ public boolean registerProcedure(@Name("name") String name, @Name("statement") S
                         Mode.valueOf(mode.toUpperCase()), admin, null, new String[0], description, null, false, true
                 );
                 procedures.register(new CallableProcedure.BasicProcedure(signature) {
+                    private CustomStatementRegistry registry = self;
                     @Override
                     public RawIterator<Object[], ProcedureException> apply(org.neo4j.kernel.api.proc.Context ctx, Object[] input, ResourceTracker resourceTracker) throws ProcedureException {
+                        GraphPropertiesProxy properties = getProperties(api);
+                        long proceduresLastUpdate = getProceduresLastUpdate(properties);
+                        if (proceduresLastUpdate > PROCEDURES_LAST_UPDATE.get()) {
```
jexp: can we move the check into a separate function.
they should always be called not just in the if case.

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -403,13 +438,33 @@ public void unavailable() {
                 Map<String, Object> previous = (value == null) ? procData.remove(name) : procData.put(name, value);
                 if (value != null || previous != null) {
                     properties.setProperty(APOC_CUSTOM, Util.toJson(data));
+                    if (PROCEDURES.equals(type)) {
+                        properties.setProperty(APOC_CUSTOM + PROCEDURES_LAST_UPDATE_KEY, System.currentTimeMillis());
```
jexp: create a constant for the key

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -403,13 +438,33 @@ public void unavailable() {
...
+
+                    }
                 }
                 tx.success();
                 return previous;
             }
         }
 
-        private static Map<String, Map<String,Map<String, Object>>> readData(GraphProperties properties) {
+        public static long getProceduresLastUpdate(GraphProperties properties) {
```
jexp: just pass teh constant to the method, no need to have two

src/main/java/apoc/util/Util.java
```diff
@@ -611,6 +611,7 @@ public static boolean isWriteableInstance(GraphDatabaseAPI db) {
                 /* ignore */
             }
             String role = db.execute("CALL dbms.cluster.role()").<String>columnAs("role").next();
+            System.out.println("role = " + role);
```
jexp: remove?

src/test/java/apoc/util/TestContainerUtil.java
```diff
@@ -23,28 +24,27 @@ private TestContainerUtil() {}
...
+    public static TestcontainersCausalCluster createEnterpriseCluster(int numOfCoreInstances, int numberOfReadReplica, Map<String, Object> neo4jConfig) {
+        return TestcontainersCausalCluster.create(numOfCoreInstances, numberOfReadReplica, Duration.ofMinutes(4), neo4jConfig);
+    }
 
     public static Neo4jContainerExtension createEnterpriseDB(boolean withLogging) {
-        // We build the project, the artifact will be placed into ./build/libs
-        executeGradleTasks("clean", "shadow");
         // We define the container with external volumes
-        Neo4jContainerExtension neo4jContainer = new Neo4jContainerExtension()
+        Neo4jContainerExtension neo4jContainer = new Neo4jContainerExtension("neo4j:3.5.3-enterprise")
```
jexp: should this come from project info ?

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -111,11 +125,27 @@ public void asFunction(@Name("name") String name, @Name("statement") String stat
...
+        }
+
+        private boolean needsUpdate(long procedureLastUpdate, AtomicLong lastCheck, AtomicLong lastUpdate) {
+            long now = System.currentTimeMillis();
+            long lastCk = lastCheck.getAndSet(now);
+            if (now - lastCk < UPDATE_FREQ) {
+                return false;
+            }
+            if (procedureLastUpdate > lastUpdate.get()) {
+                PROCEDURES_LAST_UPDATE.set(procedureLastUpdate);
```
jexp: needs to be: `lastUpdate.set()`

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -111,11 +125,27 @@ public void asFunction(@Name("name") String name, @Name("statement") String stat
...
             this.api = api;
             procedures = api.getDependencyResolver().resolveDependency(Procedures.class);
             this.log = log;
+            self = this;
+        }
+
+        private boolean needsUpdate(long procedureLastUpdate, AtomicLong lastCheck, AtomicLong lastUpdate) {
+            long now = System.currentTimeMillis();
+            long lastCk = lastCheck.getAndSet(now);
+            if (now - lastCk < UPDATE_FREQ) {
```
jexp: I think I might have gotten this wrong.
As it leads to starvation, if you always check within 60s then you will never update.
I think it needs to set `lastCheck.set(now)` when it doesn't return false. because then it really checks.

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -111,11 +125,27 @@ public void asFunction(@Name("name") String name, @Name("statement") String stat
...
+                return false;
+            }
+            if (procedureLastUpdate > lastUpdate.get()) {
```
jexp: In general, I try to return early then you can even combine the two checks above.

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -403,13 +459,26 @@ public void unavailable() {
...
                 }
                 tx.success();
                 return previous;
             }
         }
 
-        private static Map<String, Map<String,Map<String, Object>>> readData(GraphProperties properties) {
+        public static long getLastUpdate(GraphProperties properties, String key) {
+            try (Transaction tx = properties.getGraphDatabase().beginTx()) {
+                tx.success();
```
jexp: should put this after the read.

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -364,30 +370,76 @@ public static GraphPropertiesProxy getProperties(GraphDatabaseAPI api) {
...
-            stored.get(PROCEDURES).forEach((name, data) -> {
-                registry.registerProcedure(name, (String) data.get("statement"), (String) data.get("mode"),
-                        (List<List<String>>) data.get("outputs"), (List<List<String>>) data.get("inputs"), (String) data.get("description"));
-            });
+            timer.scheduleAtFixedRate(new TimerTask() {
+                private AtomicLong highWaterMark = new AtomicLong(0);
+                private static final String LAST_UPDATE_KEY = "lastUpdate";
+                @Override
+                public void run() {
+                    Map<String, Map<String, Map<String, Object>>> storage = readData(properties);
```
jexp: can we extract this into a number of methods to make the code more readable

jexp: also can we keep the simple approach from before where we have one method for functions and one for procedures.

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -364,30 +370,76 @@ public static GraphPropertiesProxy getProperties(GraphDatabaseAPI api) {
...
+                    int updated = storage.entrySet().stream()
+                            .filter(entry -> CypherProcedures.PROCEDURES.equals(entry.getKey()) || CypherProcedures.FUNCTIONS.equals(entry.getKey()))
+                            .flatMap(storageEntry -> {
+                                String type = storageEntry.getKey();
+                                Map<String, Map<String, Object>> map = storageEntry.getValue();
+                                boolean isProcedure = CypherProcedures.PROCEDURES.equals(type);
+                                return map.entrySet().stream()
+                                        // `!e.getValue().containsKey("lastUpdate")` is for backwards compatibility for already present procedures/functions
+                                        .filter(e -> !e.getValue().containsKey(LAST_UPDATE_KEY) || (long) e.getValue().get(LAST_UPDATE_KEY) > highWaterMark.get())
```
jexp: this whole code block looks too complicated for just handling an potentially absent update key.

PR #1179, opened 2019-04-19
fixes #1178: Make the `dbms.directories.import` dir root dir for exports

### LINKED ISSUE #1178: Make the `dbms.directories.import` dir root dir for exports
In case the property `apoc.import.file.use_neo4j_config` is set to `true` use the `dbms.directories.import` dir as root for `apoc.export.*` procedures in order to unify the behaviour with the `apoc.load.*` procedures.

conker84:
Fixes #1178 

Make the `dbms.directories.import` dir root dir for exports 

  - changed the behaviour of `FileUtils#getOutputStream` method in order to manage the property `apoc.import.file.use_neo4j_config`, in case it's `true` it takes the value `dbms.directories.import` as root dir
  - fixed all the test
  - updated the documentation

### DIFF
--- a/src/main/java/apoc/util/FileUtils.java
+++ b/src/main/java/apoc/util/FileUtils.java
@@ -8,6 +8,7 @@ import apoc.util.s3.S3URLConnection;
 import org.apache.commons.io.output.WriterOutputStream;
 
 import java.io.*;
+import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.util.Arrays;
@@ -86,7 +87,7 @@ public class FileUtils {
     }
 
     public static String changeFileUrlIfImportDirectoryConstrained(String url) throws IOException {
-        if (isFile(url) && ApocConfiguration.isEnabled("import.file.use_neo4j_config")) {
+        if (isFile(url) && isImportUsingNeo4jConfig()) {
             if (!ApocConfiguration.isEnabled("import.file.allow_read_from_filesystem"))
                 throw new RuntimeException("Import file "+url+" not enabled, please set dbms.security.allow_csv_import_from_file_urls=true in your neo4j.conf");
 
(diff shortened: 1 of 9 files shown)


src/main/java/apoc/util/FileUtils.java
```diff
@@ -138,7 +138,20 @@ public static OutputStream getOutputStream(String fileName, OutputStream out) th
                 throw new RuntimeException(e);
             }
         } else {
-            outputStream = fileName.equals("-") ? out : new FileOutputStream(fileName);
+            if (fileName.equals("-")) {
+                outputStream = out;
+            } else {
+                boolean enabled = ApocConfiguration.isEnabled("import.file.use_neo4j_config");
```
jexp: can we encapsulate this?

jexp:
If possible try not to have unrelated changes in a single PR.
So if you see something that you want to clean up, please take note and clean it up in a separate PR.

jexp:
It's ok for now just for the future.

src/main/java/apoc/util/FileUtils.java
```diff
@@ -138,7 +138,20 @@ public static OutputStream getOutputStream(String fileName, OutputStream out) th
...
+                if (enabled) {
+                    String importDir = ApocConfiguration.get("dbms.directories.import", "import");
```
jexp: do we have constants for these? or move them into a method in apocconfig or fileutils

conker84:
> It's ok for now just for the future.

Ok

src/main/java/apoc/util/FileUtils.java
conker84: ok

src/main/java/apoc/util/FileUtils.java
conker84: ok, we put them into two methods into `FileUtils`

PR #1180, opened 2019-04-19
#1177 keep the mongodb connection open until stream is closed

### LINKED ISSUE #1177: MongoDB Connection State is not 'open'
## How  the Problem was Produced

The source mongodb has 500K matching trip records.
The following returns one of these records without issue.
```
CALL apoc.mongodb.first('mongodb://localhost:27017', 'mobilytics', 'mds_trips', {provider_name: 'Lime'}, false) YIELD value as trip;
```
The following request produces the error.

```
CALL apoc.mongodb.find('mongodb://localhost:27017', 'mobilytics', 'mds_trips', {provider_name: 'Lime'}, null, null, false, 0, 0) YIELD value as trip ;
```

Failed to invoke procedure `apoc.mongodb.find`: Caused by: java.lang.IllegalStateException: state should be: open

### Versions

- OS: Ubuntu 18.04
- Neo4j: 3.5.3
- Neo4j-Apoc: apoc-3.5.0.3-all.jar 

- bson-3.5.0.jar  
-  mongo-java-driver-3.5.0.jar
-  mongodb-driver-3.5.0.jar
-  mongodb-driver-core-3.5.0.jar

- jackson-annotations-2.9.8.jar
-  jackson-core-2.9.8.jar
-  jackson-databind-2.9.8.jar

phreed:
In apoc/mongodb/MongoDB.java ...
```
 try (Coll coll = getMongoColl(hostOrKey, db, collection, compatibleValues)) {
            return coll.all(query, skip, limit).map(MapResult::new); // .onClose(coll::safeClose);
        } catch (Exception e) {
            log.error("...");
        }
```
When leaving the try block the `coll` goes out of scope and its `close()` is called.
apoc.mongodb.MongoDBColl
```
   @Override
    public void close() throws IOException { 
        mongoClient.close();
    }
```
However, the returned stream (let's call it `res_stream`) needs for 
the mongo connection associated with `coll`  to remain open.

The closing of `res_stream` is handled by the .onClose() stream modifier
apoc.mongodb.MongoDBColl
```
 private Stream<Map<String, Object>> asStream(FindIterable<Document> result) {
        logger.info("making a result stream");
        Iterable<Document> it = () -> result.iterator();
        return StreamSupport
                .stream(it.spliterator(), false)
                .map(doc -> this.documentToPackableMap(doc))
                .onClose( () -> {
                        result.iterator().close();
                    } );
    }
```
The closing of the `mongoClient.close()` needs to be contingent 
upon the closing of `res_stream` and not the `coll` going out of scope.

Of course in those situations where a stream is not returned the `mongoClient.close()` 
may need to be called when `coll` goes out of scope.
I suppose a `door_stop` flag could be used to hold the `mongoClient` open
in those cases where things like stream are returned.
Seems kind of a hack (I generally feel that way about flags).

### DIFF
--- a/src/main/java/apoc/mongodb/MongoDB.java
+++ b/src/main/java/apoc/mongodb/MongoDB.java
@@ -55,7 +55,8 @@ public class MongoDB {
                                  @Name("db") String db,
                                  @Name("collection") String collection,
                                  @Name("query") Map<String, Object> query,
-                                 @Name(value = "compatibleValues", defaultValue = "false") boolean compatibleValues, @Name(value = "skip", defaultValue = "0") Long skip,
+                                 @Name(value = "compatibleValues", defaultValue = "false") boolean compatibleValues, 
+                                 @Name(value = "skip", defaultValue = "0") Long skip,
                                  @Name(value = "limit", defaultValue = "0") Long limit) {
         try (Coll coll = getMongoColl(hostOrKey, db, collection, compatibleValues)) {
             return coll.all(query, skip, limit).map(MapResult::new); // .onClose(coll::safeClose);
--- a/src/main/java/apoc/mongodb/MongoDBColl.java
+++ b/src/main/java/apoc/mongodb/MongoDBColl.java
@@ -31,6 +31,7 @@ class MongoDBColl implements MongoDB.Coll {
     private MongoCollection<Document> collection;
     private MongoClient mongoClient;
     private boolean compatibleValues = false;
+    private boolean doorStop = false;
 
     public MongoDBColl(String url, String db, String coll) {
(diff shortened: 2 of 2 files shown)


conker84:
Hi @phreed thanks for fixing this!
Can you also add a test case that recreates the problem?

conker84:
Hi @phreed any updates on this?

phreed:
@conker84 I did not see your earlier request for a test case.
The problem with testing this is that small tests seem to generally avoid the race condition.
Apparently I am not the first to not see how to test this issue as I see code fragments as comments which are working around it.

phreed:
Incidentally my load of 5M docs from mongodb into neo4j ran to completion so the fix seems to work.

PR #1183, opened 2019-04-26
Added apoc.merge.node.eager() and apoc.merge.relationship.eager()

### LINKED ISSUE #1182: apoc.merge.node() does not update node properties when matching to existing node
On an empty database using APOC 3.5.0.3 execute the following:

CREATE (p:Person{ssid:'123', name:'Jim'}) 
CALL apoc.merge.node(['Person'],{ssid:'123'}, {name:'John'}) YIELD node RETURN node 

## Expected Behavior

Since the identProps argument is `ssid:'123'`, this should match to the just-created Jim node. The name property should be updated to 'John', as we would expect if running this Cypher equivalent:

```
MERGE (node:Person {ssid:'123'})
SET node.name = 'John'
```

The MERGE indeed matched to the existing Jim node (no new node was created), but the name property was not updated.  

## Other Notes

Looking at the PR for this, it looks like this behavior was intentional (see [testMergeNodeWithPreExisting()](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/508/files#diff-0357b9c1505ecd881eaaa564ca7e2107R47), though there doesn't seem to be any documentation about this behavior, nor does it seem intuitive (this would behave like an ON CREATE SET).

### Versions

  - OS: Mac OS 10.12.6
  - Neo4j: 3.5.4 Enterprise
  - Neo4j-Apoc: 3.5.0.3

InverseFalcon:
Fixes #1182 
New procedures added, old procedures unchanged (some refactoring, but no changes to behavior)

These both accept a config map that can take 'onMatchProps' and 'onCreateProps' map properties.

These procs can also match to multiple existing nodes unlike apoc.merge.node() and apoc.merge.relationship().

Documentation to be checked in separately.

### DIFF
--- a/src/main/java/apoc/merge/Merge.java
+++ b/src/main/java/apoc/merge/Merge.java
@@ -1,7 +1,7 @@
 package apoc.merge;
 
 import apoc.result.*;
-import apoc.util.Util;
+import com.google.common.collect.Lists;
 import org.neo4j.graphdb.*;
 import org.neo4j.helpers.collection.Iterators;
 import org.neo4j.procedure.*;
@@ -25,7 +25,8 @@ public class Merge {
         String labels = labelNames.stream().map(s -> wrapInBacktics(s)).collect(Collectors.joining(":"));
 
         Map<String, Object> params = new HashMap<>();
-        params.put("props", Util.merge(identProps, props));
+        params.put("identProps", identProps);
+        params.put("props", props);
         String identPropsString = buildIdentPropsString(identProps);
 
         final String cypher = "MERGE (n:" + labels + "{" + identPropsString + "}) ON CREATE SET n += $props RETURN n";
@@ -33,18 +34,40 @@ public class Merge {
         return Stream.of(new NodeResult(node));
     }
 
-    private String wrapInBacktics(String s) {
-        return "`" + s + "`";
+    @Procedure(value="apoc.merge.node.eager", mode = Mode.WRITE, eager = true)
+    @Description("apoc.merge.node.eager(['Label'], identProps:{key:value, ...}, config:{onCreateProps:{key:value,...}, onMatchProps:{key:value,...}}) - merge node with dynamic labels, with support for setting properties ON CREATE or ON MATCH")
+    public Stream<NodeResult> nodeEager(@Name("label") List<String> labelNames, @Name("identProps") Map<String, Object> identProps, @Name(value="config", defaultValue = "{}") Map<String, Object> config) {
+        if ((identProps==null) || (identProps.isEmpty())) {
+            throw new IllegalArgumentException("you need to supply at least one identifying property for a merge");
+        }
+
+        validateKeys(config.keySet(), Lists.newArrayList("onCreateProps", "onMatchProps"));
+
+        String labels = labelNames.stream().map(s -> wrapInBacktics(s)).collect(Collectors.joining(":"));
+
+        Map<String, Object> params = new HashMap<>();
+
+        Map<String, Object> onCreateProps = (Map<String, Object>) config.getOrDefault("onCreateProps", Collections.emptyMap());
+        Map<String, Object> onMatchProps = (Map<String, Object>) config.getOrDefault("onMatchProps", Collections.emptyMap());
+
+        params.put("identProps", identProps);
+        params.put("onCreateProps", onCreateProps);
+        params.put("onMatchProps", onMatchProps);
+        String identPropsString = buildIdentPropsString(identProps);
+
+        final String cypher = "MERGE (n:" + labels + "{" + identPropsString + "}) ON CREATE SET n += $onCreateProps ON MATCH SET n += $onMatchProps RETURN n";
+        return db.execute(cypher, params ).columnAs("n").stream().map(node -> new NodeResult((Node) node));
     }
 
     @Procedure(mode = Mode.WRITE)
-    @Description("apoc.merge.relationship(startNode, relType,  {key:value, ...}, {key:value, ...}, endNode) - merge relationship with dynamic type")
+    @Description("apoc.merge.relationship(startNode, relType,  identProps:{key:value, ...}, {key:value, ...}, endNode) - merge relationship with dynamic type")
     public Stream<RelationshipResult> relationship(@Name("startNode") Node startNode, @Name("relationshipType") String relType,
                                                    @Name("identProps") Map<String, Object> identProps, @Name("props") Map<String, Object> props, @Name("endNode") Node endNode) {
         String identPropsString = buildIdentPropsString(identProps);
 
         Map<String, Object> params = new HashMap<>();
-        params.put("props", Util.merge(identProps, props));
+        params.put("identProps", identProps);
+        params.put("props", props != null ? props : Collections.emptyMap());
         params.put("startNode", startNode);
         params.put("endNode", endNode);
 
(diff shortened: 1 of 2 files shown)


jexp:
Are we good with this?

src/main/java/apoc/merge/Merge.java
```diff
@@ -25,26 +25,49 @@
...
+    @Procedure(value="apoc.merge.node.eager", mode = Mode.WRITE, eager = true)
+    @Description("apoc.merge.node.eager(['Label'], identProps:{key:value, ...}, config:{onCreateProps:{key:value,...}, onMatchProps:{key:value,...}}) - merge node with dynamic labels, with support for setting properties ON CREATE or ON MATCH")
+    public Stream<NodeResult> nodeEager(@Name("label") List<String> labelNames, @Name("identProps") Map<String, Object> identProps, @Name(value="config", defaultValue = "{}") Map<String, Object> config) {
+        if ((identProps==null) || (identProps.isEmpty())) {
+            throw new IllegalArgumentException("you need to supply at least one identifying property for a merge");
+        }
+
+        validateKeys(config.keySet(), Lists.newArrayList("onCreateProps", "onMatchProps"));
+
+        String labels = labelNames.stream().map(s -> wrapInBacktics(s)).collect(Collectors.joining(":"));
```
jexp: there is Util.quote which does a better job of this.

src/main/java/apoc/merge/Merge.java
```diff
@@ -25,26 +25,49 @@
...
     }
 
     @Procedure(mode = Mode.WRITE)
-    @Description("apoc.merge.relationship(startNode, relType,  {key:value, ...}, {key:value, ...}, endNode) - merge relationship with dynamic type")
+    @Description("apoc.merge.relationship(startNode, relType,  identProps:{key:value, ...}, {key:value, ...}, endNode) - merge relationship with dynamic type")
     public Stream<RelationshipResult> relationship(@Name("startNode") Node startNode, @Name("relationshipType") String relType,
                                                    @Name("identProps") Map<String, Object> identProps, @Name("props") Map<String, Object> props, @Name("endNode") Node endNode) {
         String identPropsString = buildIdentPropsString(identProps);
 
         Map<String, Object> params = new HashMap<>();
```
jexp: easier with Util.map(args)

src/main/java/apoc/merge/Merge.java
```diff
@@ -53,18 +76,47 @@ private String wrapInBacktics(String s) {
         return Stream.of(new RelationshipResult(rel));
     }
 
+    @Procedure(value = "apoc.merge.relationship.eager", mode = Mode.WRITE, eager = true)
+    @Description("apoc.merge.relationship.eager(startNode, relType,  identProps:{key:value, ...}, config:{onCreateProps:{key:value, ...}, onMatchProps:{key:value, ...}}, endNode) - merge relationship with dynamic type, with support for setting properties ON CREATE or ON MATCH")
+    public Stream<RelationshipResult> relationshipEager(@Name("startNode") Node startNode, @Name("relationshipType") String relType,
+                                                   @Name("identProps") Map<String, Object> identProps, @Name("config") Map<String, Object> config, @Name("endNode") Node endNode) {
+        validateKeys(config.keySet(), Lists.newArrayList("onCreateProps", "onMatchProps"));
```
jexp: shouldn't this just delegate to the non-eager version?

src/main/java/apoc/merge/Merge.java
```diff
@@ -25,26 +25,49 @@
...
         Node node = Iterators.single(db.execute(cypher, params ).columnAs("n"));
         return Stream.of(new NodeResult(node));
     }
 
-    private String wrapInBacktics(String s) {
-        return "`" + s + "`";
+    @Procedure(value="apoc.merge.node.eager", mode = Mode.WRITE, eager = true)
+    @Description("apoc.merge.node.eager(['Label'], identProps:{key:value, ...}, config:{onCreateProps:{key:value,...}, onMatchProps:{key:value,...}}) - merge node with dynamic labels, with support for setting properties ON CREATE or ON MATCH")
+    public Stream<NodeResult> nodeEager(@Name("label") List<String> labelNames, @Name("identProps") Map<String, Object> identProps, @Name(value="config", defaultValue = "{}") Map<String, Object> config) {
+        if ((identProps==null) || (identProps.isEmpty())) {
```
jexp: shouldn't this just delegate to the non-eager version?

src/main/java/apoc/merge/Merge.java
InverseFalcon: The non-eager version can only merge a single result (see `Iterators.single()` usage) and doesn't handle ON MATCH SET.

src/main/java/apoc/merge/Merge.java
InverseFalcon: The non-eager version can only merge a single result (see `Iterators.single()` usage) and doesn't handle ON MATCH SET.

src/main/java/apoc/merge/Merge.java
InverseFalcon: In place of the wrapInBacktics() call? Okay, will replace.

src/main/java/apoc/merge/Merge.java
InverseFalcon: For building out the `params` map? Got it.

src/main/java/apoc/merge/Merge.java
jexp: but then we should add both there

src/main/java/apoc/merge/Merge.java
```diff
@@ -22,49 +23,88 @@
...
         params.put("endNode", endNode);
 
-        final String cypher = "WITH $startNode as startNode, $endNode as endNode MERGE (startNode)-[r:"+ wrapInBacktics(relType) +"{"+identPropsString+"}]->(endNode) ON CREATE SET r+= $props RETURN r";
+        final String cypher = "WITH $startNode as startNode, $endNode as endNode MERGE (startNode)-[r:"+ Util.quote(relType) +"{"+identPropsString+"}]->(endNode) ON CREATE SET r+= $props RETURN r";
         Relationship rel = Iterators.single(db.execute(cypher, params ).columnAs("r"));
         return Stream.of(new RelationshipResult(rel));
     }
 
+    @Procedure(value = "apoc.merge.relationship.eager", mode = Mode.WRITE, eager = true)
+    @Description("apoc.merge.relationship.eager(startNode, relType,  identProps:{key:value, ...}, config:{onCreateProps:{key:value, ...}, onMatchProps:{key:value, ...}}, endNode) - merge relationship with dynamic type, with support for setting properties ON CREATE or ON MATCH")
```
jexp: I wouldn't put the extra props into config but into optional parameters which default to empty maps.

src/main/java/apoc/merge/Merge.java
```diff
@@ -22,49 +23,88 @@
...
         final String cypher = "MERGE (n:" + labels + "{" + identPropsString + "}) ON CREATE SET n += $props RETURN n";
         Node node = Iterators.single(db.execute(cypher, params ).columnAs("n"));
         return Stream.of(new NodeResult(node));
     }
 
-    private String wrapInBacktics(String s) {
-        return "`" + s + "`";
+    @Procedure(value="apoc.merge.node.eager", mode = Mode.WRITE, eager = true)
+    @Description("apoc.merge.node.eager(['Label'], identProps:{key:value, ...}, config:{onCreateProps:{key:value,...}, onMatchProps:{key:value,...}}) - merge node with dynamic labels, with support for setting properties ON CREATE or ON MATCH")
+    public Stream<NodeResult> nodeEager(@Name("label") List<String> labelNames, @Name("identProps") Map<String, Object> identProps, @Name(value="config", defaultValue = "{}") Map<String, Object> config) {
```
jexp: I wouldn't put the extra props into config but into optional parameters which default to empty maps.

src/main/java/apoc/merge/Merge.java
```diff
@@ -22,49 +23,88 @@
             throw new IllegalArgumentException("you need to supply at least one identifying property for a merge");
         }
 
-        String labels = labelNames.stream().map(s -> wrapInBacktics(s)).collect(Collectors.joining(":"));
+        String labels = labelNames.stream().map(s -> Util.quote(s)).collect(Collectors.joining(":"));
```
jexp: there is a method in Util for this.

jexp:
Merged manually with some changes.
Removed the config maps.
Simplified the call to redirect.

PR #1192, opened 2019-05-06
fixes #1188: headers incorrect order after apoc csv export in enterprise edition 3.5.4, 3.5.5

### LINKED ISSUE #1188: headers incorrect order after apoc csv export in enterprise edition 3.5.4, 3.5.5
- Neo4j version: >= 3.5.4 - only happens in the enterprise edition, not the community edition! (also happens on 3.5.5)
- Operating system: Tested on MacOSX, Ubuntu
- API/Driver: (for example Cypher/Java API/Python driver vX.Y.Z) - apoc-3.5.0.3-all.jar
- steps to reproduce - 

1) We have a specific database here when we do an apoc query similar to this: 
`CALL apoc.export.csv.query("MATCH (n:Document {pk:'5921569'}) return n.pk as pk, n.copyright as copyright", "/tmp/out.csv", {})` - the headers in the export are reversed, 
e.g. in the file out.csv we see:
Note: Using the apoc.export.csv.query is required to reproduce the bug.

```
"copyright","pk"
"5921569","
(c) 2018 Hovsepian, Albanese, et al. ""ASCB(r),"" ""The American Society for Cell Biology(r),"" and ""Molecular Biology of the Cell(r)"" are registered trademarks of The American Society for Cell Biology.
2018

This article is distributed by The American Society for Cell Biology under license from the author(s). Two months after publication it is available to the public under an Attribution-Noncommercial-Share Alike 3.0 Unported Creative Commons License.

"
```

2) Afterwards, when we query the Neo4j via Browser, we can see the same output without APOC call e.g. just MATCH (n:Document {pk:'5921569'}) return n.pk as pk, n.copyright as copyright in the browser

```
╒═══════════╤══════════════════════════════════════════════════════════════════════╕
│"copyright"│"pk"                                                                  │
╞═══════════╪══════════════════════════════════════════════════════════════════════╡
│"5921569"  │"
(c) 2018 Hovsepian, Albanese, et al. "ASCB(r)," "The American Societ│
│           │y for Cell Biology(r)," and "Molecular Biology of the Cell(r)" are reg│
│           │istered trademarks of The American Society for Cell Biology.
2018

Thi│
│           │s article is distributed by The American Society for Cell Biology unde│
│           │r license from the author(s). Two months after publication it is avail│
│           │able to the public under an Attribution-Noncommercial-Share Alike 3.0 │
│           │Unported Creative Commons License.

"                                 │
└───────────┴──────────────────────────────────────────────────────────────────────┘
```
Note: the data here contains a " and \n in the output,but using similar nodes with similar data, was unable to reproduce it.
3) if we restart the neo4j service, the query in browser returns the header in the correct order
- Expected behavior
Headers are displayed correctly in export and browser (even after using apoc export)
- Actual behavior
Headers are in the wrong order for certain pieces of data (after using the apoc export)

(copied from same issue at https://github.com/neo4j/neo4j/issues/12210, not sure if there's a better way to link it there)

conker84:
Fixes #1188 

Removed sorting for `apoc.export.csv.query` and added for `apoc.export.csv.all/data/graph`

  - Added sorting in exporting a subgraph. As described here:

https://github.com/conker84/neo4j-apoc-procedures/blob/fb316cce65daab4198a1d348fa9f39f08138f849/src/main/java/apoc/export/csv/CsvFormat.java#L245-L261

In particular, now we have this general structure for the headers:

`_id,_labels,<list_nodes_properties_naturally_sorted>,_start,_end,_type,<list_rel_properties_naturally_sorted>`

So for instance:

`_id,_labels,age,city,kids,male,name,street,_start,_end,_type,bar,foo`

  - Fixed tests related to `apoc.export.csv.query`
  - Updated the documentation

### DIFF
--- a/src/test/java/apoc/export/csv/ExportCsvTest.java
+++ b/src/test/java/apoc/export/csv/ExportCsvTest.java
@@ -19,6 +19,7 @@ import org.neo4j.test.TestGraphDatabaseFactory;
 import java.io.File;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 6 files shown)


src/test/java/apoc/export/csv/ExportCsvTest.java
```diff
@@ -203,7 +204,7 @@ public void testExportGraphCsvWithoutQuotes() throws Exception {
     @Test
     public void testExportQueryCsv() throws Exception {
         File output = new File(directory, "query.csv");
-        String query = "MATCH (u:User) return u.age, u.name, u.male, u.kids, labels(u)";
+        String query = "MATCH (u:User) return labels(u), u.age, u.kids, u.male, u.name";
```
jexp: the columns shouldn't be in sort order for this test otherwise we don't know it actually keeps the query return order.

src/test/java/apoc/export/csv/ExportCsvTest.java
```diff
@@ -357,7 +358,7 @@ private void assertResults(File output, Map<String, Object> r, final String sour
     }
 
     @Test public void testCypherCsvStreamingWithAlwaysQuotes() throws Exception {
-        String query = "MATCH (a:Address) return a.name, a.city, a.street, labels(a)";
```
jexp: should stay name, city

jexp:
Cool, thanks a lot!

Issue #562, opened 2017-08-23
Improve thread pool management to prevent blocking and waiting when using a combination of apoc components that use thread pools


bradnussbaum:

jexp:
Do you have some more details here?
Esp. thread-dumps or other logs that show the blocking.

We've seen some of this in the graph-algos library when a ton of tasks got scheduled, the thread-queue was full and the tasks had to be rescheduled after a short wait.

bradnussbaum:
@jexp Just talking to Alex, sounds like you have reached a consensus about what is going to be added. Can you provide an update of what will get in?

alex-price:
@jexp Can you clarify that parallel is indeed bugged and deprecated? Or am I misinterpreting this result being incorrect?

with timestamp() as start
call apoc.cypher.parallel('call apoc.util.sleep(1000) return timestamp() as end', { x: range(1, 100) }, 'x') yield value
return collect(value.end - start) as execution_time

[1016, 2023, 3024, 4028, 5030, 6032, 7038, 8044, 9050, 10051, 11053, 12054, 13057, 14062, 15067, 16072, 17073, 18078, 19082, 20085, 21086, 22088, 23091, 24095, 25100, 26104, 27106, 28107, 29108, 30114, 31115, 32119, 33121, 34126, 35129, 36135, 37138, 38144, 39145, 40148, 41154, 42160, 43164, 44170, 45171, 46176, 47179, 48184, 49186, 50191, 51197, 52200, 53204, 54210, 55212, 56215, 57217, 58222, 59223, 60227, 61229, 62233, 63240, 64244, 65250, 66253, 67258, 68261, 69266, 70270, 71275, 72277, 73282, 74286, 75292, 76294, 77298, 78298, 79303, 80309, 81311, 82315, 83320, 84321, 85323, 86329, 87333, 88335, 89338, 90344, 91349, 92351, 93355, 94356, 95361, 96367, 97367, 98369, 99373, 100374]

bradnussbaum:
Looks like parallel and parallel2 have slightly different behavior when it comes to projecting the partition objects. See below:

### Working parallel:
`with timestamp() as start, range(1, 100) AS range
unwind range as num
with collect({num:num, start:start}) AS partitions, start
with { partitions:partitions , start:start } as params, start
[...]

Issue #104, opened 2016-08-18
fix streaming in cypher parallel


jexp:

fbiville:
Assuming it was fixed with https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1193

Issue #27, opened 2016-05-07
parallel execution


ikwattro:
Would improve timings of some queries like : 

```
MATCH (user:User) WHERE id(user) = {id}
MATCH (user)-[:BOUGHT]->(product)<-[:BOUGHT]-(other)-[:BOUGHT]->(reco:Product)
USING SCAN reco:Product
USING JOIN ON other
RETURN reco, count(*) as score
ORDER BY score DESC LIMIT 10;
```

where you could parallel the process on the collaborators for eg

jexp:
Can you try this with:

```
MATCH (user:User) WHERE id(user) = {id}
MATCH (user)-[:BOUGHT]->(product)
WITH user,collect(product) as products
CALL apoc.cypher.parallelMap("MATCH (_)<-[:BOUGHT]-(other)-[:BOUGHT]->(reco:Product) USING SCAN reco:Product USING JOIN ON other RETURN reco,count(*) as freq",null,products) yield value
RETURN id(value.reco) as reco, sum(value.freq) as score
ORDER BY score DESC LIMIT 10;

```

jexp:
Ping @ikwattro ?

ikwattro:
Didn't have time to test it yet with apoc + this current dataset. Will ping during the week

jexp:
Did you test it? There is now also `apoc.periodic.iterate` and `apoc.periodic.commit`

Besides `apoc.cypher.*`

ikwattro:
@jexp Just tested with latest version of apoc and neo4j 3.0.3

```
MATCH (u:User) WHERE u.id = 463
MATCH (u)-[:LIKES]->(movie)
WITH u, collect(movie) as movies
CALL apoc.cypher.mapParallel("MATCH (_)<-[:LIKES]-(other)-[:LIKES]->(reco:Movie) USING SCAN reco:Movie USING JOIN ON other RETURN reco,count(*) as freq",{},movies) yield value
RETURN id(value.reco) as reco, sum(value.freq) as score
ORDER BY score DESC LIMIT 10;
```

CPU is burning and the query never returns

ikwattro:
Same query in plain cypher is taking 350ms : 

```
MATCH (u:User) WHERE u.id = 463
MATCH (u)-[:LIKES]->(movie)<-[:LIKES]-(other)
WITH distinct other
MATCH (other)-[:LIKES]->(movie)
RETURN movie as reco, count(*) as score
[...]

PR #1193, opened 2019-05-06
Avoid using recursion for re-submitting rejected tasks in the Pools.

### LINKED ISSUE #965: parallel2 stackoverflow error
Parallel2 works fine but gives stackoverflow error sometimes

image

chrisvest:
Fixes #965

The RejectedExecutionHandler used by the APOC Pools for all sorts of asynchronisity and parallelism, was attempting to block the submitting thread until the task was accepted by the executor.
This was done by letting the rejected execution handler submit the rejected task to the executor again. If the task was once again rejected, the handler would be called again, and so on.
The troubling part of this is that this happened recursively. This could produce some enourmous stack depths for many threads.
Some threads would get StackOverflowErrors when trying to submit tasks, which could break invariants and compromise the integrity of the system.

The solution is to make use of the blocking APIs on the internal task queue of the executor, to manually enqueue tasks in a blocking way.
This is done using the blocking APIs that take a timeout, such that we can still periodically check if the pool has been shut down.

Maybe the following issues are also related to this. I did not investigate those further: #27, #104, #562.

### DIFF
--- a/src/main/java/apoc/Pools.java
+++ b/src/main/java/apoc/Pools.java
@@ -1,15 +1,27 @@
 package apoc;
 
 import apoc.util.Util;
-import org.neo4j.graphdb.GraphDatabaseService;
-import org.neo4j.graphdb.Transaction;
-import org.neo4j.scheduler.JobScheduler;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jexp:
Thanks a lot Chris !

PR #1196, opened 2019-05-08
fixes #1194 - Adds ignoreQuotations configuration to apoc.load.csv  - Branch 3.5

### LINKED ISSUE #1194: Add ignoreQuotations configuration to apoc.load.csv
### Feature description
The apoc.load.csv method lacks the ability to completely ignore quotes. This is problematic when you want to load in a CSV row which opens a quote but does not close it, as doing this results in an error. This occurs even when the quote is escaped using the escape character. This feature would allow CSV rows with quotes to be loaded in, ignoring the quotes entirely. If one would still like the quotes to be present in the loaded row, but not act as a special quote character, they can then use the escape character.

A solution to this would make use of the CSVReaderBuilder's built in ignoreQuotations by adding a ignoreQuotations field to the LoadCsvConfig.

#### Alternative
An alternative solution I originally thought of would be using an empty string in the quoteChar field. However, doing this automatically assigns the default quote character which is the double quote, so it does not work.

alexiudice:
Fixes #1194

Adds ignoreQuotations to the apoc.load.csv configuration. This will allow the load csv operation to treat quote characters as regular characters, preventing issues if a quote is unclosed. When the row is loaded quote characters are NOT included. You can include quote characters by placing an escape character in front of them.

### DIFF
--- a/src/main/java/apoc/load/LoadCsv.java
+++ b/src/main/java/apoc/load/LoadCsv.java
@@ -47,6 +47,7 @@ public class LoadCsv {
                     .withCSVParser(new CSVParserBuilder()
                             .withQuoteChar(config.getQuoteChar())
                             .withSeparator(config.getSeparator())
+                            .withIgnoreQuotations( config.isIgnoreQuotations() )
                             .build())
                     .build();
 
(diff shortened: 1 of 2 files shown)


jexp:
Thanks a lot.

Please add a test.

PR #1201, opened 2019-05-11
fixes #1177: MongoDB Connection State is not 'open'

### LINKED ISSUE #1177: MongoDB Connection State is not 'open'

conker84:
Fixes #1177

Changed the way that the MongoDB procedures are executed.

  - created a new method `executeMongoQuery` which removes the `try-with-resource` and manage close of the `Coll` when the stream is closed with the `Stream#onClose` method

### DIFF
--- a/src/main/java/apoc/mongodb/MongoDB.java
+++ b/src/main/java/apoc/mongodb/MongoDB.java
@@ -14,6 +14,8 @@ import java.io.Closeable;
 import java.lang.reflect.InvocationTargetException;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.stream.Stream;
 
 /*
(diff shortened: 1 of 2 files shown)


jexp:
Looks good to me, thanks!

phreed:
Combine with #1180

PR #1202, opened 2019-05-13
fixes #1181: APOC 3.5.0.3 changes import file URL paths too aggressively

### LINKED ISSUE #1181: APOC 3.5.0.3 changes import file URL paths too aggressively

```
call apoc.metrics.get("neo4j.bolt.sessions_started");
```

Should return metric data in Neo4j Desktop installations.

```
Neo.ClientError.Procedure.ProcedureCallFailed: Failed to invoke procedure `apoc.metrics.get`: 

Caused by: java.io.FileNotFoundException: 
/Users/davidallen/Library/Application Support/Neo4j Desktop/Application/neo4jDatabases/database-6df59434-8fe6-405c-ab17-dd0a913b1498/installation-3.5.4/import/Users/davidallen/Library/Application Support/Neo4j Desktop/Application/neo4jDatabases/database-6df59434-8fe6-405c-ab17-dd0a913b1498/installation-3.5.4/metrics/neo4j.bolt.sessions_started.csv 
(No such file or directory)
```

(Run above query)

## Cause

In `apoc.metrics.Metrics.loadCsvForMetric`, great care is taken to build an absolute directory reference to the location of metrics, taking into account neo4j config.  This is done by `FileUtils.getMetricsDirectory`.    But when this absolute URL gets passed to the LoadCsv methods, it appears to use `FileUtils.changeFileUrlIfImportDirectoryConstrained` which seems to be appending the absolute directory as a relative to the import path.

The result is an ultra-long local URL which is wrong, causing the file not to exist, causing the metrics load to fail.

conker84:
Fixes #1181

Fixed the bug and created a Test case with Testcontainers

  - fixed the bug
  - created the testing class `MetricsTest`

### DIFF
--- a/src/main/java/apoc/load/LoadCsv.java
+++ b/src/main/java/apoc/load/LoadCsv.java
@@ -22,12 +22,14 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import static apoc.load.util.LoadCsvConfig.DEFAULT_ARRAY_SEP;
+import static apoc.load.util.LoadCsvConfig.Results;
+import static apoc.util.FileUtils.closeReaderSafely;
 import static apoc.util.Util.cleanUrl;
 import static apoc.util.Util.parseCharFromConfig;
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
-import static apoc.load.util.LoadCsvConfig.*;
 
 public class LoadCsv {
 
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/metrics/Metrics.java
```diff
@@ -162,10 +166,18 @@ public boolean test(LoadCsv.CSVResult o) {
                     "https://neo4j.com/docs/operations-manual/current/monitoring/metrics/expose/#metrics-csv");
         }
 
-        String url = new File(metricsDir, metricName + ".csv").toURI().toString();
```
jexp: What is the difference that makes it not fail with the new, more restrictive config setting?

src/main/java/apoc/util/FileUtils.java
```diff
@@ -260,4 +260,10 @@ public static boolean inNeo4jOwnedDirectory(File f) throws IOException {
             "dbms.directories.tx_log",
             "unsupported.dbms.directories.neo4j_home"
     );
+
+    public static void closeReaderSafely(CountingReader reader) {
```
jexp: couldn't have a try-with-resource done the same? just curious

jexp:
Not 100% sure what changed to make it work now, but otherwise looks good to me.

src/main/java/apoc/metrics/Metrics.java
```diff
@@ -162,10 +166,18 @@ public boolean test(LoadCsv.CSVResult o) {
...
+        String url = new File(metricsDir, metricName + ".csv").getAbsolutePath();
+        CountingReader reader = null;
+        try {
+            reader = FileUtils.readFile(url);
+            return new LoadCsv()
+                    .streamCsv(url, new LoadCsvConfig(config), reader)
+                    .filter(Metrics.duplicatedHeaderRows)
+                    .map(csvResult -> new GenericMetric(metricName, Util.toLong(csvResult.map.get("t")), csvResult.map));
+        } catch (Exception e) {
+            closeReaderSafely(reader);
```
moxious: does reader get closed if an exception was not thrown?   Seems like this would be better in finally?

src/main/java/apoc/load/LoadCsv.java
```diff
@@ -39,30 +41,35 @@
...
+        CSVReader csv = new CSVReaderBuilder(reader)
+                .withCSVParser(new CSVParserBuilder()
+                        .withQuoteChar(config.getQuoteChar())
+                        .withSeparator(config.getSeparator())
+                        .build())
+                .build();
+
+        String[] header = getHeader(csv, config);
+        boolean checkIgnore = !config.getIgnore().isEmpty() || config.getMappings().values().stream().anyMatch(m -> m.ignore);
+        return StreamSupport.stream(new CSVSpliterator(csv, header, url, config.getSkip(), config.getLimit(),
```
moxious: not in scope of this PR but yikes that's a lot of arguments CSVSpliterator takes.  Might want to figure out how to simplify that object at some point.

src/test/java/apoc/metrics/MetricsTest.java
```diff
@@ -0,0 +1,69 @@
...
+            neo4jContainer.close();
+        }
+        cleanBuild();
+    }
+
+    @Test
+    public void shouldGetMetrics() {
+        session.readTransaction(tx -> tx.run("RETURN 1 AS num;"));
+        String metricKey = "neo4j.bolt.sessions_started";
+        testResult(session, "CALL apoc.metrics.get({metricKey})",
```
moxious: nice!  This is basically the test I wish I could have written with the original code, but we needed testcontainers.

src/main/java/apoc/load/LoadCsv.java
jexp: probably just pass the config :)

src/main/java/apoc/util/FileUtils.java
conker84: good question my initial implementation was with a `try-with-resource`:
https://github.com/conker84/neo4j-apoc-procedures/commit/e41b338778432ceb1f7c050044ff4e4a08ab2af7#diff-fc68dee7028a27c02a83473c5e03ca44R167

But a test did by @moxious returned this exception:
```
Neo.ClientError.Procedure.ProcedureCallFailed: Failed to invoke procedure `apoc.metrics.get`: Caused by: java.io.IOException: Stream closed
```

So in order to fix it, I removed the `try-with-resource` and closed it as the stream closes

src/main/java/apoc/metrics/Metrics.java
conker84: good question, it's only here because of the `streamCsv` method closes the reader as the stream closes:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/62cb555123757e4384280246233ca0ce5539feb5/src/main/java/apoc/load/LoadCsv.java#L70

So I need to manage it only in case of the exception.
But we can change it if you want

src/main/java/apoc/load/LoadCsv.java
conker84: yep

src/main/java/apoc/metrics/Metrics.java
conker84: The old way leverages directly the `LoadCsv#csv` method which uses `FileUtils#readerFor` and it internally uses:
`FileUtils#changeFileUrlIfImportDirectoryConstrained` which is the cause of the issue; so I created the `LoadCsv#streamCsv` method that thakes a reader in input.

Issue #931, opened 2018-09-29
XXE in Xml.java


ProDigySML:
# The Issue

An XML External Entity attack is a type of attack against an application that parses XML input. This attack occurs when XML input containing a reference to an external entity is processed by a weakly configured XML parser. This attack may lead to the disclosure of confidential data, denial of service, server side request forgery, port scanning from the perspective of the machine where the parser is located, and other system impacts.

# Where the Issue Occurred

The following code snippets display the usage of `DocumentBuilderFactory` without securely disabling entities:

https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/591f8edba053ee92b65968f76cb68af1aece9898/src/main/java/apoc/load/Xml.java#L74

PR #1208, opened 2019-05-22
allow loading of xml with DTD references, fixes #1185

### LINKED ISSUE #1185: apoc.load.xml not working on specific API's

WITH 'https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=pmc&id=212403' as url
CALL apoc.load.xml(url, '',{}, false) yield value
return value

should return a xml 

returns error:

Neo.ClientError.Procedure.ProcedureCallFailed: Failed to invoke procedure `apoc.load.xml`: Caused by: org.xml.sax.SAXParseException; lineNumber: 2; columnNumber: 10; DOCTYPE is disallowed when the feature "http://apache.org/xml/features/disallow-doctype-decl" set to true.

## Specifications (Mandatory)

Using the same query in a previous version of neo4j and apoc will return the values.

Currently used versions

neo4j v3.5.0.4 
apoc v3.5.0.3

taseroth:
loading of xml files was previously forbidden (#931). With this change,
an dummy entity resolver is registered that does nothing, e.g. does not
load external files.

Fixes #1185

allow parsing of xml files containing dtd references.

  - removed the parser configuration that prohibited loading xml with dtd
  - added a EntityResolver that does not load dtd (does nothing)
  - added unit test and example xml file
  - removed XXE unit test and associated xml file


taseroth:
I' m new to travis: do have anything to do to fix this? I did ran all test locally (via gradle) without error

jexp:
Looks good thanks a lot!

PR #1222, opened 2019-06-12
fixes #1161 - Adds apoc.util.compress and apoc.util.decompress functions

### LINKED ISSUE #1161: add function to compress/decompress binary data
apoc.util.decompress(byte[], [{options)]) -> string ?
apoc.util.compress(string, [{options}]) -> byte[]

alexiudice:
Fixes #1161 

Adds apoc.util.compress and apoc.util.decompress functions.

Uses a `{config}` map for configuration so function header does not have to change incase we want to add more options at a later time.

### DIFF
--- a/src/main/java/apoc/util/CompressionUtils.java
+++ b/src/main/java/apoc/util/CompressionUtils.java
@@ -1,0 +1,63 @@
+package apoc.util;
+
+import org.apache.commons.lang.ArrayUtils;
+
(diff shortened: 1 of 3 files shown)


jexp:
I think you missed the idea :)

it was meant to be an efficient way of turning text into gzip compressed byte[]

I had already started on these (also just limiting it to UTF-8) just haven't pushed the branch yet.

PR #1224, opened 2019-06-13
fixes #1159 - Adds apoc.coll.extract function and unit tests.

### LINKED ISSUE #1159: Add apoc.coll.extract(coll, 'property')
## Feature description
Function to extract a property from a collection of maps, nodes or relationships.
`apoc.coll.extract(coll, 'property')`

This would be a cleaner way for extracting properties from a collection rather than using  pattern comprehension.

alexiudice:
Fixes #1159 

Adds apoc.coll.extract function and unit tests.

Can only extract from collections of Maps, Nodes, and Relationships. If one of the objects in the collection is missing the property then a null value is used.

### DIFF
--- a/src/main/java/apoc/coll/Coll.java
+++ b/src/main/java/apoc/coll/Coll.java
@@ -1,6 +1,7 @@
 package apoc.coll;
 
 import org.apache.commons.math3.util.Combinations;
+import org.neo4j.graphdb.Entity;
 import org.neo4j.graphdb.Path;
 import org.neo4j.helpers.collection.Pair;
 import org.neo4j.kernel.impl.util.statistics.IntCounter;
@@ -873,5 +874,28 @@ public class Coll {
         }
 
         return newList;
+    }
+
+    @UserFunction
+    @Description("apoc.coll.extract(coll, property) - extracts the property from each element in the coll and returns a list of the values ")
+    public List<Object> extract(@Name("coll") List<Object> coll, @Name( "property" ) String property) {
+        if (coll == null || coll.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        return coll.stream().map( o -> {
+            if ( o instanceof Map )
+            {
+                return ((Map) o).getOrDefault( property, null );
+            }
+            else if ( o instanceof Entity )
+            {
+                return ((Entity) o).getProperty( property, null );
+            }
+            else
+            {
+                return null;
+            }
+        } ).collect( Collectors.toList() );
     }
 }
--- a/src/test/java/apoc/coll/CollTest.java
+++ b/src/test/java/apoc/coll/CollTest.java
@@ -841,5 +841,38 @@ public class CollTest {
                 });
     }
 
+    @Test
+    public void testExtractMap() throws Exception {
+        testCall(db, "WITH [{a: 1},{b: 2},{a: 3}] AS coll" +
+                        " RETURN apoc.coll.extract(coll, 'a') AS ext",
+                (row) -> {
+                    assertEquals(asList(1L, null, 3L), (List<Long>) row.get( "ext" ));
+                });
+    }
+
+    @Test
+    public void testExtractNode() throws Exception {
+        testCall(db, "CREATE (n1:Node {prop: 1}) CREATE (n2:Node) CREATE (n3:Node {prop: 3}) WITH [n1,n2,n3] AS coll RETURN apoc.coll.extract(coll, 'prop') AS ext",
+                (row) -> {
+                    assertEquals(asList(1L, null, 3L), (List<Long>) row.get( "ext" ));
+                });
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


src/main/java/apoc/coll/Coll.java
```diff
@@ -874,4 +875,33 @@ public boolean different(@Name("values") List<Object> values) {
 
         return newList;
     }
+
+    @UserFunction
+    @Description("apoc.coll.extract(coll, property) - extracts the property from each element in the coll and returns a list of the values ")
+    public List<Object> extract(@Name("coll") List<Object> coll, @Name( "property" ) String property) {
+        if (coll == null || coll.size() == 0) {
```
jexp: coll.isEmpty()

src/main/java/apoc/coll/Coll.java
```diff
@@ -874,4 +875,33 @@ public boolean different(@Name("values") List<Object> values) {
...
+            return Collections.emptyList();
+        }
+        List<Object> extraction = new ArrayList<>(  );
```
jexp: initialize with size

src/main/java/apoc/coll/Coll.java
```diff
@@ -874,4 +875,33 @@ public boolean different(@Name("values") List<Object> values) {
...
+        Object o = coll.get( 0 );
```
jexp: We should check it per instance in the loop, and possibly add a null if it's neither map nor graph entity.

src/main/java/apoc/coll/Coll.java
```diff
@@ -874,4 +875,33 @@ public boolean different(@Name("values") List<Object> values) {
...
+        }
+        else if (o instanceof Node || o instanceof Relationship) {
+            for(Object c : coll)
+            {
+                extraction.add(((Entity) c).getProperty( property, null ));
+            }
+        }
+        else
+        {
+            throw new RuntimeException( "Function apoc.coll.extract requires a collection of Maps, Nodes, or Relationships.");
```
jexp: I would just do the null thing. Which is how most functions in cypher work.

src/main/java/apoc/coll/Coll.java
```diff
@@ -874,4 +875,33 @@ public boolean different(@Name("values") List<Object> values) {
...
+    @Description("apoc.coll.extract(coll, property) - extracts the property from each element in the coll and returns a list of the values ")
+    public List<Object> extract(@Name("coll") List<Object> coll, @Name( "property" ) String property) {
+        if (coll == null || coll.size() == 0) {
+            return Collections.emptyList();
+        }
+        List<Object> extraction = new ArrayList<>(  );
+        Object o = coll.get( 0 );
+
+
+        if (o instanceof Map) {
```
jexp: it would be nicer to do a `coll.stream().map().collect(toList())`

src/main/java/apoc/coll/Coll.java
```diff
@@ -874,4 +875,33 @@ public boolean different(@Name("values") List<Object> values) {
...
+            for(Object c : coll)
+            {
+                extraction.add(((Map) c).getOrDefault( property, null ));
+            }
+        }
+        else if (o instanceof Node || o instanceof Relationship) {
```
jexp: Check for Entity should be good enough.

src/test/java/apoc/coll/CollTest.java
```diff
@@ -841,5 +841,31 @@ public void testDropNeighboursDifferentTypes() throws Exception {
                 });
     }
 
+    @Test
+    public void testExtractMap() throws Exception {
+        testCall(db, "WITH {a: 1} AS m1, {b: 2} AS m2, {a: 3} AS m3" +
+                        " WITH [m1,m2,m3] AS coll" +
```
jexp: you can inline the maps

src/test/java/apoc/coll/CollTest.java
```diff
@@ -841,5 +841,31 @@ public void testDropNeighboursDifferentTypes() throws Exception {
...
+    }
+
+    @Test
+    public void testExtractRelationship() throws Exception {
+        testCall(db, "CREATE (n1:Node) CREATE (n2:Node) CREATE (n1)-[r1:Rel {prop: 1}]->(n2) CREATE (n1)-[r2:Rel]->(n2) CREATE (n1)-[r3:Rel {prop: 3}]->(n2) WITH [r1,r2,r3] AS coll RETURN apoc.coll.extract(coll, 'prop') AS ext",
+                (row) -> {
+                    assertEquals(asList(1L, null, 3L), (List<Long>) row.get( "ext" ));
+                });
+    }
+
```
jexp: add a check for mixed collections also with non-valid elements

alexiudice:
@jexp I've made the requested changes/additions.

jexp:
Thanks a lot, we should also add some docs for it.

PR #1227, opened 2019-06-16
 fixes rounding issues in min, max calculations of statistics aggrega…

### LINKED ISSUE #788: max value in apoc.agg.statistics does not match either apoc.coll.max nor builtin Max() in Cypher
I tried with the the set values given in the text file.
[datasource.txt](https://github.com/neo4j-contrib/neo4j-apoc-procedures/files/1917361/datasource.txt)

the result of Cypher's `Max()` and `apoc.coll.max()` are the same
but the max value returned by `apoc.agg.statistics()` is different from both `Max()` and `apoc.coll.max()`

vboulaye:
Fixes #788

Compute Min/Max properties of the apoc.agg.statistics aggregation using standard comparison instead of the HdrHistogram

## Proposed Changes 

HdrHistogram is used to compute statistics but it introduces some differences on min/max properties when it uses double computations. HdrHistogram use is replaced by keeping a reference to the actual min and max values.

The minNonZero calculation is still coming from HdrHistogram, as I am not sure if we should also handle it outside of the histogram.

### DIFF
--- a/src/main/java/apoc/agg/Statistics.java
+++ b/src/main/java/apoc/agg/Statistics.java
@@ -5,7 +5,6 @@ import org.HdrHistogram.Histogram;
 import org.HdrHistogram.HistogramUtil;
 import org.neo4j.procedure.*;
 
-import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/agg/Statistics.java
```diff
@@ -42,21 +43,27 @@ public void aggregate(@Name("value") Number value, @Name(value = "percentiles",
...
             boolean empty = totalCount == 0;
-            Map<String,Number> result = new LinkedHashMap<>(percentiles.size()+6);
-            result.put("min",values != null ? (Number)values.getMinValue() : (Number)doubles.getMinValue());
-            result.put("minNonZero",values != null ? (Number)values.getMinNonZeroValue() : (Number)doubles.getMinNonZeroValue());
-            result.put("max",values != null ? (Number)values.getMaxValue() : (Number)doubles.getMaxValue());
-            result.put("total",totalCount);
-            result.put("mean",values != null ? values.getMean() : doubles.getMean());
-            result.put("stdev",values != null ? values.getStdDeviation() : doubles.getStdDeviation());
+            Map<String, Number> result = new LinkedHashMap<>(percentiles.size() + 6);
+            result.put("min", minValue != null ? minValue : 0L);
```
jexp: I think we can and should put null there if we had no rows.

vboulaye: I implemented it this way because before the change there was always a non-null value in the map as the getMinValue()/getMaxValue() from Histogram answer with `long` values (and I always worry about storing null values anyway).
but I don't mind changing this!

jexp: yes please change.

vboulaye: Hi, I did it in the second commit

Issue #1229, opened 2019-06-17
Create maxItems() and minItems() aggregations


InverseFalcon:
## Feature description 
Working with `min()` and `max()` can be troublesome, as often we aren't actually looking for just the min or max values, but any items associated with those values. To get these, we often need to resort to more complex Cypher.

In the event that there are no other groupings, and we only want one minValue result, we can collect() the items by the value, ORDER BY and LIMIT 1:

```
MATCH (p:Person)
WHERE p.born >= 1930
WITH p.born as born, collect(p) as persons
ORDER BY born ASC LIMIT 1
RETURN born, [person in persons | person.name] as persons
```

But when we want to do this per row across multiple rows, we need a more complicated alternative since we can't use LIMIT:

```
MATCH (p:Person)-[:ACTED_IN]->(m:Movie)
WHERE m.title STARTS WITH 'The Matrix'
WITH m, p.born as born, collect(p) as persons
WITH m, min(born) as minBorn, collect({born:born, persons:persons}) as bornInfoList
UNWIND [info in bornInfoList WHERE info.born = minBorn] as bornInfo
RETURN m.title as title, bornInfo.born as born, [person in bornInfo.persons | person.name] as persons
```

It would be great to have an aggregation function where you pass an item, pass a value, and get back a structure with the min (or max) value and the list of items with that value. We could also optionally limit the list.

## Considered alternatives
We may want to consider what to do about null values. In the Cypher above, if we mistype the property name we get all results aggregated for a value of null. I don't think we want to have that behavior. It may be best to ignore any null values and remove them, as that is likely to be costly, and aggregations by null when there are no values to use for max or min are unlikely.

## How this feature can improve the project?

Usage could be as follows:

```
MATCH (p:Person)-[:ACTED_IN]->(m:Movie)
WHERE m.title STARTS WITH 'The Matrix'
WITH m, apoc.agg.minItems(p, p.born) as minResult
RETURN m.title as title, minResult.value as born, [person in minResult.items | person.name] as persons
```

With a good implementation, we could have better memory handling than existing Cypher approaches, as we can only concentrate on items with the max or min value and optionally limit the results.

InverseFalcon:
Request fulfilled, closing.

PR #1230, opened 2019-06-17
Added apoc.agg.maxItems() and apoc.agg.minItems()

InverseFalcon:
Implements #1229 

Adds aggregation functions for collecting items associated with a minimal or maximal value.

Adds aggregation functions for collecting items associated with a minimal or maximal value. Includes the ability to optionally limit the results per value.


InverseFalcon:
Looks like the failure is unrelated to the code I checked in

PR #1242, opened 2019-07-05
fixes #1200: better error handling for apoc.graph.fromDocument

### LINKED ISSUE #1200: better error handling for `apoc.graph.fromDocument`
e.g. list the different types of objects that are missing label/id-fields

Neo.ClientError.Procedure.ProcedureCallFailed: Failed to invoke procedure `apoc.graph.fromDocument`: Caused by: java.lang.RuntimeException: every object must have item as `id` field name

conker84:
Fixes #1200 

Improved the error handling for `apoc.graph.fromDocument` procedure by adding the JSON representation in the error message 

  - changed `Doc2Graph#validate`  method in order to include the he JSON representation of the document with the error
  - simplified the `apoc.graph.validateDocument` procedure in order to show only documents with errors
  - improved the documentation of `apoc.graph.validateDocument`


src/main/java/apoc/graph/document/builder/DocumentToGraph.java
```diff
@@ -28,10 +29,15 @@ public DocumentToGraph(GraphDatabaseService db, GraphsConfig config) {
...
-            throw new RuntimeException("every object must have " + config.getIdField() + " as `id` field name");
-        } else if (!map.containsKey(config.getLabelField())) {
-            throw new RuntimeException("every object must have " + config.getLabelField() + " as `label` field name");
+            messages.add("`" + config.getIdField() + "` as id-field name");
+        }
+        if (!map.containsKey(config.getLabelField())) {
+            messages.add("`" + config.getLabelField() + "` as label-field name");
+        }
+        if (!messages.isEmpty()) {
+            throw new RuntimeException("The object " + JsonUtil.writeValueAsString(map) + " must have " + String.join(" and ", messages));
```
jexp: we have to be careful, if the document is 100MB, either take a substring or render just the first few fields.
perhaps write to a string writer that's limited to a 1000 char buffer and then stop writing.

PR #1244, opened 2019-07-06
fixes #1243: Improving apoc.graph.fromDocument: manage missing id and label

### LINKED ISSUE #1243: Improving apoc.graph.fromDocument: manage missing id and labels
Sub-tasks taken from #1198

TODO:
- [x] be smarter about missing id's assign a UUID
- [x] Validate unique ids and labels
- [x] provide a default label from config in case is missing in the data

conker84:
Fixes #1243 

Added new config params in order to manage auto-generated ids and the default label

- [x] be smarter about missing id's assign a UUID
- [x] Validate unique ids (we only check duplicates on ids, not labels)
- [x] provide a default label from config in case is missing in the data

### DIFF
--- a/src/main/java/apoc/graph/Graphs.java
+++ b/src/main/java/apoc/graph/Graphs.java
@@ -5,8 +5,8 @@ import apoc.graph.document.builder.DocumentToGraph;
 import apoc.graph.util.GraphsConfig;
 import apoc.result.RowResult;
 import apoc.result.VirtualGraph;
-import apoc.util.JsonUtil;
 import apoc.util.Util;
+import org.apache.commons.lang3.StringUtils;
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.Path;
@@ -15,7 +15,6 @@ import org.neo4j.helpers.collection.Iterables;
 import org.neo4j.procedure.*;
 
 import java.util.*;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Stream;
 
 /**
@@ -117,39 +116,20 @@ public class Graphs {
     @Description("apoc.graph.fromDocument({json}, {config}) yield graph - transform JSON documents into graph structures")
     @Procedure(mode = Mode.WRITE)
     public Stream<VirtualGraph> fromDocument(@Name("json") Object document, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) throws Exception {
-        Collection<Map> coll = getDocumentCollection(document);
         DocumentToGraph documentToGraph = new DocumentToGraph(db, new GraphsConfig(config));
-        return Stream.of(documentToGraph.create(coll));
-    }
-
-    public Collection<Map> getDocumentCollection(@Name("json") Object document) {
-        Collection<Map> coll;
-        if (document instanceof String) {
-            document  = JsonUtil.parse((String) document, null, Object.class);
-        }
-        if (document instanceof Collection) {
-            coll = (Collection) document;
-        } else {
-            coll = Collections.singleton((Map) document);
-        }
-        return coll;
+        return Stream.of(documentToGraph.create(document));
     }
 
     @Description("apoc.graph.validateDocument({json}, {config}) yield row - validates the json, return the result of the validation")
     @Procedure(mode = Mode.READ)
     public Stream<RowResult> validateDocument(@Name("json") Object document, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) {
-        DocumentToGraph documentToGraph = new DocumentToGraph(db, new GraphsConfig(config));
-        AtomicLong index = new AtomicLong(-1);
-        return getDocumentCollection(document).stream()
-                .map(elem -> {
-                    long line = index.incrementAndGet();
-                    try {
-                        documentToGraph.validate(elem);
-                        return null;
-                    } catch (Exception e) {
-                        return new RowResult(Util.map("index", line, "message", e.getMessage()));
-                    }
-                })
-                .filter(elem -> elem != null);
+        GraphsConfig graphConfig = new GraphsConfig(config);
+        DocumentToGraph documentToGraph = new DocumentToGraph(db, graphConfig);
+        Map<Long, List<String>> dups = documentToGraph.validateDocument(document);
+
+        return dups.entrySet().stream()
+                .map(e -> new RowResult(Util.map("index", e.getKey(), "message", String.join(StringUtils.LF, e.getValue()))));
     }
+
+
 }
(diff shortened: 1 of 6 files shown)


src/main/java/apoc/graph/Graphs.java
```diff
@@ -137,17 +138,53 @@ else if (data instanceof Iterator) {
...
+        Set<String> skipLine = new HashSet<>();
+        return documentCollection.stream()
                 .map(elem -> {
+                    long line = index.incrementAndGet();
                     try {
+                        // id, label validation
+                        documentToGraph.prepareData(elem);
                         documentToGraph.validate(elem);
-                        return new RowResult(Util.map("index", index.incrementAndGet(), "message", "Valid"));
+                        Object id = elem.get(graphConfig.getIdField());
```
jexp: please extract the code in a well named method so it's at the same abstraction as `documentToGraph.prepareData(elem);`

src/test/java/apoc/graph/GraphsTest.java
```diff
@@ -588,6 +611,41 @@ public void testCreateVirtualSimpleNodeFromCypherMap() {
         });
     }
 
+    @Test
```
jexp: please also add a test for nested children with missing id's / labels

jexp:
@conker84 can you rebase this one?

src/main/java/apoc/graph/Graphs.java
```diff
@@ -117,37 +118,60 @@ else if (data instanceof Iterator) {
...
-                        documentToGraph.validate(elem);
-                        return new RowResult(Util.map("index", index.incrementAndGet(), "message", "Valid"));
-                    } catch (Exception e) {
-                        return new RowResult(Util.map("index", index.incrementAndGet(), "message", e.getMessage()));
-                    }
-                });
+    public Stream<RowResult> validateDocument(@Name("json") Object document, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) {
+        GraphsConfig graphConfig = new GraphsConfig(config);
+        DocumentToGraph documentToGraph = new DocumentToGraph(db, graphConfig);
+        Collection<Map<String, Object>> documentCollection = getDocumentCollection(document);
```
jexp: shouldn't this be a method on documentToGraph?

src/main/java/apoc/graph/Graphs.java
```diff
@@ -117,37 +118,60 @@ else if (data instanceof Iterator) {
...
+
+        Map<Long, List<String>> dups = regroupDuplicates(documentToGraph, documentCollection);
```
jexp: Same here.

src/main/java/apoc/graph/Graphs.java
```diff
@@ -117,37 +118,60 @@ else if (data instanceof Iterator) {
...
+        Map<Long, String> invalids = documentToGraph.validate(documentCollection);
+
+        for (Map.Entry<Long, String> invalid : invalids.entrySet()) {
```
jexp: this code-block including the mapping below should also be in a method

src/main/java/apoc/graph/document/builder/DocumentToGraph.java
```diff
@@ -129,11 +175,55 @@ private boolean isSimpleType(Map.Entry<String, Object> e) {
...
+                })
+                .collect(Collectors.groupingBy(Map.Entry::getKey,
+                        Collectors.mapping(Map.Entry::getValue, Collectors.toList())))
+                .entrySet()
+                .stream()
+                .filter(e -> e.getValue().size() > 1)
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    private List<Map<String, Object>> flatMapFields(Map<String, Object> map) {
```
jexp: can just return a stream if you use it always as streams :)
and concat the `Stream.of(map).cconcat(map.values().stream().filter(...)).flatMap())` with to combine the streams

src/main/java/apoc/graph/document/builder/LabelBuilder.java
```diff
@@ -16,7 +16,7 @@ public LabelBuilder(GraphsConfig config) {
     }
 
     public Label buildLabel(Map<String, Object> obj) {
-        String label = "DocNode"; // Default label
+        String label = "";
```
jexp: are we making sure that we never end up with `""` as a label, otherwise Neo4j would fail on that.

src/main/java/apoc/graph/util/GraphsConfig.java
```diff
@@ -10,14 +10,18 @@
...
     private String idField;
+    private boolean autoGenerateID;
+    private String defaultLabel;
 
     public GraphsConfig(Map<String, Object> config) {
         if (config == null) {
             config = Collections.emptyMap();
         }
         write = toBoolean(config.getOrDefault("write", false));
+        autoGenerateID = toBoolean(config.getOrDefault("autoGenerateID", true));
```
jexp: `"generateId"` as config option.

src/main/java/apoc/graph/Graphs.java
conker84: ok

src/main/java/apoc/graph/document/builder/DocumentToGraph.java
conker84: done

src/main/java/apoc/graph/document/builder/LabelBuilder.java
conker84: We prepare the data with the `prepareData` method so if the user provide a default Label we'll use it. Otherwise it will fail.

conker84: The property `defaultLabel` in `GraphConfig` is empty I can use the `DocNode` as default value

src/main/java/apoc/graph/Graphs.java
conker84: ok

src/main/java/apoc/graph/util/GraphsConfig.java
conker84: done

src/main/java/apoc/graph/document/builder/DocumentToGraph.java
```diff
@@ -30,21 +33,23 @@ public DocumentToGraph(GraphDatabaseService db, GraphsConfig config) {
...
+                        msgs.add("`" + config.getLabelField() + "` as label-field name");
+                    }
+                    return new AbstractMap.SimpleEntry<>(elem, msgs);
+                })
+                .filter(elem -> !elem.getValue().isEmpty())
+                .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
     }
 
-    private String formatDocument(Map map) {
+    public String formatDocument(Map map) {
```
jexp: make the 100 a parameter and rename the method to indicate that it's only for reporting
like "renderPartialDocument"

jexp: Can also be done later.

PR #1247, opened 2019-07-08
Fix test failures & JDK 11 compatibility

sarmbruster:


conker84:
LGTM,
I got the following message in Travis for the [apoc build](https://travis-ci.org/conker84/neo4j-apoc-procedures), in the last days:
`Expected feature release number in range of 9 to 14, but got: 8`
So I don't know if we also need to update the `travis.yml` file

sarmbruster:
Weird, the CI build for this PR went green. @conker84 maybe you could try to flush caches in travis?

PR #1248, opened 2019-07-08
 fixes #1245: Improving apoc.graph.fromDocument: storing value objects

### LINKED ISSUE #1245: Improving apoc.graph.fromDocument: storing value objects
Sub-task of #1198

figure out a ok/configurable way to store "value-objects" like: size: { h: 14, w: 21, uom: "cm" } which is not an entity, either as JSON or as two field: one for keys and one for values (as string), e.g. size_keys:["h","w","uom"],size_values:["14","21","cm"]

conker84:
Fixes #1245

Add support to store value object via `valueObjects` map. The properties of the value objects will be flattened with this general rule:

A json like this:
```json
{
    "id": 1,
    "type": "Person",
    "name": "Andrea",
    "sizes": {
        "weight": {
            "value": 70,
            "um": "Kg"
        },
        "height": {
            "value": 174,
            "um": "cm"
        }
    }
}
```

Will be flattened as follows:

```json
{
    "id": 1,
    "type": "Person",
    "name": "Andrea",
    "sizes.weight.value": 70,
    "sizes.weight.um": "Kg",
    "sizes.height.value": 174,
    "sizes.height.um": "cm"
}
```

 - added the support via the property configuration `valueObjects`
 - updated the documentation

### DIFF
--- a/docs/asciidoc/virtual/virtual-graph.adoc
+++ b/docs/asciidoc/virtual/virtual-graph.adoc
@@ -75,6 +75,8 @@ The config is composed by the following parameters:
 * *write*, _type boolean_: persist the graph otherwise return a Virtual Graph, default *false*
 * *labelField*, _type String_: the field name that became the label of the node, default *type*
 * *idField*, _type String_: the document field name that will become the id field of the created nodes (used for node resolution when you create relationships between nodes), default *id*
+* *autoGenerateID*, _type boolean_: in case of missing id-field value it generates an UUID for it, default *true*
+* *defaultLabel*, _type String_: in case of missing label-field value is uses the provided default label, default is empty
 
 
 [source, json]
--- a/docs/asciidoc/virtual/virtual-graph.adoc
+++ b/docs/asciidoc/virtual/virtual-graph.adoc
@@ -199,7 +199,11 @@ RETURN graph
 
 As a result we have a virtual graph with two nodes and one relationship:
 
+<<<<<<< HEAD
 image::apoc.graph.fromDocument_1[scaledwidth="100%"]
+=======
+image::{img}/apoc.graph.fromDocument_1[width=800]
+>>>>>>> fixes #1200: better error handling for apoc.graph.fromDocument
 
 You can also do a pre-validation over the document with the `apoc.graph.validateDocument` procedure that will return the
 record with invalid data.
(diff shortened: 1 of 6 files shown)


docs/asciidoc/virtual-graph.adoc
```diff
@@ -181,3 +185,64 @@ CALL apoc.graph.fromDocument("{'id': 1,'type': 'artist','name':'Genesis','member
...
+    "sizes.height.value": 174,
+    "sizes.height.um": "cm"
+}
+----
+
+As specified you can also provide a set of value-object properties for a Label:
+
+[source,cypher]
+----
+call apoc.graph.validateDocument(<json>, {valueObjects: {Person: ['sizes', 'anotherPropertyKey']}})
```
jexp: best use the config we talked about here, so we don't publish an API that we need to support and which is immediately deprecated

https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1267#issuecomment-517628589

jexp:
Needs rebase.

docs/asciidoc/virtual/virtual-graph.adoc
```diff
@@ -75,7 +75,38 @@ The config is composed by the following parameters:
 * *write*, _type boolean_: persist the graph otherwise return a Virtual Graph, default *false*
```
jexp: we should make this a table

docs/asciidoc/virtual/virtual-graph.adoc
```diff
@@ -75,7 +75,38 @@ The config is composed by the following parameters:
...
 * *labelField*, _type String_: the field name that became the label of the node, default *type*
 * *idField*, _type String_: the document field name that will become the id field of the created nodes (used for node resolution when you create relationships between nodes), default *id*
+* *generateId*, _type boolean_: in case of missing id-field value it generates an UUID for it, default *true*
+* *defaultLabel*, _type String_: in case of missing label-field value is uses the provided default label, default is empty
+* *valueObjects*, _type Map<String, Object>_: for each provided Label as key you can specify a value or a set of values
```
jexp: should we keep `valueObjects` if we have `mappings`?

jexp:
Looks good, can you look at my two comments.
Not sure if mappings is too complex for simple "valueObjects" or if we can drop valueObjects?

i.e. valueObjects: {`Person`: [addresses]} vs. mappings: ":Person { *, @addresses}" 

and also what happens if they contradict each other :)

conker84:
Sorry @jexp I forgot to delete the old property from the doc there is no `valueObjects` anymore

conker84:
I updated the doc

jexp:
Thanks a lot

PR #1255, opened 2019-07-15
fixes #1246: Incorrect headers order after apoc csv export in community edition 3.5.6

### LINKED ISSUE #1246: Incorrect headers order after apoc csv export in community edition 3.5.6
CSV headers seem to misalign with values in the rows when exporting with APOC. The issue could be related to https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1188, however the issue is now happening on the community version. Also, the bug appeared after upgrading to apoc 3.5.0.4 and neo4j 3.5.6 but was non-existent with apoc 3.4.0.4 and neo4j 3.4.5.

## Expected Behavior
Headers/columns are displayed correctly in csv export.

## Actual Behavior
Headers/columns are in the wrong order for certain pieces of data. For example, _id, _labels and _key are correct but _style should include an object with css styles instead of 'Field'.

## Steps to reproduce
Running "CALL apoc.export.csv.all(null, {stream:true,batchSize:100}) YIELD data RETURN data;", gives:

|                                                                                                                                                                             | 
|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------| 
| "_id","_labels","_key","_style","_type","allowed","name","_start","_end","_type","_key","_style" | 
| "0",":SomeLabel","name","Field","someType","","",,,,,                                                                                                               | 
| "20",":SomeOtherLabel","step","Field","someOtherType","","",,,,,                                                                                                    | 

## Versions

  - OS: MacOSX, Ubuntu
  - Neo4j: >= 3.5.6 community edition
  - Neo4j-Apoc: apoc-3.5.0.4-all.jar

conker84:
Fixes #1246

Fixed the bug.

  - Fixed the bug in the export

### DIFF
--- a/src/main/java/apoc/export/csv/CsvFormat.java
+++ b/src/main/java/apoc/export/csv/CsvFormat.java
@@ -28,6 +28,9 @@ public class CsvFormat implements Format {
     private final GraphDatabaseService db;
     private boolean applyQuotesToAll = true;
 
+    private static final String[] NODE_HEADER_FIXED_COLUMNS = {"_id:id", "_labels:label"};
+    private static final String[] REL_HEADER_FIXED_COLUMNS = {"_start:id", "_end:id", "_type:label"};
+
     public CsvFormat(GraphDatabaseService db) {
         this.db = db;
     }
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/export/csv/CsvFormat.java
```diff
@@ -129,8 +129,8 @@ public void writeAll(SubGraph graph, Reporter reporter, ExportConfig config, CSV
         out.writeNext(header.toArray(new String[header.size()]), applyQuotesToAll);
         int cols = header.size();
 
-        writeNodes(graph, out, reporter, nodePropTypes, cols, config.getBatchSize(), config.getDelim());
-        writeRels(graph, out, reporter, relPropTypes, cols, nodeHeader.size(), config.getBatchSize(), config.getDelim());
+        writeNodes(graph, out, reporter, nodeHeader.subList(2, nodeHeader.size()), cols, config.getBatchSize(), config.getDelim());
```
jexp: this looks very implict, can we make it explicit, what it means? the offset of 2 or 3

src/main/java/apoc/export/csv/CsvFormat.java
```diff
@@ -266,16 +266,16 @@ private void writeNodes(SubGraph graph, CSVWriter out, Reporter reporter, Export
         String[] header = nodeHeader.toArray(new String[nodeHeader.size()]);
         out.writeNext(header, applyQuotesToAll); // todo types
         int cols = header.length;
-        writeNodes(graph, out, reporter, nodePropTypes, cols, config.getBatchSize(), config.getDelim());
+        writeNodes(graph, out, reporter, nodeHeader.subList(2, nodeHeader.size()), cols, config.getBatchSize(), config.getDelim());
```
jexp: same here

conker84:
@jexp updated with your feedback.
Thanks a lot!

PR #1259, opened 2019-07-17
Upgraded com.jayway.jsonpath:json-path from version 2.2.0 to 2.4.0

DirkMahler:

Upgraded com.jayway.jsonpath:json-path from version 2.2.0 to 2.4.0 as the older version comes with a dependency to net.minidev:accessors-smart (version 1.1) which contains a shaded but outdated version of ASM, see https://github.com/netplex/json-smart-v2/pull/35. 

(caused troubles using Neo4j/APOC embedded in jQAssistant)


jexp:
Thanks a lot

PR #1296, opened 2019-09-20
fixes: #1253: Add support for gs:// file prefix for Google Cloud Storage

### LINKED ISSUE #1253: Add support for gs:// file prefix for Google Cloud Storage
Apoc supports s3:// to read/write files to  AWS S3 storage but it doesn’t support the equivalent Google Cloud Storage gs://

Both cloud storage services provide java libraries that use the hdfs interface so it should be not too hard to add gs://

There is more info on gs:// here:
https://cloud.google.com/dataproc/docs/concepts/connectors/cloud-storage

conker84:
Fixes #1253 

  - Added support for `gs://` protocol

### DIFF
--- a/docs/asciidoc/import/webapis.adoc
+++ b/docs/asciidoc/import/webapis.adoc
@@ -6,7 +6,7 @@
 This section gives an overview of procedures that can be used to load data from Web-APIs into Neo4j.
 --
 
-Supported protocols are `file`, `http`, `https`, `s3`, `hdfs` with redirect allowed.
+Supported protocols are `file`, `http`, `https`, `s3`, `gs`, `hdfs` with redirect allowed.
 
 If no procedure is provided, this procedure will try to check whether the URL is actually a file.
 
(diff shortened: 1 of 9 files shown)


src/main/java/apoc/util/google/cloud/GCStorageURLConnection.java
```diff
@@ -0,0 +1,95 @@
...
+        return blob.getContentType();
+    }
+
+    @Override
+    public InputStream getInputStream() throws IOException {
+        if (!connected) {
+            connect();
+        }
+        byte[] data;
+        if (blob == null) {
```
jexp: is there no streaming way too read the content?

jexp: e.g. via 
https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/storage/Blob.html#downloadTo-java.io.OutputStream-com.google.cloud.storage.Blob.BlobSourceOption...-

or 

https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/storage/Blob.html#reader-com.google.cloud.storage.Blob.BlobSourceOption...-

docs/asciidoc/import/webapis.adoc
```diff
@@ -62,6 +62,52 @@ or
 * `s3://endpoint:port/bucket/key?accessKey=accessKey&secretKey=secretKey`
 
 
+== Using Google Cloud Storage
+
+In order to use Google Cloud Storage you need that the following jars are included into the plugin dir:
+
+* api-common-1.8.1.jar
```
jexp: lots of jars, can we look into providing a module that generates an uber jar for these? that we can make available?

src/main/java/apoc/util/google/cloud/GCStorageURLConnection.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @VisibleForTesting
+    public Storage getStorage(URI uri) {
+        Storage storage;
+        Map<String, String> queryParams = getQueryParams(uri);
+        AuthType authenticationType = AuthType.valueOf(queryParams.getOrDefault("authenticationType", AuthType.NONE.toString()));
+        switch (authenticationType) {
+            case SERVICE:
+                String googleAppCredentialsEnv = System.getenv("GOOGLE_APPLICATION_CREDENTIALS");
```
jexp: can't it be passed in other means? e.g. in server setups those might not be settable?

src/test/java/apoc/load/LoadGoogleCloudStorageTest.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final String BUCKET_NAME = System.getenv("APOC_GC_BUCKET_NAME");
+
+    @BeforeClass
+    public static void before() {
+        Assume.assumeNotNull(BUCKET_NAME);
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        db = new TestGraphDatabaseFactory().newImpermanentDatabaseBuilder()
```
jexp: do we need file import enabled for remote loads, I'm not sure how we do it with other options?

src/test/java/apoc/load/LoadGoogleCloudStorageTest.java
```diff
@@ -0,0 +1,70 @@
...
+import static apoc.load.LoadCsvTest.assertRow;
+import static apoc.util.MapUtil.map;
+import static apoc.util.TestUtil.testResult;
+import static org.junit.Assert.assertFalse;
+
+public class LoadGoogleCloudStorageTest {
+
+    private GraphDatabaseService db;
+
+    private static final String BUCKET_NAME = System.getenv("APOC_GC_BUCKET_NAME");
```
jexp: where does this come from?

jexp: Is this something we can set up dynamically?

docs/asciidoc/import/webapis.adoc
conker84: I created this issue #1297

src/main/java/apoc/util/google/cloud/GCStorageURLConnection.java
conker84: As the first implementation, I tried to stick to the best practice, but I added the authentication type because as you underscored there can be someone that uses an alternative way to authenticate to the service, so we can improve that in a community-driven way. WDYT?

src/main/java/apoc/util/google/cloud/GCStorageURLConnection.java
conker84: Ok!

src/test/java/apoc/load/LoadGoogleCloudStorageTest.java
conker84: It's not necessary, I'll remove it

src/test/java/apoc/load/LoadGoogleCloudStorageTest.java
conker84: Unfortunately (as the result of my investigation) there is no Docker container or java library that allows testing the connection to a Google Storage Bucket. The only way to do that at this moment is to connect to a real bucket. So I used my private bucket to test it. I thought that we can use an env variable so into Travis we can easily inject a bucket.

jexp: ok

PR #1302, opened 2019-09-27
fixes #1288: Mongo connector chokes on ObjectId.

### LINKED ISSUE #1288: Mongo connector chokes on ObjectId.
I'd expect a Mongo ObjectId to be converted to a simple string.

apoc chokes on any ObjectId apart from the primary key "_id". 

Stick any ObjectId in a field that's not namned "_id".  See thread on forum 
https://community.neo4j.com/t/how-to-insert-objectids-from-mongodb-with-apoc

  - OS: OSX
  - Neo4j: 3.5.9
  - Neo4j-Apoc: 3.5.0.4

conker84:
Fixes #1288

The Mongo ObjectId is now managed as String.

 - ObjectId managed as String
 - added a new property `extractReferences` that recursively extract documents if finds `ObjectId` fields into a document (except for the `_id` field)
 - added support for new neo4j datatypes
 - added a round-trip test with `apoc.graph.fromDocument` in order to better test the `compatibleValues` with the new data-types

### DIFF
--- a/src/main/java/apoc/mongodb/MongoDBColl.java
+++ b/src/main/java/apoc/mongodb/MongoDBColl.java
@@ -1,23 +1,23 @@
 package apoc.mongodb;
 
-import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.mongodb.MongoClient;
 import com.mongodb.MongoClientURI;
 import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoCollection;
-import com.mongodb.client.MongoCursor;
 import com.mongodb.client.MongoDatabase;
 import com.mongodb.client.result.DeleteResult;
 import com.mongodb.client.result.UpdateResult;
 import org.bson.Document;
+import org.bson.types.Binary;
+import org.bson.types.ObjectId;
 
 import java.io.IOException;
-import java.util.List;
-import java.util.Map;
-import java.util.Spliterator;
-import java.util.Spliterators;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.*;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/mongodb/MongoDBColl.java
```diff
@@ -67,31 +74,76 @@ public void close() throws IOException {
...
+        }
+        if (data instanceof Map) {
+            Map<String, Object> map = (Map<String, Object>) data;
+            return map.entrySet().stream()
+                    .map(e -> {
+                        Object value;
+                        if (ID.equals(e.getKey())) { // avoid circular conversions
+                            if (compatibleValues && objectIdAsMap) {
+                                try {
+                                    value = jsonMapper.readValue(jsonMapper.writeValueAsBytes(e.getValue()), Map.class);
```
jexp: will we lose the object-id here? so that e.g. extractReference won't work anymore?
Or will it be a map of _id to object-id ? would probably good to add some more examples to the docs perhaps in a separate PR

conker84: That behavoiur is related on how we convert the `_id` field:

Follwing an example that could help to understand the behaviour:

Given the following collections:

```
// Product
...
{"_id": ObjectId("product1"), "name": "Product 1", "price": 100}
{"_id": ObjectId("product3"), "name": "Product 2", "price": 200}
{"_id": ObjectId("product3"), "name": "Product 3", "price": 300}
...
```

```
// Person
...
{"_id": ObjectId("person"), "name": "Andrea", "bought": [ObjectId("product1"), ObjectId("product3")]}
...
```

With the `extractReferences=true`, `compatibleValues=true` and `objectIdAsMap=false`:

```
{
  "_id": "person",
  "name": "Andrea",
  "bought": [
    {"_id": "product1", "name": "Product 1", "price": 100},
    {"_id": "product3", "name": "Product 3", "price": 300}
  ]
}
```


With the `extractReferences=true`, `compatibleValues=true` and `objectIdAsMap=true`:

```
{
  "_id": {
  	"timestamp": <...>,
	"machineIdentifier": <...>,
	"processIdentifier": <...>,
	"counter": <...>,
  },
  "name": "Andrea",
  "bought": [
    {
      "_id": {
	  	"timestamp": <...>,
		"machineIdentifier": <...>,
		"processIdentifier": <...>,
		"counter": <...>,
	  },
	  "name": "Product 1",
	  "price": 100
	},
    {
      "_id": {
	  	"timestamp": <...>,
		"machineIdentifier": <...>,
		"processIdentifier": <...>,
		"counter": <...>,
	  },
	  "name": "Product 3",
	  "price": 300
	},
  ]
}
```


With the `extractReferences=false`, `compatibleValues=false` and `objectIdAsMap=false`:

```
{
  "_id": "person",
  "name": "Andrea",
  "bought": ["product1", "product3"]
}
```

jexp: ok, thanks, can you add the example to the docs?

jexp:
I'm still torn with adding all the extra parameters.

conker84:
> I'm still torn with adding all the extra parameters. Vs. adding a config map.

I feel in the same way, I think that we should provide a new set of procedures (`apoc.mongo`?) with a config map, wdyt?

src/main/java/apoc/mongodb/MongoDBColl.java
conker84: sure

jexp:
Yes I felt the same. 


But we should deprecate these already now.

conker84:
@jexp documentation updated

conker84:
> But we should deprecate these already now

In this PR?

jexp:
The deprecation yes, the new methods, if you get to them quickly.

conker84:
I added the deprecation in this, let's do the new procedures in a separate PR, because I'd like also leverage the Mongo DB connection string that includes the db (now we are using a specific input param): 

https://docs.mongodb.com/manual/reference/connection-string/

I create the #1308 to track it

jexp:
Thank you Andrea!

PR #1307, opened 2019-10-02
apoc.datetime.parse like apoc.date.parse but returns a datetime

### LINKED ISSUE #1298: apoc.datetime.parse
A companion to `apoc.date.parse` that has the same format but instead of a number returns a datetime

karmakaze:
Fixes #1298

A companion to apoc.date.parse which returns a datetime rather than a Long.

  - new class apoc.datetime.DateTime
  - new method apoc.datetime.DateTime.parse which takes date/time string, format, and timezone (with defaults)
  - returns a ZonedDateTime

This preview PR just added the method into apoc.date.Date with a different function signature which will be moved to DateTime class and adding tests if this approach is acceptable.

It wasn't quite clear in the issue description what type of DateTime it should return. I'm not yet very familiar with Neo4j and couldn't see any other DateTime types to go by.

Also the implementation assumes Java8 or later.

### DIFF
--- a/src/main/java/apoc/date/Date.java
+++ b/src/main/java/apoc/date/Date.java
@@ -181,6 +181,13 @@ public class Date {
 	}
 
 	@UserFunction
+	@Description("apoc.date.parse('2012-12-23','yyyy-MM-dd') parse date string using the specified format at the specified timezone")
+	public ZonedDateTime parse(@Name("time") String time, @Name(value = "format",defaultValue = DEFAULT_FORMAT) String format, final @Name(value = "timezone", defaultValue = "UTC") String timezone) {
+		Long value = parseOrThrow(time, getFormat(format, timezone));
+		return value == null ? null : Instant.ofEpochMilli(value).atZone(ZoneId.of(timezone));
+	}
+
+	@UserFunction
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/date/Date.java
```diff
@@ -180,6 +180,13 @@ public Long parse(@Name("time") String time, @Name(value = "unit", defaultValue
 		return value == null ? null : unit(unit).convert(value, TimeUnit.MILLISECONDS);
 	}
 
+	@UserFunction
+	@Description("apoc.date.parse('2012-12-23','yyyy-MM-dd') parse date string using the specified format at the specified timezone")
```
conker84: @karmakaze Please, can you add the `timezone` to the signature?

jexp:
@karmakaze please also sign our CLA https://neo4j.com/developer/cla

Thank you!

conker84:
@karmakaze can you add test cases for the new procedure?
One with a call for the mandatory field, one with also the format and one also with the time zone.
Thanks.
Andrea

conker84:
@karmakaze and also update the docs, please!

conker84:
@karmakaze ping

conker84:
@karmakaze ping

PR #1310, opened 2019-10-05
fix #1304 removeRelProperties expects only 2 params + plural property…

### LINKED ISSUE #1304: removeRelProperties doesn't work
## Guidelines
I use the  named "CALL apoc.create.removeRelProperties( [rel,id,ids,rels], [keys], [values])"  procedures . The apoc's vesion is  "3.5.0.4".  The cypher is ：
```xml
match (n)-[serve:Serve]->(m) WITH serve CALL apoc.create.removeRelProperties(serve,['name']) YIELD rel RETURN rel
```
I expected that it can remove the name Propertity .
The name property doesn't remove it The property is available for the send time .

### Steps (Mandatory)

  1.
  1.
  1.

image
## Specifications (Mandatory)
For this problem,I  read the  document with linked "https://neo4j.com/docs/labs/apoc/current/graph-updates/data-creation/".I find that need three parameters 。So, I use the three parameters and the cypher follows:
```xml
match (n)-[serve:Serve]->(m) WITH serve CALL apoc.create.removeRelProperties(serve,['name'],['张三']) YIELD rel RETURN COUNT(rel) 
```
  Next,is responses that ,as  follows:
```xml
[Neo.ClientError.Statement.SyntaxError]:Procedure call does not provide the required number of arguments: got 3 expected 2.
Procedure apoc.create.removeRelProperties has signature: apoc.create.removeRelProperties(rels :: ANY?, keys :: LIST? OF STRING?) :: rel :: RELATIONSHIP?
meaning that it expects 2 arguments of type ANY?, LIST? OF STRING?
Description: apoc.create.removeRelProperties( [rel,id,ids,rels], [keys], [values]) - removes the given property from the relationship(s) (line 2, column 6 (offset: 45))
"     CALL apoc.create.removeRelProperties(serve,['name'],['张三']) "
```

Currently used versions

### Versions

  - OS: win10 enterprise
  - Neo4j:3.5
  - Neo4j-Apoc:3.5.0.4

vboulaye:
Fixes #1304

The removeRelProperties expect 2 parameters, but 3 are described in the doc

## Proposed Changes

  - removed the values parameter from the removeRelProperties description
  - changed the description of the *Properties procedure to apply the plural form there too
  -


jexp:
Please edit the asciidoc instead not the html as the html is generated.

vboulaye:
I guess I need to do it in the java @description too.

I did not understand properly where the doc did came from, sorry!

I'll redo the PR from the 3.5 branch.

PR #1311, opened 2019-10-05
fixes #1270 add optional recursive-parameter to map.removeKeys

### LINKED ISSUE #1270: add optional recursive-parameter to map.removeKeys
As we often need to export "deeply nested" maps as JSON within our application, it would be great to have recursion optionally inside the map.removeKeys() function.
Therefore we could suppress unnecessary attributes inside our queries and not within the API.

Kind regards,
Marc

vboulaye:
Fixes #1270 

add optional recursive-parameter to map.removeKeys

## Proposed Changes

  - added the "recursive" optional parameter with default set to false to the map.removeKey and map.removeKeys procedures (similar to what is done for coll.flatten)

### DIFF
--- a/src/main/java/apoc/map/Maps.java
+++ b/src/main/java/apoc/map/Maps.java
@@ -186,7 +186,7 @@ public class Maps {
         Map<String, Object> res = new LinkedHashMap<>(map);
         res.remove(key);
         Map<String, Object> checkedConfig = config == null ? Collections.emptyMap() : config;
-        if (checkedConfig.get("recursive") == Boolean.TRUE) {
+        if (Util.toBoolean(config.getOrDefault("recursive", false))) {
             for (Map.Entry<String, Object> entry : res.entrySet()) {
                 if (entry.getValue() instanceof Map) {
                     Map<String, Object> updatedMap = removeKey((Map<String, Object>) entry.getValue(), key, checkedConfig);
@@ -211,7 +211,7 @@ public class Maps {
         Map<String, Object> res = new LinkedHashMap<>(map);
         res.keySet().removeAll(keys);
         Map<String, Object> checkedConfig = config == null ? Collections.emptyMap() : config;
-        if (checkedConfig.get("recursive") == Boolean.TRUE) {
+        if (Util.toBoolean(config.getOrDefault("recursive", false))) {
             for (Map.Entry<String, Object> entry : res.entrySet()) {
                 if (entry.getValue() instanceof Map) {
                     Map<String, Object> updatedMap = removeKeys((Map<String, Object>) entry.getValue(), keys, checkedConfig);
--- a/src/main/java/apoc/map/Maps.java
+++ b/src/main/java/apoc/map/Maps.java
@@ -187,18 +187,26 @@ public class Maps {
         res.remove(key);
         Map<String, Object> checkedConfig = config == null ? Collections.emptyMap() : config;
         if (Util.toBoolean(config.getOrDefault("recursive", false))) {
-            for (Map.Entry<String, Object> entry : res.entrySet()) {
+            for (Iterator<Map.Entry<String, Object>> iterator = res.entrySet().iterator(); iterator.hasNext(); ) {
+                Map.Entry<String, Object> entry = iterator.next();
                 if (entry.getValue() instanceof Map) {
                     Map<String, Object> updatedMap = removeKey((Map<String, Object>) entry.getValue(), key, checkedConfig);
-                    if (updatedMap != entry.getValue()) {
+                    if (updatedMap.isEmpty()) {
+                        iterator.remove();
+                    } else if (!updatedMap.equals(entry.getValue())) {
                         entry.setValue(updatedMap);
                     }
                 } else if (entry.getValue() instanceof Collection) {
                     Collection<Object> values = (Collection<Object>) entry.getValue();
                     List<Object> updatedValues = values.stream()
                             .map(value -> value instanceof Map ? removeKey((Map<String, Object>) value, key, checkedConfig) : value)
+                            .filter(value -> value instanceof Map ? !((Map<String, Object>) value).isEmpty() : true)
                             .collect(Collectors.toList());
-                    entry.setValue(updatedValues);
+                    if (updatedValues.isEmpty()) {
+                        iterator.remove();
+                    } else  {
+                        entry.setValue(updatedValues);
+                    }
                 }
             }
         }
@@ -212,16 +220,26 @@ public class Maps {
         res.keySet().removeAll(keys);
         Map<String, Object> checkedConfig = config == null ? Collections.emptyMap() : config;
         if (Util.toBoolean(config.getOrDefault("recursive", false))) {
-            for (Map.Entry<String, Object> entry : res.entrySet()) {
+            for (Iterator<Map.Entry<String, Object>> iterator = res.entrySet().iterator(); iterator.hasNext(); ) {
+                Map.Entry<String, Object> entry = iterator.next();
                 if (entry.getValue() instanceof Map) {
                     Map<String, Object> updatedMap = removeKeys((Map<String, Object>) entry.getValue(), keys, checkedConfig);
-                    entry.setValue(updatedMap);
+                    if (updatedMap.isEmpty()) {
+                        iterator.remove();
+                    } else if (!updatedMap.equals(entry.getValue())) {
+                        entry.setValue(updatedMap);
+                    }
                 } else if (entry.getValue() instanceof Collection) {
                     Collection<Object> values = (Collection<Object>) entry.getValue();
                     List<Object> updatedValues = values.stream()
                             .map(value -> value instanceof Map ? removeKeys((Map<String, Object>) value, keys, checkedConfig) : value)
+                            .filter(value -> value instanceof Map ? !((Map<String, Object>) value).isEmpty() : true)
                             .collect(Collectors.toList());
-                    entry.setValue(updatedValues);
+                    if (updatedValues.isEmpty()) {
+                        iterator.remove();
+                    } else {
+                        entry.setValue(updatedValues);
+                    }
                 }
             }
         }
--- a/src/main/java/apoc/map/Maps.java
+++ b/src/main/java/apoc/map/Maps.java
@@ -185,16 +185,19 @@ public class Maps {
         }
         Map<String, Object> res = new LinkedHashMap<>(map);
         res.remove(key);
-        if (config!=null && config.get("recursive")==Boolean.TRUE) {
+        Map<String, Object> checkedConfig = config == null ? Collections.emptyMap() : config;
+        if (checkedConfig.get("recursive") == Boolean.TRUE) {
             for (Map.Entry<String, Object> entry : res.entrySet()) {
                 if (entry.getValue() instanceof Map) {
-                    Map<String, Object> updatedMap = removeKey((Map<String, Object>) entry.getValue(), key, config);
-                    if (updatedMap!=entry.getValue()) {
+                    Map<String, Object> updatedMap = removeKey((Map<String, Object>) entry.getValue(), key, checkedConfig);
+                    if (updatedMap != entry.getValue()) {
                         entry.setValue(updatedMap);
                     }
                 } else if (entry.getValue() instanceof Collection) {
-                    Collection<Map<String, Object>> values = (Collection<Map<String, Object>>) entry.getValue();
-                    List<Map<String, Object>> updatedValues = values.stream().map(value -> removeKey(value, key, config)).collect(Collectors.toList());
+                    Collection<Object> values = (Collection<Object>) entry.getValue();
+                    List<Object> updatedValues = values.stream()
+                            .map(value -> value instanceof Map ? removeKey((Map<String, Object>) value, key, checkedConfig) : value)
+                            .collect(Collectors.toList());
                     entry.setValue(updatedValues);
                 }
             }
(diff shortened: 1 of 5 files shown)


src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,34 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
     }
 
     @UserFunction
-    @Description("apoc.map.removeKey(map,key)")
-    public Map<String,Object> removeKey(@Name("map") Map<String,Object> map, @Name("key") String key) {
+    @Description("apoc.map.removeKey(map,key,recursive:false) - remove the key from the map (recursively if recursive is true)")
+    public Map<String,Object> removeKey(@Name("map") Map<String,Object> map, @Name("key") String key, @Name(value="recursive", defaultValue = "false") boolean recursive) {
         Map<String, Object> res = new LinkedHashMap<>(map);
```
jexp: should we check first if the current map contains the key before creating that new map?
otherwise we can just return the original map

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,34 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
         res.remove(key);
+        if (recursive) {
+            for (Map.Entry<String, Object> entry : res.entrySet()) {
+                if (entry.getValue() instanceof Map) {
+                    Map<String, Object> updatedMap = removeKey((Map<String, Object>) entry.getValue(), key, true);
```
jexp: here we can check if the updatedMap != entry.getValue() and only then (modified) update the value

vboulaye:
I added the checks you described in the comments.

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,39 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
     }
 
     @UserFunction
-    @Description("apoc.map.removeKey(map,key)")
-    public Map<String,Object> removeKey(@Name("map") Map<String,Object> map, @Name("key") String key) {
+    @Description("apoc.map.removeKey(map,key,recursive:false) - remove the key from the map (recursively if recursive is true)")
+    public Map<String,Object> removeKey(@Name("map") Map<String,Object> map, @Name("key") String key, @Name(value="recursive", defaultValue = "false") boolean recursive) {
```
conker84: Can you turn the `recursive` argument into a map `{recursive: true/false}`, so in the future we can add more properties like the recursion `depth`.
```suggestion
    public Map<String,Object> removeKey(@Name("map") Map<String,Object> map, @Name("key") String key, @Name(value="config", defaultValue = "{}") Map<String, Object> config) {
```

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,39 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+        if (!map.containsKey(key)) {
+            return map;
+        }
         Map<String, Object> res = new LinkedHashMap<>(map);
         res.remove(key);
+        if (recursive) {
+            for (Map.Entry<String, Object> entry : res.entrySet()) {
+                if (entry.getValue() instanceof Map) {
```
conker84: we should manage also collections, what if we have the following use-case:

```
RETURN apoc.map.removeKey({a:1,b:2,c:[{a:3,b:4}, {a:4,b:5}]},'a', true)
```

I expect that the result is:

```
{b:2,c:[{b:4}, {b:5}]}
```

Please add this as test-case

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,39 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
 
     @UserFunction
-    @Description("apoc.map.removeKeys(map,keys)")
-    public Map<String,Object> removeKeys(@Name("map") Map<String,Object> map, @Name("keys") List<String> keys) {
+    @Description("apoc.map.removeKeys(map,[keys],recursive:false) - remove the keys from the map (recursively if recursive is true)")
+    public Map<String,Object> removeKeys(@Name("map") Map<String,Object> map, @Name("keys") List<String> keys, @Name(value="recursive", defaultValue = "false") boolean recursive) {
         Map<String, Object> res = new LinkedHashMap<>(map);
         res.keySet().removeAll(keys);
+        if (recursive) {
+            for (Map.Entry<String, Object> entry : res.entrySet()) {
```
conker84: also here the collection management

conker84:
@vboulaye please check my comments

src/main/java/apoc/map/Maps.java
conker84: the same for `removeKeys`

vboulaye:
I replaced the flag by a config map and added the collection processing

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,47 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
     }
 
     @UserFunction
-    @Description("apoc.map.removeKeys(map,keys)")
-    public Map<String,Object> removeKeys(@Name("map") Map<String,Object> map, @Name("keys") List<String> keys) {
+    @Description("apoc.map.removeKeys(map,[keys],{recursive:true/false}) - remove the keys from the map (recursively if recursive is true)")
+    public Map<String,Object> removeKeys(@Name("map") Map<String,Object> map, @Name("keys") List<String> keys, @Name(value="config", defaultValue = "{}") Map<String, Object> config) {
         Map<String, Object> res = new LinkedHashMap<>(map);
         res.keySet().removeAll(keys);
+        if (config!=null && config.get("recursive")==Boolean.TRUE) {
```
conker84: can you add a null check over the config?
```suggestion
        if (config!=null) {
             config = Collections.emptyMap();
        }
```

And then:
```
if (Util.toBoolean(config.getOrDefault("recursive", false))) {
....
}
```

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,47 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+            for (Map.Entry<String, Object> entry : res.entrySet()) {
+                if (entry.getValue() instanceof Map) {
+                    Map<String, Object> updatedMap = removeKeys((Map<String, Object>) entry.getValue(), keys, config);
```
conker84: what happens if the map is empty?

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,47 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+                    entry.setValue(updatedMap);
+                } else if (entry.getValue() instanceof Collection) {
+                    Collection<Map<String, Object>> values = (Collection<Map<String, Object>>) entry.getValue();
```
conker84: The cast assumes that every collection is a collection of Maps but what about strings or primitives?
Please add test for that

src/main/java/apoc/map/Maps.java
conker84: `if (config == null) {`

conker84:
@vboulaye please I added a bunch of other comments.
Thanks a lot

src/main/java/apoc/map/Maps.java
vboulaye: ok

src/main/java/apoc/map/Maps.java
vboulaye: it returns the empty map. this was the existing behavior before adding the recursive flag.
I added tests that describe this.

src/main/java/apoc/map/Maps.java
vboulaye: ok, I check the type now, and added  some unit tests

vboulaye:
Hi, I included the requested changes.
regarding the empty maps, I just added some tests: do you want empty maps to be removed completely ?

conker84:
> Hi, I included the requested changes.
> regarding the empty maps, I just added some tests: do you want empty maps to be removed completely ?

great question, I think that we should be compliant with the current behaviour for simple properties, so we should remove completely the property.

vboulaye:
Hi,
the current behavior (without the recursive flag set) is to return an empty map when the last property is removed: 
should I change this to return null when there is nothing left ? 
or should I only remove empty  map when processing child map/collections ?

conker84:
> Hi,
> the current behavior (without the recursive flag set) is to return an empty map when the last property is removed:
> should I change this to return null when there is nothing left ?
> or should I only remove empty map when processing child map/collections ?

I mean the current behaviour for simple properties `{a: 1, b: 2} = remove 'b' => {a: 1}`, please remove the property

vboulaye:
I already have a test that checks that `removeKey({a:1,b:2,c:{a:3,b:4}},'a', {recursive:true})` returns ̀{b:2,c:[b:4}}`
I added your test case, but I think it works already like that.

jexp:
Looks good. Would you mind removing the html files from the PR? As they are generated from the asciidoc (and should probably not be in the repo in the first place)

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,53 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+            return map;
+        }
         Map<String, Object> res = new LinkedHashMap<>(map);
         res.remove(key);
+        Map<String, Object> checkedConfig = config == null ? Collections.emptyMap() : config;
+        if (Util.toBoolean(config.getOrDefault("recursive", false))) {
+            for (Map.Entry<String, Object> entry : res.entrySet()) {
+                if (entry.getValue() instanceof Map) {
+                    Map<String, Object> updatedMap = removeKey((Map<String, Object>) entry.getValue(), key, checkedConfig);
+                    if (updatedMap != entry.getValue()) {
```
conker84: Use !equals

conker84:
```suggestion
                    if (!updatedMap.equals(entry.getValue())) {
```

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,53 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+                        entry.setValue(updatedMap);
```
conker84: if updatedMap is empty remove the key

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,53 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+        Map<String, Object> checkedConfig = config == null ? Collections.emptyMap() : config;
+        if (Util.toBoolean(config.getOrDefault("recursive", false))) {
+            for (Map.Entry<String, Object> entry : res.entrySet()) {
+                if (entry.getValue() instanceof Map) {
+                    Map<String, Object> updatedMap = removeKeys((Map<String, Object>) entry.getValue(), keys, checkedConfig);
+                    entry.setValue(updatedMap);
+                } else if (entry.getValue() instanceof Collection) {
+                    Collection<Object> values = (Collection<Object>) entry.getValue();
+                    List<Object> updatedValues = values.stream()
+                            .map(value -> value instanceof Map ? removeKeys((Map<String, Object>) value, keys, checkedConfig) : value)
```
conker84: add a filter that excludes empty maps

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,53 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+                            .collect(Collectors.toList());
+                    entry.setValue(updatedValues);
```
conker84: remove the entry if the array is empty

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,53 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+    @Description("apoc.map.removeKeys(map,[keys],{recursive:true/false}) - remove the keys from the map (recursively if recursive is true)")
+    public Map<String, Object> removeKeys(@Name("map") Map<String, Object> map, @Name("keys") List<String> keys, @Name(value = "config", defaultValue = "{}") Map<String, Object> config) {
         Map<String, Object> res = new LinkedHashMap<>(map);
         res.keySet().removeAll(keys);
+        Map<String, Object> checkedConfig = config == null ? Collections.emptyMap() : config;
+        if (Util.toBoolean(config.getOrDefault("recursive", false))) {
+            for (Map.Entry<String, Object> entry : res.entrySet()) {
+                if (entry.getValue() instanceof Map) {
+                    Map<String, Object> updatedMap = removeKeys((Map<String, Object>) entry.getValue(), keys, checkedConfig);
+                    entry.setValue(updatedMap);
```
conker84: if updatedMap is empty remove the key

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,53 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+            for (Map.Entry<String, Object> entry : res.entrySet()) {
+                if (entry.getValue() instanceof Map) {
+                    Map<String, Object> updatedMap = removeKey((Map<String, Object>) entry.getValue(), key, checkedConfig);
+                    if (updatedMap != entry.getValue()) {
+                        entry.setValue(updatedMap);
+                    }
+                } else if (entry.getValue() instanceof Collection) {
+                    Collection<Object> values = (Collection<Object>) entry.getValue();
+                    List<Object> updatedValues = values.stream()
+                            .map(value -> value instanceof Map ? removeKey((Map<String, Object>) value, key, checkedConfig) : value)
```
conker84: add a filter that excludes empty maps

src/main/java/apoc/map/Maps.java
```diff
@@ -178,18 +178,53 @@ public Object get(@Name("map") Map<String,Object> map, @Name("key") String key,
...
+                            .collect(Collectors.toList());
+                    entry.setValue(updatedValues);
```
conker84: remove the entry if the array is empty

src/main/java/apoc/map/Maps.java
vboulaye: ok

src/main/java/apoc/map/Maps.java
vboulaye: ok

src/main/java/apoc/map/Maps.java
vboulaye: ok

src/main/java/apoc/map/Maps.java
vboulaye: ok

vboulaye:
Hi,
I added the filtering + removed the html updates

conker84:
LGTM
@vboulaye thanks a lot!

PR #1313, opened 2019-10-06
fixes #151 coll.isEqualCollection compare 2 collections in any order

### LINKED ISSUE #151: Feature Request: expose isEqualCollection to cypher
See: [CollectionUtils#isEqualCollection](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#isEqualCollection%28java.util.Collection,%20java.util.Collection%29)

This would be very useful to check if two collections are the same. As far as I'm aware this isn't exposed to cypher.

vboulaye:
Fixes #151

This is an attempt at implementing the new function coll.isEqualCollection that returns true if the two collections contain the same elements with the same cardinality in any order

  - new UserFunction coll.isEqualCollection, internally uses an HashMap of object => cardinality to keep track of the cardinality of each element

### DIFF
--- a/src/main/java/apoc/coll/Coll.java
+++ b/src/main/java/apoc/coll/Coll.java
@@ -32,18 +32,18 @@ public class Coll {
     public Stream<ListResult> zipToRows(@Name("list1") List<Object> list1, @Name("list2") List<Object> list2) {
         if (list1.isEmpty()) return Stream.empty();
         ListIterator<Object> it = list2.listIterator();
-        return list1.stream().map((e) -> new ListResult(asList(e,it.hasNext() ? it.next() : null)));
+        return list1.stream().map((e) -> new ListResult(asList(e, it.hasNext() ? it.next() : null)));
     }
 
     @UserFunction
     @Description("apoc.coll.zip([list1],[list2])")
     public List<List<Object>> zip(@Name("list1") List<Object> list1, @Name("list2") List<Object> list2) {
-		if (list1 == null || list2 == null) return null;
-		if (list1.isEmpty() || list2.isEmpty()) return Collections.emptyList();
+        if (list1 == null || list2 == null) return null;
+        if (list1.isEmpty() || list2.isEmpty()) return Collections.emptyList();
         List<List<Object>> result = new ArrayList<>(list1.size());
         ListIterator it = list2.listIterator();
         for (Object o1 : list1) {
-            result.add(asList(o1,it.hasNext() ? it.next() : null));
+            result.add(asList(o1, it.hasNext() ? it.next() : null));
         }
         return result;
     }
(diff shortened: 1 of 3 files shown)


conker84:
@vboulaye can you push the `Coll` class with only your changes? It's difficult to review it with all the formatting stuff.
Thanks a lot.

vboulaye:
sorry about that.
I didn't see it because I ignore all blanks in the diff viewer in intellij !

src/main/java/apoc/coll/Coll.java
```diff
@@ -431,6 +431,30 @@ public boolean containsAllSorted(@Name("coll") List<Object> coll, @Name("values"
         return true;
     }
 
+    @UserFunction
+    @Description("apoc.coll.isEqualCollection(coll, values) return true if two collections contain the same elements with the same cardinality in any order (using a HashMap)")
```
conker84: Can we simplify this? The implementation is quite complex, I think that something like this should do the work:
```
        Map<Object, Long> map1 = first.stream()
                .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
        Map<Object, Long> map2 = second.stream()
                .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
        return map1.equals(map2);
```

vboulaye: this is simpler indeed!

src/main/java/apoc/coll/Coll.java
```diff
@@ -431,6 +432,17 @@ public boolean containsAllSorted(@Name("coll") List<Object> coll, @Name("values"
...
+    public boolean isEqualCollection(@Name("coll") List<Object> first, @Name("values") List<Object> second) {
+        if (first == null || first.isEmpty() || first.size() != second.size()) return false;
```
conker84: We must manage the null check for the second otherwise we get a NullPointerException at line 442. 
I think that we can also get rid of `first.size() != second.size()` because we have the count comparison for each collection element.

Please also add a

vboulaye: I did not do the null check on second because it was not done on the existing function I started from `containsAll` . Should I heck it there too ?
I added some tests for the various null/empty values.
I kept the starting size check as it allows exiting early without having to create the intermediate Maps, but if you prefer I'll remove it.

conker84: In that case makes sense because:
1. there is a null check on `coll`
2. the `call.containsAll` method provides a null check internally.
Please add the null check.

src/main/java/apoc/coll/Coll.java
```diff
@@ -431,6 +432,18 @@ public boolean containsAllSorted(@Name("coll") List<Object> coll, @Name("values"
         return true;
     }
 
+    @UserFunction
+    @Description("apoc.coll.isEqualCollection(coll, values) return true if two collections contain the same elements with the same cardinality in any order (using a HashMap)")
+    public boolean isEqualCollection(@Name("coll") List<Object> first, @Name("values") List<Object> second) {
+        if (first == second) return true;
```
conker84: @vboulaye please remove this line or use `equals` after the null check

conker84:
@vboulaye please look at my last comment and we're done!
Thanks a lot for committing on this!

conker84:
@vboulaye ping

src/main/java/apoc/coll/Coll.java
vboulaye: the problem is that if I remove this test, I have to add an explicit test to handle the case when both entries are null:  I believe  ̀apoc.coll.isEqualCollection(null,null)`  should return true .
I added 
```
        if (first == null && second == null) return true;
```
but I am not sure it is much better!

src/main/java/apoc/coll/Coll.java
vboulaye: I could not find the check on call.containsAll:I added a test and `apoc.coll.containsAll([1,2,3],null) ` throws a NullPointerException

src/main/java/apoc/coll/Coll.java
vboulaye: I just saw that in containsAllSorted the null check is done  with   if (coll == null || values == null) return false;
so containsAllSorted(null,null) is false

maybe this should be the same for  isEqualCollection ?

conker84:
LGTM,
@vboulaye do you signed the CLA https://neo4j.com/developer/cla ?

PR #1320, opened 2019-10-22
fixes #1268: apoc.periodic.repeat doesn't provide any feedback when query is bad

### LINKED ISSUE #1268: apoc.periodic.repeat doesn't provide any feedback when query is bad
I’m doing this in cypher.  What I think this is doing is re-executing my cypher every 2 min (120 seconds).   APOC isn’t doing that though - if I do apoc.periodic.list() immediately after submitting this I can see that my job is there - but in under 5 seconds it disappears and no ongoing execution happens.   

```
CALL apoc.periodic.repeat(
    'detectCommunities',
    '   MATCH (p:Party) where p.louvainCommunity IS NOT NULL
        SET p.louvainCommunity = null
        WITH count(p) as erased 
        CALL apoc.periodic.iterate("
            CALL algo.louvain.stream(\'Party\', \'ASSOCIATED\', 
               {concurrency:8, graph:\'heavy\'}) 
            YIELD nodeId, community", 
        "MATCH (n) where id(n) = nodeId set n.louvainCommunity = community",
        {batchSize:2000, parallel:false})
    ',
    120)
```

There's nothing in the logs to tell why this happens and no feedback - but the reaosn the query is failing is because apoc.periodic.iterate inside doesn't have a YIELD and RETURN clause.

The specifics of the interior query don't matter, the issue is just that apoc.periodic.repeat will swallow all of this and the user will think it did something when it didn't, because the query was bad.

conker84:
Fixes #1268
Added query validation in each `apoc.periodic.*` procedure

  - Added the validation in each procedure
  - Added test for each procedure/arguments

### DIFF
--- a/src/main/java/apoc/periodic/Periodic.java
+++ b/src/main/java/apoc/periodic/Periodic.java
@@ -56,6 +56,7 @@ public class Periodic {
     @Procedure(mode = Mode.WRITE)
     @Description("apoc.periodic.commit(statement,params) - runs the given statement in separate transactions until it returns 0")
     public Stream<RundownResult> commit(@Name("statement") String statement, @Name(value = "params", defaultValue = "") Map<String,Object> parameters) throws ExecutionException, InterruptedException {
+        validateQuery(statement);
         Map<String,Object> params = parameters == null ? Collections.emptyMap() : parameters;
         long total = 0, executions = 0, updates = 0;
         long start = nanoTime();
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/periodic/Periodic.java
```diff
@@ -227,7 +235,9 @@ public static JobInfo schedule(String name, Runnable task, long delay, long repe
             @Name("cypherIterate") String cypherIterate,
             @Name("cypherAction") String cypherAction,
             @Name("batchSize") long batchSize) {
-
+        validateQuery(cypherLoop);
```
jexp: for the places where we have several checks we should collect the errors and return them together.
so perhaps a varags parameter and iterate over them and collecting the errors per statement

conker84: done

PR #1322, opened 2019-10-23
fix 1184

### LINKED ISSUE #1184: apoc.export.csv.query() does not register duration.inSeconds formatting
## Guidelines

`CALL apoc.export.csv.query(" ... ... RETURN duration.inSeconds(a.timeStamp, b.timeStamp) ..." )`
should return "P0M0DT2651S" in the csv file (only seconds)

wrongly returns "PT44M11S" in csv file (minutes and seconds)
(when queried directly on Neo4j, it correctly returns "P0M0DT930S" on the dashboard)

### Simple Dataset (where it's possible)

```

```

### Steps (Mandatory)

  1.
  1.
  1.

## Screenshots (where it's possible)

Screen Shot 2019-04-28 at 11 31 00 AM

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS:
  - Neo4j:
  - Neo4j-Apoc:

enriquezrene:
Fixes #1184
Adding a format method for DurationValue data types

  - Add a new method to format DurationValue data types

### DIFF
--- a/src/main/java/apoc/export/util/FormatUtils.java
+++ b/src/main/java/apoc/export/util/FormatUtils.java
@@ -5,9 +5,11 @@ import apoc.util.Util;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import org.neo4j.graphdb.*;
 import org.neo4j.graphdb.spatial.Point;
+import org.neo4j.values.storable.DurationValue;
 
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


jexp:
Perhaps rather add an separate user defined function for your output-format and then you can use that during your export?

enriquezrene:
Hey @conker84 you're right, but the statement is asking for a value represented using seconds only.
@jexp I think that's why `duration.inSeconds` was added in the statement, or do you think the user should provide any other metadata for it?

jexp:
@enriquezrene ah I see what you're saying sorry I overlooked that.
That then seems to be an issue in Neo4j itself that we have to differentiate with the duration values.

jexp:
@conker84 what's the state on this PR?

fbiville:
Closing this as we have not heard from the author. Feel free to reopen if needed.

PR #1318, opened 2019-10-16
WIP: Tables for labels metadata, mirroring CALL db.schema.(nodeTypeProperties|relTypeProperties)

moxious:
Implement in APOC a version of `db.schema.nodeTypeProperties()` and `db.schema.relTypeProperties()` that permit flexible probabilistic sampling, rather than the full DB scan that those others do.

These procedures are key because they provide metadata output in a very specific format that permits the "Tables for Labels" mapping.   T4L can be thought of as a basic/naive Graph->Tables mapping.  This has lots of uses, the most particular one I have in mind is for use with the BI driver that's in progress.   Tools can call these functions to build a virtual RDBMS schema over top of Neo4j which can be queried, basically.

The trouble with the product built-in procedures is that they don't use sampling, they scan the entire DB, so for large DBs this will churn the page cache and perform very badly.

This approach has been discussed with Mats & the internal Morpheus team who wrote the product versions of the stored procedures, and has generally been agreed on as the best available approach.

**This is just a code spike right now for visibility**.

A design document describing how this is supposed to work and what behavior should be is here: https://docs.google.com/document/d/1-U-L5anu50CGHXYipjdj5AwdeFprd7lvggD97fbqHeA/edit?usp=sharing

PR #1324, opened 2019-10-30
[Waiting for David feedback] fixes #1319: apoc.meta.data() reports wrong property type information

### LINKED ISSUE #1319: apoc.meta.data() reports wrong property type information
Replication that worked for me:

```
create (:A { x: 1 }), (:A { x: 'Foo' });
create (:B { x: 'Foo' }), (:B { x: 'Foo' }), (:B { x: 1 });
CALL apoc.meta.data();
```

This setup is intentionally to confuse APOC.   For both labels A and B, the property X is typed both string and long.

The APOC call:

Screen Shot 2019-10-17 at 2 48 47 PM

This procedure uses a sampling configuration as specified by the config parameters, which is a MetaConfig object.  The sampling appears respected, but the property type information stored in  the MetaResult class is "last type in wins".  

I would say this behavior is incorrect.  There are two possible correct behaviors:

1. Return an array of different types the property could have, as observed in the sample ["Long", "String"]
2. Perform some type-coercion, and always report "String" as a single property type, because it is the widest of all observed types.

conker84:
Fixes #1319

Added new `types` field that provides a list of the sampled field types, if the list size is > 1 the `type` field is coerced to `STRING`
 
  - added `types` field in `apoc.meta.data`
  - added `types` field in `apoc.meta.schema`

### DIFF
--- a/src/main/java/apoc/meta/Meta.java
+++ b/src/main/java/apoc/meta/Meta.java
@@ -134,6 +134,7 @@ public class Meta {
         public List<String> other = new ArrayList<>();
         public List<String> otherLabels = new ArrayList<>();
         public String elementType;
+        public List<String> types = new ArrayList<>();
 
         public MetaResult addLabel(String label) {
             this.otherLabels.add(label);
@@ -167,7 +168,17 @@ public class Meta {
         }
 
         public MetaResult type(String type) {
-            this.type = type;
+            if (type == null || type.equals(this.type)) {
+                return this;
+            }
+            if (!this.types.contains(type)) {
+                this.types.add(type);
+            }
+            if (!type.equals(this.type) && this.type != null) {
+                this.type = Types.STRING.toString();
+            } else {
+                this.type = type;
+            }
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


conker84:
> Ultimately a different solution, adopted by (#1318) is to list every type encountered in the sampling process.

In the `types` field we have every type encountered in the sampling process.

> I think the right approach for this is to have a type lattice describing these "type domination relationships.

Do you think can we leave for the `type` field the behaviour as-is (so the last type wins)? And then in a future PR create the "type domination relationships"

src/main/java/apoc/meta/Meta.java
```diff
@@ -167,7 +168,17 @@ public MetaResult other(List<String> labels) {
...
         public MetaResult type(String type) {
-            this.type = type;
+            if (type == null || type.equals(this.type)) {
+                return this;
+            }
+            if (!this.types.contains(type)) {
+                this.types.add(type);
+            }
+            if (!type.equals(this.type) && this.type != null) {
+                this.type = Types.STRING.toString();
```
jexp: perhaps it's better to set it null then?
like "undefined" if we use STRING we also need to add STRING to the list?
something else that might make sense if we have INTEGER and FLOAT to coerce to FLOAT

conker84: But what if we have a field which is a `timestamp` for some values and `localdate` for others? Maybe we can keep the current behaviour as-is as `deprecation` and highlight that in the docs

Issue #1338, opened 2019-11-09
String multiplication function apoc.text.repeat


tomasonjo:
We would like to add string multiplication option in APOC, similar to python code `"a"*3 `, that returns "aaa"

Example

`return apoc.text.repeat("a",3)`

returns "aaa"

PR #1340, opened 2019-11-09
Add string mulitiplication apoc.text.repeat

tomasonjo:
Fixes https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1338

Added string multiplication with `apoc.text.repeat(item,count)`

### DIFF
--- a/src/main/java/apoc/text/Strings.java
+++ b/src/main/java/apoc/text/Strings.java
@@ -576,7 +576,7 @@ public class Strings {
 
     @UserFunction
     @Description("apoc.text.repeat(item, count) - string multiplication")
-    public StringBuilder repeat(@Name("item") String item, @Name("count") long count) {
+    public String repeat(@Name("item") String item, @Name("count") long count) {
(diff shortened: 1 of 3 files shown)


tomasonjo:
Have to fix the test error :)

tomasonjo:
IMO is ok now.

src/main/java/apoc/text/Strings.java
```diff
@@ -573,4 +573,10 @@ public String toCypher(@Name("value") Object value, @Name(value = "config",defau
         }
         return null;
     }
+
+    @UserFunction
+    @Description("apoc.text.repeat(item, count) - string multiplication")
+    public String repeat(@Name("item") String item, @Name("count") long count) {
+        return new String(new char[(int) count]).replace("\0", item);
```
jexp: better just use a pre-sized StringBuilder and a loop with append.
as the item might be longer than length = 1

tomasonjo:
switched to stringbuilder

Issue #1339, opened 2019-11-09
populating arrays with apoc.coll.fill(item, count)


tomasonjo:
To populate an array with a predefined number of the same element we could use:

`apoc.coll.fill(item, count)`

Example:

`return apoc.coll.fill("a", 3)`

returns ["a","a","a"]

PR #1341, opened 2019-11-09
Add apoc.coll.fill function

tomasonjo:
Fixes https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1339

Added apoc.coll.fill function

Proposed Changes

apoc.coll.fill(item, count) function returns a list with the "count" number of items

Example

return apoc.coll.fill('ab', 2)

returns ['ab', 'ab']

### DIFF
--- a/src/main/java/apoc/coll/Coll.java
+++ b/src/main/java/apoc/coll/Coll.java
@@ -891,7 +891,6 @@ public class Coll {
     @UserFunction
     @Description("apoc.coll.fill(item, count) - returns a list with the given count of items")
     public List<Object> fill(@Name("item") String item, @Name("count") long count) {
-        List<Object> newList = new ArrayList<Object>(Collections.nCopies((int) count, item));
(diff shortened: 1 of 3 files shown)


src/main/java/apoc/coll/Coll.java
```diff
@@ -887,4 +887,11 @@ public boolean different(@Name("values") List<Object> values) {
 
         return newList;
     }
+
+    @UserFunction
+    @Description("apoc.coll.fill(item, count) - returns a list with the given count of items")
+    public List<Object> fill(@Name("item") String item, @Name("count") long count) {
+        List<Object> newList = new ArrayList<Object>(Collections.nCopies((int) count, item));
```
jexp: no need for the new ArrayList() just return the result of nCopies

tomasonjo: Fixed

PR #1344, opened 2019-11-12
Recreate index schema

mneedham:
This procedure will drop and recreate all indexes and constraints in the database. It's useful if you want to stop using legacy index providers, but don't want to manually drop and create them

Paired on with @JMHReif


src/main/java/apoc/schema/ExceptionUtil.java
```diff
@@ -0,0 +1,70 @@
...
+     * @return the root cause.
+     * @throws IllegalArgumentException if the provided exception is null.
+     */
+    public static Throwable rootCause(Throwable caughtException) {
+        if (null == caughtException) {
+            throw new IllegalArgumentException("Cannot obtain rootCause from (null)");
+        }
+        Throwable root = caughtException;
+        while (root.getCause() != null) {
+            root = root.getCause();
```
jexp: cause might be the exception it self so this might end up in an endless loop.

src/main/java/apoc/schema/Schemas.java
```diff
@@ -36,6 +36,23 @@
     @Context
     public KernelTransaction tx;
 
+    @Procedure(value = "apoc.schema.recreate", mode = Mode.SCHEMA)
+    @Description("apoc.schema.recreate() - recreates indexes and constraints")
+    public Stream<AssertSchemaResult> schemaRecreate(@Name(value = "config", defaultValue = "{}") Map<String, Object> config) throws ExecutionException, InterruptedException, TimeoutException {
+        long allIndexesTimeout = (long) config.getOrDefault("retrieveIndexesTimeout", 10000L);
```
jexp: the timeouts are bit low
esp. on large graphs those can be hours, so I would increase them massively (which doesn't hurt)

jexp: where do the current numbers come from?

src/main/java/apoc/schema/Schemas.java
```diff
@@ -36,6 +36,23 @@
...
+        long dropAllIndexesTimeout =  (long) config.getOrDefault("dropAllIndexesTimeout", 30000L);
+        long createAllIndexesTimeout = (long) config.getOrDefault("createAllIndexesTimeout", 30000L);
+
+        long allConstraintsTimeout = (long) config.getOrDefault("retrieveConstraintsTimeout", 10000L);
+        long dropAllConstraintsTimeout =  (long) config.getOrDefault("dropAllConstraintsTimeout", 30000L);
+        long createAllConstraintsTimeout = (long) config.getOrDefault("createAllConstraintsTimeout", 30000L);
```
jexp: we can just have a config object instead of individual timeout

src/main/java/apoc/schema/Schemas.java
```diff
@@ -117,6 +135,101 @@ private AssertSchemaResult createUniqueConstraint(Schema schema, String lbl, Str
         return new AssertSchemaResult(lbl, key).unique().created();
     }
 
+    public List<AssertSchemaResult> recreateIndexes(long allIndexesTimeout, long dropAllIndexesTimeout, long recreateAllIndexesTimeout) throws ExecutionException, InterruptedException, IllegalArgumentException, TimeoutException {
+        Schema schema = db.schema();
+        List<AssertSchemaResult> result = Collections.synchronizedList(new ArrayList<>());
+        ExecutorService executorService = Pools.SINGLE;
```
jexp: why not Pools.DEFAULT?

src/main/java/apoc/schema/Schemas.java
```diff
@@ -117,6 +135,101 @@ private AssertSchemaResult createUniqueConstraint(Schema schema, String lbl, Str
         return new AssertSchemaResult(lbl, key).unique().created();
     }
 
+    public List<AssertSchemaResult> recreateIndexes(long allIndexesTimeout, long dropAllIndexesTimeout, long recreateAllIndexesTimeout) throws ExecutionException, InterruptedException, IllegalArgumentException, TimeoutException {
+        Schema schema = db.schema();
+        List<AssertSchemaResult> result = Collections.synchronizedList(new ArrayList<>());
```
jexp: we can just use the results of the future of the 2nd call.
no need for synchronized collection to be passed in

src/main/java/apoc/schema/Schemas.java
```diff
@@ -117,6 +135,101 @@ private AssertSchemaResult createUniqueConstraint(Schema schema, String lbl, Str
...
+        ExecutorService executorService = Pools.SINGLE;
+        List<IndexDefinition> indexesToRecreate = Util.inTxFuture(executorService, db, () -> {
+            List<IndexDefinition> idx = new ArrayList<>();
+            Iterable<IndexDefinition> outerIndexes = schema.getIndexes();
```
jexp: not sure if we need to pull the schema from the db in the separate tx?

jexp: does it have to be in a separate tx just reading them?

src/main/java/apoc/schema/Schemas.java
mneedham: I don't think the operations are synchronous though. The actual command to drop or create should happen straight away I think? And then in the background there would be work going on in an async fashion

src/main/java/apoc/schema/Schemas.java
mneedham: It does! This line:

```
schema.getIndexes();
```

Takes a lock on all the indexes and it's only released when the transaction is closed. So you end up in a deadlock if you do that in the procedure's implicit transaction as it would still be holding the lock when the other thread is trying to drop it

src/main/java/apoc/schema/ExceptionUtil.java
mneedham: Sorry this was added by mistake. Copied across from algos

src/main/java/apoc/schema/Schemas.java
mneedham: I'm not sure what a config object is - could you elaborate?

src/main/java/apoc/schema/Schemas.java
mneedham: Wanted it to be single threaded. As it is it woudn't make much difference as do blocking futures anyway

src/main/java/apoc/schema/Schemas.java
mneedham: Is your suggestion that each call builds up a `List<AssertSchemaResult>` and then we combine them together afterwards?

src/main/java/apoc/schema/Schemas.java
jexp: right for indexes, for constraints it blocks

jexp: sorry I was referring to db.awaitIndexes() timeouts

src/main/java/apoc/schema/Schemas.java
jexp: just a object that has all those timeouts (encapsulate parameters)

src/main/java/apoc/schema/Schemas.java
jexp: ok

src/main/java/apoc/schema/Schemas.java
jexp: yes

src/main/java/apoc/schema/Schemas.java
jexp: I meant I was not sure if in the separate tx we actually have to call `db.schema()...` there to have the right tx connected to the schema object.

src/main/java/apoc/schema/Schemas.java
jexp: just a regular java object that we can pass to the methods.

src/main/java/apoc/schema/Schemas.java
```diff
@@ -117,6 +135,101 @@ private AssertSchemaResult createUniqueConstraint(Schema schema, String lbl, Str
...
+                info.dropped();
+                result.add(info);
+            }
+            return null;
+        }).get(dropAllIndexesTimeout,TimeUnit.MILLISECONDS);
+
+        Util.inTxFuture(executorService, db, () -> {
+            for (IndexDefinition definition : indexesToRecreate) {
+                List<String> propertyKeys = new ArrayList<>();
+                for (String propertyKey : definition.getPropertyKeys()) {
```
jexp: little bit inconsistent with above :)

jexp: And isn't that info already in AssertSchema result? so we can just use that?

src/main/java/apoc/schema/Schemas.java
```diff
@@ -117,6 +135,101 @@ private AssertSchemaResult createUniqueConstraint(Schema schema, String lbl, Str
...
+            return idx;
+        }).get(allIndexesTimeout, TimeUnit.MILLISECONDS);
+
+        Util.inTxFuture(executorService, db, () -> {
+            for (IndexDefinition definition : indexesToRecreate) {
+                String label = definition.getLabel().name();
+                List<String> keys = new ArrayList<>();
+                definition.getPropertyKeys().forEach(keys::add);
+
+                AssertSchemaResult info = new AssertSchemaResult(label, keys);
```
jexp: should also contain the type of index/constraint

jexp: and compound vs. single

src/main/java/apoc/schema/Schemas.java
```diff
@@ -117,6 +135,101 @@ private AssertSchemaResult createUniqueConstraint(Schema schema, String lbl, Str
...
+            for (ConstraintDefinition index : outerIndexes) {
+                idx.add(index);
+            }
+            return idx;
+        }).get(allConstraintsTimeout, TimeUnit.MILLISECONDS);
+
+        Util.inTxFuture(executorService, db, () -> {
+            for (ConstraintDefinition constraint : constraintsToRecreate) {
+                String label = constraint.isConstraintType(ConstraintType.RELATIONSHIP_PROPERTY_EXISTENCE) ? constraint.getRelationshipType().name() : constraint.getLabel().name();
+                AssertSchemaResult info = new AssertSchemaResult(label, Iterables.asList(constraint.getPropertyKeys())).unique();
```
jexp: third variant of transforming `Iterables.asList(constraint.getPropertyKeys())` into a List :) we should pick one and stay consistent.

src/main/java/apoc/schema/Schemas.java
```diff
@@ -170,7 +283,7 @@ private AssertSchemaResult createSinglePropertyIndex(Schema schema, String lbl,
 
     private AssertSchemaResult createCompoundIndex(String label, List<String> keys) {
         List<String> backTickedKeys = new ArrayList<>();
-        keys.forEach(key->backTickedKeys.add(String.format("`%s`", key)));
+        keys.forEach(key -> backTickedKeys.add(String.format("`%s`", key)));
```
jexp: should be just Util.quote

src/main/java/apoc/schema/Schemas.java
```diff
@@ -296,15 +409,15 @@ private Boolean constraintsExistsForRelationship(String type, List<String> prope
...
                 Iterable<ConstraintDescriptor> allConstraints = () -> schemaRead.constraintsGetAll();
-                constraintsIterator = StreamSupport.stream(allConstraints.spliterator(),false)
+                constraintsIterator = StreamSupport.stream(allConstraints.spliterator(), false)
                         .filter(constraint -> Arrays.stream(constraint.schema().getEntityTokenIds()).noneMatch(id -> {
                             try {
                                 return excludeLabels.contains(tokenRead.nodeLabelName(id));
                             } catch (LabelNotFoundKernelException e) {
                                 return false;
                             }
-                            }))
```
jexp: lots of unrelated whitespace changes :)

src/test/java/apoc/schema/SchemasTest.java
```diff
@@ -41,6 +48,54 @@ public void tearDown() {
...
+        db.execute("CREATE INDEX ON :Foo(baz)").close();
+
+        List<String> initialIndexNames;
+        try (Transaction tx = db.beginTx()) {
+            initialIndexNames = Iterables.asList(db.schema().getIndexes()).stream().map(Object::toString).collect(Collectors.toList());
+        }
+
+        testResult(db, "CALL apoc.schema.recreate()", (result) -> {
+            while(result.hasNext()) {
+                System.out.println("result.next() = " + result.next());
```
jexp: asserts?

src/test/java/apoc/schema/SchemasTest.java
```diff
@@ -41,6 +48,54 @@ public void tearDown() {
...
+        db.execute("CREATE CONSTRAINT ON (m:Movie) ASSERT m.title IS UNIQUE").close();
+
+        List<String> initialIndexNames;
+        try (Transaction tx = db.beginTx()) {
+            initialIndexNames = Iterables.asList(db.schema().getIndexes()).stream().map(IndexDefinition::getName).collect(Collectors.toList());
+        }
+
+        testResult(db, "CALL apoc.schema.recreate()", (result) -> {
+            while(result.hasNext()) {
+                System.out.println("result.next() = " + result.next());
```
jexp: asserts

src/main/java/apoc/schema/Schemas.java
mneedham: We didn't change anything in this function

src/main/java/apoc/schema/Schemas.java
mneedham: so for this it's fine as it is?

src/main/java/apoc/schema/Schemas.java
```diff
@@ -117,6 +135,101 @@ private AssertSchemaResult createUniqueConstraint(Schema schema, String lbl, Str
...
+                constraint.drop();
+                info.dropped();
+                result.add(info);
+            }
+            return null;
+        }).get(dropAllConstraintsTimeout,TimeUnit.MILLISECONDS);
+
+        Util.inTxFuture(executorService, db, () -> {
+            for (ConstraintDefinition constraint : constraintsToRecreate) {
+                List<Object> propertyKeys = new ArrayList<>();
```
jexp: consistency around property keys :)

src/main/java/apoc/schema/Schemas.java
```diff
@@ -117,6 +135,101 @@ private AssertSchemaResult createUniqueConstraint(Schema schema, String lbl, Str
         return new AssertSchemaResult(lbl, key).unique().created();
     }
 
+    public List<AssertSchemaResult> recreateIndexes(long allIndexesTimeout, long dropAllIndexesTimeout, long recreateAllIndexesTimeout) throws ExecutionException, InterruptedException, IllegalArgumentException, TimeoutException {
+        Schema schema = db.schema();
+        List<AssertSchemaResult> result = Collections.synchronizedList(new ArrayList<>());
+        ExecutorService executorService = Pools.SINGLE;
+        List<IndexDefinition> indexesToRecreate = Util.inTxFuture(executorService, db, () -> {
+            List<IndexDefinition> idx = new ArrayList<>();
```
jexp: can we extract these inner closures into methods so that we can just call them as a method (handle)

or simplify them more into a chain:

1. collect indexes + constraints into AssertResults
2. go over them and drop them
3. go over them and recreate them

src/main/java/apoc/schema/Schemas.java
mneedham: I'm not sure what you mean. The reason we have a separate thread to collect all the indexes and constraints is the following code:

src/main/java/apoc/schema/Schemas.java
mneedham: This class is copied from the existing `apoc.schema.assert`procedure, we haven't done anything different here

src/main/java/apoc/schema/Schemas.java
mneedham: This one is different because it's a `List<Object>` rather than `List<String>` because that's what the downstream functions take as an argument. So we can't use the Collections approach unless we change the downstream function, and we were trying to change minimal parts of the code.

jexp:
@mneedham what's the state on this?

PR #1347, opened 2019-11-14
fixes #1342: APOC load xls range problem with columns after Z

### LINKED ISSUE #1342: APOC load xls range problem with columns after Z
I am trying to use apoc.load.xls to import from a large spreadsheet. It seems as if the range has to be between A and Z. But this spreadsheet goes all the way to AY.
Here is an example that works:

CALL apoc.load.xls('GQL_Inputs_and_comparative_analysis-2019-11-07.xlsx', '1.1 Basic operations!A1:Z10') yield lineNo, list
return *

And here is an example that does not work:

CALL apoc.load.xls('GQL_Inputs_and_comparative_analysis-2019-11-07.xlsx', '1.1 Basic operations!A1:AY10') yield lineNo, list
return *

Error message: 

Failed to invoke procedure `apoc.load.xls`: Caused by: java.lang.IllegalStateException: Header at position 51 doesn't have a value

See screendumps here: https://community.neo4j.com/t/apoc-load-xls-range-problem-with-columns-after-z/11782?u=thomasf

Neo4j Browser version: 3.2.24

Neo4j Server version: 3.5.8 (enterprise)

The Apache libraries were downloaded a couple of weeks ago.

The spreadsheet can be found here: https://docs.google.com/spreadsheets/d/1zpLDPWQYwACgphPz8MxcKvtbOjF3Ye-z5d2jSnNa1GI/edit#gid=1358645973

Best regards

Thomas

conker84:
Fixes #1342 

Fixed the bug

  - Changed the column index resolution algorithm


conker84:
> The new test file `GQL_ Inputs and comparative analysis.xlsx` should have a name without spaces. Also it would be good if we can reduce its file size - it's currently 341 KB. I suspect a simple long header and one data line should be good enough for testing.

Done! Thanks a lot for the feedback!

PR #1355, opened 2019-11-20
Neo4j MqTT client

bzupan:
Option to publish / subscribe to MqTT broker from the Neo4j DB.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ hadoop
 progress.csv
 logs/
 store_lock
+/.nb-gradle/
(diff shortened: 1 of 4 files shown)


sarmbruster:
First of all thanks for the PR. Some feedback:

* you seem to use different settings regarding tab/spaces. Lots of changes in `build.gradle` and `.gitignore` are based on this. Please do only change the relevant things there.
* Can you please add unit tests? Please understand that we won't merge new functionality not being covered by tests.

conker84:
I'm closing this since we got no feedback from the user. We can also plan to add this into Neo4j Streams instead of APOC.
Thank you so much for the contribution

PR #1359, opened 2019-11-25
fixes #1335: Using apoc.export.csv.graph with bulkImport:true produces wrong result for id property

### LINKED ISSUE #1335: Using apoc.export.csv.graph with bulkImport:true produces wrong result for "id" property
When the node has a property named "id" and the node will be exported with the bulkImport:true option, then the value of the "id" property will be replaced with the value of the node's internal ID.
The code is probably wrong in the CsvFormat.java file, in the writeNodesBulkImport procedure.

conker84:
Fixes #1335

Removed the `id` name for the property as default name for the `:ID` field

  - Fixed the bug 
  - added a test case

### DIFF
--- a/src/main/java/apoc/export/csv/CsvFormat.java
+++ b/src/main/java/apoc/export/csv/CsvFormat.java
@@ -25,6 +25,7 @@ import static apoc.util.Util.joinLabels;
  * @since 22.11.16
  */
 public class CsvFormat implements Format {
+    public static final String ID = "id";
     private final GraphDatabaseService db;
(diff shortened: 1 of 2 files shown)


JMHReif:
Was this updated for business id fields? It looks like it still sets the user's id field to the internal Neo4j id value, rather than a business key id value. I think that's what the user was after on the initial issue.

conker84:
@JMHReif please look at my comment:

https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1335#issuecomment-572529311

lemme know what you think about it!

PR #1361, opened 2019-11-29
[WIP] fixes #1336: Create apoc.export.cypher format and outputs optimised for passing to apoc.bolt.execute

### LINKED ISSUE #1336: Create apoc.export.cypher format and outputs optimised for passing to apoc.bolt.execute

I would like to be able to do this (using a new "apoc-bolt" format):

```
call apoc.export.cypher.all(null,{format:'apoc-bolt', streamStatements:true}) YIELD cypherStatements, cypherParams
CALL apoc.bolt.execute(
  "bolt+routing://neo4j:<password>@<database url>:7687",
  cypherStatement, cypherParams) 
YIELD row RETURN *
```

and I would like it to be fast (e.g. by allowing me to use the unwind parameters optimisations).

Currently I can just about make this work like so:

```
call apoc.export.cypher.all(null,{format:'plain', streamStatements:true, batchSize:1, separateFiles:true, useOptimizations: {type: 'None'}}) YIELD nodeStatements, relationshipStatements, schemaStatements, cleanupStatements
WITH CASE cleanupStatements
 WHEN NULL THEN [schemaStatements, nodeStatements, relationshipStatements]
  ELSE  [schemaStatements, nodeStatements, relationshipStatements] + [ x in split(cleanupStatements, ";\n") | trim(x)]
END as cypherStatementList
UNWIND cypherStatementList as cypherStatement
WITH cypherStatement
WHERE cypherStatement is not NULL and size(cypherStatement) > 0
CALL apoc.bolt.execute(
"bolt+routing://neo4j:<password>@2bd3b1d2.databases.neo4j.io:7687",
  cypherStatement) YIELD row RETURN *
```
but it is insanely slow, doesn't use parameters and doesn't use any optimisations.

This will make it easier to copy (sub)graphs between neo4j databases.

conker84:
Fixes #1336

This (WIP) PR proposes two procedures in order to address the request or the related issue:

- `apoc.bolt.load.fromLocal($url, $localStatement, $remoteStatement, $config)`: which allow to execute a statement into `local` Neo4j instance with the `localStatement` variable and use the result of the execution in order use them in combination with the `remoteStatement` in the `remote` Neo4j instance. **The procedure can be useful when the user needs to call `apoc.bolt.load` more than once (maybe in an unwind) because on each call we create a new Driver (& Session) instance that can slow down the performances, while into the `fromLocal` procedure we create them once**
Please look at the related tests `testLoadFromLocal` and `testLoadFromLocalStream` into the `BoltTest` class
It the `config` contains `streamStatements` as true, the procedure assumes that the output of the `localStatement` execution is a table rows of two columns:
  - `statement`: a cypher string executed into the remote statement (mandatory)
  - `params`: a map of params related to the cypher statement (optional)
- `apoc.bolt.clone.all($url)`: under-the-hood uses the `ExportCypher` business logic in order to stream statements to the remote instance.

So while the first procedure is more general, the second one is focused on cloning the database. The first idea with @mneedham was to provide only the `apoc.bolt.clone.all` but when I start to work on it I found that maybe a general approach can be useful for other use-cases.

I created this as WIP because I want to chat with you guys about the approach.

  - Implemented the procedures
  - Added tests

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -91,7 +91,7 @@ dependencies {
     testCompile group: 'com.sun.mail', name: 'javax.mail', version: '1.6.0'
     compile group: 'com.jayway.jsonpath', name: 'json-path', version: '2.4.0'
     compile group: 'org.hdrhistogram', name: 'HdrHistogram', version: '2.1.9'
-    compile group: 'org.neo4j.driver', name: 'neo4j-java-driver', version: '1.7.3'
+    compile group: 'org.neo4j.driver', name: 'neo4j-java-driver', version: '1.7.5'
     compile group: 'com.novell.ldap', name: 'jldap', version: '2009-10-07'
 
     def withoutAntlr =  {
--- a/src/main/java/apoc/bolt/Bolt.java
+++ b/src/main/java/apoc/bolt/Bolt.java
@@ -1,15 +1,27 @@
 package apoc.bolt;
 
 import apoc.Description;
+import apoc.export.cypher.ExportCypher;
+import apoc.result.MapResult;
 import apoc.result.RowResult;
 import apoc.result.VirtualNode;
 import apoc.result.VirtualRelationship;
+import apoc.util.JsonUtil;
 import apoc.util.UriResolver;
 import apoc.util.Util;
+import org.apache.commons.lang3.StringUtils;
 import org.neo4j.driver.internal.InternalEntity;
 import org.neo4j.driver.internal.InternalPath;
 import org.neo4j.driver.internal.logging.JULogging;
-import org.neo4j.driver.v1.*;
+import org.neo4j.driver.internal.summary.InternalSummaryCounters;
+import org.neo4j.driver.v1.Config;
+import org.neo4j.driver.v1.Driver;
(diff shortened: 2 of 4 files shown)


conker84:
@eastlondoner can you please take a look at this? I'd like to know what do you think about?

I also prepared a jar for you so you can do a test:
[apoc-3.5.0.6-all.jar.zip](https://github.com/neo4j-contrib/neo4j-apoc-procedures/files/3904978/apoc-3.5.0.6-all.jar.zip)

Thanks a lot!

conker84:
@eastlondoner ping

eastlondoner:
thanks! I am still off on parental leave but I will try and look at this at the beginning of Feb

JMHReif:
@conker84, is the `apoc.bolt.load.fromLocal` going FROM a local instance TO a remote instance? Or is it going from a remote to a local? The name of the proc indicates it's going from, but in your text description in this PR (`allow to execute a statement into local Neo4j instance`), it seems like it goes to local. Just want to clarify.

mneedham:
@conker84 can you rebase this one and then we'll merge it in?

PR #1367, opened 2019-12-13
fixes #1366: apoc.export.cypher.* bug with multiple labels sorted

### LINKED ISSUE #1366: apoc.export.cypher.* bug with multiple labels sorted
Given the provided dataset the expected exported query is:

```
UNWIND [{name:\"A\", properties:{}}] AS row
CREATE (n:Bar{name: row.name}) SET n += row.properties SET n:Baz;
```

But we have

```
UNWIND [{name:"A", _id:20, properties:{}}] AS row
CREATE (n:Bar{name: row.name, `UNIQUE IMPORT ID`: row._id}) SET n += row.properties SET n:Baz;
```

`CALL apoc.export.cypher.query('MATCH (n:Baz) RETURN n', null,{stream: true, format: 'plain'})`

```
CREATE (n:Bar:Baz{name: 'A'})
```

conker84:
Fixes #1366

  - Fixed the bug

### DIFF
--- a/src/main/java/apoc/export/cypher/formatter/CypherFormatterUtils.java
+++ b/src/main/java/apoc/export/cypher/formatter/CypherFormatterUtils.java
@@ -12,6 +12,7 @@ import org.neo4j.values.storable.Values;
 import java.lang.reflect.Array;
 import java.time.temporal.Temporal;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static apoc.export.util.FormatUtils.getLabelsSorted;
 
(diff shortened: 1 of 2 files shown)


conker84:
@Dcanzano please look here

JMHReif:
This looks good to me. How can we tell which test failed in the build? Is it something we should run again?

conker84:
The failing test is `apoc.periodic.PeriodicTest > testTerminateCommit FAILED` that (I think) is not related to the PR

PR #1370, opened 2019-12-20
fixes #1064: Add procedure to compare graphs

### LINKED ISSUE #1064: Add procedure to compare graphs
Add procedures to compare graphs

are these two graphs equal (hashing neighborhoods)

if not list differences (e.g. via label + primary key to identify nodes (or all props))

optionally for remote graphs (via bolt)

conker84:
Fixes #1064

This procedure implements the changes in the related issue, with small changes in the requirement about the relationship comparison.

The `apoc.diff.graphs` compares two graphs and returns the differences in term of:

* same node count
* same count per label
* same relationship counter
* same count per rel-type

For each node in the `source` graph with a certain label, find the same node (in the `dest` graph) by keys or internal id in the other graph and if found:
* compare all labels
* compare all properties

For each relationship in the `source` graph, we'll get the two nodes of the relationship and look into the the `dest`
graph if there is a relationship with the same properties and the same start/end node.

In order to allow to use this procedure with the `apoc.bolt.*` I had to do changes into this one because the procedure didn't manage `map` and `collection` return types.

  - Implemented `apoc.diff.graphs`
  - Extended the behaviour of `apoc.bolt.*` in order to manage collections and maps
  - In `apoc.bolt.*` added a new property `withRelationshipNodeProperties` that retrieves the node properties attaching them into `VirtualRelationship` nodes.
  - Update the documentation

### DIFF
--- a/docs/asciidoc/database-integration/bolt.adoc
+++ b/docs/asciidoc/database-integration/bolt.adoc
@@ -66,7 +66,9 @@ apoc.bolt.production.url=bolt://password:test@localhost:7688
 Config available are:
 
 * `statistics`: possible values are true/false, the default value is false. This config print the execution statistics;
-* `virtual`: possible values are true/false, the default value is false. This config return result in virtual format and not in map format, in apoc.bolt.load.
+* `virtual`: possible values are true/false, the default value is false. This config return result in virtual format and not in map format. *N.B.* If `withRelationshipNodeProperties=false` the `VirtualRelationship` contains only the ids about the nodes connected by the edge
+* `readOnly`: possible values are true/false, the default value is true. Defines the operation performed over the remote instance.
+* `withRelationshipNodeProperties`: possible values are true/false, the default value is false. If `virtual=true` it returns the `VirtualRelationship` with nodes that also contains properties attached to them.
 
 == Driver configuration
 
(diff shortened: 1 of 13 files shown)


jexp:
I would not do internal id except if the user requests it (config option) otherwise there is very little chance that the internal id's are the same.

conker84:
Hi @jexp I updated the PR with a config param `findById` that works as you suggested.
Thanks a lot!

JMHReif:
It looks like the documentation states that we're pulling a bunch of extra information from the remote db to compare with just the data of the local db? Is there something that's needed with the index retrieval in the remote one? Can we make it consistent where we're pulling the same data/fields for each db to compare?

JMHReif:
Also, what would need to change to adapt for 4.0 (multi-database, system db that has schema info, etc)?

conker84:
On Thursday, we discussed about managing:
* maps like `{nodes: <node_list>, relationships: <rel_list>}` in this case the procedure should handle the keys by using the local constraints;
* strings like `bolt(neo4j)://<remote_url>:<port>?query=<QUERY>` in this case we can get rid of the `apoc.load.*` procedure invocation and build the required map internally;

jacktang:
May I ask why `apoc.diff.graphs` is missed in branch 4.4?

PR #1389, opened 2020-01-27
Tables4Labels

### LINKED ISSUE #1388: Sample-based metadata for node/relationship properties.
## Feature description
This feature will be initially used internally for a future-release component. This feature replaces built-in metadata functions to improve performance when attempting to scan the entire graph for properties. This is done by sampling 1/nth nodes instead of scanning the entire graph.

neoarchitect:
Fixes #1388

Full list of features:

1. **Tables4Labels**. Base functionality to replace nodeTypeProperties() and relTypeProperties() with sampling versions.
2. **Relationships4Labels**. Adds two columns to relTypeProperties() to show source and target labels for each relationship that has a valid property.
3. **Mandatory handling functionality change**. Forces "mandatory" flag to true only if a constraint exists for this property (only works for Enterprise Edition as constraints are a licensed function of Neo4J).
4. **Allow inclusion/exclusion of nodes by relationships**. When calling nodeTypeProperties(), include/exclude a node if it has a specifically has/doesn't have a relationship type associated).

Work-in-progress PR. Still requires:

- Complete unit tests (some are complete, some are not)
- Documentation


moxious:
What this PR is about:  an excerpt from technical requirements for the BI Driver:

The data model for the first version will be a “Tables for Labels” model.  Neo4j nodes can be given one or more labels.  Each unique combination of labels will translate to a table.  In this model, a node will only be a member of one unique label combination/table.  Nodes that do not have a label will not be accessible through the driver.

There will be two sets of tables (which can exist in different schemas if desired) Labels and Relationships.  The relationships schema contains all the relationships, which connect nodes.  The records in a relationship table will include source and target node ID as well as source and target label names.
Properties of both Nodes and Relationships will be represented as columns in their respective table.

The driver should sample records and infer consistent property structure / data types on that basis.   Neo4j already provides APOC schema extensions as well as built-in procedures that can provide this functionality if needed.   The driver should be able to handle situations where “schema-optional” Neo4j nodes may not have a totally consistent typing structure, for example a property on a certain kind of node which could be either a string or an integer.

src/main/java/apoc/meta/tablesforlabels/PropertyContainerProfile.java
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            tracker.addObservation(n.getProperty(propName));
+
+            if (isNode) {
+
+                // Check for node constraints
+
+                tracker.mandatory = false;
+                for (ConstraintDefinition cd : constraints) {
```
moxious: Each time a property container (i.e. an individual node is observed) we're looking at the constraints.   They'll never change, they should be examined once upfront and not every time we look at a node, because this will add unnecessary processing time.

Additionally -- a thing to be super careful of is that mandatory = true **only when** there is an **existence constraint**, not just any constraint.  If for example there is a uniqueness constraint, this will cause mandatory=true, which is wrong.   Here's an example of why:

```
CREATE CONSTRAINT ON (f:Foo) ASSERT f.x is unique;
create (:Foo { x: null });
create (:Foo { x: null });
```

That works because null is never equal to itself in any database, and so they are unique.  But they don't exist.  

So set mandatory=true only for existance constraints, and do it only once.

src/main/java/apoc/meta/Tables4LabelsProfile.java
```diff
@@ -0,0 +1,256 @@
...
+            for(Relationship r : n.getRelationships(type, Direction.OUTGOING)) {
+                Iterable<Label> relStartNode = r.getStartNode().getLabels();
+                Iterable<Label> relEndNode = r.getEndNode().getLabels();
+                boolean isNode = false;
+                localRelProfile.observe(r, constraints, isNode, relConstraints, relStartNode, relEndNode);
+            }
+        }
+    }
+
+    public Tables4LabelsProfile finished() {
```
moxious: It's here, when all processing is completed -- that you should consider the constraints and do it only once.   Don't observe the constraints inside of the 'observe' method that will be called potentially millions of times.  It'll also help refactor that deeply nested loop below where your comments say there's a performance concern.

src/main/java/apoc/meta/tablesforlabels/PropertyContainerProfile.java
```diff
@@ -0,0 +1,98 @@
...
+                        if (pk == propName) {
+                            tracker.mandatory = true;
+                        }
+                    }
+                }
+            } else {
+
+                // Check for relationship constraints - NOTE: Could probably improve the efficiency here a bit. Too many nested loops.
+
+                List<String> sourceNodeLabels = new ArrayList<String>();
```
moxious: sourceNodeLabels and targetNodeLabels can be initialized with an argument 4 for initial size.  99.9999% of neo4j nodes don't have more than 4 labels for reasons we can discuss later if you're interested.

Also, you can save a lot of code and two while loops here by initializing your lists from your iterators in an easier way here:  

https://www.baeldung.com/java-convert-iterator-to-list#java8

src/main/java/apoc/meta/tablesforlabels/PropertyContainerProfile.java
```diff
@@ -0,0 +1,98 @@
...
+                while (ren.hasNext()) {
+                    targetNodeLabels.add(ren.next().name());
+                }
+
+                tracker.sourceNodeLabels = sourceNodeLabels;
+                tracker.targetNodeLabels = targetNodeLabels;
+                
+                tracker.mandatory = false;
+
+                for (Map.Entry<String,Iterable<ConstraintDefinition>> entry : relConstraints.entrySet()) {
```
moxious: See above comment.   This code should be in Tables4LabelsProfile#finished(), and you need to check only for existance constraints, not for any constraint (which includes uniqueness, which does not guarantee the property is mandatory)

src/main/java/apoc/meta/tablesforlabels/PropertyTracker.java
```diff
@@ -0,0 +1,92 @@
...
+        // because while it may always be populated *when its present* it can be missing
+        // from some nodes.
+        if (totalObservations > observations) {
+            mandatory = false;
+        }
+    }
+
+    public static final Map<String,String> typeMappings = new HashMap<String,String>();
+    static {
+        typeMappings.put("java.lang.String", "String");
```
moxious: Prior to final approval, Shash please go through the types and ensure that we haven't missed anything there, or ideally make sure unit test coverage covers every possible type.

Some type documentation (or the entry point to it) is here: https://neo4j.com/docs/cypher-manual/current/functions/temporal/

The original code was written under 3.5.   I'm about 80% sure that in Neo4j 4.0 we didn't introduce any new types, but we need to make sure we have total coverage.   For each type we also need to cover the corresponding array type.

If we miss one, then the code will report the type as string.  The driver will similarly coerce it to string, and we'll end up with a bug report so we need to be sure we're 100% complete.  If you've done the review or are satisfied the unit tests cover this, then just flag a follow-up comment saying you looked into it and this point is good to go.

moxious:
Something isn't working right with source and target labels on relationships.    Simple test case for a unit test:

```
CREATE (:Person)-[:HAS]->(:Job)
CREATE (:Person)-[:HAS]->(:Hobby)
CREATE (:Person)<-[:HAS]-(:Organization)
CREATE (:Component)-[:HAS]->(:Component);

CALL apoc.meta.relTypeProperties();
```

(I'm expecting to get 4 records back, for HAS with different source and target labels).   What I actually get is a single line with sourceNodeLabels=null targetNodeLabels=null.

A query you can use to validate that your test case is right is this:

```
match (n)-[r]->(m) return labels(n) as sourceNodeLabels, labels(m) as targetNodeLabels;
```

In this case it returns:
```
Person,Job
Person,Hobby
Organization,Person
Component,Component
```

src/main/java/apoc/meta/tablesforlabels/PropertyContainerProfile.java
neoarchitect: Existence check was added a while ago, so this one is done!

src/main/java/apoc/meta/Tables4LabelsProfile.java
neoarchitect: Moved the check code to finished(), and cleaned up a lot of the passing around of constraint data.

neoarchitect:
> Something isn't working right with source and target labels on relationships. Simple test case for a unit test:
> 
> ```
> CREATE (:Person)-[:HAS]->(:Job)
> CREATE (:Person)-[:HAS]->(:Hobby)
> CREATE (:Person)<-[:HAS]-(:Organization)
> CREATE (:Component)-[:HAS]->(:Component);
> 
> CALL apoc.meta.relTypeProperties();
> ```
> 
> (I'm expecting to get 4 records back, for HAS with different source and target labels). What I actually get is a single line with sourceNodeLabels=null targetNodeLabels=null.
> 
> A query you can use to validate that your test case is right is this:
> 
> ```
> match (n)-[r]->(m) return labels(n) as sourceNodeLabels, labels(m) as targetNodeLabels;
> ```
> 
> In this case it returns:
> 
> ```
> Person,Job
> Person,Hobby
> Organization,Person
> Component,Component
> ```

This was resolved in a previous update!

src/main/java/apoc/meta/tablesforlabels/PropertyContainerProfile.java
neoarchitect: Moved to finished() as per previous comment!

src/main/java/apoc/meta/tablesforlabels/PropertyContainerProfile.java
neoarchitect: This was entirely re-written in a previous update, so may not be relevant any more..

src/main/java/apoc/meta/tablesforlabels/PropertyTracker.java
neoarchitect: As far as I can tell all types are accounted for as follows:

String
StringArray
Double
DoubleArray
Integer
IntegerArray
Long
LongArray
Point
PointArray
DateTime
DateTimeArray
Boolean
BooleanArray
Date
DateArray
LocalDateTime
LocalDateTimeArray
LocalTime
LocalTimeArray
Duration
DurationArray
Time
TimeArray

PR #1401, opened 2020-02-05
Tables4Labels - 4.x

### LINKED ISSUE #1388: Sample-based metadata for node/relationship properties.

neoarchitect:
Fixes #1388 (for 4.x)

Full list of features:

1. **Tables4Labels**. Base functionality to replace nodeTypeProperties() and relTypeProperties() with sampling versions.
2. **Relationships4Labels**. Adds two columns to relTypeProperties() to show source and target labels for each relationship that has a valid property.
3. **Mandatory handling functionality change**. Forces "mandatory" flag to true only if a constraint exists for this property (only works for Enterprise Edition as constraints are a licensed function of Neo4J).
4. **Allow inclusion/exclusion of nodes by relationships**. When calling nodeTypeProperties(), include/exclude a node if it has a specifically has/doesn't have a relationship type associated).

Work-in-progress PR. Still requires:

- Complete unit tests (some are complete, some are not)
- Documentation

### DIFF
--- a/src/main/java/apoc/meta/Meta.java
+++ b/src/main/java/apoc/meta/Meta.java
@@ -6,6 +6,7 @@ import apoc.result.VirtualNode;
 import apoc.result.VirtualRelationship;
 import apoc.util.MapUtil;
 import org.neo4j.graphdb.*;
+import org.neo4j.graphdb.schema.ConstraintType;
 import org.neo4j.graphdb.schema.ConstraintDefinition;
 import org.neo4j.graphdb.schema.IndexDefinition;
 import org.neo4j.graphdb.schema.Schema;
@@ -15,6 +16,7 @@ import org.neo4j.internal.helpers.collection.Pair;
 import org.neo4j.internal.kernel.api.Read;
 import org.neo4j.internal.kernel.api.TokenRead;
 import org.neo4j.kernel.api.KernelTransaction;
+import org.neo4j.kernel.internal.GraphDatabaseAPI;
 import org.neo4j.procedure.*;
 import org.neo4j.values.storable.DurationValue;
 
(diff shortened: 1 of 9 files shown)


sarmbruster:
Thanks for this PR. I've went through your changes - please consider a few changes:

* rebase your feature branch on upstream 4.0 - you're missing one or two commits.
* the commit msg says "3.5", but it's for 4.0
* I cannot see any unit test being added. as a rule of thumb each and every proc/function in apoc should have at least one unit tests
* also docs would be nice

I'm adding more via GH review

src/main/java/apoc/meta/Meta.java
```diff
@@ -430,6 +432,126 @@ private void collectStats(Collection<String> labelNames, Collection<String> relT
...
+     * metadata that is useful for generating "Tables 4 Labels" schema designs for RDBMSs, but in a more performant way.
+     */
+    @Procedure
+    @Description("apoc.meta.nodeTypeProperties()")
+    public Stream<Tables4LabelsProfile.NodeTypePropertiesEntry> nodeTypeProperties(@Name(value = "config",defaultValue = "{}") Map<String,Object> config) {
+        MetaConfig metaConfig = new MetaConfig(config);
+        try {
+            return collectTables4LabelsProfile(metaConfig).asNodeStream();
+        } catch (Exception exc) {
+            exc.printStackTrace();
```
sarmbruster: printStackTrace isn't good error handling. Maybe use log.warn to dump the stacetrace and rethrow it wrapped into a RuntimeException

src/main/java/apoc/meta/Meta.java
```diff
@@ -430,6 +432,126 @@ private void collectStats(Collection<String> labelNames, Collection<String> relT
...
+     * metadata that is useful for generating "Tables 4 Labels" schema designs for RDBMSs, but in a more performant way.
+     */
+    @Procedure
+    @Description("apoc.meta.relTypeProperties()")
+    public Stream<Tables4LabelsProfile.RelTypePropertiesEntry> relTypeProperties(@Name(value = "config",defaultValue = "{}") Map<String,Object> config) {
+        MetaConfig metaConfig = new MetaConfig(config);
+        try {
+            return collectTables4LabelsProfile(metaConfig).asRelStream();
+        } catch (Exception exc) {
+            exc.printStackTrace();
```
sarmbruster: same as above on error handling

src/main/java/apoc/meta/Meta.java
```diff
@@ -430,6 +432,126 @@ private void collectStats(Collection<String> labelNames, Collection<String> relT
...
+
+        Set<String> includeRels = config.getIncludesRels();
+        Set<String> excludeRels = config.getExcludeRels();
+
+        for (Label label : tx.getAllLabelsInUse()) {
+            String labelName = label.name();
+
+            if (excludes.contains(labelName)) {
+                // Skip if explicitly excluded
+                continue;
```
sarmbruster: `continue` is IMO bad practice. Instead use a `if (!badCondition) { /* do stuff */ }`

src/main/java/apoc/meta/tablesforlabels/OrderedLabels.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Abstraction on an ordered label set, used as a key for tables for labels profiles
+ */
+public class OrderedLabels {
+    List<String> labels;
+
+    public OrderedLabels(Iterable<Label> input) {
+        labels = new ArrayList<>(3);
```
sarmbruster: I guess it's more concise to use java 8 streams for mapping and sorting?

src/main/java/apoc/meta/Meta.java
neoarchitect: Done - pushed the stacktrace to debug (instead of warn) and wrapped in RuntimeException. As it's something that shouldn't really happen, I thought debug would be better than warn.

src/main/java/apoc/meta/Meta.java
neoarchitect: Done - pushed the stacktrace to debug (instead of warn) and wrapped in RuntimeException. As it's something that shouldn't really happen, I thought debug would be better than warn.

src/main/java/apoc/meta/Meta.java
neoarchitect: I changed it and wrapped the code in an if predicate. I'm not too concerned with continue's usually, but I can see where it could be an issue. Performance wise this might save a few clock cycles too (not really sure)!

src/main/java/apoc/meta/tablesforlabels/OrderedLabels.java
neoarchitect: For minimal changes between 3.5 and 4.0 I didn't change it, but if you think updating this would help, let me know. We're in a bit of a time crunch, so unless critical I'd rather leave it.

neoarchitect:
> Thanks for this PR. I've went through your changes - please consider a few changes:
> 
> * rebase your feature branch on upstream 4.0 - you're missing one or two commits.
> * the commit msg says "3.5", but it's for 4.0
> * I cannot see any unit test being added. as a rule of thumb each and every proc/function in apoc should have at least one unit tests
> * also docs would be nice
> 
> I'm adding more via GH review

- Rebase done
- The commit message actually said "from 3.5" because I was taking the code I did in 3.5 and moving it to 4.0. :)
- Will add the unit tests shortly.
- Will add the asciidocs shortly.

PR #1409, opened 2020-02-16
fixes #1257: apoc.load.json file roundtrip from apoc.export.json.all

### LINKED ISSUE #1257: apoc.load.json file roundtrip from apoc.export.json.all
See here for full detail:
https://stackoverflow.com/questions/57036051/neo4j-apoc-load-json-no-data-in-neo4j

conker84:
Fixes #1257 

Imported json files produced with `apoc.export.json.all`

  - created `apoc.import.json.all`
  - added tests
  - updated documentation

### DIFF
--- a/docs/asciidoc/import/loadjson.adoc
+++ b/docs/asciidoc/import/loadjson.adoc
@@ -195,3 +195,28 @@ You will receive the following response:
   "weight": 0.02548018842935562
 }
 ----
+
+== Import JSON
+
+If you used the `apoc.export.json.all` to export the graph, we provide the `apoc.import.json` to reimport that data.
+
+[source,cypher]
+----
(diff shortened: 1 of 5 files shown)


JMHReif:
Is the difference between this and `apoc.load.json` just that it takes maps with both nodes/rels in same json file? Should we move the procedure under the `apoc.import.*` or `apoc.load.json*` path?

jexp:
It's meant to support the roundtrip from apoc.export.json from another db.

The json package should be fine.

I don't think we need the `.all` suffix.

conker84:
> It's meant to support the roundtrip from apoc.export.json from another db.
> 
> The json package should be fine.
> 
> I don't think we need the `.all` suffix.

@jexp I removed the `.all` suffix

PR #1426, opened 2020-02-25
fixes #1423: Upgrade common beantuils from 1.9.3

### LINKED ISSUE #1423: Upgrade common beantuils from 1.9.3
Security fix, please both in 3.5 and 4.0 branch

conker84:
Fixes #1423

  - Updated BeanUtils to `1.9.4`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -197,6 +197,9 @@ dependencies {
     // there is a compatibility problem between the Guava Library used by HDFS, Google Cloud Storage and Reflections; the only version that fits for all is the following
     testCompile group: 'com.google.guava', name: 'guava', version: '20.0'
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -197,6 +197,8 @@ dependencies {
     // there is a compatibility problem between the Guava Library used by HDFS, Google Cloud Storage and Reflections; the only version that fits for all is the following
     testCompile group: 'com.google.guava', name: 'guava', version: '20.0'
 
+    compile group: 'commons-beanutils', name: 'commons-beanutils', version: '1.9.4'
```
sarmbruster: Since this is a new dependency for apoc and we don't really depend on it directly, that change should carry a comment like "force 1.9.4 of beanutils as transitive dependency due to CVE xxxx"

conker84: done
thanks a lot!

PR #1462, opened 2020-03-28
fix deadlock situation in concurrent executions

sarmbruster:
We did use the default thread pool for the thread sending eventually the tombstone object to a queue. If used in a nested way this could lead to a situation when default pool is full that the tombstone sending thread waits for the queue. Queue consumer waits as well -> deadlock situation.

By using a new thread instead we ensure that the tombstone is always sent correctly. Also QueueBasedSpliterator has been refactor and uses `take` instead of `poll`.

### DIFF
--- a/src/main/java/apoc/cypher/Cypher.java
+++ b/src/main/java/apoc/cypher/Cypher.java
@@ -74,11 +74,12 @@ public class Cypher {
     public Stream<RowResult> runFiles(@Name("file") List<String> fileNames, @Name(value = "config",defaultValue = "{}") Map<String,Object> config) {
         boolean addStatistics = Util.toBoolean(config.getOrDefault("statistics",true));
         int timeout = Util.toInteger(config.getOrDefault("timeout",10));
+        int queueCapacity = Util.toInteger(config.getOrDefault("queueCapacity",100));
         List<RowResult> result = new ArrayList<>();
         @SuppressWarnings( "unchecked" )
         Map<String,Object> parameters = (Map<String,Object>)config.getOrDefault("parameters",Collections.emptyMap());
         for (String f : fileNames) {
-            List<RowResult> rowResults = runManyStatements(readerForFile(f), parameters, false, addStatistics, timeout).collect(Collectors.toList());
+            List<RowResult> rowResults = runManyStatements(readerForFile(f), parameters, false, addStatistics, timeout, queueCapacity).collect(Collectors.toList());
             result.addAll(rowResults);
         }
         return result.stream();
@@ -95,24 +96,28 @@ public class Cypher {
     public Stream<RowResult> runSchemaFiles(@Name("file") List<String> fileNames, @Name(value = "config",defaultValue = "{}") Map<String,Object> config) {
         boolean addStatistics = Util.toBoolean(config.getOrDefault("statistics",true));
         int timeout = Util.toInteger(config.getOrDefault("timeout",10));
+        int queueCapacity = Util.toInteger(config.getOrDefault("queueCapacity",100));
         List<RowResult> result = new ArrayList<>();
         for (String f : fileNames) {
-            List<RowResult> rowResults = runManyStatements(readerForFile(f), Collections.emptyMap(), true, addStatistics, timeout).collect(Collectors.toList());
+            List<RowResult> rowResults = runManyStatements(readerForFile(f), Collections.emptyMap(), true, addStatistics, timeout, queueCapacity).collect(Collectors.toList());
             result.addAll(rowResults);
         }
         return result.stream();
     }
 
-    private Stream<RowResult> runManyStatements(Reader reader, Map<String, Object> params, boolean schemaOperation, boolean addStatistics, int timeout) {
-        BlockingQueue<RowResult> queue = new ArrayBlockingQueue<>(100);
+    private Stream<RowResult> runManyStatements(Reader reader, Map<String, Object> params, boolean schemaOperation, boolean addStatistics, int timeout, int queueCapacity) {
+        BlockingQueue<RowResult> queue = new ArrayBlockingQueue<>(queueCapacity);
         Util.inThread(() -> {
-            if (schemaOperation) {
-                runSchemaStatementsInTx(reader, queue, params, addStatistics,timeout);
-            } else {
-                runDataStatementsInTx(reader, queue, params, addStatistics,timeout);
+            try {
+                if (schemaOperation) {
+                    runSchemaStatementsInTx(reader, queue, params, addStatistics,timeout);
+                } else {
+                    runDataStatementsInTx(reader, queue, params, addStatistics,timeout);
+                }
+                return null;
+            } finally {
+                queue.put(RowResult.TOMBSTONE);
             }
-            queue.put(RowResult.TOMBSTONE);
-            return null;
         });
         return StreamSupport.stream(new QueueBasedSpliterator<>(queue, RowResult.TOMBSTONE, terminationGuard, timeout), false);
     }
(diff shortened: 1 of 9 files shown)


sarmbruster:
This PR needs to be cherry-picked to 4.0 as well.

src/main/java/apoc/cypher/Cypher.java
```diff
@@ -95,26 +112,47 @@
...
+        new Thread(() -> {
+            try {
+                action.run();
+            } finally {
+                while (true) {  // ensure we send TOMBSTONE even if there's an InterruptedException
+                    try {
+                        queue.put(tombstone);
+                        return;
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
```
jexp: can this lead into being blocked forever? Shouldn't we take the timeout into account?
int put and/or the whole method time?

src/main/java/apoc/cypher/Cypher.java
```diff
@@ -163,7 +203,7 @@ private Object executeStatement(BlockingQueue<RowResult> queue, String stmt, Map
                 queue.put(new RowResult(row++, result.next()));
             }
             if (addStatistics) {
-                queue.offer(new RowResult(-1, toMap(result.getQueryStatistics(), System.currentTimeMillis() - time, row)), timeout,TimeUnit.SECONDS);
+                queue.put(new RowResult(-1, toMap(result.getQueryStatistics(), System.currentTimeMillis() - time, row)));
```
jexp: this is now blocking forever if the queue is full? 
exactly what we wanted to avoid with the timeout?
that was the reason we used offer in the first place ...

src/main/java/apoc/util/QueueBasedSpliterator.java
```diff
@@ -6,49 +6,35 @@
...
-
-    private boolean isEnd() {
-        return entry == null || entry == tombstone;
-    }
-
-    private T poll() {
+        if (Util.transactionIsTerminated(terminationGuard) || foundTombstone) return false;
         try {
-            return queue.poll(timeout, SECONDS);
+            T element = queue.take();
```
jexp: this will now potentially block forever ...

src/main/java/apoc/util/QueueBasedSpliterator.java
```diff
@@ -6,49 +6,35 @@
...
+            if (element.equals(tombstone)) {
+                foundTombstone = true;
+            } else {
+                action.accept(element);
+            }
+            return true;
```
jexp: should this really return true if the tombstone was found?

src/main/java/apoc/util/QueueBasedSpliterator.java
```diff
@@ -6,49 +6,35 @@
...
     public boolean tryAdvance(Consumer<? super T> action) {
-        if (Util.transactionIsTerminated(terminationGuard)) return false;
-        if (isEnd()) return false;
-        action.accept(entry);
-        entry = poll();
-        return !isEnd();
-    }
-
-    private boolean isEnd() {
-        return entry == null || entry == tombstone;
```
jexp: afaik there were some places which indicated null for end of (partial-) processing, but I can't remember exactly.

src/test/java/apoc/cypher/CypherTest.java
```diff
@@ -463,4 +478,20 @@ public void testRunFileWithEmptyFile() throws Exception {
         testResult(db, "CALL apoc.cypher.runFile('src/test/resources/empty.cypher')",
                 r -> assertFalse("should be empty", r.hasNext()));
     }
+
+    @Test
+    public void lengthyRunManyShouldTerminate() {
+        String repetetiveStatement= "CALL apoc.cypher.runFile(\"src/test/resources/enrollment-incremental.cypher\",{parameters: {SubID: \"218598584\", Account_Number: \"\", AccountType: \"\",Source: \"VerizonMASnapshot\", MDN: \"\", Offering: \"\", Enroll_Date: \"\", Product_SKU: \"\", Device_Model: \"\", Device_Make: \"\", First_Name: \"\", Last_Name: \"\",Email1: \"\", Email2: \"\", Email3: \"\", Postal_CD: \"\", City: \"\", State: \"\", BillingStatus: \"\", ActionType: \"Drop\", Text_Date : \"2020-03-11\"}}) yield result return sum(result.total) as total;\n" +
```
jexp: does this have any data to run on?

sarmbruster:
when applying, please to a squash&merge

src/main/java/apoc/cypher/Cypher.java
jexp: Did you want to wrap this one too?

src/main/java/apoc/cypher/Cypher.java
jexp: Blocking forever?

src/main/java/apoc/util/QueueUtil.java
```diff
@@ -0,0 +1,73 @@
...
+     * @param queue
+     * @param item
+     * @param timeoutSeconds
+     * @param failWithExecption true if a {@link RuntimeException} should be thrown in case we couldn't add item into the queue within timeframe
+     * @param checkDuringOffering a callback supposed to throw an exception to terminate
+     * @param <T>
+     */
+    public static <T> void put(BlockingQueue<T> queue, T item, long timeoutSeconds, boolean failWithExecption, Runnable checkDuringOffering) {
+        withHandlingInterrupted("Queue offer interrupted before " + timeoutSeconds + " seconds", () -> {
+            long started = System.currentTimeMillis();
```
jexp, line 28: we can compute  `started + timeoutSeconds * 1000` here (in both methods)

src/main/java/apoc/util/QueueUtil.java
```diff
@@ -0,0 +1,73 @@
...
+            while (started + timeoutSeconds * 1000 > System.currentTimeMillis()) {
+                boolean success = queue.offer(item, WAIT, WAIT_UNIT);
+                if (success) {
+                    return null;
```
jexp, line 32: do we have to return something from this method? or can it be just void?

src/main/java/apoc/util/QueueUtil.java
```diff
@@ -0,0 +1,73 @@
...
+                if (polled != null) {
+                    return polled;
+                }
+                checkDuringPolling.run();
+            }
+            throw new RuntimeException("Error polling, timeout of " + timeoutSeconds + " seconds reached.");
+        });
+    }
+
+    public static <T> T withHandlingInterrupted(String msg, ThrowingSupplier<T, InterruptedException> consumer) {
```
jexp, line 65: not sure it's worth the indirection.

what about that Thread.currentThread().interrrupt() 

handling? I always forget which one is right :)

src/main/java/apoc/util/QueueUtil.java
sarmbruster: BlockingQueue.put doesn't return anything, therefore I choose void for the method signature. Since `withHandlingInterrupted` is using a generic return type, I have to return something - null in this case.

sarmbruster: I'm inlining that indirection anyway, see comment below.

src/main/java/apoc/util/QueueUtil.java
sarmbruster: Gonna remove the indirection. I guess `Thread.currentThread.interrupt()` is not good here since I don't know what to return for `take()` in this case. So bubbling up the exception sounds more reasonable to me.

PR #1473, opened 2020-04-02
apoc.coll.min and max comparsion operator logic like cypher

### LINKED ISSUE #1463: apoc.coll.min() and max() do not work on temporal types
APOC max() and min() functions should work on temporal types

They don't. We get an error.

No dataset needed

### Steps (Mandatory)

`return apoc.coll.max([date(), date() - duration('p5M')])`
results in the following errro:

`Failed to invoke function `apoc.coll.max`: Caused by: java.lang.ClassCastException: java.time.LocalDate cannot be cast to java.lang.Number`
  

### Versions

  - OS: Mac OS Mojave
  - Neo4j: 3.5.16
  - Neo4j-Apoc: 3.5.0.9 (and presumably everything below, likely includes 4.0.0.x)

lqst:
Fixes #1463

## comparsion operator logic like cypher

  - apoc.coll.min and max doing a reduce db call
  - Added unit test for max date/datetime

Note 1: I ran some performance test with jmh and could not see any significant impact

Note 2: There is a ton of other places where we can discuss if apoc is implementing "cypher like" operator logic.

### DIFF
--- a/src/main/java/apoc/coll/Coll.java
+++ b/src/main/java/apoc/coll/Coll.java
@@ -6,6 +6,7 @@ import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.Path;
 import org.neo4j.graphdb.Relationship;
+import org.neo4j.graphdb.Result;
 import org.neo4j.helpers.collection.Pair;
 import org.neo4j.kernel.impl.util.statistics.IntCounter;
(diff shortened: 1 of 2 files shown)


jexp:
Thanks a lot. I’m only on mobile but guess lists with null were in the regular test

lqst:
Yes I tested min and max with lists that contain null. The logic is treating null as if it is not a value (it neither smaller nor larger than any other value).

PR #1475, opened 2020-04-14
Only register with kernel if there are triggers

sarmbruster:
The kernel event handling creates a big expensive transaction snapshot
for every transaction if there are registered event handlers.

To avoid paying that overhead, unregister from the kernel if there
are no triggers.

This works as a complement to the config option, so you can deploy
large groups of databases with the config enabled but only pay the
overhead on those databases that use trigger.

This is backporting the respective commit from 4.0 branch.

  -
  -
  -

### DIFF
--- a/src/main/java/apoc/trigger/Trigger.java
+++ b/src/main/java/apoc/trigger/Trigger.java
@@ -18,6 +18,7 @@ import org.neo4j.procedure.*;
 
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Stream;
 
 import static apoc.util.Util.map;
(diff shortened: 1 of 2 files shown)


sarmbruster:
This is https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1474 ported back to 3.5 branch.

src/main/java/apoc/trigger/Trigger.java
```diff
@@ -148,80 +149,107 @@ public TriggerInfo toTriggerInfo(Map.Entry<String, Object> e) {
...
+
+        private TriggerHandler(GraphDatabaseAPI api, Log log) {
             properties = api.getDependencyResolver().resolveDependency(EmbeddedProxySPI.class).newGraphPropertiesProxy();
 //            Pools.SCHEDULED.submit(() -> updateTriggers(null,null));
             this.log = log;
+            this.db = api;
         }
 
-        public static void checkEnabled() {
+        public static TriggerHandler initialize(GraphDatabaseAPI api, Log log) {
```
jakewins, line 192: Should this be `synchronized`, in case multiple callers invoke procedures simultaneously?

jakewins: Oop, nvm, only called from start()

Issue #1489, opened 2020-04-28
Join labels configuration param for Sheet name in apoc.export.xls


conker84:
## Feature description (Mandatory)
We should provide a configuration param for the `apoc.export.xls` in order to manage the allow the user to have XLS sheets with joined labels.

This because if we have a dataset like this:

```
CREATE (u:User:Customer{name: 'Andrea'})
```

We now generate an XLS with 2 sheets (`User` and `Customer`) with exactly the same data.
Could be a good idea give to the developer the power to choose if join these two sheets into one, by simply join labels.

PR #1490, opened 2020-04-28
fixes #1488 and #1489: Export XLS improvements

### LINKED ISSUE #1488: apoc.export.xls.* breaks when the relationship has the same name as label
`apoc.export.xls.*` breaks when the relationship has the same name as label.

Should import the data

The procedure throws an exception:

```
org.neo4j.graphdb.QueryExecutionException: Failed to invoke procedure `apoc.export.xls.graph`: Caused by: java.lang.IllegalArgumentException: The workbook already contains a sheet named 'COMPANY'

```

```
CREATE (u:User{name: 'Andrea'})-[r:COMPANY{since: 2018}]->(c:Company{name: 'Larus'})
```

### Steps (Mandatory)

```java
    @Test
    public void testExportQueryXlsWithSameLabelAndRelName() throws Exception {
        db.executeTransactionally("CREATE (u:User{name: 'Andrea'})-[r:COMPANY{since: 2018}]->(c:Company{name: 'Larus'})");
        String fileName = "query.xlsx";
        TestUtil.testCall(db, "CALL apoc.graph.fromCypher($query, {}, '', {}) YIELD graph AS exportedGraph " +
                        "CALL apoc.export.xls.graph(exportedGraph, $file, {}) YIELD file, source, format, time " +
                        "RETURN *",
                map("file", fileName,
                        "query", "MATCH (u:User{name: 'Andrea'})-[r:COMPANY]->(c:Company{name: 'Larus'}) return *"),
                (r) -> {
                    assertTrue("Should get statement",r.get("source").toString().equals("graph: nodes(2), rels(1)"));
                    assertEquals(fileName, r.get("file"));
                    assertEquals("xls", r.get("format"));
                });
        assertSheets(fileName, 3);
    }
```

conker84:
Fixes #1488 and #1489

For #1488 
- I added a new config param `prefixSheetWithEntityType` that prefix the sheet with the entity type `Node-` / `Rel-`

For #1489 
- I added a new config param `joinLabels` that joint labels for the sheet name (ie `:User:Customer`, become `User,Customer`)

### DIFF
--- a/src/test/java/apoc/export/csv/ExportXlsTest.java
+++ b/src/test/java/apoc/export/csv/ExportXlsTest.java
@@ -31,58 +31,6 @@ import static org.junit.Assert.assertTrue;
 
 public class ExportXlsTest {
 
-    private static final String EXPECTED_QUERY_NODES = String.format("\"u\"%n" +
-            "\"{\"\"id\"\":0,\"\"labels\"\":[\"\"User\"\",\"\"User1\"\"],\"\"properties\"\":{\"\"name\"\":\"\"foo\"\",\"\"age\"\":42,\"\"male\"\":true,\"\"kids\"\":[\"\"a\"\",\"\"b\"\",\"\"c\"\"]}}\"%n" +
-            "\"{\"\"id\"\":1,\"\"labels\"\":[\"\"User\"\"],\"\"properties\"\":{\"\"name\"\":\"\"bar\"\",\"\"age\"\":42}}\"%n" +
-            "\"{\"\"id\"\":2,\"\"labels\"\":[\"\"User\"\"],\"\"properties\"\":{\"\"age\"\":12}}\"");
-    private static final String EXPECTED_QUERY = String.format("\"u.age\",\"u.name\",\"u.male\",\"u.kids\",\"labels(u)\"%n" +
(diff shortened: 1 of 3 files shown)


mneedham:
@conker84 why don't we just do that prefixing and joining of labels anyway and not bother with the extra config params?

src/test/java/apoc/export/csv/ExportXlsTest.java
```diff
@@ -154,9 +140,9 @@ private void assertExcelFileForGraph(String fileName) {
             Workbook wb = WorkbookFactory.create(inp);
 
             int numberOfSheets = wb.getNumberOfSheets();
-            assertEquals(Iterables.count(tx.getAllLabels()) + Iterables.count(tx.getAllRelationshipTypes()), numberOfSheets);
+            assertEquals(Iterables.count(tx.getAllLabelsInUse()) + Iterables.count(tx.getAllRelationshipTypesInUse()), numberOfSheets);
```
mneedham, line 143: is this an unrelated change?

conker84: It's releated because at line 110 I create new labels `Customer` & `Company`, this makes fail the other tests now

conker84:
> @conker84 why don't we just do that prefixing and joining of labels anyway and not bother with the extra config params?

I used the config params so it's backwards compatible, because the defaults replicate the behaviour as-is.

PR #1494, opened 2020-05-02
semaphore constrained execution

sarmbruster:
I've added couple of procedures to manage named semaphores. E.g. using apoc.concurrent.runInSemaphore you can run a given statement inside a semaphore.
By configuration or a procedure call you can setup semaphores with a defined number of permits.
This enables to control of concurrency in environments where you have "stupid" clients that potentially overload the system. Or if you want prevent locking issues you can just set a default semaphore of 1.

### DIFF
--- a/docs/asciidoc/concurrent/semaphore.adoc
+++ b/docs/asciidoc/concurrent/semaphore.adoc
@@ -1,0 +1,69 @@
+[[semaphore]]
+= Cypher execution in semaphores
+
+[abstract]
+--
+This section describes a hook to run Cypher commands after database initialization.
+--
+
+Using a semaphore you can control concurrency. You can define semaphores either via configuration or create them explicitly.
+Semaphores do have a name and a number of available permits.
+
+### Configuration
+
+In `apoc.conf` you can define multiple semaphores
+
+[source,config]
+----
+apoc.semaphore.<mysemaphoreName>=<numberOfPermits>
+----
+
+Example:
+[source,config]
+----
+apoc.semaphore.onlyOnce=1
+----
+
+### Managing semaphores
+
+Additionally to config based setup you can use:
+
+[source,cypher]
+----
+CALL apoc.concurrent.addSemaphore("onlyOnce", 1)
+----
+
+To remove semaphores:
+
+[source,cypher]
+----
+CALL apoc.concurrent.removeSemaphore("onlyOnce")  // removes that specific semaphores
+CALL apoc.concurrent.removeAllSemaphores() // removes all semaphores
+----
+
+Note that the procedure based setup will only affect this instance.
+If you use them in a cluster, you have to run them on each cluster member to stay in sync.
+
+### Using semaphores
+
+You can wrap any cypher statement into a semaphore based execution:
+
+[source,cypher]
+----
+CALL apoc.concurrent.runInSemaphore(<semaphoreName>, <cypher>, <params / optional>) YIELD value
+RETURN value.n
+----
+
+`runInSemaphores` acquires a permit from the named semaphore. If there's no permit is available it will wait.
+While waiting a check for termination is run, effectively making it possible to kill waiting queries.
+After a permit has been acquired, the statement is run with the optionally provided parameters.
+Once finishing consuming the results the permit is passed back to the semaphore.
+
+If you have a semaphore with 1 permit this means effectively serialization of all operations on this semaphore.
+
+Example:
+[source,cypher]
+----
+CALL apoc.concurrent.runInSemaphore("onlyOnce", "MATCH (n) RETURN n") YIELD value
+RETURN value.n
+----
--- a/src/main/java/apoc/concurrent/SemaphoreProcedures.java
+++ b/src/main/java/apoc/concurrent/SemaphoreProcedures.java
@@ -1,0 +1,59 @@
+package apoc.concurrent;
+
+import apoc.result.MapResult;
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.procedure.Context;
+import org.neo4j.procedure.Description;
+import org.neo4j.procedure.Mode;
+import org.neo4j.procedure.Name;
+import org.neo4j.procedure.Procedure;
+import org.neo4j.procedure.TerminationGuard;
+import org.neo4j.procedure.UserFunction;
+
+import java.util.Map;
+import java.util.stream.Stream;
+
+public class SemaphoreProcedures {
+
+    @Context
+    public Transaction tx;
+
+    @Context
+    public TerminationGuard terminationGuard;
+
+    @Context
+    public Semaphores semaphores;
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.concurrent.runInSemaphore('mySemaphoreName', '<cypher>', {somekey: 'someValue', ...}) - runs the given cypher statement inside a semaphore to control concurrent access")
+    public Stream<MapResult> runInSemaphore(
+            @Name("semaphoreName") String semaphoreName,
+            @Name("cypher") String cypher,
+            @Name(value="params", defaultValue="{}") Map<String, Object> params) {
+        return semaphores.withSemaphore(terminationGuard, semaphoreName, () -> tx.execute(cypher, params).stream().map(MapResult::new));
+    }
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.concurrent.removeAllSemaphores() - removes all registered semaphores")
+    public void removeAllSemaphores() {
+        semaphores.removeAll();
+    }
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.concurrent.addSemaphores('myName', 1) - creates a semaphore with given number of permits")
+    public void addSemaphore(@Name("semaphoreName") String semaphoreName, @Name("permits") long permits) {
+        semaphores.add(semaphoreName, (int) permits);
+    }
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.concurrent.removeSemaphore() - removes given semaphore")
+    public void removeSemaphore(@Name("semaphoreName") String semaphoreName) {
+        semaphores.remove(semaphoreName);
+    }
+
+    @UserFunction()
+    @Description("apoc.concurrent.availablePermits(semaphoreName) - function returning the available number of permits for given semaphore")
+    public long availablePermits(@Name("semaphoreName") String semaphoreName) {
+        return semaphores.availablePermits(semaphoreName);
+    }
+}
--- a/src/main/java/apoc/concurrent/Semaphores.java
+++ b/src/main/java/apoc/concurrent/Semaphores.java
@@ -1,0 +1,105 @@
+package apoc.concurrent;
+
+import apoc.ApocConfig;
+import org.neo4j.internal.helpers.collection.Iterators;
+import org.neo4j.kernel.lifecycle.LifecycleAdapter;
+import org.neo4j.logging.Log;
+import org.neo4j.logging.internal.LogService;
+import org.neo4j.procedure.TerminationGuard;
+import org.neo4j.procedure.impl.GlobalProceduresRegistry;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+/**
+ * holds a per-database level map of declared semaphores
+ */
+public class Semaphores extends LifecycleAdapter {
+
+    private final Map<String, Semaphore> semaphoreMap = new ConcurrentHashMap<>();
+
+    private final Log log;
+    private final GlobalProceduresRegistry globalProceduresRegistry;
+    private final ApocConfig apocConfig;
+
+    public Semaphores(LogService log, GlobalProceduresRegistry globalProceduresRegistry, ApocConfig apocConfig) {
+
+        this.log = log.getInternalLog(Semaphores.class);
+        this.globalProceduresRegistry = globalProceduresRegistry;
+        this.apocConfig = apocConfig;
+
+        // expose this config instance via `@Context ApocConfig config`
+        globalProceduresRegistry.registerComponent((Class<Semaphores>) getClass(), ctx -> this, true);
+        this.log.info("successfully registered Pools for @Context");
+    }
+
+    @Override
+    public void init() throws Exception {
+        Iterators.stream(apocConfig.getKeys("apoc.semaphore")).forEach(key -> {
+            int permits = apocConfig.getInt(key, -1);
+            semaphoreMap.put(key, new Semaphore(permits));
+            log.debug("created semaphore %s with %d permits", key, permits);
+        });
+    }
+
+    public <T> Stream<T> withSemaphore(TerminationGuard terminationGuard, String semaphoreName, Supplier<Stream<T>> supplier) {
+        Semaphore semaphore = getSemaphoreOrThrow(semaphoreName);
+        long now = System.currentTimeMillis();
+        boolean permitAcquired = false;
+        try {
+
+            // we can't simply use semaphore.acquire() here since it blocks and doesn't give us a chance for checking termination status
+            log.debug("about to acquire semaphore %s", semaphoreName);
+            while (!semaphore.tryAcquire(100, TimeUnit.MILLISECONDS)) {
+                terminationGuard.check();
+            }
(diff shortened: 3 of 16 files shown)


docs/asciidoc/concurrent/semaphore.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+You can wrap any cypher statement into a semaphore based execution:
+
+[source,cypher]
+----
+CALL apoc.concurrent.runInSemaphore(<semaphoreName>, <cypher>, <params / optional>) YIELD value
+RETURN value.n
+----
+
+`runInSemaphores` acquires a permit from the named semaphore. If there's no permit is available it will wait.
```
jexp, line 57: would it make sense to add an optional timeout? (perhaps even with an default?)

docs/asciidoc/concurrent/semaphore.adoc
```diff
@@ -0,0 +1,69 @@
...
+Note that the procedure based setup will only affect this instance.
+If you use them in a cluster, you have to run them on each cluster member to stay in sync.
+
+### Using semaphores
+
+You can wrap any cypher statement into a semaphore based execution:
+
+[source,cypher]
+----
+CALL apoc.concurrent.runInSemaphore(<semaphoreName>, <cypher>, <params / optional>) YIELD value
```
jexp, line 53: do you differentiate between updates and read only? b/c some users might only be able to do read-only queries?

docs/asciidoc/concurrent/semaphore.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+[abstract]
+--
+This section describes a hook to run Cypher commands after database initialization.
+--
+
+Using a semaphore you can control concurrency. You can define semaphores either via configuration or create them explicitly.
+Semaphores do have a name and a number of available permits.
+
+### Configuration
```
jexp, line 12: In Adoc: `==`

src/main/java/apoc/concurrent/SemaphoreProcedures.java
```diff
@@ -0,0 +1,59 @@
...
+        return semaphores.withSemaphore(terminationGuard, semaphoreName, () -> tx.execute(cypher, params).stream().map(MapResult::new));
+    }
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.concurrent.removeAllSemaphores() - removes all registered semaphores")
+    public void removeAllSemaphores() {
+        semaphores.removeAll();
+    }
+
+    @Procedure(mode = Mode.WRITE)
```
jexp, line 42: it would be nice to have a function / proc that lists all the semaphores, their permits and how many are currently taken.

src/main/java/apoc/concurrent/SemaphoreProcedures.java
```diff
@@ -0,0 +1,59 @@
...
+    @Description("apoc.concurrent.addSemaphores('myName', 1) - creates a semaphore with given number of permits")
+    public void addSemaphore(@Name("semaphoreName") String semaphoreName, @Name("permits") long permits) {
+        semaphores.add(semaphoreName, (int) permits);
+    }
+
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.concurrent.removeSemaphore() - removes given semaphore")
```
jexp, line 49: misses parameter name, there is no indication e.g. in results if that semaphore actually existed, and also if it still had threads waiting on it? what happens then?

src/main/java/apoc/concurrent/Semaphores.java
```diff
@@ -0,0 +1,116 @@
...
+    private final ApocConfig apocConfig;
+
+    public Semaphores(LogService log, GlobalProceduresRegistry globalProceduresRegistry, ApocConfig apocConfig) {
+
+        this.log = log.getInternalLog(Semaphores.class);
+        this.globalProceduresRegistry = globalProceduresRegistry;
+        this.apocConfig = apocConfig;
+
+        // expose this config instance via `@Context ApocConfig config`
+        globalProceduresRegistry.registerComponent((Class<Semaphores>) getClass(), ctx -> this, true);
```
jexp, line 36: we should probably also register other components that we currently access with static

src/main/java/apoc/concurrent/Semaphores.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public void init() throws Exception {
+        Iterators.stream(apocConfig.getKeys(ApocConfig.APOC_SEMAPHORE_PREFIX))
+                .filter(key -> !key.equals(ApocConfig.APOC_SEMAPHORE_DEFAULT_NAME))
+                .forEach(key -> {
+                    int permits = apocConfig.getInt(key, -1);
+                    String semaphoreName = key.substring(ApocConfig.APOC_SEMAPHORE_PREFIX.length() + 1);
+                    semaphoreMap.put(semaphoreName, new Semaphore(permits));
+                    log.debug("created semaphore %s with %d permits", key, permits);
```
jexp, line 48: could be log.info perhaps

src/main/java/apoc/concurrent/Semaphores.java
```diff
@@ -0,0 +1,116 @@
...
+        if (semaphoreName == null) {
+            return supplier.get();
+        } else {
+            Semaphore semaphore = getSemaphoreOrThrow(semaphoreName);
+            long now = System.currentTimeMillis();
+            boolean permitAcquired = false;
+            try {
+
+                // we can't simply use semaphore.acquire() here since it blocks and doesn't give us a chance for checking termination status
+                log.debug("about to acquire semaphore %s", semaphoreName);
```
jexp, line 62: if (Log.isDebugEnabled()) oder so

jexp: for all log.debug in the hot path

src/main/java/apoc/concurrent/Semaphores.java
```diff
@@ -0,0 +1,116 @@
...
+                permitAcquired = true;
+                log.debug("acquired semaphore %s in %d millis", semaphoreName, System.currentTimeMillis()-now);
+
+                return supplier.get().onClose(() -> {
+                    semaphore.release();
+                    log.debug("released semaphore %s after %d millis", semaphoreName, System.currentTimeMillis()-now);
+                });
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            } catch (RuntimeException e) {
```
jexp, line 75: what happens with other exceptions?

src/main/java/apoc/concurrent/Semaphores.java
```diff
@@ -0,0 +1,116 @@
...
+                }
+                permitAcquired = true;
+                log.debug("acquired semaphore %s in %d millis", semaphoreName, System.currentTimeMillis()-now);
+
+                return supplier.get().onClose(() -> {
+                    semaphore.release();
+                    log.debug("released semaphore %s after %d millis", semaphoreName, System.currentTimeMillis()-now);
+                });
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
```
jexp, line 74: should this also release? in case the InterruptedException didn't come from the tryAcquire but someplace else?

src/main/java/apoc/concurrent/Semaphores.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public void add(String semaphoreName, int permits) {
+        if (semaphoreMap.get(semaphoreName) != null) {
+            throw new IllegalArgumentException("already existing semaphore :" + semaphoreName + ". Use remove first");
+        }
+        semaphoreMap.put(semaphoreName, new Semaphore(permits));
+    }
+
+    public void remove(String semaphoreName) {
+        semaphoreMap.remove(semaphoreName);
```
jexp, line 98: should this also throw on failure? so that typos are spotted and semaphores don't remain that shouldn't?

jexp: we don't do anything with the blocked threads when we remove them?

src/main/java/apoc/concurrent/Semaphores.java
```diff
@@ -0,0 +1,116 @@
...
+    private Semaphore getSemaphoreOrThrow(String semaphoreName) {
+        Semaphore semaphore = semaphoreMap.get(semaphoreName);
+        if (semaphore == null) {
+            throw new IllegalArgumentException("no declared semaphore: " + semaphoreName);
+        }
+        return semaphore;
+    }
+
+    public void removeAll() {
+        semaphoreMap.clear();
```
jexp, line 114: should this return which ones were there and have been removed incl. names?

src/main/java/apoc/periodic/Periodic.java
```diff
@@ -10,10 +12,26 @@
 import org.neo4j.internal.helpers.collection.Iterators;
 import org.neo4j.internal.helpers.collection.Pair;
 import org.neo4j.logging.Log;
-import org.neo4j.procedure.*;
-
-import java.util.*;
-import java.util.concurrent.*;
+import org.neo4j.procedure.Context;
```
jexp, line 15: we shouldn't change the imports back and forth :)

src/main/java/apoc/periodic/Periodic.java
```diff
@@ -288,21 +309,26 @@ private void validateQueries(Map<String, String> fieldStatement) {
...
-        int failedParams = Util.toInteger(config.getOrDefault("failedParams", -1));
-        try (Result result = tx.execute(slottedRuntime(cypherIterate),params)) {
-            Pair<String,Boolean> prepared = prepareInnerStatement(cypherAction, iterateList, result.columns(), "_batch");
-            String innerStatement = prepared.first();
-            iterateList=prepared.other();
-            log.info("starting batching from `%s` operation using iteration `%s` in separate thread", cypherIterate,cypherAction);
-            return iterateAndExecuteBatchedInSeparateThread((int)batchSize, parallel, iterateList, retries, result,
-                    (tx, p) -> Iterators.count(tx.execute(innerStatement, merge(params, p))), concurrency, failedParams);
-        }
+        String semaphoreName = (String) config.getOrDefault("semaphore", apocConfig.getString(APOC_SEMAPHORE_DEFAULT_NAME));
```
jexp, line 312: I'm not sure about this default even when none is set ...

src/main/java/apoc/periodic/Periodic.java
```diff
@@ -350,37 +376,38 @@ public Pattern regNoCaseMultiLine(String pattern) {
...
-                                if (localCount.get() % 1000 == 0 && Util.transactionIsTerminated(terminationGuard)) {
-                                    return 0;
-                                }
-                                Map<String, Object> params = merge(p, Util.map("_count", localCount.get(), "_batch", batch));
-                                return executeAndReportErrors(txInThread, consumer, params, batch, 1, localCount, collector);
-                            }).mapToLong(n -> (Long) n).sum();
-                };
-            }
-            futures.add(Util.inTxFuture(log, pool, db, task, retries, aLong -> collector.incrementRetried(), _ignored -> collector.incrementBatches()));
+        ExecutorService pool = parallel ? pools.getDefaultExecutorService() : Executors.newSingleThreadExecutor();
```
jexp, line 379: shouldn't the pools management still stay in Pools?

src/main/java/apoc/periodic/Periodic.java
```diff
@@ -395,18 +422,28 @@ public Pattern regNoCaseMultiLine(String pattern) {
...
+            ToLongFunction<Future<Long>> toLongFunction = wasTerminated ?
+                    f -> Util.getFutureOrCancel(f, collector.getBatchErrors(), collector.getFailedBatches(), 0L) :
+                    f -> Util.getFuture(f, collector.getBatchErrors(), collector.getFailedBatches(), 0L);
+            collector.incrementSuccesses(futures.stream().mapToLong(toLongFunction).sum());
+
+            Util.logErrors("Error during iterate.commit:", collector.getBatchErrors(), log);
+            Util.logErrors("Error during iterate.execute:", collector.getOperationErrors(), log);
+            return Stream.of(collector.getResult());
+        } finally {
+            if (parallel==false) { // we need to ensure destruction of the executor we've created above
```
jexp, line 438: pools management in pools?

src/main/java/apoc/periodic/Periodic.java
jexp: perhaps a withPool(type) ? or so? and pass a lambda to it.
perhaps in a separate PR

src/main/java/apoc/periodic/Periodic.java
```diff
@@ -395,18 +422,28 @@ public Pattern regNoCaseMultiLine(String pattern) {
...
+                pool.shutdown();
+                try {
+                    pool.awaitTermination(10, TimeUnit.SECONDS);
```
jexp, line 441: what happens if the update statements are long running statements ...
either give it much more time like 10 or 30 mins
or ... configurable? or ...

the single pool was meant to avoid this creating and shutting down pools but serialize everything

src/test/java/apoc/concurrent/SemaphoreProceduresTest.java
```diff
@@ -0,0 +1,125 @@
...
+        thrown.expectMessage("java.lang.IllegalArgumentException: no declared semaphore: undeclaredSemaphore");
+        String cypher = "call apoc.concurrent.runInSemaphore($name, 'return 1')";
+
+        Map<String, Object> params = MapUtil.map("name", "undeclaredSemaphore");
+        db.executeTransactionally(cypher, params, Iterators::single);
+    }
+
+    @Test
+    public void testThatSemaphoreSerializesOperations() throws InterruptedException {
+        long duration = 200;
```
jexp, line 57: should we check that the real duration is not already twich the duration value?

src/test/java/apoc/concurrent/SemaphoreProceduresTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+        assertTrue( (System.currentTimeMillis()- now) >= 2*duration);
+        long availablePermits = TestUtil.singleResultFirstColumn(db, "return apoc.concurrent.availablePermits($semaphoreName)", params);
+        assertEquals(1l, availablePermits);
+    }
+
+    @Test
+    public void testSemaphoreReleasedOnFailure() throws InterruptedException {
+        String cypher = "call apoc.concurrent.runInSemaphore($semaphoreName, 'call does.not.exist() return 1')";
+        Map<String, Object> params = MapUtil.map("semaphoreName", "single");
```
jexp, line 88: constant

src/test/java/apoc/concurrent/SemaphoreProceduresTest.java
```diff
@@ -0,0 +1,125 @@
...
+    @Test(timeout = 10000)
+    public void testRunInSemaphoreIsKillable() throws InterruptedException {
+        Thread thread = new Thread(() -> {
+            db.executeTransactionally("call apoc.concurrent.runInSemaphore($semaphoreName, 'return 1')", Collections.singletonMap("semaphoreName", SEMAPHORE_ZERO), Iterators::single);
+        });
+        thread.start();
+        Thread.sleep(100);  // await planning
+
+        // kill all active queries
+        DependencyResolver dependencyResolver = ((GraphDatabaseAPI) db).getDependencyResolver();
```
jexp, line 116: should we test this also from a users perspective? like listQueries/killQuery?

what happens if the inner query is killed?

src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -577,6 +587,74 @@ public void testRockNRollWhileFail() {
         }
     }
 
+    @Test
+    public void testPeriodicSemaphore() throws InterruptedException {
+        long duration = 100;
```
jexp, line 592: same as above about the measurement

src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -577,6 +587,74 @@ public void testRockNRollWhileFail() {
...
+
+        Set<Thread> threads = new HashSet<>();
+        for (int i=0; i < times; i++) {
+            threads.add(new Thread(() -> {
+                try {
+                    Map<String, Object> r = db.executeTransactionally(cypher, params, Iterators::single);
+                    Map<String, Object> errorMessages = (Map<String, Object>) r.get("errorMessages");
+                    if (!errorMessages.isEmpty()) {
+                        failureInThread.set(true);
+//                        System.out.println("failed: " + errorMessages);
```
jexp, line 620: remove sysout

fbiville:
@sarmbruster would you have some time to review @jexp suggestions (and rebase)?

PR #1508, opened 2020-05-12
Apoc periodic iterate

mneedham:
This PR introduces `batchMode`, which deprecates `iterateList`. It also introduces an extra mode, which can be used to access a batch of values returned by the data-driven statement in the operations statement.

Example of what you might use that for:

Tweaking of the periodic iterate documentation to explain `batchMode`

Explanation of the new `BATCH_SINGLE` mode

### DIFF
--- a/docs/asciidoc/graph-updates/periodic-iterate.adoc
+++ b/docs/asciidoc/graph-updates/periodic-iterate.adoc
@@ -1,5 +1,5 @@
 [[commit-batching]]
-== apoc.periodic.iterate
+== Periodic Iterate
 
 ifdef::backend-html5[]
 ++++
@@ -8,14 +8,15 @@ ifdef::backend-html5[]
 endif::[]
 
 The `apoc.periodic.iterate` procedure is helpful when you need to handle large amounts of data for import, refactoring, and other cases that require large transactions.
-It provides a way to batch the data by dividing the workload into two parts - a data-driven statement and an operations statement.
+It provides a way to batch the data by dividing the workload into two parts:
 
-The data-driven statement is how you select what data needs handled.
+a data-driven statement:: This defines how you select what data needs handled.
 You can provide a Cypher statement to select from existing graph data, read external data from a file or API, or retrieve data from another datastore.
-The operations statement is what you want done to the selected data.
+
+an operation statement:: This defines what you want done to the selected data.
 You can do things like execute Cypher for updating or creating/deleting the data or run other procedures to manipulate and transform values before loading.
 
-To run `apoc.periodic.iterate`, the data-driven statement is provided as the *first*, outer statement that results in a stream of values to be processed.
+The data-driven statement is provided as the *first*, outer statement that results in a stream of values to be processed.
 The operations statement is provided as the *second*, inner statement to process *one* element at a time or (with `iterateList:true`) a batch at a time.
 The results of the outer statement are passed to the inner statement as parameters, so they are automatically made available with their names.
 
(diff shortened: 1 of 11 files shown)


docs/asciidoc/graph-updates/periodic-iterate.adoc
```diff
@@ -69,6 +112,74 @@ CALL apoc.periodic.iterate(
...
 ----
 
-In the above code, we select all the `Order` nodes that have an order date greater than `October 13, 2016` (first Cypher statement).
-Our configuration will batch those nodes into groups of 100 (`batchSize:100`) and run the batches in parallel for the second statement to process.
-The second Cypher statement takes those groups and finds the nodes that have a `HAS_ITEM` relationship to other nodes, then sums up the value of those items and sets that sum as a property (`o.value`) for the total order value.
\ No newline at end of file
+Let's break down the parameters passed to the procedure:
+
+* Our first Cypher statement selects all the `Order` nodes that have an order date greater than `October 13, 2016` (first Cypher statement).
+* Out second Cypher statement takes those groups and finds the nodes that have a `HAS_ITEM` relationship to other nodes, then sums up the value of those items and sets that sum as a property (`o.value`) for the total order value.
```
JMHReif, line 118: `Our` rather than Out? :)

docs/asciidoc/graph-updates/periodic-iterate.adoc
```diff
@@ -69,6 +112,74 @@ CALL apoc.periodic.iterate(
...
+----
+
+The contents of the `$_batch` parameter that is used in the operations statement would be as follows:
+
+[source,text]
+----
+[{p: Node<1>}, {p: Node<2>},...]
+----
+
+We an use a https://neo4j.com/docs/cypher-manual/current/syntax/lists/#cypher-list-comprehension[list comprehension^] to extract the `p` variable from each item in the list.
```
JMHReif, line 176: `We can use` rather than We an use? :)

docs/asciidoc/graph-updates/periodic-iterate.adoc
```diff
@@ -8,32 +8,68 @@ ifdef::backend-html5[]
...
 
-.configuration options
+.Config
+[options=header]
+|===
+| name | type | default | description
+| batchSize | Long | 10000 | run the specified number of operations statements in a single tx - params: {_count, _batch}
+| parallel | boolean | false | run operations statements in parallel (note that statements might deadlock if conflicting)
+| retries | Long | 0 | if the operations statement fails with an error, sleep 100ms and retry until retries-count is reached - param {_retry}
+| batchMode | String | "BATCH" a| how data-driven statements should be processed by outer statement. Valid values are:
```
JMHReif, line 30: Should it be how the data-driven statement is processed by the INNER statement? Because the data-driven statement = outer statement?

mneedham: Haha, I was getting so confused by outer and inner! I've removed those words and replaced with data-driven and operation now

src/main/java/apoc/periodic/BatchMode.java
```diff
@@ -0,0 +1,44 @@
...
+    BATCH_SINGLE,
+    SINGLE;
+
+
+    private static final Map<String, BatchMode> nameIndex =  new HashMap<>();
+    static {
+        for (BatchMode batchMode : BatchMode.values()) {
+            nameIndex.put(batchMode.name(), batchMode);
+        }
+    }
```
conker84, line 20: I don't think we need this you can use `BatchMode#valueOf` method instead

src/main/java/apoc/periodic/BatchMode.java
```diff
@@ -0,0 +1,44 @@
...
+
+   static BatchMode fromIterateList(boolean iterateList) {
+        return iterateList ? BATCH : SINGLE;
+    }
+
+    static BatchMode fromConfig(Map<String,Object> config) {
+        Object batchMode = config.get("batchMode");
+
+        if(batchMode != null) {
+            BatchMode lookedUpBatchMode = nameIndex.get(batchMode.toString().toUpperCase());
```
conker84, line 30: if we get rid of `nameIndex` you can change it into:
```suggestion
            BatchMode lookedUpBatchMode = BatchMode.valueOf(batchMode.toString().toUpperCase());
```

src/main/java/apoc/periodic/PeriodicIterate.java
```diff
@@ -0,0 +1,44 @@
...
+
+import apoc.util.Util;
+import org.neo4j.internal.helpers.collection.Pair;
+
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class PeriodicIterate {
```
conker84, line 11: Maybe a better name could be `PeriodicUtils`?

And add a private constructor so we can prevent the initialization

src/main/java/apoc/periodic/BatchMode.java
conker84: And we can also get rid of the following exception

mneedham:
@conker84 https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1508#discussion_r423841544

I wanted to return the list of valid values in the exception, which the standard exception from `valueOf()` doesn't have

conker84:
> @conker84 [#1508 (comment)](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1508#discussion_r423841544)
> 
> I wanted to return the list of valid values in the exception, which the standard exception from `valueOf()` doesn't have

Ok

mneedham:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1508#discussion_r423842994


Good idea. Have renamed and added private const

PR #1512, opened 2020-05-18
Porting ldap 4.0

mneedham:
Porting the LDAP refactoring to 4.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -109,6 +109,10 @@ dependencies {
 
     compile 'org.jsoup:jsoup:1.11.3'
 
+    compileOnly group: 'org.apache.directory.api', name: 'api-all', version: '2.0.1'
+    testImplementation group: 'org.apache.directory.api', name: 'api-all', version: '2.0.1'
+    testCompile group: 'org.apache.directory.server', name: 'apacheds-test-framework', version: '2.0.0.AM26', { exclude group: 'org.apache.directory.api', module: 'api-ldap-schema-data' }
+
     compile group: 'org.roaringbitmap', name: 'RoaringBitmap', version: '0.7.17'
     compile group: 'org.apache.commons', name: 'commons-configuration2', version: '2.7'
     testCompile group: 'com.github.stefanbirkner', name: 'system-rules', version: '1.19.0'
--- a/docs/asciidoc/database-integration/loadldap.adoc
+++ b/docs/asciidoc/database-integration/loadldap.adoc
@@ -93,9 +93,9 @@ foreach (member in entry.uniqueMember |
 
 === Credentials
 
-To protect credentials, you can configure aliases in `conf/neo4j.conf`:
+To protect credentials, you can configure aliases in `conf/apoc.conf`:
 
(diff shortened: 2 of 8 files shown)


mneedham:
@ tuftedocelot I've tried to port this to 4.0. Main thing I'm not sure about is the name of the properties in apoc.conf. I'm wondering whether it would be simpler to just use the existing static value storage functions/procedures rather than building it into the LDAP procedure itself - https://neo4j.com/docs/labs/apoc/current/misc/static-values/

tuftedocelot:
oho, I didn't know the static value storage was a thing! This does look way easier.

I'll start on a new PR that will adopt that usage instead and then your 4.0 port can be layered atop that?

tuftedocelot:
Actually, probably a good thing that everything was reverted. The old-style configs will continue to be the only compatibility thing to maintain for a while

tuftedocelot:
@mneedham, making sure I understand correctly, you are envisioning something like this going forward?

```
WITH apoc.static.getAll("myldapconfig") AS config
CALL apoc.load.ldapurl(config)
```

where the `config` map has the `url`, `username`, `password`, etc keys/values set

tuftedocelot:
Thoughts on how this matches (roughly?) up with what you had in mind? https://github.com/tuftedocelot/neo4j-apoc-procedures/commit/2afe02f3789a5d6610f100f6e65cd675952853f5

I like this much better than the original approach. I think the next logical step is to do away with the concept of "aliases" for the procedure and always take in the Map returned from `apoc.static.getAll`.

If that seems like best way to go, I'll sort out a rebase of my branch against 3.5 and see if the test hanging goes away (https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1429#issuecomment-630607525). Strange that the hangs occurred once the branch was pulled in considering prior tests passed

mneedham:
> @mneedham, making sure I understand correctly, you are envisioning something like this going forward?
> 
> ```
> WITH apoc.static.getAll("myldapconfig") AS config
> CALL apoc.load.ldapurl(config)
> ```
> 
> where the `config` map has the `url`, `username`, `password`, etc keys/values set

Yup, exactly that

mneedham:
> Thoughts on how this matches (roughly?) up with what you had in mind? [tuftedocelot@2afe02f](https://github.com/tuftedocelot/neo4j-apoc-procedures/commit/2afe02f3789a5d6610f100f6e65cd675952853f5)
> 
> I like this much better than the original approach. I think the next logical step is to do away with the concept of "aliases" for the procedure and always take in the Map returned from `apoc.static.getAll` or whatever the user passed in.
> 
> If that seems like best way to go, I'll sort out a rebase of my branch against 3.5 and see if the test hanging goes away ([#1429 (comment)](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1429#issuecomment-630607525)). Strange that the hangs occurred once the branch was pulled in considering prior tests passed

Yep that looks good. I think that'll make the code much simpler

PR #1516, opened 2020-05-21
fixes #1336: Create apoc.export.cypher format and outputs optimised for passing to apoc.bolt.execute

### LINKED ISSUE #1336: Create apoc.export.cypher format and outputs optimised for passing to apoc.bolt.execute

conker84:
Fixes #1336

- `apoc.bolt.load.fromLocal($url, $localStatement, $remoteStatement, $config)`: which allow to execute a statement into `local` Neo4j instance with the `localStatement` variable and use the result of the execution in order use them in combination with the `remoteStatement` in the `remote` Neo4j instance. **The procedure can be useful when the user needs to call `apoc.bolt.load` more than once (maybe in an unwind) because on each call we create a new Driver (& Session) instance that can slow down the performances, while into the `fromLocal` procedure we create them once**
Please look at the related tests `testLoadFromLocal` and `testLoadFromLocalStream` into the `BoltTest` class
It the `config` contains `streamStatements` as true, the procedure assumes that the output of the `localStatement` execution is a table rows of two columns:
  - `statement`: a cypher string executed into the remote statement (mandatory)
  - `params`: a map of params related to the cypher statement (optional)

I'm closing the #1361 which is related to the branch `3.5`

  - Implemented the procedure
  - Added tests

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/database-integration/bolt-neo4j.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/database-integration/bolt-neo4j.adoc
@@ -11,6 +11,8 @@ Bolt procedures allows to accessing other databases via bolt protocol.
 |===
 | CALL apoc.bolt.execute(urlOrKey, statement, params, config) YIELD row  | access to other databases via bolt for read and write
 | CALL apoc.bolt.load(urlOrKey, statement, params, config) YIELD row | access to other databases via bolt for read
+| CALL apoc.bolt.fromLocal(urlOrKey, localStatement, remoteStatement, config) YIELD row | executes the `localStatement`
+on the local instance and uses the results a parameters for the `remoteStatement` executed on the remote instance
 |===
 
 **urlOrKey** param allows users to decide if send url by apoc or if put it into apoc.conf file.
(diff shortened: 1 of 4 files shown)


mneedham:
@conker84 needs rebasing

conker84:
@mneedham done

conker84:
n.b. the commit should be cherry-picked into `4.1`

PR #1529, opened 2020-05-27
Re-add the LoadLdap procedure post-reversion

### LINKED ISSUE #1428: Overhaul LoadLdap procedure with Apache LDAP API 2.0
The current version of LoadLdap has a number of drawbacks mostly due to the use of the jLDAP library. Additionally, LDAP URLs are a easier method of specifying LDAP servers and search parameters and [was noted in the initial comments](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/537#pullrequestreview-54337093). Storing predefined LDAP URLs and credentials in APOC's configuration is also not possible

- jLDAP doesn't like nonconformant schemas which impacts 389DS, SunONE, and ODSEE directory servers
- No support for paging was included which can impact servers which implement result size limits
- jLDAP hasn't seen updates and [is effectively abandoned](https://directory.apache.org/api/vision.html)
- Apparently issues with searching from the root of a domain #1150 

## Considered alternatives
Implementing the paging, researching the issue behind searching from the domain root, and adding support for LDAP URLs are possibilities. However, with the use of jLDAP, APOC will be fundamentally limited in the LDAP servers that it supports.

By using Apache's LDAP API 2.0 and rewriting the module, it should be possible to:

- Support a wider range of LDAP servers
- Allow looser LDAP searches from the root of a domain
- Store LDAP URLs/searches and associated credentials in APOCs configuration
- Use paged searches to avoid server-imposed search size limits

tuftedocelot:
Fixes #1428, improves upon #1429, and a step towards #1512 

Rewrite the LoadLdap procedure to incorporate new features and enhancements and use APOC static values to store the configurations

- Support a wider range of LDAP servers
- Allow looser LDAP searches from the root of a domain
- Store LDAP URLs/searches and associated credentials in APOCs configuration
- Use paged searches to avoid server-imposed search size limits
- Handle Active Directory's use of ranged attribute retrieval

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,8 @@ dependencies {
     compile group: 'com.jayway.jsonpath', name: 'json-path', version: '2.4.0'
     compile group: 'org.hdrhistogram', name: 'HdrHistogram', version: '2.1.9'
     compile group: 'org.neo4j.driver', name: 'neo4j-java-driver', version: '1.7.3'
-    compile group: 'com.novell.ldap', name: 'jldap', version: '2009-10-07'
+    compile group: 'org.apache.directory.api', name: 'api-all', version: '2.0.1'
+    testCompile group: 'org.apache.directory.server', name: 'apacheds-test-framework', version: '2.0.0.AM26', { exclude group: 'org.apache.directory.api', module: 'api-ldap-schema-data' }
 
     def withoutAntlr =  {
         exclude group: 'org.glassfish'
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,8 @@ dependencies {
     compile group: 'com.jayway.jsonpath', name: 'json-path', version: '2.4.0'
     compile group: 'org.hdrhistogram', name: 'HdrHistogram', version: '2.1.9'
     compile group: 'org.neo4j.driver', name: 'neo4j-java-driver', version: '1.7.3'
-    compile group: 'org.apache.directory.api', name: 'api-all', version: '2.0.1'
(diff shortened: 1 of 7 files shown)


tuftedocelot:
@mneedham, this is interesting. Running the tests locally, they all pass fine. In Travis CI, the tests failed and you can't see/download all of the output because it's too long.

I ran the LoadLdapTest suite this time redirecting the output to a file. 87MB output! Seems a bit excessive. It looks like the DEBUG output level is very verbose when it comes to the library. In the last [successful job](https://travis-ci.org/github/neo4j-contrib/neo4j-apoc-procedures/builds/676923432), I don't see any such DEBUG level output. Nothing looked obviously different in build.gradle when it came to log settings, but I won't pretend to be a gradle expert either

mneedham:
yeh I don't really know why it doesn't run on CI.  But one difference will be that on CI the tests are running inside a Docker container, so any network requests might be referring to the wrong network addresses?

tuftedocelot:
I wouldn't think so? The LDAP server which gets spun up for tests is just on localhost

tuftedocelot:
Well, still fails :slightly_frowning_face: 

Nothing still seems obviously wrong aside from the fact that it appears DEBUG level logging starts happening. Anything the CI image or CI configuration that changed in the recent past? Nothing that I can see has changed when it comes to the gradle settings either before or after the merge was attempted. If that is indeed the case, then what's changed in the CI?

My first thought was that there was something in the version bumps for the api-all and DS, but those happened before these symptoms started appearing

fbiville:
Closing this PR as it has remained open for quite a long time. Feel free to reopen if needed.

Issue #1458, opened 2020-03-25
apoc.load.jdbcUpdate doesn't work with apoc.periodic.iterate


conker84:
_From the community slack channel_

Hi All,
I have the following code (removed and changed column names for privacy reasons, so it may not look completely proper), using apoc.periodic.iterate, it doesn't give any errors, but the data is not found in the sqlserver database. If I run the code without the apoc.periodic.iterate, the database is filled properly.

```
// Read project
CALL apoc.periodic.iterate("
MATCH (p:Project)
WHERE p.p6_id = 'LVL1-NEO4J' 
WITH p.p6_wbs_object_id AS col2
                ,p.p6_object_id AS col3
// Find all LVL1 activities
MATCH (n:LVL1) WITH collect(n) AS b, col2, col3 UNWIND range(0, size(b)-1) AS pos WITH b[pos] AS n, col2, col3 RETURN n{.*, col2, col3}","CALL apoc.load.jdbcUpdate('jdbc:sqlserver://<connection_settings>','INSERT INTO activities (col1, col2, col3) values(?,?,?)',[n.col1, n.col2, n.col3]) YIELD row RETURN 'Done'", {batchSize: 100})
YIELD committedOperations, failedOperations, failedBatches, errorMessages
RETURN *
```

jexp:
Did they try 

Return ... as n

jexp:
Do the current modes make the planning ignore the procs?

conker84:
I found @jexp the problem is that the procedure gets not executed because of the proc Mode. See the #1458

jexp:
@conker84 why did we have to change load.jdbc too?
It always returns data so it should not be optimized away?

It seems that:

>>> Write procedures are eagerly loaded into memory and from just glancing at the procedure it doesn't look like it needs the @Write ?

PR #1531, opened 2020-05-28
Change load.jdbc back to mode=READ

jexp:
as it always returns data it should never be optimized away

left loadJdbcUpdate as DBMS to avoid that there, although WRITE might be more correct there.

### DIFF
--- a/src/main/java/apoc/load/Jdbc.java
+++ b/src/main/java/apoc/load/Jdbc.java
@@ -56,7 +56,7 @@ public class Jdbc {
         }
     }
 
-    @Procedure(mode = Mode.WRITE)
+    @Procedure(mode = Mode.READ)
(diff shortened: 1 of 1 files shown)


jexp:
Feedback from engineering:

It seems that:

            Write procedures are eagerly loaded into memory and from just glancing at the procedure it doesn't look like it needs the @Write ?

https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1458

conker84:
@jexp if we switch it to `READ` it will be not used in combination with `apoc.periodic.*` procedures maybe we can load it as `DBMS`?

jexp:
The was afaik with json update And it not returning columns?

/cc @pontusmelke

jexp:
ping @conker84 
we can switch it back to read for load.jdbc
but leave DBMS for jdbcUpdate

PR #1538, opened 2020-06-02
Ensure periodic.iterate data is released once the computation completes

fickludd:
## Problem 

Fix issue with `apoc.periodic.iterate` keeping full intermediate result in memory until the procedure completed. 

## Proposed Changes

Ensure periodic.iterate data is released once the computation completes by using custom classes to capture the data needed for batch processing, and explicitly releasing these on batch completion.


fickludd:
@sarmbruster np, great that it was discovered. Should I do anything for this fix to also get added to the `4.1` branch or how does that work in apoc?

sarmbruster:
no need to worry, we've started a 4.1 branch right after merging this one.

Issue #1558, opened 2020-06-12
Issues with being able create multiple user defined functions with same name and signature


dfgitn4j:
## Expected Behavior (Mandatory)

Allow only one function name / signature combination to be created

## Actual Behavior (Mandatory)
Creates duplicates, can't remove them

## How to Reproduce the Problem

`apoc.custom.asFunction` allows duplicate function name / signature combinations.  Error when try and remove with `apoc.custom.removeFunction`

### Simple Dataset (where it's possibile)

```
CALL apoc.custom.asFunction('tooManyLifeUniverseEverythings','RETURN 42');
CALL apoc.custom.asFunction('tooManyLifeUniverseEverythings','RETURN 42');

```

### Steps (Mandatory)

  1. create same function twice with apoc.custom.asFunction
  1. try and remove function(s) with apoc.custom.removeFunction
 

## Screenshots (where it's possibile)

```
CALL apoc.custom.list
╒══════════╤════════════════════════════════╤═════════════╤══════╤═══════════╤═══════════════════════╤══════════════╤═════════════╕
│"type"    │"name"                          │"description"│"mode"│"statement"│"inputs"               │"outputs"     │"forceSingle"│
╞══════════╪════════════════════════════════╪═════════════╪══════╪═══════════╪═══════════════════════╪══════════════╪═════════════╡
│"function"│"tooManyLifeUniverseEverythings"│""           │null  │"RETURN 42"│[["params","map","{}"]]│"list? of map"│false        │
├──────────┼────────────────────────────────┼─────────────┼──────┼───────────┼───────────────────────┼──────────────┼─────────────┤
│"function"│"tooManyLifeUniverseEverythings"│""           │null  │"RETURN 42"│[["params","map","{}"]]│"list? of map"│false        │
└──────────┴────────────────────────────────┴─────────────┴──────┴───────────┴───────────────────────┴──────────────┴─────────────┘
```

Gives
```
call apoc.custom.removeFunction('tooManyLifeUniverseEverythings')

Neo.ClientError.Procedure.ProcedureCallFailed
Failed to invoke procedure `apoc.custom.removeFunction`: Caused by: java.util.NoSuchElementException: More than one element in java.util.Spliterators$1Adapter@235d2118. First element is 'Node[28]' and the second element is 'Node[30]'

```

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: Catalina 10.5.15
  - Neo4j: 3.x 4.x
[...]

PR #1559, opened 2020-06-12
Fixes #1555 and #1558 

### LINKED ISSUE #1555: apoc.custom.asProcedure doesn't overwrite existing procedure by same name
## To reproduce
I'll use the example of the documentation

Execute this twice
`CALL apoc.custom.asProcedure('neighbours',
  'MATCH (n:Person {name:$name})-->(nb) RETURN nb as neighbour','read',
  [['neighbour','NODE']],[['name','STRING']], 'get neighbours of a person');
`
then `CALL apoc.custom.list`

## Expected Behavior
(Assuming clean db) lists one procedure with name 'neighbours'

## Actual Behavior
 (Assuming clean db) lists two procedures with name 'neighbours'

##
You also can't remove the procedure with `CALL apoc.custom.removeProcedure('neighbours')` because that expects there to only be one record for that name and it freaks out when many are returned.

Also when restarting the db errors get logged, I suspect, for attempts of registering procedure of the same name repeatedly.

So there's two things to fix:

- `apoc.custom.asProcedure` to owerwrite existing procedures
- `apoc.custom.removeProcedure` to delete all "procedures" of name (to allow affected users fix their databases)

P.S. Maybe `apoc.custom.asFunction` has same issue, haven't checked.

### Versions

  - OS: Ubuntu 18.04
  - Neo4j: 4.0.4
  - Neo4j-Apoc: 4.0.0.15

augonis:
Fixes #1555 and #1558 
Fixes the issue and enables repairs for affected users

## Proposed Changes
 - Fix typo in `Util.mergeNode` (cause of the issues)
 - Let `apoc.custom.removeProcedure` and `apoc.custom.removeFunction` handle existence of multiple records of procedure/function with same name

### DIFF
--- a/src/main/java/apoc/custom/CypherProceduresHandler.java
+++ b/src/main/java/apoc/custom/CypherProceduresHandler.java
@@ -598,30 +598,32 @@ public class CypherProceduresHandler implements AvailabilityListener {
 
     public void removeProcedure(String name) {
         withSystemDb(tx -> {
-            Node node = Iterators.single(tx.findNodes(SystemLabels.ApocCypherProcedures,
+            tx.findNodes(SystemLabels.ApocCypherProcedures,
                     SystemPropertyKeys.database.name(), api.databaseName(),
                     SystemPropertyKeys.name.name(), name
-            ).stream().filter(n -> n.hasLabel(SystemLabels.Procedure)).iterator());
-            ProcedureDescriptor descriptor = procedureDescriptor(node);
-            registerProcedure(descriptor.getSignature(), null);
-            registeredProcedureSignatures.remove(descriptor.getSignature());
-            node.delete();
-            setLastUpdate(tx);
+            ).stream().filter(n -> n.hasLabel(SystemLabels.Procedure)).forEach(node -> {
+                ProcedureDescriptor descriptor = procedureDescriptor(node);
+                registerProcedure(descriptor.getSignature(), null);
+                registeredProcedureSignatures.remove(descriptor.getSignature());
+                node.delete();
+                setLastUpdate(tx);
+            });
             return null;
         });
     }
 
     public void removeFunction(String name) {
         withSystemDb(tx -> {
-            Node node = Iterators.single(tx.findNodes(SystemLabels.ApocCypherProcedures,
+            tx.findNodes(SystemLabels.ApocCypherProcedures,
                     SystemPropertyKeys.database.name(), api.databaseName(),
                     SystemPropertyKeys.name.name(), name
-            ).stream().filter(n -> n.hasLabel(SystemLabels.Function)).iterator());
-            UserFunctionDescriptor descriptor = userFunctionDescriptor(node);
-            registerFunction(descriptor.getSignature(), null, false);
-            registeredUserFunctionSignatures.remove(descriptor.getSignature());
-            node.delete();
-            setLastUpdate(tx);
+            ).stream().filter(n -> n.hasLabel(SystemLabels.Function)).forEach(node -> {
+                UserFunctionDescriptor descriptor = userFunctionDescriptor(node);
+                registerFunction(descriptor.getSignature(), null, false);
+                registeredUserFunctionSignatures.remove(descriptor.getSignature());
+                node.delete();
+                setLastUpdate(tx);
+            });
             return null;
         });
 
--- a/src/test/java/apoc/custom/CypherProceduresTest.java
+++ b/src/test/java/apoc/custom/CypherProceduresTest.java
@@ -259,6 +259,22 @@ public class CypherProceduresTest  {
     }
 
     @Test
+    public void shouldOverwriteAndRemoveCustomProcedure() throws Exception {
+        db.executeTransactionally("call apoc.custom.asProcedure('answer','RETURN 42')");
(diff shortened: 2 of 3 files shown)


src/main/java/apoc/custom/CypherProceduresHandler.java
```diff
@@ -598,30 +598,32 @@ private AnyValue convertToValueRecursive(Object... toConverts) {
 
     public void removeProcedure(String name) {
         withSystemDb(tx -> {
-            Node node = Iterators.single(tx.findNodes(SystemLabels.ApocCypherProcedures,
+            tx.findNodes(SystemLabels.ApocCypherProcedures,
```
conker84, line 601: This should be not necessary.

src/main/java/apoc/custom/CypherProceduresHandler.java
```diff
@@ -598,30 +598,32 @@ private AnyValue convertToValueRecursive(Object... toConverts) {
...
+                setLastUpdate(tx);
+            });
             return null;
         });
     }
 
     public void removeFunction(String name) {
         withSystemDb(tx -> {
-            Node node = Iterators.single(tx.findNodes(SystemLabels.ApocCypherProcedures,
+            tx.findNodes(SystemLabels.ApocCypherProcedures,
```
conker84, line 617: This should be not necessary as well

augonis:
The changes to `removeProcedure` and `removeFunction` are needed to allow users already affected by the issue fix their damaged databases. Can you propose an alternative way of removing the existing duplicate procedure/function?

src/main/java/apoc/custom/CypherProceduresHandler.java
augonis: We can no longer assume there will no more than one result, because there are databases out in the wilds affected by #1555 that have multiple records with same name

src/main/java/apoc/custom/CypherProceduresHandler.java
augonis: We can no longer assume there will no more than one result, because there are databases out in the wilds affected by #1558 that have multiple records with same name

src/test/java/apoc/custom/CypherProceduresTest.java
```diff
@@ -258,6 +258,22 @@ public void shouldRemoveTheCustomFunction() throws Exception {
         TestUtil.count(db, "return custom.answer()");
     }
 
+    @Test
+    public void shouldOverwriteAndRemoveCustomProcedure() throws Exception {
+        db.executeTransactionally("call apoc.custom.asProcedure('answer','RETURN 42')");
+        db.executeTransactionally("call apoc.custom.asProcedure('answer','RETURN 42')");
+        assertEquals("Expecting one procedure listed", 1, TestUtil.count(db, "call apoc.custom.list()"));
```
conker84, line 265: Can you also add a test for the system db that checks if there is more than 1 node?

src/test/java/apoc/custom/CypherProceduresTest.java
```diff
@@ -258,6 +258,22 @@ public void shouldRemoveTheCustomFunction() throws Exception {
...
+        db.executeTransactionally("call apoc.custom.removeProcedure('answer')");
+    }
+
+    @Test
+    public void shouldOverwriteAndRemoveCustomFunction() throws Exception {
+        db.executeTransactionally("call apoc.custom.asFunction('answer','RETURN 42','long')");
+        db.executeTransactionally("call apoc.custom.asFunction('answer','RETURN 42','long')");
+        assertEquals("Expecting one function listed", 1, TestUtil.count(db, "call apoc.custom.list()"));
```
conker84, line 273: Same as above

src/test/java/apoc/custom/CypherProceduresTest.java
augonis: You mean like a separate test (what exatcly I would be testing? Unit test for `Util.mergeNode?`) or just a new assertions in these tests?

conker84: No I mean a new assertion for this test that checks into the system db if there is just one node for the procedure/function as we removed the `Iterators.single` now would be better have a direct check into the system db

augonis: `assertEquals("Expecting one procedure listed", 1, TestUtil.count(db, "call apoc.custom.list()"));` already does that, just indirectly.
And to do the check I would be more or less copying the implementation of `CypherProceduresHandler.readSignatures`, I'm not that comfortable querying db in code (no cypher) to write a bespoke query for the test.

conker84: ok so let's leave it as is

conker84:
LGTM thank you so much @augonis for this

philastrophist:
What do I have to do if I need to remove a duplicated procedure with `apoc-4.2.0.0-all` on my live database? Is there a way to do this without apoc?

Thanks

augonis:
Well, looks like this pull request was merged into a dead end branch (I can't find any release including it)  .
You could build your own patched version of apoc to include my fix.

philastrophist:
ah, that's why...

Ok, I'll submit a issue. In the mean time, I'll do as you say. But is there really no way to remove an apoc procedure without apoc?

PR #1573, opened 2020-06-29
fixes #1547: Add a new after-async or async phase to triggers

### LINKED ISSUE #1547: Add a new after-async or async phase to triggers
So that locks that are not released yet in `after` are gone and writes don't affect the current transaction.

conker84:
Fixes #1547

Added the `afterAsync` phase

  - Added the `afterAsync` phase
  - Added test
  - Updated documentation


mneedham:
This looks good to me, what do you think @jexp 

I think we need to add something more to the documentation advising users when they should be used the `after` phase and when they should be using the `afterAsync` phase. At the moment it says use the latter for 'heavy operations', but if we could give a more specific example I think it'll make the feature more useable

docs/asciidoc/job-management/trigger.adoc
```diff
@@ -202,3 +204,16 @@ CALL apoc.trigger.add('lowercase-by-label','UNWIND apoc.trigger.nodesByLabel({as
...
+
+.Trigger Phase Table
+.Helper Functions
+[cols="1m,5"]
+|===
+| Phase | Description
+| before | The trigger will be activate right `before` the commit. If no phase is specified, it's the default.
+| rollback | The trigger will be activate right after the  `rollback`
+| after | The trigger will be activate right `after` the commit
+| afterAsync | The trigger will be activate right `after` the commit and inside a new transaction that will not
```
jexp, line 216: also in a new thread!?

docs/asciidoc/job-management/trigger.adoc
```diff
@@ -15,7 +15,9 @@ Enable `apoc.trigger.enabled=true` in `$NEO4J_HOME/conf/apoc.conf` first.
 
 [cols="1m,5"]
 |===
-| CALL apoc.trigger.add(name, statement, selector) yield name, statement, installed | add a trigger statement under a name, in the statement you can use {createdNodes}, {deletedNodes} etc., the selector is {phase:'before/after/rollback'} returns previous and new trigger information
+| CALL apoc.trigger.add(name, statement, selector) yield name, statement, installed | add a trigger statement under a name,
+in the statement you can use {createdNodes}, {deletedNodes} etc., the selector is {phase:'before/after/rollback/afterAsync'}
```
jexp, line 19: change syntax to `$createdNodes` etc.

jexp:
Looks good to me, please merge and forward merge.

PR #1575, opened 2020-06-30
SystemDB procs check admin

jexp:
The systemdb procedures should check if the current user is admin and fail otherwise.


eastlondoner:
Looks good to me

eastlondoner:
please don't forget to forward-port this to 4.1

PR #1590, opened 2020-07-13
fixes #1589: Some settings don't work with Docker containers

### LINKED ISSUE #1589: Some settings don't work with Docker containers
Properties like `apoc.initializer.cypher` don't work with docker containers via env variables (i.e. `NEO4J_apoc_initializer_cypher`) because they're not defined in `ApocSettings`.

conker84:
Fixes #1589

Added settings into `ApocSettings.java`

### DIFF
--- a/src/main/java/apoc/ApocSettings.java
+++ b/src/main/java/apoc/ApocSettings.java
@@ -39,6 +39,14 @@ public class ApocSettings implements SettingsDeclaration {
 
     public static final Setting<Boolean> apoc_uuid_enabled = newBuilder(APOC_UUID_ENABLED, BOOL, false ).build();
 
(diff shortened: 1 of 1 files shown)


ivangreene:
@conker84 having a problem with this and came across this issue/PR. It looks like this will resolve `apoc.initializer.cypher` not working, but isn't going to solve `apoc.initializer.cypher.1`, `apoc.initializer.cypher.2`, etc. I'm not sure how a wildcard configuration option like this can work

sarmbruster:
Neo4 4.x does no longer support dynamic configurations. You either put them into `conf/apoc.conf` or pass them directly as env variables e.g. with docker `-e apoc.initializer.cypher.1=value`

ivangreene:
@sarmbruster thanks, putting these in `conf/apoc.conf` instead of in `conf/neo4j.conf` fixed my problem

PR #1593, opened 2020-07-14
fixes #1581: Error in query - CALL apoc.schema.nodes()

### LINKED ISSUE #1581: Error in query - CALL apoc.schema.nodes()
Neo4j Vsersion : 3.5.19
Apoc Versions: 3.5.0.4 and 3.5.0.12

An issue in APOC procedure CALL apoc.schema.nodes(), It runs fine when there is NO index but It got fails with below error when there is any type of index in the database

```
Neo.ClientError.Procedure.ProcedureCallFailed: Failed to invoke procedure `apoc.schema.nodes`: Caused by: java.lang.IncompatibleClassChangeError: Found interface org.neo4j.storageengine.api.schema.PopulationProgress, but class was expected
```
IndexOutput

Requesting a fix for this bug and need a updated apoc version to share with the customer please

conker84:
Fixes #1581

This PR contains two changes the one related to #1518 (because otherwise I had compilation problems) and the changes required to fix the problem raised in the #1581 
Fix the problem has been not quite straightforward because seems there is a bug in the Neo4j so I had to provide the workaround that you can find in `Schemas.java` from line 438 to 451.

  - Fixed the bug

### DIFF
--- a/src/main/java/apoc/schema/Schemas.java
+++ b/src/main/java/apoc/schema/Schemas.java
@@ -2,26 +2,43 @@ package apoc.schema;
 
 import apoc.result.AssertSchemaResult;
 import apoc.result.ConstraintRelationshipInfo;
-import apoc.result.GraphResult;
 import apoc.result.IndexConstraintNodeInfo;
 import apoc.util.Util;
 import org.apache.commons.lang3.StringUtils;
-import org.neo4j.graphdb.*;
+import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.graphdb.Label;
+import org.neo4j.graphdb.RelationshipType;
 import org.neo4j.graphdb.schema.ConstraintDefinition;
 import org.neo4j.graphdb.schema.ConstraintType;
 import org.neo4j.graphdb.schema.IndexDefinition;
 import org.neo4j.graphdb.schema.Schema;
 import org.neo4j.helpers.collection.Iterables;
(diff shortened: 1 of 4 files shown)


src/main/java/apoc/schema/Schemas.java
```diff
@@ -417,6 +435,20 @@ private IndexConstraintNodeInfo nodeInfoFromIndexDefinition(IndexReference index
...
         Arrays.stream(indexReference.properties()).forEach((i) -> properties.add(tokens.propertyKeyGetName(i)));
         try {
+            float populationProgressValue;
+            try {
+                final PopulationProgress populationProgress = schemaRead.indexGetPopulationProgress(indexReference);
+                populationProgressValue = populationProgress.getCompleted() / populationProgress.getTotal() * 100;
+            } catch (IncompatibleClassChangeError e) {
+                // Workaround because in Neo4j 3.5.19 we have the following error
+                // Failed to invoke procedure `apoc.schema.nodes`: Caused by: java.lang.IncompatibleClassChangeError: Found interface org.neo4j.storageengine.api.schema.PopulationProgress, but class was expected
+                populationProgressValue = db.execute("CALL db.indexes() YIELD tokenNames, properties, progress " +
```
mneedham, line 445: @conker84 I think you said it will work across all versions if we use the procedure approach? If so, let's do that and not bother calling the API in the lines above.

PR #1621, opened 2020-08-14
fixes #1592: apply setDaemon(true) everywhere a new thread is created

### LINKED ISSUE #1592: Make sure that none of the threads we create is a non-daemon thread
Esp in cypher initjalizer and other delays
Or in periodic iterate/semaphores/tx-utilities

sarmbruster:
please cherry pick this one to 4.1 and 4.2


jexp:
@mneedham can you do the merge + cherry-pick?

PR #1632, opened 2020-08-26
Document Typo apoc.path.subgraph - Config parameters

ajomathew:
Looks like there is a typo in the documentation.

Looking in the actual code, the "List" in "blackListNodes" should have a lower-case "L". That is, it should be "blacklistNodes".

  -
  -
  -


ajomathew:
Raising an Isssue

PR #1649, opened 2020-09-04
Escape attribute reference

### LINKED ISSUE #1648: Espace AsciiDoc attribute references to suppress warnings when building the documentation with Antora
Currently, when we build the documentation with Antora from https://github.com/neo4j-documentation/docs-refresh, Asciidoctor prints almost 300 warnings because the documentation contains unresolved attribute references.

For reference, the syntax of an attribute reference in AsciiDoc is: `{myattribute}`.

The main issue is that the description of the `apoc.*` procedures frequently contain this syntax.
For instance:

```
apoc.cypher.runTimeboxed('cypherStatement',{params}, timeout) - abort kernelTransaction after timeout ms if not finished
```

As you can see in this example, AsciiDoc will try to resolve the attribute reference `{params}`.
One way to solve this issue is to escape the attribute reference with `\`:

```
apoc.cypher.runTimeboxed('cypherStatement',\{params}, timeout) - abort kernelTransaction after timeout ms if not finished
```

We could also prevent substitution using a single plus symbol. While doing so we could also use a monospaced font formatting:

```
`+apoc.cypher.runTimeboxed('cypherStatement',{params}, timeout)+` - abort kernelTransaction after timeout ms if not finished
```

In my opinion, the result will be more readable:

**Current**

apoc.cypher.runTimeboxed('cypherStatement',{params}, timeout) - abort kernelTransaction after timeout ms if not finished

**After**

`apoc.cypher.runTimeboxed('cypherStatement',{params}, timeout)` - abort kernelTransaction after timeout ms if not finished

However, this solution is more complex to implement because we need to accurately find the start and the end of the procedure (and if we guess wrong it will produce unexpected results).

Mogztter:
Fixes #1648

This pull request contains a lot of changes but most of them were automatically generated from `DocsTest.java`.

## Proposed Changes

  - Use `\` to espace attribute references in the documentation.
  - Put table cell `|` on a newline to improve readability


jexp:
We should instead move to the new parameter syntax, which is `$param`

Mogztter:
> We should instead move to the new parameter syntax, which is $param

That would be even better :+1: 
Though we can keep the escaping logic just in case the description contains `{something}`.

PR #1651, opened 2020-09-05
db specific initializers

sarmbruster:
Adding support for db specific cypher initializers.

Side note: also added a custom TestRule that allows for test-method specific env settings via annotations.

Please cherry pick this one for 4.2 as well.

### DIFF
--- a/core/src/main/java/apoc/ApocConfig.java
+++ b/core/src/main/java/apoc/ApocConfig.java
@@ -58,6 +58,12 @@ public class ApocConfig extends LifecycleAdapter {
     public static final String APOC_CONFIG_JOBS_POOL_NUM_THREADS = "apoc.jobs.pool.num_threads";
     public static final String APOC_CONFIG_JOBS_QUEUE_SIZE = "apoc.jobs.queue.size";
     public static final String APOC_CONFIG_INITIALIZER = "apoc.initializer";
+
+    /**
+     * @deprecated
+     * This has been replaced by database-specific initialisers.
+     * Use apoc.initializer.<database name> instead.
+     */
     @Deprecated
     public static final String APOC_CONFIG_INITIALIZER_CYPHER = APOC_CONFIG_INITIALIZER + ".cypher";
 
--- a/core/src/main/java/apoc/ApocConfig.java
+++ b/core/src/main/java/apoc/ApocConfig.java
@@ -57,7 +57,9 @@ public class ApocConfig extends LifecycleAdapter {
     public static final String APOC_CONFIG_JOBS_SCHEDULED_NUM_THREADS = "apoc.jobs.scheduled.num_threads";
     public static final String APOC_CONFIG_JOBS_POOL_NUM_THREADS = "apoc.jobs.pool.num_threads";
     public static final String APOC_CONFIG_JOBS_QUEUE_SIZE = "apoc.jobs.queue.size";
-    public static final String APOC_CONFIG_INITIALIZER_CYPHER = "apoc.initializer.cypher";
+    public static final String APOC_CONFIG_INITIALIZER = "apoc.initializer";
+    @Deprecated
+    public static final String APOC_CONFIG_INITIALIZER_CYPHER = APOC_CONFIG_INITIALIZER + ".cypher";
 
     private static final List<Setting> NEO4J_DIRECTORY_CONFIGURATION_SETTING_NAMES = new ArrayList<>(Arrays.asList(
             data_directory,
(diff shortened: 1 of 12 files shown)


core/src/main/java/apoc/ApocConfig.java
```diff
@@ -57,7 +57,9 @@
     public static final String APOC_CONFIG_JOBS_SCHEDULED_NUM_THREADS = "apoc.jobs.scheduled.num_threads";
     public static final String APOC_CONFIG_JOBS_POOL_NUM_THREADS = "apoc.jobs.pool.num_threads";
     public static final String APOC_CONFIG_JOBS_QUEUE_SIZE = "apoc.jobs.queue.size";
-    public static final String APOC_CONFIG_INITIALIZER_CYPHER = "apoc.initializer.cypher";
+    public static final String APOC_CONFIG_INITIALIZER = "apoc.initializer";
+    @Deprecated
```
eastlondoner, line 61: It's nice if you're deprecating things to add some info on what the replacement is:
e.g.
```
/**
     * @deprecated
     * This has been replaced by database-specific initialisers.
     * Use apoc.initializer.<database name> instead.
     */
```

sarmbruster: good catch, added your suggestions.

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -46,23 +52,31 @@ public void available() {
...
+                final boolean isSystemDatabase = db.databaseName().equals(GraphDatabaseSettings.SYSTEM_DATABASE_NAME);
+                if (!isSystemDatabase) {
+                    awaitApocProceduresRegistered();
+                }
                 Configuration config = dependencyResolver.resolveDependency(ApocConfig.class).getConfig();
 
-                TreeMap<String, String> initializers = Iterators.stream(config.getKeys(ApocConfig.APOC_CONFIG_INITIALIZER_CYPHER))
-                        .collect(Collectors.toMap(k -> k, k -> config.getString(k),
-                                (v1, v2) -> {
-                                    throw new RuntimeException(String.format("Duplicate key for values %s and %s", v1, v2));
```
jexp, line 55: does this break the server startup? if so, do we want that? or rather juts log an error?

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -46,23 +52,31 @@ public void available() {
...
-                                },
-                                TreeMap::new));
+                TreeMap<String, String> initializers = new TreeMap<>();
+
+                config.getKeys(ApocConfig.APOC_CONFIG_INITIALIZER + "." + db.databaseName()).forEachRemaining(key -> initializers.put(key, config.getString(key)));
+
+                if (!isSystemDatabase) {
```
jexp, line 65: can we move this higher as a guard clause and just return from the method?

also can this whole collecting initializers go into a separate method?

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -46,23 +52,31 @@ public void available() {
...
                     String query = initializer.toString();
-                    try {
-                        db.executeTransactionally(query);
-                        userLog.info("successfully initialized: " + query);
-                    } catch (Exception e) {
-                        userLog.warn("error upon initialization, running: " + query, e);
+                    if (!query.isEmpty()) {
+                        try {
+                            // we need to apply a retry strategy here since in systemdb we potentially conflict with
+                            // creating contraints which could cause our query to fail with a transient error.
```
jexp, line 74: what is an example of this?

sarmbruster: when not having the retry strategy I found that an initializer of `apoc.initializer.system=create user dummy set password 'abc'` resulted in a transaction failed since a concurrent transaction was setting up a constraint.

core/src/main/java/apoc/cypher/CypherInitializer.java
sarmbruster: moving higher doesn't make a lot of sense IMO. We do want to execute initializers for any db (including system). Systemdb is just treated differently.
Collecting initializers is now in a separate method, see my commit below.

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -37,32 +39,32 @@ public boolean isFinished() {
...
 
         // run initializers in a new thread
         // we need to wait until apoc procs are registered
         // unfortunately an AvailabilityListener is triggered before that
         Util.newDaemonThread(() -> {
-
             try {
-                awaitApocProceduresRegistered();
+                final boolean isSystemDatabase = db.databaseName().equals(GraphDatabaseSettings.SYSTEM_DATABASE_NAME);
+                if (!isSystemDatabase) {
```
jexp, line 55: but if `isSystemDatabase` is true none of the code is executed, not the await for apoc nor collecting the config (it will return an empty list)

so we can also just return at the beginning of `available()` for the systemdb

jexp (approved):
Looks good to me, please merge into the 4.0, 4.1, 4.2 branches.

PR #1654, opened 2020-09-06
remove code duplication

sarmbruster:
core/extended split caused code duplications. This PR removes one of them. ApocExendedInit was basically a copy of ApocConfig with additionally initializing JDBC driver classes. This has been refactored into JdbcRegistererInitFactory.

Additionally all classes used by JVM service loader are now using `@ServiceProvider` to generate `META-INF/services` files. No need to maintain hand crafted `META-INF/services` stuff any more.

Please cherry-pick that PR to 4.2.


conker84:
Thank you @sarmbruster !

PR #1658, opened 2020-09-07
Update init script documentation

tomasonjo:
Adding multiple cypher call example and change the config option to be less ambiguous as it references to the Neo4j configuration file.

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/operational/init-script.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/operational/init-script.adoc
@@ -7,7 +7,7 @@
 Apoc optionally allows you to run a cypher command after database initialization is finished.
 This can e.g. be used to ensure indexes/constraints are created up front.
 
-To use this feature, add the following line to Neo4j configuration file:
+To use this feature, add the following line to one of the https://neo4j.com/labs/apoc/4.1/config/#_location_of_config_options[configuration options]:
 
 [source,config]
 ----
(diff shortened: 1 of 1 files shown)


docs/asciidoc/modules/ROOT/pages/operational/init-script.adoc
```diff
@@ -7,7 +7,7 @@
 Apoc optionally allows you to run a cypher command after database initialization is finished.
 This can e.g. be used to ensure indexes/constraints are created up front.
 
-To use this feature use a config option:
+To use this feature, add the following line to Neo4j configuration file:
```
sarmbruster, line 10: it doesn't have to be necessarily in the config file. We can use either `apoc.conf`, env variables or system properties

tomasonjo: words are hard

tomasonjo: I have added the link to the configuration page

tomasonjo: Obviously I don't know how to link properly, if you could help me with that would be great!

tomasonjo:
I have just added the cypher.1 and cypher.2 options and removed the linking:

@sarmbruster

conker84:
Thank you @tomasonjo !

PR #1682, opened 2020-10-06
Handle Optional.empty() function descriptions

jakewins:
This handles loading a stored APOC custom state where
a function or procedure had an Optional.empty() description.

Jackson serialized those to {"present::false}, which then broke
Apocs assumptions about the schema here.

This makes it so we handle any in-the-wild incarnations of this,
and modifies our write path to always ensure the value is a string

When APOC serializes custom function/procedure state, it does this:

```
Map<String, Object> data = map("statement", statement, "mode", signature.mode().toString(), "signature", signature.toString(), "description", signature.description());
```

Which assumes `signature.description()` won't be `Optional.empty()` (it seems that's been changed to be Optional since the original apoc code was written). In cases where it *is* empty, the end-result JSON will be `{"present":false}`, rather than a string, which then breaks APOC loading the state back again.

This changes the serialization behavior to always result in a string, and also modifies the deserialization behavior to handle existing in-the-wild stores that have ran into this issue.

### DIFF
--- a/src/main/java/apoc/custom/CypherProcedures.java
+++ b/src/main/java/apoc/custom/CypherProcedures.java
@@ -575,7 +575,7 @@ public class CypherProcedures {
                 }
                 return (String)rawDescription;
             }
-            return "";
+            return null;
         }
 
         @Override
@@ -593,7 +593,7 @@ public class CypherProcedures {
         }
 
         public static Map<String, Object> storeProcedure(GraphDatabaseAPI api, ProcedureSignature signature, String statement) {
-            Map<String, Object> data = map("statement", statement, "mode", signature.mode().toString(), "signature", signature.toString(), "description", signature.description().orElse(""));
+            Map<String, Object> data = map("statement", statement, "mode", signature.mode().toString(), "signature", signature.toString(), "description", signature.description().orElse(null));
             return updateCustomData(getProperties(api), signature.name().toString(), PROCEDURES, data);
         }
         public static Map<String, Object> storeFunction(GraphDatabaseAPI api, String name, String statement, String output, List<List<String>> inputs, boolean forceSingle, String description) {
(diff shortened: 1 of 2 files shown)


src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -565,6 +566,18 @@ private void restoreProcedures() {
...
+        private String parseStoredDescription(Object rawDescription) {
+            // Description was changed to be an Optional in the Neo4j internal API. Optional.None serialized to
+            // JSON objects in various stores around the world; hence, deal with this value not being a string
+            if(rawDescription instanceof String) {
+                if("null".equals(rawDescription)) {
+                    return null;
+                }
+                return (String)rawDescription;
+            }
+            return "";
```
jexp, line 578: should probably return null as well

jexp (approved):
Looks good to me, please check that one comment.

src/main/java/apoc/custom/CypherProcedures.java
```diff
@@ -589,7 +602,7 @@ public void unavailable() {
         }
 
         public static Map<String, Object> storeFunction(GraphDatabaseAPI api, UserFunctionSignature signature, String statement, boolean forceSingle) {
-            Map<String, Object> data = map("statement", statement, "forceSingle", forceSingle, "signature", signature.toString(), "description", signature.description());
+            Map<String, Object> data = map("statement", statement, "forceSingle", forceSingle, "signature", signature.toString(), "description", signature.description().orElse(""));
```
jexp, line 605: sorry I overlooked this b/c it was too far on the right :) can you use `.orElse(null)` so that we're consistent? in the 2 places?

jakewins: :+1:

Issue #1705, opened 2020-11-02
Cypher export leaves property key with special character (here: $) unquoted


Phoosha:
## Expected Behavior

When a node has a property e.g. set to `$lock = true`, exporting to cypher (using the format `cypher-shell`) should create valid cypher, which actually can be imported (using `cypher-shell`).

More specifically, I'd probably expect one of the following:
1. Identifiers containing special characters are quoted only if needed, e.g. `$lock` is quoted but not `name`.
2. All identifiers containing special characters are always quoted.

## Actual Behavior

- An identifier such as `$lock` is not quoted in the cypher output. Therefore the cypher output is invalid and cannot be re-imported.
- If I manually quote all invalid identifiers (i.e. I replace `$lock` by `` `$lock` ``), the cypher output can be correctly imported.
- Passing the config option `quotes: always` (or `quotes: ifNeeded`) does not have any effect (which I expected because there's no documentation of it pertaining to the cypher export).
- I believe, the export code checks for needed quoting. But at least for `$` it does not apply quotes. This is the line, which I believe should handle the quoting (I did not verify whether it's actually hit at runtime though): 
  - *3.5 branch:* https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/6ad36abddff10a9fef89f023c1d5520370fd92e9/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java#L345
  - *4.1 branch:* https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/12f66b76a296094feeb3021e1ce88bc09dd6a5d7/core/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java#L354

## How to Reproduce the Problem

Run for example this query:
```
MERGE (n:TestNode)
SET n.`$lock` = true
WITH n
CALL apoc.export.cypherData([n], [], "export.cypher", {format: 'cypher-shell'})
YIELD file, time, nodes, relationships, properties, rows, batches, batchSize
RETURN file, time, nodes, relationships, properties, rows, batches, batchSize;
```

### Resulting content of `export.cypher`

Notice how on line 5 `$lock` is not quoted.
```
:begin
CREATE CONSTRAINT ON (node:`UNIQUE IMPORT LABEL`) ASSERT (node.`UNIQUE IMPORT ID`) IS UNIQUE;
:commit
:begin
UNWIND [{_id:261, properties:{$lock:true}}] AS row
[...]

PR #1707, opened 2020-11-03
fixes #1706: Match relationship end nodes consistently

### LINKED ISSUE #1706: Cypher export outputs unsatisfiable filter for relationships with multiple unique constraints via multiple labels
There's an inconsistency between two methods used to generate the `MATCH` clauses for the end nodes of relationships:
1. https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/12f66b76a296094feeb3021e1ce88bc09dd6a5d7/core/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java#L434
2. https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/12f66b76a296094feeb3021e1ce88bc09dd6a5d7/core/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java#L421

Specifically the 2. method tries to filter on multiple unique constraints whereas the 1. method (writing the id values to filter by) only uses one unique constraint. This happens when there are:
1. two labels (e.g. `Bar` and `Base`)
2. each label has one unique constraint (e.g. `Bar` for `name` and `Base` for `id`)
3. a node which has both labels

cmp. the simple reproducer below

## Expected Behavior

The generated cypher should import the relationship. I guess, one of the following options should be chosen:
1. Use in both methods the same single unique constraint
2. Use in both methods all unique constraints _(seems like overkill)_

For the simple reproducer below I'd expect the export to yield in case 1.:
```
UNWIND [{start: {name:"barista"}, end: {_id:4}, properties:{}}] AS row    // <-- handled by method 1.
MATCH (start:Bar{name: row.start.name})                                   // <-- handled by method 2.
MATCH (end:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row.end._id})
CREATE (start)-[r:KNOWS]->(end) SET r += row.properties;
```

## Actual Behavior

The map in the `UNWIND` clause only mentions the `name` property of `Bar`, while the `MATCH` clause also filters on the `id` property of `Base`, whose value was not defined in the `UNWIND` clause:
```
UNWIND [{start: {name:"barista"}, end: {_id:4}, properties:{}}] AS row
MATCH (start:Bar{name: row.start.name, id: row.start.id})    // <-- Notice id here but not above in UNWIND!
MATCH (end:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row.end._id})
CREATE (start)-[r:KNOWS]->(end) SET r += row.properties;
```

```
CREATE CONSTRAINT ON (b:Bar) ASSERT b.name IS UNIQUE
CREATE CONSTRAINT ON (b:Base) ASSERT b.id IS UNIQUE
CREATE (b:Bar:Base {id:'waBfk3z', name:'barista'})-[r:KNOWS]->(f:Foo)
WITH b, r, f
CALL apoc.export.cypherData([b, f], [r], "export.cypher", {format: 'cypher-shell'})
YIELD file, time, nodes, relationships, properties, rows, batches, batchSize
RETURN file, time, nodes, relationships, properties, rows, batches, batchSize;
```

### Versions

 - OS: Linux
 - Neo4j: 3.5.22
 - Neo4j-Apoc: 3.5.0.15
 - Cypher-Shell: 1.1.13

Phoosha:
Fixes #1706

Fix the inconsistency described in in #1706 when exporting a node with multiple labels with one unique property constraint for each label.

## Proposed Changes

1. Extract the logic from `writeRelationshipNodeIds` to `getUniqueConstrainedProperties`
2. Use the same consistent logic in `writeRelationshipMatchAsciiNode`
3. Add two test cases, which both previously failed

The logic from `writeRelationshipNodeIds` is sufficient because matching by one unique constrained property is sufficient to achieve unambiguity.

Feel free to let me know, if you want something handled differently.


conker84:
@Phoosha Can you please add a similar test for compound keys in `ExportCypherEnterpriseFeaturesTest`?

Phoosha:
> @Phoosha Can you please add a similar test for compound keys in `ExportCypherEnterpriseFeaturesTest`?

@conker84 I added a similar test. It's a little less straightforward though, because with the test containers database content is kept between tests.

conker84 (approved):
Thank you @Phoosha!

mneedham:
@Phoosha I think merging your other commit has caused a conflict on this one. I tried to resolve it, but I'm not sure exactly how to do that, but I expect you do!

Phoosha:
@mneedham Conflict is resolved and it should be good to go.

On a different matter, is there any chance these fixes for #1705 & #1706 will make it into the 3.5 branch?

If it helps, I've already got a version of the fixes laying around for 3.5.: https://github.com/neo4j-contrib/neo4j-apoc-procedures/compare/3.5...Phoosha:3-5-gh-1706-gh-1705

mneedham:
@Phoosha it's a bit of a pain backporting these commit as we completely changed the layout of the repository in 4.1

Would you be able to create a PR for the changes in 3.5? And then I can use that commit to merge these changes into 4.0 as well.

Thanks!
Mark

PR #1713, opened 2020-11-09
fixes #1712

### LINKED ISSUE #1712: apoc.util.validate 'sister function' apoc.util.validatePredicate
## Feature description
Internally at neo4j, we are working on implementing Auth for a new GraphQL library. We are using `apoc.util.validate` to enforce fine-grained authorization rules however we have come across a roadblock where this procedure cant be used in predicates. This feature request is asking for a new UserFunction `apoc.util.validatePredicate`. The function should be very similar to `apoc.util.validate` but also return `true` so it can be easily used inside a WHERE subclause. 

```graphql
type Product @auth(rules: [{allow: {id: "jwt.allowedProducts"}}]){
  colors: [Color] @relationship(...)
}

type Color @auth(rules: [{allow: {id: "jwt.allowedColors"}}]){
  name: String
}

# =============================
query {
  Products {
    colors {
      name
    }
  }
}
```

>  Example generated cypher

```
MATCH (product:Product)
CALL apoc.util.validate(NOT(color.id IN $allowedProducts), "Forbidden", [0])
RETURN product { 
  colors: [ 
       (product)-[:HAS_COLOR]->(color:Color) 
        WHERE apoc.util.validatePredicate(NOT(color.id IN $allowedColors), "Forbidden", [0])
        | color { .name }
  ]
} as product
```

### Why
One may be asking why do you need to write the cypher like this **Desired Usage** ^^? This is due to the fact the cypher is autogenerated from a GraphQL selection set & is recursive. The need for this becomes more apparent when we introduce a nested/recursive hierarchy. 

>  Arbitrary GraphQL for context

```graphql
type Product @auth(rules: [{allow: {id: "jwt.allowedProducts"}}]){
  photos: [Photo] @relationship(...)
}


type Photo @auth(rules: [{allow: {id: "jwt.allowedPhotos"}}]){
  description: String
  colors: [Color] @relationship(...)
}

type Color @auth(rules: [{allow: {id: "jwt.allowedColors"}}]){
  name: String
}

# =============================
query {
  Products {
    photos {
      description
       colors {
          name
       }
    }
  }
}
```

>  Example generated cypher

```cypher
MATCH (product:Product)
CALL apoc.util.validate(NOT(product.id IN $allowedProducts), "Forbidden", [0])
RETURN product { 
  photos: [
      (product)-[:HAS_PHOTO]->(photo:Photo) WHERE apoc.util.validatePredicate(NOT(photo.id IN $allowedPhotos), "Forbidden", [0]) | photo {
          .description,
          colors: [
              (photo)-[:OF_COLOR]->(color:Color) WHERE apoc.util.validatePredicate(NOT(color.id IN $allowedColors), "Forbidden", [0]) | color { 
                  .name
              }
          ]
      }
  ]
} as product
````

Using the shape of the selection set we generate a similar projection.

### PR DESCRIPTION
**Pull request by danstarns**

### CODE REVIEW

File: core/src/main/java/apoc/util/Utils.java [hunk 1]
```diff
@@ -76,4 +76,15 @@ public void validate(@Name("predicate") boolean predicate, @Name("message") Stri
             throw new RuntimeException(message);
         }
     }
+
+    @UserFunction
+    @Description("apoc.util.validatePredicate(predicate, message, params) | if the predicate yields to true raise an exception else returns true, for use inside WHERE subclauses")
```
jexp, line 81: shouldn't it fail on false? otherwise it's confusing and afaik also the opposite of the existing functionality?

jexp (changes requested):
Please have a  look at my comment.

We also need to add it to the docs.

core/src/main/java/apoc/util/Utils.java
danstarns: The current validate function will raise an exception if the predicate yields to true, I do find it confusing to reason about, but for consistency, I kept to this paradigm.

danstarns:
> Please have a look at my comment.
> 
> We also need to add it to the docs.

My bad I thought they were autogenerated.

PR #1718, opened 2020-11-11
Fix broken GCS support that's throwing NPEs

voutilad:
The call to the `getHost()` method on the `URI` instance returns `null` for a GCS uri like `gs://my_bucket/my_object`. Checking the URI docs from Oracle (https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URI.html) we should be using `getAuthority()`.

This commit changes the method call, thus resolving the `NullPointerException` I discovered when trying to load data from a publicly accessible object in a GCS bucket:

```
java.lang.NullPointerException
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:770)
	at com.google.cloud.storage.BlobId.of(BlobId.java:102)
	at apoc.util.google.cloud.GCStorageURLConnection.connect(GCStorageURLConnection.java:70)
	at apoc.util.google.cloud.GCStorageURLConnection.getInputStream(GCStorageURLConnection.java:85)
	at apoc.util.google.cloud.GCStorageTest.testGcsPublicBucket(GCStorageTest.java:18)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
[...]

PR #1720, opened 2020-11-12
Support passive environmental service account authentication in Google Cloud

voutilad:
This PR adds the ability to leverage passive authentication if Neo4j is running in Google Cloud, like in the case of running in GCE on a VM host or as a container instance. This simplifies and improves APOC's Google Cloud Storage support by:

- not requiring users to generate private key files for authenticating as a service account, which is generally [frowned upon](https://medium.com/@jryancanty/stop-downloading-google-cloud-service-account-keys-1811d44a97d9) because they do not expire and are hard to track.
- fixes a fundamental logic error causing NullPointerExceptions preventing accessing Google Storage objects (due to calling `getHost()` and not `getAuthority()` on the `URI` instance)
- updates the docs accordingly for the new `authenticationType` options (`NONE`, `PRIVATE_KEY`, `GCP_ENVIRONMENT`)
  - `PRIVATE_KEY` replaces the previous behavior used by `SERVICE`, i.e. using the `GOOGLE_APPLICATION_CREDENTIALS` approach of pointing the client libraries at the private key json on the file system
  - `GCP_ENVIRONMENT` lets the client libraries naturally identify the service account and fetch a token for authenticating with Cloud Storage

This PR supersedes #1718 (of which I will close).

Note: Ultimately this should be cherry-picked on the 4.2 branch when that's ready.

### DIFF
--- a/core/src/main/java/apoc/util/google/cloud/GCStorageURLConnection.java
+++ b/core/src/main/java/apoc/util/google/cloud/GCStorageURLConnection.java
@@ -67,7 +67,7 @@ public class GCStorageURLConnection extends URLConnection {
             if (StringUtils.isBlank(uri.getPath())) {
                 throw new RuntimeException("Please provide the file name");
             }
-            blob = getStorage(uri).get(BlobId.of(uri.getHost(), uri.getPath().substring(1)));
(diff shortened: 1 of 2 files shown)


sarmbruster (approved):
Looks good for me.

PR #1724, opened 2020-11-19
Update invert-relationship.adoc

gitclem:
I suggest that you have an example that's is more concrete than using the meaningless "foo" and "bar".  An example where we got the relationship backwards will stick in people's minds better.

That is, it's more obvious that the relationship CAR DRIVES a PERSON is wrong and needs to be reversed to be a PERSON DRIVES a CAR.

I'm not sure how to use fix the pictures though....

Use a more concrete example than meaningless "foo" and "bar"

Replace Labels `Foo` with `Car` and `Bar` with `Person`.  Replace relationship `:FOOBAR` with `:DRIVES`

The pictures need to be updated although I'm not sure how to do that.

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/graph-updates/graph-refactoring/invert-relationship.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/graph-updates/graph-refactoring/invert-relationship.adoc
@@ -20,7 +20,7 @@ The example below will help us learn how to use this procedure.
 
 [source,cypher]
 ----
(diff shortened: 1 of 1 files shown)


mneedham:
Hi @gitclem,

Could you fill in the CLA (https://neo4j.com/developer/cla/) so that I can merge your commit?

I can fix the pictures after we merge it.

Cheers, Mark

gitclem:
It’s a bit awkward but I’ll be sending you the CLA via my work email:

cwang@conductiv.com

conker84 (approved):

PR #1733, opened 2020-11-30
Backport quote #1707 and export relationship #1708 fixes to 3.5

### LINKED ISSUE #1705: Cypher export leaves property key with special character (here: $) unquoted

Phoosha:
Fixes #1705 #1706

### DIFF
--- a/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java
+++ b/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java
@@ -9,10 +9,7 @@ import org.neo4j.graphdb.*;
 import org.neo4j.helpers.collection.Iterables;
 
 import java.io.PrintWriter;
-import java.util.AbstractMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -311,10 +308,10 @@ abstract class AbstractCypherFormatter implements CypherFormatter {
 	private void closeUnwindRelationships(String relationshipClause, String setClause, Map<String, Set<String>> uniqueConstraints, ExportConfig exportConfig, PrintWriter out, String start, String end, Map<String, Object> path, Relationship last) {
 		writeUnwindEnd(exportConfig, out);
 		// match start node
-		writeRelationshipMatchAsciiNode(last.getStartNode(), out, start, path, uniqueConstraints);
+		writeRelationshipMatchAsciiNode(last.getStartNode(), out, start, uniqueConstraints);
 
 		// match end node
-		writeRelationshipMatchAsciiNode(last.getEndNode(), out, end, path, uniqueConstraints);
+		writeRelationshipMatchAsciiNode(last.getEndNode(), out, end, uniqueConstraints);
 
 		out.append(StringUtils.LF);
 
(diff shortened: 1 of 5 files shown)


conker84:
@Phoosha can you please take a look to the PR because you missed the right directory there is no `core` directory inside branch `3.5`

Phoosha:
@conker84 Ooops :blush: I meant not to commit those files. Should be fixed. Thanks :+1:

conker84 (approved):
LGTM

PR #1745, opened 2020-12-04
Trying to simplify TTL tests by testing the config on its own

mneedham:
The current tests in TTLMultiDbTest are super flaky. I think a better way to test this functionality is to check that the config used for scheduling is correct.

So I've pulled out the config and created a function that exposes the config for the current database. This makes the tests more deterministic, which will stop them failing almost every build!


conker84 (approved):

conker84:
LGTM thank you!

PR #1746, opened 2020-12-04
Fixes #1161: add function to compress/decompress binary data

### LINKED ISSUE #1161: add function to compress/decompress binary data

vga91:
Fixes #1161

I created `apoc.util.compress` and `apoc.util.decompress` procedures using the same approach as the procedure `apoc.text.bytes`, then compressing the `String` in `List<Long>` and vice versa.


core/src/main/java/apoc/util/CompressionAlgo.java
```diff
@@ -0,0 +1,60 @@
...
+        Constructor<?> constructor = compressor.getConstructor(OutputStream.class);
+        try (ByteArrayOutputStream stream = new ByteArrayOutputStream()) {
+            try (OutputStream outputStream = (OutputStream) constructor.newInstance((OutputStream) stream)) {
+                outputStream.write(string.getBytes(charset));
+            }
+            return convertFromBytesToList(stream.toByteArray());
+        }
+    }
+
+    public String decompress(List<Long> byteArray, Charset charset) throws Exception {
```
jexp, line 52: this should be a byte[]

core/src/main/java/apoc/util/Util.java
```diff
@@ -913,4 +914,19 @@ public static Thread newDaemonThread(Runnable target) {
         thread.setDaemon(true);
         return thread;
     }
+
+    public static List<Long> convertFromBytesToList(byte[] bytes) {
+        List<Long> result = new ArrayList<>(bytes.length);
```
jexp, line 919: this is a big waste of memory we should not do that

PR #1766, opened 2021-01-14
Fixes #1617: Logging for apoc.periodic.*

### LINKED ISSUE #1617: Logging for apoc.periodic.*
It would be great to have more detail on execution of the apoc.periodic.* - specifically periodic.iterate and probably parallel apocs too, could we generate additional logging to show progress of batches e.g.

Processing [batch x] - [size y]?

Not sure whether we'd benefit start/end messages for each batch (to give information on length of time a batch takes? or just enough to know there is progress

## Considered alternatives
Currently we can add apoc.log.* to the periodic iterate command to get some logging

vga91:
Fixes #1617

  - Added unique identifier to periodic procedures
  - Added logs when in `neo4j.conf` we have `dbms.logs.debug.level=DEBUG`

### DIFF
--- a/core/src/main/java/apoc/periodic/Periodic.java
+++ b/core/src/main/java/apoc/periodic/Periodic.java
@@ -63,7 +63,10 @@ public class Periodic {
         Map<String,Long> commitErrors = new ConcurrentHashMap<>();
         AtomicInteger failedBatches = new AtomicInteger();
         Map<String,Long> batchErrors = new ConcurrentHashMap<>();
-
+        String periodicId = UUID.randomUUID().toString();
+        if (log.isDebugEnabled()) {
+            log.debug("Starting periodic commit from `%s` in separate thread with id: `%s`", statement, periodicId);
+        }
         do {
             Map<String, Object> window = Util.map("_count", updates, "_total", total);
             updates = Util.getFuture(pools.getScheduledExecutorService().submit(() -> {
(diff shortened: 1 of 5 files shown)


core/src/main/java/apoc/periodic/Periodic.java
```diff
@@ -64,6 +64,10 @@
         AtomicInteger failedBatches = new AtomicInteger();
         Map<String,Long> batchErrors = new ConcurrentHashMap<>();
 
+        String periodicId = UUID.randomUUID().toString();
+        if (log.isDebugEnabled()) {
+            log.debug("Starting periodic commit from `%s` in separate thread with id: `%s`", statement, periodicId);
```
fbiville, line 69: nit: might be worth removing the `if` and use the `{}` placeholder syntax of slf4j if applicable

fbiville (approved):

core/src/main/java/apoc/periodic/Periodic.java
vga91: Removed `if` statements, with `{}` it doesn't seem to work with `neo4j.logging.Log`..

fbiville: If no access to that, might be better to keep those `if`s for consistency? I'm not sure it really affects performance too much, as it just created a few String instances here and there, but I haven't monitored that all

vga91: You're right, better keep them. Restored

PR #1779, opened 2021-01-21
Fixes #1715: When registering custom function `a.b.c` to access via `custom.a.b.c`, it instead registers at `custom.c`

### LINKED ISSUE #1715: When registering custom function `a.b.c` to access via `custom.a.b.c`, it instead registers at `custom.c`
```
CALL apoc.custom.asFunction('a.b.c', 'RETURN 42')
CALL apoc.custom.list()
```

image

image

### Versions

  - OS: docker (official `neo4j:latest` image) on ubuntu 20.04
  - Neo4j: 4.1.13
  - Neo4j-Apoc: 4.1.0.2

vga91:
Fixes #1715

Improved logic of `apoc.custom.removeFunction()`, `apoc.custom.removeProcedure()` and `apoc.custom.list()` to work with names containing dots.


full/src/test/java/apoc/custom/CypherProceduresTest.java
```diff
@@ -240,6 +240,121 @@ public void shouldRemoveTheCustomProcedure() throws Exception {
...
+        // when
+        db.executeTransactionally("call apoc.custom.removeProcedure('a.b.c')");
+        db.executeTransactionally("call db.clearQueryCaches()");
+
+        // then
+        TestUtil.count(db, "call custom.a.b.c()");
+    }
+
+    @Test
+    public void shouldOverrideCustomFunctionWithDotInNameOnlyIfWithSameNamespaceAndFinalName() throws Exception {
```
fbiville, line 326: might be worth testing list and removal of such functions (e.g.: list displays both, removing 1 does not remove the other...)

full/src/test/java/apoc/custom/CypherProceduresTest.java
```diff
@@ -240,6 +240,121 @@ public void shouldRemoveTheCustomProcedure() throws Exception {
...
+        TestUtil.testCall(db, "return custom.a.b.name() as row", (row) -> assertEquals(34L, ((Map)((List)row.get("row")).get(0)).get("answer")));
+
+        db.executeTransactionally("call apoc.custom.asFunction('x.z.name','RETURN 12 as answer')");
+        TestUtil.testCall(db, "return custom.x.z.name() as row", (row) -> assertEquals(12L, ((Map)((List)row.get("row")).get(0)).get("answer")));
+        TestUtil.testCall(db, "return custom.a.b.name() as row", (row) -> assertEquals(34L, ((Map)((List)row.get("row")).get(0)).get("answer")));
+
+    }
+
+    @Test
+    public void shouldOverrideCustomProcedureWithDotInNameOnlyIfWithSameNamespaceAndFinalName() throws Exception {
```
fbiville, line 343: same as above

fbiville (approved):

full/src/test/java/apoc/custom/CypherProceduresTest.java
vga91: Well yes, in this way I found another case that I had not considered :( . 
Now I have updated the code and the test case.

full/src/test/java/apoc/custom/CypherProceduresTest.java
vga91: Updated

PR #1780, opened 2021-01-21
Fixes #1639: add apoc.load.directory(...) yield value

### LINKED ISSUE #1639: add apoc.load.folder("pattern", recursive=true/false) yield url
Allow to load the "import" folder recursively so if there are multiple nested files in directories you can get them all

so you could then use 

```
call apoc.load.folder("*.csv", true) yield url
load csv with headers from url as row ....

```

vga91:
Fixes #1639

  - Added `apoc.load.directory('pattern', 'urlDir', \{config}) YIELD value` to get a list of all files in folder specified by "urlDir" or in import folder if "urlDir" string is empty or not specified


full/src/main/java/apoc/load/LoadDirectory.java
```diff
@@ -0,0 +1,58 @@
+package apoc.load;
+
+import apoc.Extended;
+import apoc.result.StringResult;
+import apoc.util.FileUtils;
+import apoc.util.Util;
+import org.apache.commons.io.filefilter.TrueFileFilter;
+import org.apache.commons.io.filefilter.WildcardFileFilter;
+import org.neo4j.logging.Log;
+import org.neo4j.procedure.*;
```
fbiville, line 10: nit: I tend to avoid wildcards (not grep-friendly)

full/src/main/java/apoc/load/LoadDirectory.java
```diff
@@ -0,0 +1,58 @@
...
+import static org.eclipse.jetty.util.URIUtil.encodeSpaces;
+
+@Extended
+public class LoadDirectory {
+
+    @Context
+    public Log log;
+
+    @Procedure
+    @Description("apoc.load.directory('pattern', 'urlDir', {config}) YIELD value - List of all files in folder specified by urlDir or in import folder if urlDir string is empty or not specified")
```
fbiville, line 29:
```suggestion
    @Description("apoc.load.directory('pattern', 'urlDir', {config}) YIELD value - Loads list of all files in folder specified by urlDir or in import folder if urlDir string is empty or not specified")
```

full/src/main/java/apoc/load/LoadDirectory.java
```diff
@@ -0,0 +1,58 @@
...
+        String dirImport = apocConfig().getString("dbms.directories.import", "import");
+
+        urlDir = urlDir.isEmpty()
+                ? dirImport
+                : FileUtils.changeFileUrlIfImportDirectoryConstrained(urlDir);
+
+        boolean isRecursive = Util.toBoolean(config.getOrDefault("recursive", true));
+
+        Collection<File> files = org.apache.commons.io.FileUtils.listFiles(
+                Paths.get(URI.create(encodeSpaces(urlDir)).getPath()).toFile(),
```
fbiville, line 45: encoding only spaces is probably not enough... I think you can get all kinds of weird characters that would need to be escaped

full/src/test/java/apoc/load/LoadDirectoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Test
+    public void testWithFilterAllRecursiveFalse() {
+        apocConfig().setProperty(APOC_IMPORT_FILE_USE_NEO4J_CONFIG, true);
+        testResult(db, "CALL apoc.load.directory('*', '', {recursive: false}) YIELD value RETURN value", result -> {
+                    List<Map<String, Object>> rows = Iterators.asList(result.columnAs("value"));
+                    assertTrue(rows.contains("TestCsv3.csv"));
+                    assertTrue(rows.contains("TestJson1.json"));
+                    assertEquals(5, rows.size());
```
fbiville, line 93: what about asserting the 3 other file names explicitly?

full/src/test/java/apoc/load/LoadDirectoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+import java.io.File;
+import java.util.List;
+import java.util.Map;
+
+import static apoc.ApocConfig.*;
+import static apoc.util.TestUtil.testResult;
+import static org.junit.Assert.*;
+
+
+public class LoadDirectoryTest {
```
fbiville, line 21: it would be good to add some tests with `recursive: true`

fbiville (changes requested):

full/src/test/java/apoc/load/LoadDirectoryTest.java
vga91: certainly, I added them

full/src/test/java/apoc/load/LoadDirectoryTest.java
vga91: Added 2 test with `{recursive: true}`

full/src/main/java/apoc/load/LoadDirectory.java
vga91: Of course, I improved the encoding and changed `IMPORT_DIR` and `SUBFOLDER` Strings in `LoadDirectoryTest` with some not-standard characters

full/src/test/java/apoc/load/LoadDirectoryTest.java
```diff
@@ -39,12 +39,12 @@ public static void setUp() throws Exception {
...
         temporaryFolder.newFile("Foo.csv");
         temporaryFolder.newFile("Bar.csv");
-        temporaryFolder.newFile("Baz.xls");
+        temporaryFolder.newFile("Baz.xsl");
         temporaryFolder.newFolder(IMPORT_DIR + File.separator + SUBFOLDER);
         temporaryFolder.newFile(IMPORT_DIR + File.separator + "TestCsv1.csv");
         temporaryFolder.newFile(IMPORT_DIR + File.separator + "TestCsv2.csv");
         temporaryFolder.newFile(IMPORT_DIR + File.separator + "TestCsv3.csv");
-        temporaryFolder.newFile(IMPORT_DIR + File.separator + "TestXls1.xsl");
+        temporaryFolder.newFile(IMPORT_DIR + File.separator + "Testxsl1.xsl");
```
fbiville, line 49: not that it matters too much, but I think `xls` was actually the correct extension

PR #1791, opened 2021-01-28
fixes #1788: create a metagraph of the results of a cypher query (viz similar to apoc.meta.graph's output)

### LINKED ISSUE #1788: create a metagraph of the results of a cypher query (viz similar to apoc.meta.graph's output)
Feature request:   function to create a metagraph of the results of a cypher query, in the same way that apoc.meta.graph does for the entire graph.  

Reference: https://community.neo4j.com/t/best-way-to-create-a-metagraph-of-the-results-from-a-cypher-query-that-returns-paths/32527

Alternative?  I'm currently attempting to do this with the non-optimal Cypher statement below (note: this cypher only works for paths, the function/procedure should I believe work for the results of any query that returns a graph like result consisting of paths, nodes/rels, at least it should without too many contortions to finagle the input)

MATCH p= (p1:Supplier {supplierID:'20'})-[*..2]->(p2)    // PUT QUERY HERE - nodes
UNWIND nodes(p) as n
with distinct n as dn
WITH labels(dn) as l
WITH apoc.create.vNode(l,{name:head(l), count:count(l)}) as metaNodes
WITH apoc.map.groupBy(collect(metaNodes),'name') as ml
MATCH p=(p1:Supplier {supplierID:'20'})-[*..2]->(p2)  // PUT QUERY HERE TOO - relationships
unwind relationships(p) as r
WITH head(labels(startNode(r))) as cFrom, head(labels(endNode(r))) as cTo, count(*) as count, ml, type(r) as tr
RETURN ml[cFrom] as from, ml[cTo] as to, apoc.create.vRelationship(ml[cFrom],tr,{count:count},ml[cTo]) as rel

This features enables an efficient look at the results of a query "at a glance", counts of nodes and relationships, it would be particularly helpful 1. in providing a summary view of the entire result set to a user when it is necessary to clip/truncate the output for display 2. while using cypher iteratively to explore high degree heterogeneous graphs, with this one can see the overall quantities at a glance, and it would be very helpful in seeing what is being found with any given query. (at a high level, quickly, without returning thousands of nodes)

Maybe a screenshot or design?
See reference URL above for a simple example with images.

conker84:
Fixes #1788

Added two procedures:

 * `apoc.meta.data.of(<input>, <config>)`
 * `apoc.meta.graph.of(<input>, <config>)`

that allows computing the metagraph/metadata of the results of:

* a cypher query
* a Virtual graph returned by `apoc.graph.*`
* a map with two field `nodes` (a list of nodes, *mandatory*), `relationships` (a list of relationships)

  - Added the feature
  - Added the documentation
  - Added tests


core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -931,4 +948,42 @@ public void testNodeTypePropertiesEquivalenceTypeMapping() throws Exception {
...
+                "(p)-[:VIEWED]->(m), (p)-[:BOUGHT{quantity: 10}]->(pr)");
+
+        TestUtil.testResult(db, "CALL apoc.meta.graph.of('MATCH p = ()-[:BOUGHT]->() RETURN p')",
+                (r) -> {
+                    Map<String, Object> row = r.next();
+                    List<Node> nodes = (List<Node>) row.get("nodes");
+                    List<Relationship> relationships = (List<Relationship>) row.get("relationships");
+                    assertEquals(2,nodes.size());
+                    assertEquals(1,relationships.size());
+                    boolean hasLabels = nodes.stream().allMatch(n -> n.hasLabel(Label.label("Person")) || n.hasLabel(Label.label("Product")));
```
fbiville, line 984: maybe rewrite this assertions, as it would currently pass if you had 2 `:Person` nodes or 2 `:Product` nodes

core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -931,4 +948,42 @@ public void testNodeTypePropertiesEquivalenceTypeMapping() throws Exception {
...
+
+        TestUtil.testResult(db, "CALL apoc.meta.data.of('MATCH p = ()-[:BOUGHT]->() RETURN p')",
+                (r) -> {
+                    Set<Map<String, Object>> result = r.stream().collect(Collectors.toSet());
+                    assertEquals(expectedResult, result);
+                });
+    }
+
+    @Test
+    public void testMetaGraphOf() throws Exception {
```
fbiville, line 973: could you add test for the 2 other supported input type (`Map` and `Virtual Graph`)?

fbiville (changes requested):

core/src/test/java/apoc/meta/MetaTest.java
conker84: done!

conker84:
@fbiville thank you for the review, I updated the code base with your feedback!

fbiville (approved):

fbiville (approved):

Issue #1769, opened 2021-01-18
Concurrency parameter to apoc.periodic.iterate is ignored


meiamsome:
## Expected Behavior (Mandatory)
For the cypher provided, the statements should be executed sequentially for a total execution time of approximately 5 seconds.

## Actual Behavior (Mandatory)
The cypher executes all steps in parallel regardless of the `concurrency` parameter and completes in approximately 1 second.

## How to Reproduce the Problem

The `apoc.jobs.pool.num_threads` option must be set to be > 1 (Preferrably > 5).
Execute the following cypher:
```cypher
CALL apoc.periodic.iterate(
    "
        UNWIND range(1, 5) AS x
        RETURN x
    ",
    "
        CALL apoc.util.sleep(1000)
        RETURN 0
    ",
    {
      batchSize: 1,
      parallel: true,
      concurrency: 1
    }
 )
```

## Specifications (Mandatory)

Currently used versions

### Versions
  - OS: Mac OS X
  - Neo4j: 4.0.3
  - Neo4j-Apoc: 4.0.0.10

## Analysis

It appears the code for handling this parameter is currently commented out:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/fd5963c1bf1bff87b228338c8683db89dcfc9cb7/core/src/main/java/apoc/periodic/Periodic.java#L282-L295

voutilad:
@meiamsome Just had a PR merged that should fix this...care to validate?

conker84:
this should be fixed, feel free to reopen in case it come back

PR #1794, opened 2021-01-29
fix apoc.periodic.iterate not obeying concurrency

voutilad:
We've noticed that `apoc.periodic.iterate` has not been obeying the `concurrency` parameter setting. 

Looking into the code, it seems the procedure uses the default pooled executor when running with parallel=true, but the pool is `#cpus * 2`...so the concurrency ends up actually being **2x** the cpu cores and not what the user sets.

This PR changes the behavior to use a set cap via an `AtomicInteger` and keeps the same Future creation/submission logic but adds the decrement of the counter to the future's callback. 

If `parallel == false`, the gating is short-circuited at the expense of some small overhead for manipulating the `AtomicInteger` counter (for now?) and the single-threaded executor will force concurrency to zero.

### DIFF
--- a/core/src/main/java/apoc/periodic/Periodic.java
+++ b/core/src/main/java/apoc/periodic/Periodic.java
@@ -16,6 +16,7 @@ import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.LockSupport;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 import java.util.function.ToLongFunction;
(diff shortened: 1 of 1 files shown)


voutilad:
This isn't ready to merge...should have set it to a Draft PR...need to validate the cancellation behavior. Right now my manual testing has me questioning how the batch counters are being incremented during a cancellation.

voutilad:
Ok, it might be my work is just uncovering other existing issues with `apoc.periodic.iterate` such as the inner transactions working on the batches don't get properly interrupted if the outer-most transaction is asked to terminate.

voutilad:
Seems someone recently reported this problem: https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1769

conker84:
@voutilad you should be able to convert it to draft if you want

PR #1798, opened 2021-02-01
Fixes #1714: Weirdness around custom functions with node return types

### LINKED ISSUE #1714: Weirdness around custom functions with node return types
## How to Reproduce

1. Set up test nodes
	```
	UNWIND range(1, 5) as val
 	CREATE (i:Target {value: val});
	```
2. Declare a simple function that returns a node 
(a) `CALL apoc.custom.declareFunction('n(val :: INTEGER) :: NODE ', "MATCH (t:Target {value : $val}) RETURN t")` or (b) `CALL apoc.custom.asFunction('n', "MATCH (t:Target {value : $val}) RETURN t", 'NODE', [['val', 'INTEGER']])`

3. Call the function `RETURN custom.n(2)`. Go to the "table" view and note how node is nested within a key "t". 
4. Try to de-nest with `RETURN custom.n(3)["t"]` or `WITH custom.n(3) AS ans RETURN ans.t`

## Actual v/s Expected Behavior

Buggy behaviour in both 2a and 2b: the nesting within `t`. The return type is `NODE` and not `MAP OF NODE`. I believe this nesting doesn't happen for simple return types like "string" in the latest versions (but was indeed happening before I updated from the older neo4j `4.0.7`)

Buggy behaviour in 2a: step 3 fails with "Unknown function 'custom.n'", i.e. you can't even de-nest it as a workaround.

### Versions

  - OS: docker (official `neo4j:latest` image) on ubuntu 20.04
  - Neo4j: 4.1.13
  - Neo4j-Apoc: 4.1.0.2

vga91:
Fixes #1714

  - Procedures/functions now return node when the output is 'NODE'
  - Added IllegalStateException when the procedure name has only one char

### DIFF
--- a/full/src/test/java/apoc/custom/CypherProceduresStorageTest.java
+++ b/full/src/test/java/apoc/custom/CypherProceduresStorageTest.java
@@ -8,6 +8,7 @@ import org.junit.rules.TemporaryFolder;
 import org.neo4j.configuration.GraphDatabaseSettings;
 import org.neo4j.dbms.api.DatabaseManagementService;
 import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.graphdb.Node;
 import org.neo4j.test.TestDatabaseManagementServiceBuilder;
 
 import java.io.IOException;
@@ -143,5 +144,13 @@ public class CypherProceduresStorageTest {
                 "[['int','int'],['float','float'],['string','string'],['map','map'],['list int','list int'],['bool','bool'],['date','date'],['datetime','datetime'],['point','point']], true)");
         restartDb();
         TestUtil.testCall(db, "return custom.answer(42,3.14,'foo',{a:1},[1],true,date(),datetime(),point({x:1,y:2})) as data", (row) -> assertEquals(9, ((List)row.get("data")).size()));
+    }
+
+    @Test
+    public void testIssue1714() throws Exception {
+        db.executeTransactionally("CREATE (i:Target {value: 2});");
+        db.executeTransactionally("CALL apoc.custom.asFunction('fun', 'MATCH (t:Target {value : $val}) RETURN t', 'NODE', [['val', 'INTEGER']])");
+        restartDb();
+        TestUtil.testCall(db, "RETURN custom.fun(2) as row", (row) -> assertEquals(2L, ((Node) row.get("row")).getProperty("value")));
     }
 }
--- a/full/src/test/java/apoc/custom/CypherProceduresStorageTest.java
+++ b/full/src/test/java/apoc/custom/CypherProceduresStorageTest.java
@@ -147,10 +147,10 @@ public class CypherProceduresStorageTest {
     }
 
     @Test
-    public void testIssue1714() throws Exception {
+    public void testIssue1714WithRestartDb() throws Exception {
         db.executeTransactionally("CREATE (i:Target {value: 2});");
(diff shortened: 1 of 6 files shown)


fbiville:
@vga91 @conker84 is it ready for review?

vga91:
> @vga91 @conker84 is it ready for review?

@fbiville Yep, it's ready. Thank you.

full/src/test/java/apoc/custom/CypherProceduresTest.java
```diff
@@ -9,10 +9,7 @@
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
-import org.neo4j.graphdb.GraphDatabaseService;
-import org.neo4j.graphdb.Node;
-import org.neo4j.graphdb.QueryExecutionException;
-import org.neo4j.graphdb.Transaction;
+import org.neo4j.graphdb.*;
```
fbiville, line 12: nit: keep single-class import if possible, please

full/src/test/java/apoc/custom/CypherProceduresStorageTest.java
```diff
@@ -144,4 +145,12 @@ public void testAllParameterTypesFunction() throws Exception {
         restartDb();
         TestUtil.testCall(db, "return custom.answer(42,3.14,'foo',{a:1},[1],true,date(),datetime(),point({x:1,y:2})) as data", (row) -> assertEquals(9, ((List)row.get("data")).size()));
     }
+
+    @Test
+    public void testIssue1714WithRestartDb() throws Exception {
+        db.executeTransactionally("CREATE (i:Target {value: 2});");
+        db.executeTransactionally("CALL apoc.custom.asFunction('n', 'MATCH (t:Target {value : $val}) RETURN t', 'NODE', [['val', 'INTEGER']])");
+        restartDb();
+        TestUtil.testCall(db, "RETURN custom.n(2) as row", (row) -> assertEquals(2L, ((Node) row.get("row")).getProperty("value")));
```
fbiville, line 154: did the test fail before the fixes?

full/src/test/java/apoc/custom/CypherProceduresTest.java
```diff
@@ -114,6 +194,13 @@ public void registerSimpleStatementFunction() throws Exception {
         TestUtil.testCall(db, "return custom.answer2() as row", (row) -> assertEquals(42L, row.get("row")));
     }
 
+    @Test
+    public void registerSimpleStatementFunctionWithOneChar() throws Exception {
+        db.executeTransactionally("call apoc.custom.asFunction('a','RETURN 42 as answer')");
+        TestUtil.testCall(db, "return custom.a() as row", (row) -> assertEquals(42L, ((Map)((List)row.get("row")).get(0)).get("answer")));
+        TestUtil.testFail(db, "CALL apoc.custom.declareFunction('b() :: STRING','RETURN 42 as answer')", QueryExecutionException.class);
```
fbiville, line 201: might be useful to test the error message

fbiville (approved):

full/src/test/java/apoc/custom/CypherProceduresStorageTest.java
vga91: Yes, with a `ClassCastException: class java.util.HashMap cannot be cast to class org.neo4j.graphdb.Node`.

full/src/test/java/apoc/custom/CypherProceduresTest.java
vga91: Changed

PR #1804, opened 2021-02-04
Added missing documentation for round mode

gitclem:
I copied the documentation for rounding modes from https://docs.oracle.com/javase/7/docs/api/java/math/RoundingMode.html

I also added that UNNECESSARY will throw a Java Exception

  - added explanation of Mode
  - added quotes around the strings that are passed to Mode.
  -


fbiville (changes requested):
Hello, the details documentation of this function is in https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.2/docs/asciidoc/modules/ROOT/pages/mathematical/math-functions.adoc. Could you please move your description there?

fbiville:
Hi @gitclem, have you had a chance to look at the comment above?

conker84:
@gitclem ping :)

gitclem:
Sorry for late reply, my inbox overflowed.

1. The quotes around the option keywords are needed, because the call won't work without them.
2. I'm a bit confused by how things are set up in github for this project.  This links are 404:
a. https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/26b4b865c83872f35117556b9167bc5d65f289b0/docs/asciidoc/modules
3. Do you mean fix this page?
https://neo4j.com/labs/apoc/4.2/overview/apoc.math/apoc.math.round/

Also on that page, there is a link:
More documentation of [apoc.math.round](https://neo4j.com/labs/apoc/4.2/mathematical/math-functions/)
This link is useless and confusing because it doesn't actually describe apoc.math.round at all.

I think somebody has in mind what they want, but I'm not sure how to make it they way they want it...  Sorry for being obtuse...

I think it would be easier if somebody were to just copy my changes (text I took from Java Docs) and put them where they want it.

fbiville:
@gitclem actually, my previous comment was wrong. The documentation is defined in [the Java method](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/c1861fc37293a9e303c4a6cbdc55bb9c6738a5f9/core/src/main/java/apoc/math/Maths.java#L16) directly.
Regarding your other points:

1. that's right, `mode` is a `String` and its value must be enclosed in double quotes
2. not sure how you got that link?
3. this page is deployed from the documentation generated from this repository. That documentation is generated from the source (see above link).

We'll take care of the change. Thanks for starting the work!

fbiville:
Superseded by https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1837.
@gitclem we opened https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1836 to make sure documentation contributions are more accessible.

PR #1811, opened 2021-02-17
fixes neo4j container tests in case the container not exists

conker84:
Fixes TeamCI build for branch 4.3

### DIFF
--- a/core/src/test/java/apoc/StartupTest.java
+++ b/core/src/test/java/apoc/StartupTest.java
@@ -1,11 +1,10 @@
 package apoc;
 
 import apoc.util.Neo4jContainerExtension;
+import apoc.util.TestContainerUtil;
 import apoc.util.TestUtil;
 import org.junit.Test;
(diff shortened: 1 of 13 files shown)


core/src/test/java/apoc/StartupTest.java
```diff
@@ -39,8 +38,11 @@ public void test() {
 
             neo4jContainer.close();
         } catch (Exception ex) {
-            ex.printStackTrace();
-            fail("Should not have thrown exception when trying to start Neo4j: " + ex);
+            // if Testcontainers wasn't able to retrieve the docker image we ignore the test
+            if (TestContainerUtil.isDockerImageAvailable(ex)) {
```
fbiville, line 42: What about using [JUnit assumptions](https://junit.org/junit4/javadoc/4.13/org/junit/Assume.html) instead? This will skip the test entirely.

fbiville (approved):

PR #1816, opened 2021-02-23
Fixes #28: Remove items from linked lists and reconnect procedure

### LINKED ISSUE #28: Remove items from linked lists and reconnect procedure
Based on the following example : 

http://stackoverflow.com/questions/37088730/delete-several-items-from-a-linked-list

vga91:
Fixes #28

  - added procedure `apoc.refactor.deleteAndReconnect(Path, ListOfNodesToRemove, {config})`
  - config `relationshipSelectionStrategy`: if `incoming`, the incoming relationship will be attached to the next node, if `outcoming` the outcoming relationship will be attached, if `merge` will be attacched a merge of incoming and outcoming relationships. With the same property name, takes precedence the incoming node.
  - config `properties`: Will be considered only if `relationshipSelectionStrategy` is merge. If `discard`, start node wins, vice versa if `overwrite` or `override`, or `combine` that create an array of incoming+outcoming properties (if both exist)


core/src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -452,6 +458,76 @@ public void categorize(
...
+    @Procedure(mode = Mode.WRITE)
+    @Description("apoc.refactor.deleteAndReconnect([pathLinkedList], [nodesToRemove], {config}) - Removes some nodes from a linked list")
+    public Stream<GraphResult> deleteAndReconnect(@Name("path") Path path, @Name("nodes") List<Node> nodesToRemove, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) {
+
+        RefactorConfig refactorConfig = new RefactorConfig(config);
+
+        List<Node> nodes = IterableUtils.toList(path.nodes());
+        Set<Relationship> rels = Iterables.asSet(path.relationships());
+
+        if (!nodes.containsAll(nodesToRemove)) {
```
fbiville, line 470: should the nodes just be filtered out instead of raising an error here?
If none of the nodes are in the list, then, do nothing?
WDYT?

core/src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -452,6 +458,76 @@ public void categorize(
...
+            throw new RuntimeException("Some nodes in nodesToRemove not present in the path");
+        }
+
+        BiFunction<Node, Direction, Relationship> filterRel = (node, direction) -> StreamSupport
+                .stream(node.getRelationships(direction).spliterator(), false)
+                .filter(rels::contains)
+                .findFirst()
```
fbiville, line 477: to be consistent with the first strict check, shouldn't an error be raised if more for than 1 relationship matches?

core/src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -452,6 +458,76 @@ public void categorize(
...
+                .orElse(null);
+
+        nodesToRemove.forEach(node -> {
+
+            Relationship relationshipIn = filterRel.apply(node, Direction.INCOMING);
+            Relationship relationshipOut = filterRel.apply(node, Direction.OUTGOING);
+
+            // if terminal node
```
fbiville, line 485:
```suggestion
            // if initial or terminal node
```

core/src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -452,6 +458,76 @@ public void categorize(
...
+            if (relationshipIn == null || relationshipOut == null) {
+                rels.remove(relationshipIn == null ? relationshipOut : relationshipIn);
```
fbiville, line 487: the `remove` call *may* fail depending on the actual `Set` concrete type of `rels` when the linked list is a single, disconnected node

core/src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -452,6 +458,76 @@ public void categorize(
...
+                        break;
+
+                    case OUTCOMING:
+                        newRelType = relationshipOut.getType();
+                        newRelProps.putAll(relationshipOut.getAllProperties());
+                        break;
+
+                    default:
+                        newRelType = RelationshipType.withName(relationshipIn.getType() + "_" + relationshipOut.getType());
+                        newRelProps.putAll(relationshipIn.getAllProperties());
```
fbiville, line 510: what about the properties of `relationshipOut`? 
and then, what about property naming conflicts between the two relations?

fbiville: oh, it's done afterwards... any change to move that here?

core/src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -452,6 +458,76 @@ public void categorize(
...
+
+                if (strategy == MERGE) {
+                    mergeProperties(relationshipOut.getAllProperties(), relCreated, refactorConfig);
+                }
+
+                rels.add(relCreated);
+                rels.removeAll(List.of(relationshipIn, relationshipOut));
+            }
+
+            tx.execute("WITH $node as n DETACH DELETE n", Map.of("node", node));
```
fbiville, line 524: where is the Cypher query that connects the previous and next nodes together?

docs/asciidoc/modules/ROOT/partials/usage/apoc.refactor.deleteAndReconnect.adoc
```diff
@@ -0,0 +1,82 @@
+The examples in this section are based on the following sample graph:
+
+Let's suppose we have a simple dataset like:
```
fbiville, line 3:
```suggestion
Let's suppose we have a simple data set like:
```

docs/asciidoc/modules/ROOT/partials/usage/config/apoc.refactor.deleteAndReconnect.adoc
```diff
@@ -0,0 +1,18 @@
+The procedure support the following config parameters:
```
fbiville, line 1:
```suggestion
The procedure supports the following config parameters:
```

docs/asciidoc/modules/ROOT/partials/usage/config/apoc.refactor.deleteAndReconnect.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+.Config parameters
+[opts=header, cols="1,1,1,5"]
+|===
+| name | type | default | description
+| relationshipSelectionStrategy | Enum[incoming, outcoming, merge] | true | if `incoming`, the incoming relationship will be attached to the next node, if `outcoming`, the outcoming relationship will be attached, if `merge` will be attacched a merge of incoming and outcoming relatioships. With the same property name, takes precedence the incoming node.
```
fbiville, line 7:
```suggestion
| relationshipSelectionStrategy | Enum[incoming, outcoming, merge] | true | if `incoming`, the incoming relationship will be attached to the next node. If `outcoming`, the outcoming relationship will be attached. If `merge`, a merge of the incoming and outcoming relationships will be attached. If they share some property names, the incoming relationship properties have higher precedence by default.
```

core/src/main/java/apoc/refactor/GraphRefactoring.java
fbiville: Forget my 2nd question, the answer is in the doc :)

fbiville (changes requested):

core/src/main/java/apoc/refactor/GraphRefactoring.java
vga91: On line 513, `Relationship relCreated = nodeIncoming.createRelationshipTo(nodeOutcoming, newRelType);`
This line deletes the intermediate node.
So with a path like `(A)-->(B)-->(C)`, if i want to delete node `B`, first I create connection between `A` and `C`, and then I delete `B`.

core/src/main/java/apoc/refactor/GraphRefactoring.java
vga91: About the last question, maybe I don't think it is possible. 
I first need to create the relation on line `513`, with `relType` based on `switch case`. 
And then I can merge the properties of `relationshipOut` using the `PropertiesManager.mergeProperties(..)` method.

core/src/main/java/apoc/refactor/GraphRefactoring.java
vga91: Do you mean in a situation like this? 
`MATCH p=(f:Three) WITH p, [f] as list CALL apoc.refactor.deleteAndReconnect(p, list)`.
In this case the procedure continues to work correctly, I added a test on line `119` with this case

core/src/main/java/apoc/refactor/GraphRefactoring.java
vga91: Yep, I agree. Changed

core/src/main/java/apoc/refactor/GraphRefactoring.java
vga91: Now I have removed the first error. Therefore the consistence (that is, raise a `RuntimeException` or other) would no longer exist. I dunno, maybe it's better not to throw the exception here?

core/src/main/java/apoc/refactor/GraphRefactoring.java
fbiville: Ok, then it appears the particular runtime type of the `Set` allows to call `remove` with `null`.
I'd probably avoid calling remove if `relationshipIn == null ? relationshipOut : relationshipIn` yields `null`, but it's good to have a test with the single node scenario!

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -58,6 +61,268 @@ public void tearDown() {
...
+                    assertEquals("b", rel1.getProperty("a"));
+                    Relationship rel2 = rels.get(1);
+                    assertEquals("GAMMA", rel2.getType().name());
+                    assertTrue(rel2.getAllProperties().isEmpty());
+                    assertNotNull(row.get("nodes"));
+                });
+
+        TestUtil.testCall(db, Util.NODE_COUNT, (row) -> assertEquals(3L, row.get("result")));
+
+        // - terminal node
```
fbiville, line 98: nit: I'd move that to a separate test

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -58,6 +61,268 @@ public void tearDown() {
...
+                    Relationship rel2 = rels.get(1);
+                    assertEquals("GAMMA", rel2.getType().name());
+                    assertNotNull(row.get("nodes"));
+                });
+
+        TestUtil.testCall(db, Util.NODE_COUNT, (row) -> assertEquals(3L, row.get("result")));
+    }
+
+    @Test
+    public void deleteAndReconnectWithOutcomingRelConfig() throws Exception {
```
fbiville, line 192:
```suggestion
    public void deleteAndReconnectWithOutgoingRelConfig() throws Exception {
```

core/src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -452,6 +458,76 @@ public void categorize(
...
+                Map<String, Object> newRelProps = new HashMap<>();
+
+                final RefactorConfig.RelationshipSelectionStrategy strategy = refactorConfig.getRelationshipSelectionStrategy();
+                switch (strategy) {
+                    case INCOMING:
+                        newRelType = relationshipIn.getType();
+                        newRelProps.putAll(relationshipIn.getAllProperties());
+                        break;
+
+                    case OUTCOMING:
```
fbiville, line 503:
```suggestion
                    case OUTGOING:
```

core/src/main/java/apoc/refactor/util/RefactorConfig.java
```diff
@@ -12,6 +12,8 @@
  */
 public class RefactorConfig {
 
+	public enum RelationshipSelectionStrategy {INCOMING, OUTCOMING, MERGE}
```
fbiville, line 15:
```suggestion
	public enum RelationshipSelectionStrategy {INCOMING, OUTGOING, MERGE}
```

core/src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -452,6 +458,76 @@ public void categorize(
...
+            Relationship relationshipIn = filterRel.apply(node, Direction.INCOMING);
+            Relationship relationshipOut = filterRel.apply(node, Direction.OUTGOING);
+
+            // if initial or terminal node
+            if (relationshipIn == null || relationshipOut == null) {
+                rels.remove(relationshipIn == null ? relationshipOut : relationshipIn);
+
+            } else {
+                Node nodeIncoming = relationshipIn.getStartNode();
+                Node nodeOutcoming = relationshipOut.getEndNode();
```
fbiville, line 491:
```suggestion
                Node nodeOutgoing = relationshipOut.getEndNode();
```

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -58,6 +61,268 @@ public void tearDown() {
...
+        db.executeTransactionally("CREATE (f:One)-[:ALPHA {a:'b'}]->(b:Two)-[:BETA {c:'d', e:'f'}]->(c:Three)-[:GAMMA]->(d:Four)-[:DELTA {aa: 'bb', cc: 'dd', ee: 'ff'}]->(e:Five {foo: 'bar', baz: 'baa'})");
+
+        TestUtil.testCall(db, Util.NODE_COUNT, (row) -> assertEquals(5L, row.get("result")));
+
+        TestUtil.testCall(db, "MATCH p=(f:One)-->(b:Two)-->(c:Three)-->(d:Four)-->(e:Five) WITH p, [b,d] as list CALL apoc.refactor.deleteAndReconnect(p, list, {relationshipSelectionStrategy: 'outcoming'}) YIELD nodes, relationships RETURN nodes, relationships",
```
fbiville, line 197:
```suggestion
        TestUtil.testCall(db, "MATCH p=(f:One)-->(b:Two)-->(c:Three)-->(d:Four)-->(e:Five) WITH p, [b,d] as list CALL apoc.refactor.deleteAndReconnect(p, list, {relationshipSelectionStrategy: 'outgoing'}) YIELD nodes, relationships RETURN nodes, relationships",
```

fbiville (changes requested):
s/outcoming/outgoing/ but the rest looks good!

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
vga91: Moved

fbiville (approved):

PR #1817, opened 2021-02-24
Fixes #1793: apoc.load.html procedure fail silently

### LINKED ISSUE #1793: apoc.load.html procedure fail silently
Currently, the `apoc.load.html` procedure returns an error and rolls back the transaction if any error occurs. The error messaging is not really helpful, you can look at https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.1/full/src/main/java/apoc/load/LoadHtml.java#L54.
In my opinion, it would be better to fail silently and just return an empty map with defined keys. Something along the line of:

```
  query.keySet().stream().forEach(key -> {
      output.put(key, empty map);
  });
```

vga91:
Fixes #1793

  - added `failsSilently: true/false` config (default `false`). With `true`, will be returned the list of the values, except for values not correctly parsed. In case of errors, will be printed a `log.warn`
  - improved the error returned, with `UnsupportedEncodingException`, `FileNotFoundException` and, in case of incorrect parse, with a message containing the incorrect tag

### DIFF
--- a/full/src/main/java/apoc/load/LoadHtml.java
+++ b/full/src/main/java/apoc/load/LoadHtml.java
@@ -15,6 +15,8 @@ import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Name;
 import org.neo4j.procedure.Procedure;
 
+import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.util.*;
 import java.util.stream.Stream;
 
@@ -44,14 +46,19 @@ public class LoadHtml {
 
             Map<String, Object> output = new HashMap<>();
 
-            query.keySet().stream().forEach(key -> {
-                Elements elements = document.select(query.get(key));
+            query.keySet().forEach(key -> {
+                try {
+                    Elements elements = document.select(query.get(key));
 
-                output.put(key, getElements(elements, config));
+                    output.put(key, getElements(elements, config));
+
+                } catch (Exception e) {
+                    output.put(key, Collections.emptyMap());
+                }
             });
 
-            return Stream.of( new MapResult(output) );
-        } catch(Exception e){
+            return Stream.of(new MapResult(output));
+        } catch(Exception e) {
             throw new RuntimeException("Can't read the HTML from: "+ url);
         }
     }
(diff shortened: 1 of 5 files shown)


full/src/main/java/apoc/load/LoadHtml.java
```diff
@@ -34,48 +37,61 @@
...
+
+                        result.put("children", getElements(element.children(), config));
+                    }
+                    else {
+                        if(element.hasText()) result.put("text", element.text());
+                    }
+
+                    elementList.add(result);
+            } catch (Exception e) {
+                final String parseError = "Error during parsing element: " + element;
```
fbiville, line 88: maybe errors could be added to the output, in a reserved key? (maybe not for the scope of this PR/issue though)

full/src/test/java/apoc/load/LoadHtmlTest.java
```diff
@@ -122,4 +125,65 @@ public void testQueryWithChildren() {
...
+            throw e;
+        }
+    }
+
+    @Test(expected = QueryExecutionException.class)
+    public void testQueryWithExceptionIfIncorrectUrl() {
+
+        final String filePath = new File("src/test/resources/wikipedia1.html").toURI().toString();
+        try {
+            String query = "CALL apoc.load.html('" + filePath + "',{a:'a'})";
```
fbiville, line 165: might be worth adding the variant with `failSilently: true` (just to "prove" it does not have an effect on unrecoverable errors)
same comment for the other test below

fbiville (approved):

full/src/main/java/apoc/load/LoadHtml.java
vga91: So, you mean, a structure like {`a: [listOfTagA]`, `h1: [list]`, .... `errors: [listOfErrorsIfExist]`} , right?
In this case maybe the `log.warn` at line 90 is useless

vga91: Perhaps, we could leave both the log and the list and transform `failSilently` from boolean to enum, for example `FALSE, WITH_LOG, WITH_LIST`, and with `WITH_LOG`, we remains the `log.warn`, and with `WITH_LIST` we create the `errors: [listOfErrorsIfExist]`

fbiville: sounds good!

vga91: @fbiville Changed as previously described

PR #1821, opened 2021-03-01
fixes #1612: Neo4j 4 on Aura permissions adjustment due to PUBLIC role

### LINKED ISSUE #1612: Neo4j 4 on Aura permissions adjustment due to PUBLIC role
For Neo4j 4 running on Aura, the neo4j user was adjusted to have the role PUBLIC, with some special exceptions.   The reasons why this is the case are complicated, but that's how it works.

APOC has some security checks in it (I think) which is causing some functions to fail specifically on Neo4j 4 running on Aura.

Specific example, this succeeds in the "Neo4j 4 on Aura" environment:

```
CREATE INDEX location_type FOR (l:Location) ON (l.type)
```

But this fails:

```
CALL apoc.cypher.doIt("CREATE INDEX whatever FOR (w:Whatever) ON (w.type)", {}) YIELD value RETURN value
```

```
Schema operations are not allowed for user 'neo4j' with roles [PUBLIC] restricted to TOKEN_WRITE.
```

vga91:
Fixes #1612

  - added `apoc.cypher.runSchema` procedure, like the `apoc.cypher.doIt`, with `mode=SCHEMA`
  - added `apoc.cypher.runWrite`, alias for `apoc.cypher.doIt` (based on https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1612#issuecomment-683685245)


core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -242,6 +243,18 @@ public static String withParamMapping(String fragment, Collection<String> keys)
         return runCypherQuery(tx, statement, params);
     }
 
+    @Procedure(mode = WRITE)
+    @Description("apoc.cypher.runWrite(fragment, params) yield value - alias for apoc.cypher.doIt")
```
fbiville, line 247:
```suggestion
    @Description("apoc.cypher.runWrite(statement, params) yield value - alias for apoc.cypher.doIt")
```

core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -242,6 +243,18 @@ public static String withParamMapping(String fragment, Collection<String> keys)
...
+    public Stream<MapResult> runWrite(@Name("cypher") String statement, @Name("params") Map<String, Object> params) {
+        return doIt(statement, params);
+    }
+
+    @Procedure(mode = SCHEMA)
+    @Description("apoc.cypher.runSchema(fragment, params) yield value - executes query schema fragment with the given parameters")
```
fbiville, line 253:
```suggestion
    @Description("apoc.cypher.runSchema(statement, params) yield value - executes query schema fragment with the given parameters")
```

core/src/test/java/apoc/cypher/CypherTest.java
```diff
@@ -66,12 +71,41 @@ public static void setUp() {
...
         try (Transaction tx = db.beginTx()) {
-            tx.schema().getIndexes().forEach(IndexDefinition::drop);
             tx.schema().getConstraints().forEach(ConstraintDefinition::drop);
+            tx.schema().getIndexes().forEach(IndexDefinition::drop);
             tx.commit();
         }
     }
 
+    @Test
+    public void testDoIt() throws Exception {
```
fbiville, line 81:
```suggestion
    public void testDoItAndRunWrite() throws Exception {
```

fbiville: What about parameterizing the test with a `functionName` parameter. The values would be `"doIt"` and `"runWrite"`. That way, you'd run the exact same test & assertions for both (as suggested by @JMHReif).
You could then get rid of the second half of this test.

core/src/test/java/apoc/cypher/CypherTest.java
```diff
@@ -66,12 +71,41 @@ public static void setUp() {
...
+        testCall(db, "CALL apoc.cypher.runWrite('Match (n:TestTwo) return n',{})",
+                r -> assertEquals("TestTwo", Iterables.single(((Node)((Map) r.get("value")).get("n")).getLabels()).name()));
+
+        testFail(db, "CALL apoc.cypher.runWrite('CREATE INDEX test FOR (w:TestTwo) ON (w.foo)',{})", QueryExecutionException.class);
+    }
+
+    @Test
+    public void testRunSchema() throws Exception {
+        testCallEmpty(db, "CALL apoc.cypher.runSchema('CREATE INDEX test FOR (w:TestOne) ON (w.name)',{})", Collections.emptyMap());
+        testCallEmpty(db, "CALL apoc.cypher.runSchema('CREATE CONSTRAINT testConstraint ON (w:TestTwo) ASSERT w.baz IS UNIQUE',{})", Collections.emptyMap());
```
fbiville, line 101: maybe add a `testFail` operation with a Cypher write statement?

fbiville (approved):

core/src/test/java/apoc/cypher/CypherTest.java
vga91: Imho it might not be worth it, because using the `@Parameterized.Parameters` annotation, all test would be executed twice. Instead with `@ParameterizedTest` in the single test, we should change all the test to make it compatible with junit5, so `@beforeClass`, `@ClassRule` etc. Maybe for 3 lines it's not worth it. However, I created a common function `runWriteAndDoItCommons` with this 2 use cases.

core/src/test/java/apoc/cypher/CypherTest.java
vga91: Added

core/src/test/java/apoc/cypher/CypherTest.java
fbiville: Might be time to migrate to JUnit 5 😇

core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -242,6 +243,21 @@ public static String withParamMapping(String fragment, Collection<String> keys)
...
+    @Procedure(mode = WRITE)
+    @Description("apoc.cypher.runWrite(statement, params) yield value - alias for apoc.cypher.doIt")
+    public Stream<MapResult> runWrite(@Name("cypher") String statement, @Name("params") Map<String, Object> params) {
+        return doIt(statement, params);
+    }
+
+    @Procedure(mode = SCHEMA)
+    @Description("apoc.cypher.runSchema(statement, params) yield value - executes query schema statement with the given parameters")
+    public Stream<MapResult> runSchema(@Name("cypher") String statement, @Name("params") Map<String, Object> params) {
+        if (!isSchemaOperation(statement)) {
```
fbiville, line 255: I know this function was already there, but parsing Cypher is considered a bad practice, I'd remove the check.

What if there is no explicit `create/drop index/constraint` in the statement?
It could delegate to another procedure doing schema operations e.g.

fbiville (changes requested):
I missed one detail last time

core/src/main/java/apoc/cypher/Cypher.java
vga91: @fbiville It's strange, but removing this condition, the `testFail(db, "CALL apoc.cypher.runSchema('CREATE (n:Test {a: 1})',{})", QueryExecutionException.class);` in `CypherTest.java` doesn't work.
It seems like the `mode = SCHEMA` considers also WRITE operations.

To avoid duplicate code, maybe I can wait for the merge of the `validateQuery` that I created in this other pr
https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1843/files#diff-0a04de30ec5509e502658b09de532ed90b645feae019d677dd83eb9e81a061db

In this way, with `validateQuery(db, statement, SCHEMA_WRITE)` it works. Wdyt?

fbiville: @JMHReif and I just double-checked Neo4j RBAC https://neo4j.com/docs/operations-manual/current/authentication-authorization/access-control/ and it seems that the default `architect` has schema **and** read/write privileges.

It seems technically possible to define a custom role with only schema privileges, but how common is that?
Does it make sense to enforce that `runSchema` can be run only with schema-only statements?

If you think so, the approach you took in the other PR makes sense (i.e. using `EXPLAIN`).

In short: should we drop the check or rely on `EXPLAIN`?

vga91: Thinking about it.. it makes little sense to impose this restriction.
Ok, I removed the `isSchemaOperation`

fbiville (approved):

PR #1832, opened 2021-03-23
Fixes #1337: apoc.truncate

### LINKED ISSUE #1337: apoc.truncate

This should wipe a database without causing an OOM. It will therefore need to not be transactional i.e. if it encounters an error during execution it could result in a partially deleted graph rather than a totally deleted graph.

under the hood it should do something like this:
```
call apoc.periodic.iterate("cypher runtime=slotted match ()-[r]->() return r","delete r", {batchSize:10000,iterateList:true})
call apoc.periodic.iterate("cypher runtime=slotted match (n) return n","delete n", {batchSize:10000,iterateList:true})
```

vga91:
Fixes #1337 

  - added `apoc.periodic.truncate({config})` that calls the `apoc.periodic.iterate` procedure
  - all configs in `apoc.periodic.iterate` procedure are allowed, besides `dropSchema` that drops all indexes and constraints


core/src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -420,6 +423,38 @@ public void testIterateQueryFail() {
...
+        db.executeTransactionally("CREATE CONSTRAINT ON (a:Actor) ASSERT a.name IS UNIQUE");
+
+        final int expectedNodes = 99999 * 3;
+        final int expectedRels = 99999 * 2;
+        assertCountEntitiesAndIndexes(expectedNodes, expectedRels, 4,2);
+
+        TestUtil.testCallEmpty(db, "CALL apoc.periodic.truncate", Collections.emptyMap());
+        assertCountEntitiesAndIndexes(0, 0, 4,2);
+
+        db.executeTransactionally("UNWIND range(1,99999) AS x CREATE (:One{name:'Person_'+x})-[:FOO {id: x}]->(:Two {surname: 'Two'+x})<-[:BAR {idBar: x}]-(:Three {other: x+'Three'})");
```
fbiville, line 442: nit: I'd move the `dropSchema:true` scenario to a separate test, in order to make sure that someone does not introduce a regression such that `dropSchema:true` would e.g. only drop the schema and not the data (which the current test would not catch).

core/src/main/java/apoc/periodic/Periodic.java
```diff
@@ -40,6 +43,21 @@
     @Context public Pools pools;
     @Context public Transaction tx;
 
+    @Admin
+    @Procedure(mode = Mode.SCHEMA)
+    @Description("apoc.periodic.truncate($config) - removes all entities (and optionally indexes) from db using the apoc.periodic.iterate under the hood")
```
fbiville, line 48:
```suggestion
    @Description("apoc.periodic.truncate($config) - removes all entities (and optionally indexes and constraints) from db using the apoc.periodic.iterate under the hood")
```

fbiville (approved):

fbiville (approved):

PR #1834, opened 2021-03-24
Added auto cherry-picking action: branch 4.2

conker84:
Same as #1833 but for branch `4.2`

### DIFF
--- a/.github/workflows/auto-cherry-pick.yml
+++ b/.github/workflows/auto-cherry-pick.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         target-branch: ['4.1', '4.3']
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-cherry-pick.yml
```diff
@@ -3,13 +3,13 @@ on:
...
     branches:
-      - '4.1'
+      - '4.2'
 jobs:
   cherry-pick:
     strategy:
       fail-fast: false
       matrix:
-        target-branch: ['4.2', '4.3']
+        target-branch: ['4.1', '4.3']
```
fbiville: isn't there a risk of cascading cherry-picks?
Like 4.1 cherry-picks to 4.2, triggering another cherry-pick to 4.1 etc?

PR #1842, opened 2021-03-29
Fixes #1685: apoc.convert.toTree() returns incorrect results in 4.1.x.x

### LINKED ISSUE #1685: apoc.convert.toTree() returns incorrect results in 4.1.x.x
After passing in a collection of paths, `apoc.convert.toTree()` should display a tree populated by all elements found within those paths.

The tree results seem to only be from a single path. It does not return to the depths explored in the paths, and most of the data that should be in the tree is missing.

The movies graph from `:play movies`

### Steps (Mandatory)

  1. Load the movies graph
  2. Execute the following:

```
match path = (k:Person {name:'Keanu Reeves'})-[*..5]-(x)
with collect(path) as paths
call apoc.convert.toTree(paths) yield value
return value
```

Observe that the results seem to be only from a single path with a shallow depth. I observed this with a depth of 1 and 1 movie:

```
{
  "acted_in.roles": [
    "Neo"
  ],
  "born": 1964,
  "_type": "Person",
  "name": "Keanu Reeves",
  "acted_in": [
    {
      "acted_in.roles": [
        "Neo"
      ],
      "_type": "Movie",
      "tagline": "Everything that has a beginning has an end",
      "_id": 10,
      "title": "The Matrix Revolutions",
      "released": 2003
    }
  ],
  "_id": 1
}
```

If you try with a different variable-length limit, you may get different results, but still it looks like it's only using a single path.

with `*..6` I observed this with a depth of 2, but only 1 movie and 1 coactor:

```
{
  "acted_in.roles": [
    "Neo"
  ],
  "born": 1964,
  "_type": "Person",
  "name": "Keanu Reeves",
  "acted_in": [
    {
      "acted_in.roles": [
        "Neo"
      ],
      "_type": "Movie",
      "acted_in": [
        {
          "_type": "Person",
          "name": "Carrie-Anne Moss",
          "_id": 2,
          "acted_in.roles": [
            "Trinity"
          ],
          "born": 1967
        }
      ],
      "tagline": "Everything that has a beginning has an end",
      "_id": 10,
      "title": "The Matrix Revolutions",
      "released": 2003
    }
  ],
  "_id": 1
}
```

## Specifications (Mandatory)

The bug only seems to be present in 4.1.x when using APOC 4.1.x.x.
Incorrect behavior was observed in APOC 4.1.0.0 and later, up to 4.1.0.2 (latest).
Correct behavior was observed in APOC 4.0.0.18 and earlier.

### Versions

  - OS: Mac OS 10.14
  - Neo4j: 4.1.x
  - Neo4j-Apoc: 4.1.0.0 and later (correct behavior was observed in 4.0.0.18 and earlier)

vga91:
Fixes #1685

Now the paths are sorted by length in decreasing order, 
so the longest paths are retrieved first.

### DIFF
--- a/core/src/main/java/apoc/convert/Json.java
+++ b/core/src/main/java/apoc/convert/Json.java
@@ -141,7 +141,8 @@ public class Json {
         Map<String, List<String>> rels = conf.getRels();
 
         Map<Long, Map<String, Object>> maps = new HashMap<>(paths.size() * 100);
-        for (Path path : paths) {
+
+        paths.stream().sorted(Comparator.comparingInt(Path::length).reversed()).forEach(path -> {
             Iterator<Entity> it = path.iterator();
             while (it.hasNext()) {
                 Node n = (Node) it.next();
@@ -165,7 +166,8 @@ public class Json {
                     }
                 }
             }
-        }
+        });
+
         return paths.stream()
                 .map(Path::startNode)
                 .distinct()
--- a/core/src/test/java/apoc/convert/ConvertJsonTest.java
+++ b/core/src/test/java/apoc/convert/ConvertJsonTest.java
@@ -227,6 +227,24 @@ public class ConvertJsonTest {
                 (row) -> assertEquals(asList(1L,2L,3L), row.get("value")) );
     }
 
+    @Test public void testToTreeIssue1685() throws Exception {
+        String movies = Util.readResourceFile("movies.cypher");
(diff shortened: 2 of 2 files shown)


fbiville:
For a later PR maybe, https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1855

fbiville (approved):

PR #1846, opened 2021-03-30
fixes #1840: [TRIGGER] createdRelationships after crash and afterAsync do nothing

### LINKED ISSUE #1840: [TRIGGER] createdRelationships "after" crash and "afterAsync" do nothing

When we add a trigger on `createdRelationships`, it should be executed wether we are in before, after or afterAsync phase.

The trigger on the phase before work, the same trigger on the phase after loop forever and on the phase afterAsync it does nothing.

I have created a new db with dedicated node/relationship to reproduce the issue. See below :

```
// TEST 1. populate DB
CREATE (:A {name: "A"})-[:R1]->(:Z {name: "Z"})

// TEST 2.1 add trigger before
CALL apoc.trigger.add('trigger-before', 'UNWIND $createdRelationships AS r
MATCH (a:A)-[r]->(z:Z)
WHERE (r:R2 OR r:R3)
MATCH (a)-[r1:R1]->(z)
SET r1.triggerBefore = true', {phase: 'before'})

// TEST 2.2 add trigger after
CALL apoc.trigger.add('trigger-after', 'UNWIND $createdRelationships AS r
MATCH (a:A)-[r]->(z:Z)
WHERE (r:R2 OR r:R3)
MATCH (a)-[r1:R1]->(z)
SET r1.triggerAfter = true', {phase: 'after'})

// TEST 2.3 add trigger after async
CALL apoc.trigger.add('trigger-after-async', 'UNWIND $createdRelationships AS r
MATCH (a:A)-[r]->(z:Z)
WHERE (r:R2 OR r:R3)
MATCH (a)-[r1:R1]->(z)
SET r1.triggerAfterAsync = true', {phase: 'afterAsync'})

// TEST 3. (:A)-[:R2]->(:Z)
MATCH (a:A {name: "A"})-[:R1]->(z:Z {name: "Z"})
MERGE (a)-[:R2]->(z)

// TEST 4. get so data
MATCH (n1)-[r]->(n2) RETURN r, n1, n2 LIMIT 25

// TEST 5. clear DB
MATCH (n) DETACH DELETE n

// TEST 6. remove triggers
CALL apoc.trigger.removeAll()
```

### Steps (Mandatory)

  1. Populate the db TEST 1.
  1. Add 1 (or many) of the trigger provided in TEST 2.X
  1. execute TEST 3. to create the new relationship
  1. check the data, clean the db and remove triggers to retry

Testing before :
image

Testing after :
image

Testing afterAsync : 
image

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: Windows 10 Family 19041.867
  - Neo4j: 4.2.4
  - Neo4j-Apoc: 4.2.0.2

conker84:
Fixes #1840

The `afterAsync` phase needs to rebind the entities

  - Fixed the bug
  - Added test

### DIFF
--- a/core/src/main/java/apoc/trigger/TriggerMetadata.java
+++ b/core/src/main/java/apoc/trigger/TriggerMetadata.java
@@ -1,0 +1,215 @@
+package apoc.trigger;
+
+import apoc.convert.Convert;
+import apoc.result.VirtualNode;
+import apoc.result.VirtualRelationship;
+import apoc.util.Util;
+import org.neo4j.graphdb.Entity;
+import org.neo4j.graphdb.Label;
+import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.Relationship;
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.graphdb.event.LabelEntry;
+import org.neo4j.graphdb.event.PropertyEntry;
+import org.neo4j.graphdb.event.TransactionData;
+import org.neo4j.internal.helpers.collection.Iterables;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static apoc.util.Util.map;
+
+public class TriggerMetadata {
+    private final long transactionId;
+    private final long commitTime;
+    private final List<Node> createdNodes;
+    private final List<Relationship> createdRelationships;
+    private final List<Node> deletedNodes;
+    private final List<Relationship> deletedRelationships;
+    private final Map<String, List<Node>> removedLabels;
+    private final Map<String, List<PropertyEntryContainer<Node>>> removedNodeProperties;
+    private final Map<String, List<PropertyEntryContainer<Relationship>>> removedRelationshipProperties;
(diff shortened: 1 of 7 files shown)


core/src/main/java/apoc/trigger/TriggerMetadata.java
```diff
@@ -0,0 +1,196 @@
...
+            txId = -1L;
+        }
+        try {
+            commitTime = txData.getCommitTime();
+        } catch (Exception ignored) {
+            commitTime = -1L;
+        }
+        List<Node> createdNodes = Convert.convertToList(txData.createdNodes());
+        List<Relationship> createdRelationships = Convert.convertToList(txData.createdRelationships());
+        List<Node> deletedNodes = Convert.convertToList(txData.deletedNodes());
```
sarmbruster, line 76: I'm not confident that rebind on deleted entities actually works. Can we please ensure we have a test case that runs deletions and validates rebind on those workds?

conker84: You're right, for these cases I converted them to Virtual Entities, so we can use them

sarmbruster: LGFM now, please feel free to merge. Thanks.

PR #1847, opened 2021-04-01
Fixes #1797: Update couchbase client to 3.1.3

### LINKED ISSUE #1797: Update couchbase client to 3.1.1
Can we please update the couchbase client to the latest version - https://mvnrepository.com/artifact/com.couchbase.client/java-client/3.1.1

I think this will meant that we can't support Couchbase v4 anymore, but I reckon that's ok since it's not even shown on the Couchbase website anymore.

vga91:
Fixes #1797

 - updated `com.couchbase.client` from `2.7.2` to `3.1.3`
  - removed `getNodes()` and `bootstrapHttpDirectPort ` in `CouchbaseManager.java`. Replaced with Set<SeedNode> in `CouchbaseConnection.java`
  - removed config `socketConnectTimeout` and `computationPoolSize` (not available in sdk 3)
  - removed v4 and ignored tests

Tested also with a non-dockerized Couchbase instance (v 6.6.0)


full/src/main/java/apoc/couchbase/CouchbaseConnection.java
```diff
@@ -233,31 +244,15 @@ public JsonDocument replace(String documentId, String json) {
      * @param parameterNames  the placeholders' names in kernelTransaction
      * @param parameterValues the values for the named placeholders in kernelTransaction
      * @return the list of {@link JsonObject}s retrieved by this query
-     * @see N1qlQuery#parameterized(Statement, JsonObject)
+     * @see Cluster#query(String, QueryOptions)
      */
     public List<JsonObject> executeParametrizedStatement(String statement, List<String> parameterNames,
```
fbiville, line 249:
```suggestion
    public List<JsonObject> executeParameterizedStatement(String statement, List<String> parameterNames,
```

full/src/main/java/apoc/couchbase/CouchbaseManager.java
```diff
@@ -138,57 +143,32 @@ protected static URI checkAndGetURI(String hostOrKey) {
...
         PasswordAuthenticator passwordAuthenticator = getPasswordAuthenticator(hostOrKey);
-        List<String> nodes = getNodes(hostOrKey);
-        DefaultCouchbaseEnvironment env = getEnv(hostOrKey);
 
+        // hostOrKey no longer necessary because bootstrapHttpDirectPort is configured via SeedNode in CouchbaseConnection
+        ClusterEnvironment env = getEnv();
 
-        String[] bucketCredentials = bucketName.split(":");
-        return new CouchbaseConnection(nodes, passwordAuthenticator, bucketCredentials[0], bucketCredentials.length == 2 ? bucketCredentials[1] : null, env);
+        // The minimum cluster version supported by SDK 3 is Server 5.0, so is not supported bucket-level passwords
```
fbiville, line 154:
```suggestion
        // The minimum cluster version supported by SDK 3 is Server 5.0, so bucket-level passwords are not supported anymore
```

full/src/main/java/apoc/couchbase/document/CouchbaseJsonDocument.java
```diff
@@ -1,58 +1,73 @@
...
   public Map<String, Object> content;
 
-  public CouchbaseJsonDocument(JsonDocument jsonDocument) {
-    this.id = jsonDocument.id();
-    this.expiry = jsonDocument.expiry();
-    this.cas = jsonDocument.cas();
-    this.mutationToken = CouchbaseUtils.convertMutationTokenToMap(jsonDocument.mutationToken());
-    this.content = (jsonDocument.content() != null) ? this.content = jsonDocument.content().toMap()
-        : new HashMap<String, Object>();
+  public CouchbaseJsonDocument(String id, long expiry, long cas, Map<String, Object> mutationToken, Map<String, Object> content) {
```
fbiville, line 53: nit: do you need the "full" constructor public?

fbiville (changes requested):

full/src/main/java/apoc/couchbase/CouchbaseConnection.java
```diff
@@ -32,182 +43,182 @@
...
-     * @return the updated {@link Document}
-     * @see Bucket#append(Document)
+     * @return the updated {@link MutationResult}
+     * @see BinaryCollection#append(String, byte[])
      */
-    public JsonDocument append(String documentId, String json) {
-        this.bucket.append(JsonDocument.create(documentId, JsonObject.fromJson(json)));
-        return this.get(documentId);
+    public MutationResult append(String documentId, String json) {
+        return binaryCollection.append(documentId, JsonObject.fromJson(json).toBytes());
```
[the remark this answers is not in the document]
vga91: @fbiville oh, sorry about that, I implemented sdk 3 version, assuming the old version was right, that is with a json. Actually, the `Bucket.append(document)` of the sdk2 says: "Note that JsonDocuments in all forms are not supported", so it wasn't right before. 

Now I have changed append and prepend from `JsonObject.fromJson(json).toBytes()` to `json.getBytes()` and added test. 

Does it make sense, at this point, to also change the name of the procedure parameter from `"json"` to `"content"` (consistent with `BinaryCollection.append(..)`) and to directly accept `byte[]` instead of `String`?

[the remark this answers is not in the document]
fbiville: Good idea, `content` is a better name indeed.

fbiville (approved):

full/src/main/java/apoc/couchbase/document/CouchbaseJsonDocument.java
```diff
@@ -1,110 +1,50 @@
...
-    final StringBuilder sb = new StringBuilder(this.getClass().getSimpleName() + " {");
-    sb.append("id='").append(getId()).append('\'');
-    sb.append(", cas=").append(getCas());
-    sb.append(", expiry=").append(getExpiry());
-    sb.append(", content=").append(getContent());
-    sb.append(", mutationToken=").append(getMutationToken());
-    sb.append('}');
-    return sb.toString();
-  }
+public class CouchbaseJsonDocument extends CouchbaseObjectDocument<Map<String, Object>> {
```
fbiville, line 19: is it still used?

fbiville (approved):

full/src/main/java/apoc/couchbase/document/CouchbaseJsonDocument.java
vga91: @fbiville 
Yes, now there are `CouchbaseJsonDocument` and `CouchbaseByteArrayDocument` (the second for `append` and `prepend` procedures).

They both extend the class `CouchbaseObjectDocument` which is almost the same as the old `CouchbaseJsonDocument`.

The difference in this case is that the two extended classes have a different return value of the `content` (`Map<String, Object>` and `byte[]`), and therefore the signature of the procedures, ie `MAP` and `BYTEARRAY`, since with a generic `"Object content"` it would have been `ANY`.

PR #1848, opened 2021-04-01
Fixes #1765: apoc.refactor.mergeNodes option to not create self-relationships

### LINKED ISSUE #1765: apoc.refactor.mergeNodes option to not create self-relationships

vga91:
Fixes #1765

  - added `config.isSelfRel()` in `RefactorUtil.java`
  - added `config.putIfAbsent("selfRel", false);` in `Nodes.java` to reuse `selfRel` configuration

### DIFF
--- a/core/src/main/java/apoc/nodes/Nodes.java
+++ b/core/src/main/java/apoc/nodes/Nodes.java
@@ -189,6 +189,7 @@ public class Nodes {
         if (nodes == null || nodes.isEmpty()) return Stream.empty();
         if (nodes.size() == 1) return Stream.of(new VirtualPathResult(nodes.get(0), null, null));
         Set<Node> nodeSet = new LinkedHashSet<>(nodes);
+        config.putIfAbsent("selfRel", false);
         RefactorConfig conf = new RefactorConfig(config);
         VirtualNode first = createVirtualNode(nodeSet, conf);
         if (first.getRelationships().iterator().hasNext()) {
--- a/core/src/main/java/apoc/nodes/Nodes.java
+++ b/core/src/main/java/apoc/nodes/Nodes.java
@@ -189,7 +189,6 @@ public class Nodes {
         if (nodes == null || nodes.isEmpty()) return Stream.empty();
         if (nodes.size() == 1) return Stream.of(new VirtualPathResult(nodes.get(0), null, null));
         Set<Node> nodeSet = new LinkedHashSet<>(nodes);
-        config.putIfAbsent("selfRel", false);
         RefactorConfig conf = new RefactorConfig(config);
         VirtualNode first = createVirtualNode(nodeSet, conf);
         if (first.getRelationships().iterator().hasNext()) {
--- a/core/src/test/java/apoc/refactor/GraphRefactoringTest.java
+++ b/core/src/test/java/apoc/refactor/GraphRefactoringTest.java
@@ -1008,8 +1008,13 @@ MATCH (a:A {prop1:1}) MATCH (b:B {prop2:99}) CALL apoc.refactor.mergeNodes([a, b
                 (r) -> {
                     Node node = (Node) r.get("node");
                     Iterator<Relationship> relIterator = node.getRelationships().iterator();
-                    assertSelfRel(relIterator.next());
-                    assertSelfRel(relIterator.next());
+                    final String expectedRelType = "TEST_REL";
+                    final Relationship firstRel = relIterator.next();
+                    assertSelfRel(firstRel, expectedRelType);
+                    assertEquals(Map.of("prop", "two"), firstRel.getAllProperties());
+                    final Relationship secondRel = relIterator.next();
+                    assertSelfRel(secondRel, expectedRelType);
+                    assertEquals(Map.of("prop", "one"), secondRel.getAllProperties());
                     assertFalse(relIterator.hasNext());
                 });
     }
@@ -1030,16 +1035,21 @@ MATCH (a:A {prop1:1}) MATCH (b:B {prop2:99}) CALL apoc.refactor.mergeNodes([a, b
                 (r) -> {
                     Node node = (Node) r.get("node");
                     Iterator<Relationship> relIterator = node.getRelationships().iterator();
-                    assertSelfRel(relIterator.next());
-                    assertSelfRel(relIterator.next());
+                    final String expectedRelType = "TEST_REL";
+                    final Relationship firstRel = relIterator.next();
+                    assertSelfRel(firstRel, expectedRelType);
+                    assertTrue(firstRel.getAllProperties().isEmpty());
+                    final Relationship secondRel = relIterator.next();
+                    assertSelfRel(secondRel, expectedRelType);
+                    assertTrue(secondRel.getAllProperties().isEmpty());
                     assertFalse(relIterator.hasNext());
                 });
     }
 
     @Test
     public void testMergeNodeShouldNotCancelOtherRelsWithSelfRelsTrue() {
-        db.executeTransactionally("CREATE (a:A {a:'a'})-[:KNOWS]->(b:B {a:'b'})-[:KNOWS]->(c:C {a:'c'})\n" +
-                "WITH a,b,c CREATE (a)-[:KNOWS]->(a) WITH a,b,c CREATE (a)-[:KNOWS]->(c) WITH c,b CREATE (c)-[:KNOWS]->(b);");
+        db.executeTransactionally("CREATE (a:A {a:'a'})-[:KNOWS {foo: 'bar'}]->(b:B {a:'b'})-[:KNOWS {baz: 'baa'}]->(c:C {a:'c'})\n" +
+                "WITH a,b,c CREATE (a)-[:KNOWS {self: 'rel'}]->(a) WITH a,b,c CREATE (a)-[:KNOWS {one: 'two'}]->(c) WITH c,b CREATE (c)-[:KNOWS {three: 'four'}]->(b);");
         testCall(db, "MATCH (n:A), (m:B) WITH [n,m] as nodes CALL apoc.refactor.mergeNodes(nodes, {mergeRels: true, produceSelfRel: true}) yield node return node",
                 (r) -> {
                     Node node = (Node) r.get("node");
@@ -1054,21 +1064,26 @@ MATCH (a:A {prop1:1}) MATCH (b:B {prop2:99}) CALL apoc.refactor.mergeNodes([a, b
                     assertEquals("A", firstSelfRel.getStartNode().getLabels().iterator().next().name());
                     assertEquals("A", firstSelfRel.getEndNode().getLabels().iterator().next().name());
                     assertEquals("KNOWS", firstSelfRel.getType().name());
+                    assertEquals(Map.of("foo", "bar"), firstSelfRel.getAllProperties());
 
                     Relationship secondSelfRel = relationships.get(1);
                     assertEquals("A", secondSelfRel.getStartNode().getLabels().iterator().next().name());
                     assertEquals("A", secondSelfRel.getEndNode().getLabels().iterator().next().name());
                     assertEquals("KNOWS", secondSelfRel.getType().name());
+                    assertEquals(Map.of("self", "rel"), secondSelfRel.getAllProperties());
 
+                    // two A-C rels created with merge (with combined properties)
                     Relationship firstNotSelfRel = relationships.get(2);
                     assertEquals("A", firstNotSelfRel.getStartNode().getLabels().iterator().next().name());
                     assertEquals("C", firstNotSelfRel.getEndNode().getLabels().iterator().next().name());
                     assertEquals("KNOWS", firstNotSelfRel.getType().name());
+                    assertEquals(Map.of("one", "two", "baz", "baa"), firstNotSelfRel.getAllProperties());
 
                     Relationship secondNotSelfRel = relationships.get(3);
                     assertEquals("C", secondNotSelfRel.getStartNode().getLabels().iterator().next().name());
                     assertEquals("A", secondNotSelfRel.getEndNode().getLabels().iterator().next().name());
                     assertEquals("KNOWS", secondNotSelfRel.getType().name());
+                    assertEquals(Map.of("three", "four"), secondNotSelfRel.getAllProperties());
                 });
     }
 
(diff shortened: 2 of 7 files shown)


core/src/main/java/apoc/nodes/Nodes.java
```diff
@@ -189,6 +189,7 @@ public boolean connected(@Name("start") Node start, @Name("start") Node end, @Na
         if (nodes == null || nodes.isEmpty()) return Stream.empty();
         if (nodes.size() == 1) return Stream.of(new VirtualPathResult(nodes.get(0), null, null));
         Set<Node> nodeSet = new LinkedHashSet<>(nodes);
+        config.putIfAbsent("selfRel", false);
```
fbiville, line 192: could you elaborate why we need a different default here (compared to the global one in `RefactorConfig`)?

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +998,148 @@ public void testRefactorCategorizeNoDups() {
         assertEquals(8, relsCount);
         db.executeTransactionally("DROP CONSTRAINT ON (n:`" + label + "`) ASSERT n.`" + targetKey + "` IS UNIQUE");
     }
+
+    @Test
+    public void testMergeNodeShouldNotCreateSelfRelationshipsInPreExistingSelfRel() {
+        db.executeTransactionally("CREATE (a:TestNode {a:'a'})-[:TEST_REL]->(b:TestNode {a:'b'})-[:TEST_REL]->(c:TestNode {a:'c'})\n" +
+                "WITH a, c CREATE (a)-[:TEST_REL]->(a) WITH c CREATE (c)-[:TEST_REL]->(c);");
+        testCall(db, "MATCH (n:TestNode) WITH collect(n) as nodes CALL apoc.refactor.mergeNodes(nodes, {mergeRels: true, selfRel: false}) yield node return node",
```
fbiville, line 1006: There should probably be a distinction between:

1. existing self-rels
2. creating new self-rels when you merge

The default current named `selfRel` should probably be renamed to `produceSelfRel` or something similar and account only for the self-rels described in point 2.
Self-rels in point 1 should by default be preserved. Maybe another config flag like `preserveExistingSelfRels` could be added (with default value `true`)?

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +998,148 @@ public void testRefactorCategorizeNoDups() {
...
+    public void testMergeNodeShouldCreateSelfRelationshipsInPreExistingSelfRel() {
+        db.executeTransactionally("CREATE (a:TestNode {a:'a'})-[:TEST_REL]->(b:TestNode {a:'b'})-[:TEST_REL]->(c:TestNode {a:'c'})\n" +
+                "WITH a, c CREATE (a)-[:TEST_REL]->(a) WITH c CREATE (c)-[:TEST_REL]->(c);");
+        testCall(db, "MATCH (n:TestNode) WITH collect(n) as nodes CALL apoc.refactor.mergeNodes(nodes, {mergeRels: true, selfRel: true}) yield node return node",
+                (r) -> {
+                    Node node = (Node) r.get("node");
+                    Iterator<Relationship> relIterator = node.getRelationships().iterator();
+                    String relName = relIterator.next().getType().name();
+                    assertEquals("TEST_REL", relName);
+                    assertFalse(relIterator.hasNext());
```
fbiville, line 1023: variation of the comment above: shouldn't there be two relationships by default? There would be the existing self-rel from `a` to `a`, and a second one resulting from the relationship merge?

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +998,148 @@ public void testRefactorCategorizeNoDups() {
...
+    @Test
+    public void testMergeNodeShouldNotCancelOtherRelsWithSelfRelsTrue() {
+        db.executeTransactionally("CREATE (a:A {a:'a'})-[:KNOWS]->(b:B {a:'b'})-[:KNOWS]->(c:C {a:'c'})\n" +
+                "WITH a,b,c CREATE (a)-[:KNOWS]->(a) WITH a,b,c CREATE (a)-[:KNOWS]->(c) WITH c,b CREATE (c)-[:KNOWS]->(b);");
+        testCall(db, "MATCH (n:A), (m:B) WITH [n,m] as nodes CALL apoc.refactor.mergeNodes(nodes, {mergeRels: true, selfRel: true}) yield node return node",
+                (r) -> {
+                    Node node = (Node) r.get("node");
+                    List<Relationship> relationships = IteratorUtils.toList(node.getRelationships().iterator());
+
+                    assertEquals(3, relationships.size());
```
fbiville, line 1036: same remark as above: shouldn't there be 4 relationship (as A to A should be preserved by default)?

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +998,148 @@ public void testRefactorCategorizeNoDups() {
...
+                    relationships.sort(Comparator.comparingLong(Relationship::getStartNodeId)
+                            .thenComparingLong(Relationship::getEndNodeId));
+
+                    Relationship firstRel = relationships.get(0);
+                    assertEquals("A", firstRel.getStartNode().getLabels().iterator().next().name());
+                    assertEquals("A", firstRel.getEndNode().getLabels().iterator().next().name());
+                    assertEquals("KNOWS", firstRel.getType().name());
```
fbiville, line 1073: why is this relationship preserved?

fbiville (changes requested):

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
vga91: @fbiville Good idea, I considered only the new `selfRel`,
leaving unchanged the previous behavior, that is to cancel the existing relationships. 
Now I changed `selfRel` to `produceSelfRel` and added the second one property.

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
vga91: Changed as above

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
vga91: Because in this case it merged only one node, this test is to demonstrate that the rels (besides nodes) will not change with 1 node.

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +999,170 @@ public void testRefactorCategorizeNoDups() {
         assertEquals(8, relsCount);
         db.executeTransactionally("DROP CONSTRAINT ON (n:`" + label + "`) ASSERT n.`" + targetKey + "` IS UNIQUE");
     }
+
+    @Test
+    public void testMergeNodeShouldNotCreateSelfRelationshipsInPreExistingSelfRel() {
+        db.executeTransactionally("CREATE (a:TestNode {a:'a'})-[:TEST_REL]->(b:TestNode {a:'b'})-[:TEST_REL]->(c:TestNode {a:'c'})\n" +
+                "WITH a, c CREATE (a)-[:TEST_REL {prop: 'one'}]->(a), (a)-[:TEST_REL {prop: 'two'}]->(a) WITH c CREATE (c)-[:TEST_REL]->(c);");
+        testCall(db, "MATCH (n:TestNode) WITH collect(n) as nodes CALL apoc.refactor.mergeNodes(nodes, {mergeRels: true, produceSelfRel: false}) yield node return node",
```
fbiville, line 1007: Shouldn't there be an explicit order clause in this `MATCH` query, given `mergeNodes` picks the first node of the list?

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +999,170 @@ public void testRefactorCategorizeNoDups() {
...
+                (r) -> {
+                    Node node = (Node) r.get("node");
+                    Iterator<Relationship> relIterator = node.getRelationships().iterator();
+                    assertSelfRel(relIterator.next());
```
fbiville, line 1011: might be worth checking the self-relationship type and property here (and below)

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +999,170 @@ public void testRefactorCategorizeNoDups() {
...
+    @Test
+    public void testMergeNodeShouldCreateSelfRelationshipsInPreExistingSelfRel() {
+        db.executeTransactionally("CREATE (a:TestNode {a:'a'})-[:TEST_REL]->(b:TestNode {a:'b'})-[:TEST_REL]->(c:TestNode {a:'c'})\n" +
+                "WITH a, c CREATE (a)-[:TEST_REL]->(a) WITH c CREATE (c)-[:TEST_REL]->(c);");
+        testCall(db, "MATCH (n:TestNode) WITH collect(n) as nodes CALL apoc.refactor.mergeNodes(nodes, {mergeRels: true, produceSelfRel: true}) yield node return node",
+                (r) -> {
+                    Node node = (Node) r.get("node");
+                    Iterator<Relationship> relIterator = node.getRelationships().iterator();
+                    assertSelfRel(relIterator.next());
+                    assertSelfRel(relIterator.next());
```
fbiville, line 1034: it would be good to assert what the properties are (it will be a good illustration of how merge works)

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +999,170 @@ public void testRefactorCategorizeNoDups() {
...
+
+                    assertEquals(4, relationships.size());
+                    relationships.sort(Comparator.comparingLong(Relationship::getStartNodeId)
+                            .thenComparingLong(Relationship::getEndNodeId));
+
+                    // two A-A rels: the existing one and the new one after merge
+                    Relationship firstSelfRel = relationships.get(0);
+                    assertEquals("A", firstSelfRel.getStartNode().getLabels().iterator().next().name());
+                    assertEquals("A", firstSelfRel.getEndNode().getLabels().iterator().next().name());
+                    assertEquals("KNOWS", firstSelfRel.getType().name());
```
fbiville, line 1056: asserting the properties would help differentiate the existing self-rel from the new one

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +999,170 @@ public void testRefactorCategorizeNoDups() {
...
+                            map(i-> i.getType().name()).collect(Collectors.toList());
+                    assertThat(relNodeList, Matchers.containsInAnyOrder("ASD", "ZXC"));
+                });
+    }
+
+    @Test
+    public void testMergeNodeShouldCreateSelfRelationshipsWithPathWithOtherRels() {
+        db.executeTransactionally("CREATE (a:One)-[:TEST_REL1]->(b:Two)-[:TEST_REL2]->(c:Three)\n" +
+                "WITH b, c CREATE (b)-[:ASD]->(q:Four), (b)-[:ZXC]->(w:Five) WITH b, c CREATE (b)-[:QWE]->(c)");
+        testCall(db, "match (a:One),(b:Two),(c:Three), ()-[rel:ASD]->() with [a,b,c] as nodes, rel CALL apoc.refactor.mergeNodes(nodes, {mergeRels: true}) yield node return node, rel",
```
fbiville, line 1146:
```suggestion
        testCall(db, "match (a:One),(b:Two),(c:Three) with [a,b,c] as nodes CALL apoc.refactor.mergeNodes(nodes, {mergeRels: true}) yield node return node",
```
nit: `rel` does not seem needed here

core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -994,5 +999,170 @@ public void testRefactorCategorizeNoDups() {
...
+                (r) -> {
+                    Node node = (Node) r.get("node");
+                    List<String> relNodeList = IteratorUtils.toList(node.getRelationships().iterator()).stream().
+                            map(i-> i.getType().name()).collect(Collectors.toList());
+                    assertThat(relNodeList, Matchers.containsInAnyOrder("ASD", "QWE", "ZXC", "TEST_REL1", "TEST_REL2"));
+
+                    final Relationship relTestRel1 = node.getRelationships(RelationshipType.withName("TEST_REL1")).iterator().next();
+                    final Relationship relTestRel2 = node.getRelationships(RelationshipType.withName("TEST_REL2")).iterator().next();
+                    final Relationship relQwe = node.getRelationships(RelationshipType.withName("QWE")).iterator().next();
```
fbiville, line 1155: nit: you can use `assertSelfRel` on the above three rels, right?

fbiville: if you remove the rel type assertion first from `assertSelfRel`, that is 😅

docs/asciidoc/modules/ROOT/pages/graph-updates/graph-refactoring/merge-nodes.adoc
```diff
@@ -41,6 +41,8 @@ In addition, mergeNodes supports the following config properties:
 |===
 | type | operations
 | mergeRels | true/false: give the possibility to merge relationships with same type and direction.
+| produceSelfRel | true/false: is valid only with `mergeRels:true`. If `true` (default), the self-relationships will be created into new node, if exist, and vice versa.
```
fbiville, line 44: I think `produceSelfRel` can be used regardless of `mergeRels` value.
Let's say we have this graph:
```
CREATE (a:A), (b:B)
CREATE (a)-[:T]->(b)
CREATE (a)-[:T]->(b)
```
We call mergeNodes on `A` and `B`, with:

1. `mergeRels: false` and `produceSelfRel: false`, we would get 0 `T` relationship
2. `mergeRels: false` and `produceSelfRel: true`, we would get 2 `T` relationships going from `A` to itself
3. `mergeRels: true` and `produceSelfRel: false`, we would get 0 `T` relationship (?)
4. `mergeRels: true` and `produceSelfRel: true`, we would get 1 `T` relationship

So, it seems `produceSelfRel` has it uses regardless of `mergeRels` value.

Side note: @JMHReif and I discussed point 3 and that particular combination is not easy to reason about. We should clearly document what happens in that case.

fbiville (approved):

PR #1849, opened 2021-04-07
Use newly introduced DatabaseReadOnlyChecker to check if database is read only

MishaDemianenko:
Database read-only property is now dynamic and controlled by set of dynamic settings.
This PR adapts `apoc.monitor.kernel` procedure to use a new API instead of the old one that is gone now.


conker84:
@MishaDemianenko thank you so much!

PR #1852, opened 2021-04-09
Corrected TTL procedure signatures in examples

stuartlaurie:

- Updated examples to use updated procedure names
- Updated expireIn example to time to match description
- Updated expireIn example to use the DIRECTED relationship as if both examples are followed then the second would not work as PRODUCED had already been removed by TTL process


conker84 (approved):
Thank you @stuartlaurie

PR #1859, opened 2021-04-19
fixes #1564: Add Apache arrow read / write support

### LINKED ISSUE #1564: Add Apache arrow read / write support
To better interop with other frameworks Like spark, kafka, etc. Align format with Graphistry 

Consider also streaming the raw format back. 

https://www.kdnuggets.com/2018/01/supercharging-visualization-apache-arrow.html

conker84:
Fixes #1564

This procedure adds the support for Apache Arrow export/load

  - Added three export procedures that streams a list of byte[] one per each batch: `apoc.export.arrow.stream.all`, `apoc.export.arrow.stream.graph`, `apoc.export.arrow.stream.query`
  - Added three procedures export an Arrow file: `apoc.export.arrow.all`, `apoc.export.arrow.graph`, `apoc.export.arrow.query`
  - Added one load procedure `apoc.load.arrow.stream` that reads an Arrow byte[] and returns a map for each row
  - Added one load procedure `apoc.load.arrow` that reads an Arrow file and returns a map for each row

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -135,10 +135,10 @@ dependencies {
 
     compile group: 'xerces', name: 'xercesImpl', version: '2.12.1'
 
-    compile group: 'org.apache.arrow', name: 'arrow-vector', version: '4.0.0'
-    compile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '4.0.0'
-    testCompile group: 'org.apache.arrow', name: 'arrow-vector', version: '4.0.0'
-    testCompile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '4.0.0'
+    compile group: 'org.apache.arrow', name: 'arrow-vector', version: '6.0.0'
+    compile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '6.0.0'
+    testCompile group: 'org.apache.arrow', name: 'arrow-vector', version: '6.0.0'
+    testCompile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '6.0.0'
 
     configurations.all {
         exclude group: 'org.slf4j', module: 'slf4j-nop'
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -135,6 +135,11 @@ dependencies {
 
     compile group: 'xerces', name: 'xercesImpl', version: '2.12.1'
 
+    compileOnly group: 'org.apache.arrow', name: 'arrow-vector', version: '3.0.0'
+    compileOnly group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '3.0.0'
+    testCompile group: 'org.apache.arrow', name: 'arrow-vector', version: '3.0.0'
+    testCompile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '3.0.0'
+
     configurations.all {
         exclude group: 'org.slf4j', module: 'slf4j-nop'
     }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -93,8 +93,8 @@ dependencies {
     compileOnly group: 'org.neo4j', name: 'neo4j', version: neo4jVersionEffective
     compileOnly group: 'org.codehaus.jackson', name: 'jackson-mapper-asl', version: '1.9.7'
 
-    testCompile 'org.mock-server:mockserver-netty:3.10.8'
-    testCompile 'org.mock-server:mockserver-client-java:3.10.8'
+    testCompile 'org.mock-server:mockserver-netty:5.6.0'
+    testCompile 'org.mock-server:mockserver-client-java:5.6.0'
 
     compileOnly group: 'com.amazonaws', name: 'aws-java-sdk-comprehend', version: '1.11.683'
     testImplementation group: 'com.amazonaws', name: 'aws-java-sdk-comprehend', version: '1.11.683'
@@ -135,10 +135,10 @@ dependencies {
 
     compile group: 'xerces', name: 'xercesImpl', version: '2.12.1'
 
-    compile group: 'org.apache.arrow', name: 'arrow-vector', version: '3.0.0'
-    compile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '3.0.0'
-    testCompile group: 'org.apache.arrow', name: 'arrow-vector', version: '3.0.0'
-    testCompile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '3.0.0'
+    compile group: 'org.apache.arrow', name: 'arrow-vector', version: '4.0.0'
+    compile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '4.0.0'
+    testCompile group: 'org.apache.arrow', name: 'arrow-vector', version: '4.0.0'
+    testCompile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '4.0.0'
 
     configurations.all {
         exclude group: 'org.slf4j', module: 'slf4j-nop'
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -135,8 +135,8 @@ dependencies {
 
     compile group: 'xerces', name: 'xercesImpl', version: '2.12.1'
 
-    compileOnly group: 'org.apache.arrow', name: 'arrow-vector', version: '3.0.0'
-    compileOnly group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '3.0.0'
+    compile group: 'org.apache.arrow', name: 'arrow-vector', version: '3.0.0'
+    compile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '3.0.0'
     testCompile group: 'org.apache.arrow', name: 'arrow-vector', version: '3.0.0'
     testCompile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '3.0.0'
 
--- a/core/src/main/java/apoc/convert/Json.java
+++ b/core/src/main/java/apoc/convert/Json.java
@@ -25,7 +25,7 @@ public class Json {
     public static String NODE = "node";
     public static String RELATIONSHIP = "relationship";
 
-    private Object writeJsonResult(Object value) {
+    public static Object writeJsonResult(Object value) {
         Meta.Types type = Meta.Types.of(value);
         switch (type) {
             case NODE:
@@ -37,7 +37,7 @@ public class Json {
                         .map(i-> i instanceof Node ? nodeToMap((Node) i) : relToMap((Relationship) i))
                         .collect(Collectors.toList()));
             case LIST:
-                return Convert.convertToList(value).stream().map(this::writeJsonResult).collect(Collectors.toList());
+                return Convert.convertToList(value).stream().map(Json::writeJsonResult).collect(Collectors.toList());
             case MAP:
                 return ((Map<String, Object>) value).entrySet()
                         .stream()
@@ -49,7 +49,7 @@ public class Json {
         }
     }
 
-    private Map<String,Object> relToMap(Relationship rel) {
+    private static Map<String,Object> relToMap(Relationship rel) {
         Map<String, Object> mapRel = map(
                 "id", String.valueOf(rel.getId()),
                 "type", RELATIONSHIP,
@@ -60,7 +60,7 @@ public class Json {
         return mapWithOptionalProps(mapRel, rel.getAllProperties());
     }
 
-    private Map<String,Object> nodeToMap(Node node) {
+    private static Map<String,Object> nodeToMap(Node node) {
         Map<String, Object> mapNode = map("id", String.valueOf(node.getId()));
 
         mapNode.put("type", NODE);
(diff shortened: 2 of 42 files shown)


lmeyerov:
Super cool! 

How would this interact with a use case like

* Ingest:

Python client: `nodes.arrow`, `edges.arrow` ---[arrow buffer over bolt on :7687]---> `neo4j`

* Export:

Python client: `MATCH (a)-[b]-(c) RETURN a,b,c LIMIT 1000000`  --[cypher over bolt on 7687]--> `neo4j` --[arrow buffer over bolt]--> python client


* Batch vs streaming


Long time coming!!!

conker84:
> Super cool!
> 
> How would this interact with a use case like
> 
> * Ingest:
> 
> Python client: `nodes.arrow`, `edges.arrow` ---[arrow buffer over bolt on :7687]---> `neo4j`

```python
from neo4j import GraphDatabase

uri = "neo4j://localhost:7687"
driver = GraphDatabase.driver(uri, auth=("neo4j", "password"))

def read_arrow_file(tx, file):
    tx.run("CALL apoc.load.arrow($file) YIELD value "
           "RETURN value", # here you can chain your cypher query because values is a map that contains the columns of the arrow file
           file=file)

def read_arrow_byteArray(tx, byteArray):
    tx.run("CALL apoc.load.arrow.stream($byteArray) YIELD value "
           "RETURN value", # here you can chain your cypher query because values is a map that contains the columns of the arrow byte array
           byteArray=byteArray)


with driver.session() as session:
    session.write_transaction(read_arrow_file, "nodes.arrow")
with driver.session() as session:
    session.write_transaction(read_arrow_file, "edges.arrow")

with driver.session() as session:
    byteArray = # your arrow byte array
    session.write_transaction(read_arrow_byteArray, byteArray)

driver.close()
```

> 
> * Export:
> 
> Python client: `MATCH (a)-[b]-(c) RETURN a,b,c LIMIT 1000000` --[cypher over bolt on 7687]--> `neo4j` --[arrow buffer over bolt]--> python client

In a very similar way to above you can export the query in these two ways:

As the list of byte arrays with a defined batch size:

```python
from neo4j import GraphDatabase

uri = "neo4j://localhost:7687"
driver = GraphDatabase.driver(uri, auth=("neo4j", "password"))

def export_arrow_stream(tx, query):
    byteArrays = []
    result = tx.run("CALL apoc.export.arrow.stream.query($query) YIELD value RETURN value", query=query):
    for record in result:
        byteArrays.append(record["value"]) # you can do whatever you want with the arrow binary :)
    return byteArrays 

def export_arrow_file(tx, fileName, query):
    byteArrays = []
    result = tx.run("CALL apoc.export.arrow.query($fileName, $query) YIELD value RETURN value", query=query, fileName=fileName):
    for record in result:
        byteArrays.append(record["value"]) # you can do whatever you want with the arrow binary :)
    return byteArrays 

with driver.session() as session:
    session.read_transaction(export_arrow_file, "myfile.arrow", "MATCH (a)-[b]-(c) RETURN a,b,c LIMIT 1000000")
    # you can do whatever you want with the arrow file :)

with driver.session() as session:
    arrowByteArrays = session.read_transaction(export_arrow_stream, "MATCH (a)-[b]-(c) RETURN a,b,c LIMIT 1000000")
    # you can do whatever you want with the arrow binary list :)

driver.close()
```

> 
> * Batch vs streaming

What do you mean?

> 
> Long time coming!!!

lmeyerov:
That helps! This seems quite close to something can seriously speed up our graph ml workflows!

I'm guessing `read+write_arrow_byteArray` address the network case as `read_arrow_file` would only do it when the client + neo4j are on the same server. It will probably be better for CPU<>GPU too as we can skip round-tripping over the disk.

RE:Streaming, consider the case of import/export of 100 GB of data, or a live stream. `load` seems fine -- we can send multiple smaller tables to avoid overwhelming neo4j. For `export`, does it cut up the rows into recordbatches, or outputs 1 row at a time? In client/server stuff we'll do 1-10KB batches, and in CPU<>GPU, more like 50MB-2GB per batch. (Maybe make that an optional param?)


**Ingest**

How would we get a row out of the dict of columns? (Or is it a dict of 1 row already?)

```python
def save_arrow_byteArray(tx, byteArray):
    tx.run("""
          CALL apoc.load.arrow.stream($byteArray) YIELD dictOfByteColumns
        ???? as row
         MERGE (c:Company {companyId: row.Id, companyName: row.Name});
         RETURN 1
         """",
           byteArray=byteArray)
```

**Export**

Is `value` the record batches in bytes form, or something else?

```python
import pyarrow as pa

def cypher_to_arrow_table(tx, query):

    buf = pa.BufferReader()
    result = tx.run("CALL apoc.export.arrow.stream.query($query) YIELD value RETURN value", query=query):
    for recordBatch in result:
        buf.append(record["value"])

    reader = pa.ipc.open_stream(buf)
    batches = []
    for b in reader:
      batches.append(b)
    return pa.Table.from_batches(batches, schema=reader.schema)

full_property_graph_arr = cypher_to_arrow_table(tx, "MATCH (a)-[b]-(c) RETURN a, b, c")
```

conker84:
@lmeyerov:

```

> That helps! This seems quite close to something can seriously speed up our graph ml workflows!
> 
> I'm guessing `read+write_arrow_byteArray` address the network case as `read_arrow_file` would only do it when the client + neo4j are on the same server. It will probably be better for CPU<>GPU too as we can skip round-tripping over the disk.

Why? Maybe I misunderstood but I saw that Spark (when uses Arrow under-the-hood) has the same approach to compute the partition, and they are (in most cases) not in the same server.

> 
> RE:Streaming, consider the case of import/export of 100 GB of data, or a live stream. `load` seems fine -- we can send multiple smaller tables to avoid overwhelming neo4j. For `export`, does it cut up the rows into recordbatches, or outputs 1 row at a time? In client/server stuff we'll do 1-10KB batches, and in CPU<>GPU, more like 50MB-2GB per batch. (Maybe make that an optional param?)

There is already a configuration param `batchSize` that works in two way:

- for the streaming procedures it defines the size of the arrow byte[]
- for the file procedures it defines the size of the record before each `writeBatch` Arrow call

> 
> **Ingest**
> 
> How would we get a row out of the dict of columns? (Or is it a dict of 1 row already?)

Please look at my example

```python
  def save_arrow_byteArray(tx, byteArray):
     tx.run("""
        CALL apoc.load.arrow.stream($byteArray) YIELD value
        MERGE (c:Company {companyId: value.Id, companyName: value.Name});
        RETURN 1
        """",
        byteArray=byteArray)
```

> **Export**
> 
> Is `value` the record batches in bytes form, or something else?

It's a byte[] :)

> 
> ```python
> import pyarrow as pa
> 
> def cypher_to_arrow_table(tx, query):
> 
>     buf = pa.BufferReader()
>     result = tx.run("CALL apoc.export.arrow.stream.query($query) YIELD value RETURN value", query=query):
>     for recordBatch in result:
>         buf.append(record["value"])
> 
>     reader = pa.ipc.open_stream(buf)
>     batches = []
>     for b in reader:
>       batches.append(b)
>     return pa.Table.from_batches(batches, schema=reader.schema)
> 
> full_property_graph_arr = cypher_to_arrow_table(tx, "MATCH (a)-[b]-(c) RETURN a, b, c")
> ```

lmeyerov:
I just reread the PR, I think I understand some of my initial misunderstandings

RE:files, this is solved for us via streams, so we're good. More as a comment if helpful, I've been seeing a trend to separate flows like `storage<>client<>server` from `client<>server<>storage`. For something like writes, streams mean our client can do `client --arrow bytes->server`, while URLs mean we can do `client--arrow url-->server<--arrow bytes---parallel s3 loaders`.  Same thing but different directions for reads.

RE:reads via `cypher_to_arrow_table`, the docs seem to help here. Will file a ticket in pygraphistry to prove this out!

conker84:
> I just reread the PR, I think I understand some of my initial misunderstandings
> 
> RE:files, this is solved for us via streams, so we're good. More as a comment if helpful, I've been seeing a trend to separate flows like `storage<>client<>server` from `client<>server<>storage`. For something like writes, streams mean our client can do `client --arrow bytes->server`, while URLs mean we can do `client--arrow url-->server<--arrow bytes---parallel s3 loaders`. Same thing but different directions for reads.

Sorry but I'm a little bit confused, can you elaborate a little bit?


> 
> RE:reads via `cypher_to_arrow_table`, the docs seem to help here. Will file a ticket in pygraphistry to prove this out!

lmeyerov:
Ex: Scaling ingest

1.  `client--[bytes]-->neo4j`: Client would push record batches to neo4j. To write more, open multiple write streams, potentially from multiple clients doing different write transactions. This is nice for when the client has the data and as a building block.


2. `client--[s3://...]-->neo4j<--s3`: Client would send a location of data in a byte-addressable store or a smart store, and then neo4j would launch parallel bulk reads. This is nice when the data is elsewhere, like s3, and there is faster neo4j<>storage bandwidth  than client<>neo4j. Likewise, clients often have thin bandwidth to neo4j to beginwith, like a consumer device, so doing GB/TB files is a non-starter to beginwith.

In the pydata world, we see analogous like `pd.read_csv('s3://...')` and Arrow's Flight protocol

But first things first... if we can get read/writes of byte streams w/ pygraphistry, that's killer!

conker84:
@lmeyerov yes we can add a feature in order to support different file stores such as `S3`, `HDFS` and so on.
Looking forward to hearing from you!

conker84:
@lmeyerov hi! any progress in the test?

conker84:
@lmeyerov ping!

lmeyerov:
Hi @conker84  I'm on much-delayed PTO so spending the week with family etc . When back, you'll be able to follow on the above ticket -- I'll be curious to see what it takes to outperform normal bolt for 1-100M rows!

conker84:
@lmeyerov Hi Leo any progress on this?

lmeyerov:
Still playing catchup at work

You can track here: https://github.com/graphistry/pygraphistry/issues/233

Our bolt bindings are simple, so this is more about making it work & learning where it outperforms:

* cypher -> pandas: https://github.com/graphistry/pygraphistry/blob/master/graphistry/bolt_util.py
* tests: 
  * python: https://github.com/graphistry/pygraphistry/blob/master/graphistry/tests/test_bolt_util.py
  * docker: https://github.com/graphistry/pygraphistry/tree/master/docker
  * test runner: https://github.com/graphistry/pygraphistry/blob/master/docker/test-cpu-local-neo4j-only.sh

JMHReif:
@lmeyerov, ping! Any updates on this one?

lmeyerov:
Sorry no, hands still tied and looks like a lot of benchmarking work to confirm this has the intended perf benefit + testing work to establish the proper query patterns, so can't ask others to jump off customer work :(

lmeyerov:
(As linked above, I don't think it's a big leap to show how to use this from a notebook, to at least confirm it gets data from point A to point B, so definitely still on my mind!)

conker84:
Hi @lmeyerov is there something that I can do to help you testing this?

lmeyerov:
@conker84  see above, we have a neo4j docker CI flow, so would be about adding the apoc to it and adding a roundtrip test of python->neo4j->python

we have a user who has been struggling with neo4j input/output performance when trying to use it for ML tasks, so will see if they want us to prioritize this!

fyi, having a discussion w/ Tinkerpop crowd on arrow support there, in case you're interested

core/build.gradle
```diff
@@ -137,6 +137,11 @@ dependencies {
     // there is a compatibility problem between the Guava Library used by HDFS, Google Cloud Storage and Reflections; the only version that fits for all is the following
     compile group: 'com.google.guava', name: 'guava', version: '20.0'
 
+    compile group: 'org.apache.arrow', name: 'arrow-vector', version: '4.0.0'
+    compile group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '4.0.0'
```
conker84, line 141: @JMHReif @fbiville I thought about this and maybe these two makes sense change to `compileOnly` as it is into core WDYT?

fbiville: `compileOnly` sounds like a good idea 👍

fbiville:
Any update on this?

voutilad:
I'll experiment with this next week, but I have a few questions/comments in the meantime after a very terse skim of the PR:

1. Arrow's now at v6.0.0 as of last month. Just a heads up. 😄 
2. The query examples all use heterogenous arrays and simple maps...is the intent to support just storable types or actual Cypher supported types? Example: this is valid in Cypher `[1, true, 2.3, null, { name: "Dave" }]` but cannot be stored in the database.
3. Given this puts an Arrow buffer payload into a packstream byte array, our drivers will all have to copy the data into Arrow buffers on the client side. Have you measured any performance impact of packstream + this memory copy? In my work with Arrow I've only created and kept buffers in native memory managed by Arrow so I'm not sure what to expect in terms of performance across our various driver implementations.

conker84:
Hi @voutilad lemme answer each point:

> I'll experiment with this next week, but I have a few questions/comments in the meantime after a very terse skim of the PR:
> 
> 1. Arrow's now at v6.0.0 as of last month. Just a heads up. 😄

Done thank you!!!

> 2. The query examples all use heterogenous arrays and simple maps...is the intent to support just storable types or actual Cypher supported types? Example: this is valid in Cypher `[1, true, 2.3, null, { name: "Dave" }]` but cannot be stored in the database.

To my knowledge, Arrow doesn't support mixed array types, so we cannot fully support all Cypher types but coerce that list in the list of strings. I added a test `testValidNonStorableQuery` please give it a look!

> 3. Given this puts an Arrow buffer payload into a packstream byte array, our drivers will all have to copy the data into Arrow buffers on the client side. Have you measured any performance impact of packstream + this memory copy? In my work with Arrow I've only created and kept buffers in native memory managed by Arrow so I'm not sure what to expect in terms of performance across our various driver implementations.

No, I didn't do any test in that way, but I think that if the bottleneck is the Driver implementation we cannot do so much, if not try to get around the driver itself. But if you have a quick way to test it and provide feedback it would be highly appreciated!

PR #1872, opened 2021-04-27
SecurityContext.allowExecuteAdminProcedure is changing in 4.4

mnd999:

SecurityContext.allowExecuteAdminProcedure is changing in 4.4 to provide more information than just true / false. 

- Update to reflect this change to the interface . 
- I haven't done so, but it would be possible to use this to provide a better error message if this is desirable.


fbiville:
@mnd999 is this PR still needed?

PR #1875, opened 2021-04-29
Deprecate the apoc round function

Lojjs:
Neo4j's built-in round() function has the same signature since 4.2

### DIFF
--- a/core/src/main/java/apoc/math/Maths.java
+++ b/core/src/main/java/apoc/math/Maths.java
@@ -12,7 +12,8 @@ import java.math.RoundingMode;
  * @since 12.12.16
  */
 public class Maths {
-    @UserFunction
+    @UserFunction(deprecatedBy = "Neo4j round() function. This function will be removed in version 5.0")
+    @Deprecated
     @Description("apoc.math.round(value,[prec],mode=[CEILING,FLOOR,UP,DOWN,HALF_EVEN,HALF_DOWN,HALF_UP,DOWN,UNNECESSARY])")
     public Double round(@Name("value") Double value,
                         @Name(value = "precision",defaultValue = "0") long precision,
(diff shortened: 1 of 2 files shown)


fbiville (approved):

fbiville:
That's good to go, right? cc @conker84 @JMHReif

docs/index.html
```diff
@@ -2501,7 +2501,8 @@ <h4 id="_math_functions"><a class="link" href="#_math_functions">Math Functions<
 <tbody>
 <tr>
 <td class="tableblock halign-left valign-top"><p class="tableblock"><code>apoc.math.round(value,[precision=0],mode=[HALF_UP,CEILING,FLOOR,UP,DOWN,HALF_EVEN,HALF_DOWN,DOWN,UNNECESSARY])</code></p></td>
-<td class="tableblock halign-left valign-top"><p class="tableblock">rounds value with optionally given precision (default 0) and optional rounding mode (default HALF_UP)</p></td>
+<td class="tableblock halign-left valign-top"><p class="tableblock">Rounds value with optionally given precision (default 0) and optional rounding mode (default HALF_UP).
```
conker84, line 2504: I think that instead of updating this (btw it should be generated, but I could be wrong) we need to update also `docs/asciidoc/modules/ROOT/pages/overview/apoc.math/apoc.math.round.adoc`

conker84 (changes requested):
Thank you so much! Can you please look at my comment?

docs/index.html
Lojjs: I thought that might be the wrong place, but the file you are linking to has a comment on top saying `This file is generated by DocsTest, so don't change it!` so I don't think that is right either

conker84: I think that is the right place because it affects the function, you can try to regenerate it by executing `DocsTest`. But maybe @JMHReif knows a better place!

Lojjs: I think I found a better, not generated place to add it now. It looks like it will show up as a note in the end of this page now: https://neo4j.com/labs/apoc/4.1/overview/apoc.math/apoc.math.round/

conker84: It's not the same page that I shared with you?

conker84: Oh you mean the usage one! I see!

Lojjs: Yes that one will be included in the page you shared, could see that when I ran the DocsTest

conker84 (approved):

PR #1878, opened 2021-05-03
Start deprecation for custom.asProcedure, custom.asFunction

### LINKED ISSUE #1877: Deprecate old style custom function/procedure replaced by newer ones
Mark the following procedures/functions as deprecated to be removed in a future release:
* apoc.custom.asFunction
* apoc.custom.asProcedure

They are replaced by the following (respectively):
* apoc.custom.declareFunction
* apoc.custom.declareProcedure

Also update any documentation, as needed.

JMHReif:
Fixes #1877

Begin deprecation process for old procedure/function that has since been replaced with improved syntax.


conker84 (commented):
Maybe makes sense to add a note into the documentation as well?

fbiville:
> Maybe makes sense to add a note into the documentation as well?

I think the documentation generator automatically includes a deprecation notice, or is that not the case @conker84?

conker84:
@fbiville @JMHReif honestly IDK, maybe makes sens run `DocsTest` and see what happens

PR #1886, opened 2021-05-05
Update APOC functions to work with new types of indexes

zimmre:
This PR makes existing APOC functions to not fail when new types of indexes are present in the database.


conker84 (approved):
LGTM

PR #1897, opened 2021-05-10
Fixes #1838:  apoc.map.unflatten as the reverse to apoc.map.flatten

### LINKED ISSUE #1838: apoc.map.unflatten as the reverse to apoc.map.flatten
apoc.mapp.flatten is great for loading nested json resources into neo4j. For use cases like using FHIR for healthcare data it provides a way to load data without heavy ingest mapping burden. However, if the graph is the storage point for this data it needs to be unflattened and returned to its json state for many usecases.

## Considered alternatives
I have not seen any feature that could stand in as an alternative

For a FHIR or other industry data exchange spec usecase, this feature would allow neo4j to be the landing zone for the data no matter what new data comes in and also allow it to be retrieved for other data exchange. Making neo4j a hub for the data.

vga91:
Fixes #1838

  - Added  `apoc.map.unflatten` equivalent to the inverse of `apoc.map.flatten`
  - Added round-trip tests


core/src/test/java/apoc/map/MapsTest.java
```diff
@@ -317,6 +317,36 @@ public void testFlattenWithDelimiter() {
...
 
+    @Test
+    public void testUnflattenRoundrip() {
+        List<Map<String, Object>> innerNestedListMap = List.of(map("somekey", "someValue", "somenumeric", 123), 
+                map("keyFoo", "valueFoo"),
+                map("keyBar", "valueBar"));
+        Map<String, Object> nestedMap = map("anotherkey", "anotherValue", "nested", innerNestedListMap);
+        final Map<String, Object> expectedMap = map("string", "value", "int", 10, "nested", nestedMap);
+
+        TestUtil.testCall(db, "WITH apoc.map.flatten($expectedMap) AS flattedMap RETURN apoc.map.unflatten(flattedMap) AS value",
```
fbiville, line 328:
```suggestion
        TestUtil.testCall(db, "WITH apoc.map.flatten($expectedMap) AS flattenedMap RETURN apoc.map.unflatten(flattedMap) AS value",
```

core/src/test/java/apoc/map/MapsTest.java
```diff
@@ -317,6 +317,36 @@ public void testFlattenWithDelimiter() {
...
+                map("expectedMap", expectedMap),
+                r -> assertEquals(expectedMap, r.get("value")));
+    }
+
+    @Test
+    public void testUnflattenRoundripWithCustomDelimeter() {
```
fbiville, line 334:
```suggestion
    public void testUnflattenRoundtripWithCustomDelimiter() {
```

fbiville: Nit: maybe do not test the round trip conversion here but start with an already-flattened map and call only unflatten with the custom delimiter?

fbiville (approved):

PR #1898, opened 2021-05-10
Fixes #1829: Bug in CALL apoc.metrics.storage('dbms.directories.data') due to APOC configuration check

### LINKED ISSUE #1544: apoc.storage.metrics not working for 4.0 due to a carry-over error
Bug:  in Neo4j 4.0, if you do CALL apoc.metrics.storage(null) you should get output and instead you get zero lines.

It works in 3.5.

the bug is caused by an error in carry-over of the code to 4.0.  In the 3.5 branch, the critical line that makes it work gets a storage configured directory right here: https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.5/src/main/java/apoc/metrics/Metrics.java#L43

In 4.0, an error was introduced in the equivalent line right here:  https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.0/src/main/java/apoc/metrics/Metrics.java#L43

I think the carry-over assumed that the config being gotten was an APOC config.  It was not.  The config values being gotten here are things like dbms.directories.data.   So when you prepend apoc to that, the result comes back empty and the proc fails.

This was one of those procedures which is not amenable to unit testing (hence why we didn't catch this) because the unit testing uses an in-memory database, and things like directories and sizes don't exist/aren't relevant in the unit testing context.   :/

### LINKED ISSUE #1829: Bug in CALL apoc.metrics.storage('dbms.directories.data') due to APOC configuration check

```CALL apoc.metrics.storage('dbms.directories.data'); ```

Should return one record with information about the underlying storage device (how much space is free/available)

No results returned

```CALL apoc.metrics.storage('dbms.directories.data'); ```

Tested on 4.1.

I think the root of the issue was an APOC configuration bit of code which was inappropriately introduced.  Here is a reference to where it exists: https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/f3a42f8b6e344a0ca1a2b7d497be7c02853b4ca9/full/src/main/java/apoc/metrics/Metrics.java#L46

The way the procedure should work is to take a dbms setting (dbms.directories.data) and get its value.  It should then path concat that value (let's say "data") to NEO4J_HOME to yield the fully qualified path where data resides.  With that, the rest is straightforward.

What *actually happens* is that the line of code referenced checks to see if `apoc.dbms.directories.data` exists as a config setting.  It doesn't and never would, so the file parameter becomes null.   The filter right here: https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/f3a42f8b6e344a0ca1a2b7d497be7c02853b4ca9/full/src/main/java/apoc/metrics/Metrics.java#L209 filters nulls out, and the user gets no results.

### PR DESCRIPTION
**Pull request by vga91**
Fixes #1829
Fixes #1544

- Removed `apoc.` prefix in `Metrics.java`
- Added missing dir settings in `ApocConfig.java`
- Changed `unsupported.dbms.directories.neo4j_home` to `dbms.directories.neo4j_home` and commented `dbms.directories.certificates` (not present in neo4j 4.x) in `FileUtils.java`
- Added test to `apoc.metrics.storage` for each `Directory configuration` and with `apoc.metrics.storage(null)`

### DIFF
--- a/core/src/main/java/apoc/util/FileUtils.java
+++ b/core/src/main/java/apoc/util/FileUtils.java
@@ -274,7 +274,7 @@ public class FileUtils {
     //
     // More likely, they'll be largely similar metrics.
     public static final List<String> NEO4J_DIRECTORY_CONFIGURATION_SETTING_NAMES = Arrays.asList(
-            "dbms.directories.certificates",
+//            "dbms.directories.certificates",  // not  in 4.x version
             "dbms.directories.data",
             "dbms.directories.import",
             "dbms.directories.lib",
@@ -283,7 +283,7 @@ public class FileUtils {
             "dbms.directories.plugins",
             "dbms.directories.run",
             "dbms.directories.tx_log",
-            "unsupported.dbms.directories.neo4j_home"
+            "dbms.directories.neo4j_home"
     );
 
     public static void closeReaderSafely(CountingReader reader) {
--- a/core/src/main/java/apoc/util/FileUtils.java
+++ b/core/src/main/java/apoc/util/FileUtils.java
@@ -274,7 +274,7 @@ public class FileUtils {
     //
     // More likely, they'll be largely similar metrics.
     public static final List<String> NEO4J_DIRECTORY_CONFIGURATION_SETTING_NAMES = Arrays.asList(
-//            "dbms.directories.certificates",  // not  in 4.x version
+//            "dbms.directories.certificates",  // not in 4.x version
             "dbms.directories.data",
             "dbms.directories.import",
             "dbms.directories.lib",
--- a/full/src/test/java/apoc/metrics/MetricsTest.java
+++ b/full/src/test/java/apoc/metrics/MetricsTest.java
@@ -6,12 +6,18 @@ import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.neo4j.driver.Session;
+import org.neo4j.internal.helpers.collection.Iterators;
 
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static apoc.util.FileUtils.NEO4J_DIRECTORY_CONFIGURATION_SETTING_NAMES;
 import static apoc.util.TestContainerUtil.createEnterpriseDB;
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by vga91 — COMMENTED** [review]
@moxious Please, could you review this, to check if it's all fine? Thanks in advance

**Review by moxious — APPROVED** [review]
@vga91 the tests you wrote are good; when this code was originally written in 3.5 this wasn't possible due to (Honestly I forget which reason).  The specific code changes look fine to me -- but they are all to do with changes made to neo4j since then, for example the unsupported configuration change and the APOC configuration change ("apoc.").

I cannot verify these are all correct just because I don't know, but the unit tests are good, and I think they can't pass without these other things being right, so I would approve this PR.

It looks to me like this isn't so much a bug, but that when this code was carried forward from the 3.5 series forward to something like 4.2, the carry-forward was incomplete.  Changes that happened in Neo4j in the meantime, combined with the unit tests not having been avialable I guess created the breakage.

File: core/src/main/java/apoc/util/FileUtils.java [hunk 1]
```diff
@@ -274,7 +274,7 @@ public static File getMetricsDirectory() {
     //
     // More likely, they'll be largely similar metrics.
     public static final List<String> NEO4J_DIRECTORY_CONFIGURATION_SETTING_NAMES = Arrays.asList(
-            "dbms.directories.certificates",
+//            "dbms.directories.certificates",  // not in 4.x version
```
fbiville, line 277: remove the line?

full/src/test/java/apoc/metrics/MetricsTest.java
```diff
@@ -62,6 +68,32 @@ public void shouldGetMetrics() {
...
+
+        NEO4J_DIRECTORY_CONFIGURATION_SETTING_NAMES.forEach(setting ->
+                testCall(session, "CALL apoc.metrics.storage($setting)", map("setting", setting), (res) -> {
+                    assertEquals(expectedSet, res.keySet());
+                    assertEquals(setting, res.get("setting"));
+                })
+        );
+
+        // all metrics with null
+        testResult(session, "CALL apoc.metrics.storage(null)", (res) -> {
```
fbiville, line 85: out of curiosity, why `CALL apoc.metrics.storage(null)` and not `CALL apoc.metrics.storage()`?

fbiville (approved):

PR #1904, opened 2021-05-12
Cleanup schema sampling procs

### LINKED ISSUE #1903: Cleanup apoc.meta.*TypeProperties()
Currently, the code for apoc.meta.nodeTypeProperties() and apoc.meta.relTypeProperties() can use some cleanup. Bonus: APOC does not compile under windows thanks to encoding errors. Will tackle both of these issues with a PR.

SaschaPeukert:
Fixes #1903

Cleanup apoc.meta.nodeTypeProperties() and apoc.meta.relTypeProperties()

## Proposed Changes

  - Make APOC compile on windows
  - Code cleanup

### DIFF
--- a/core/src/main/java/apoc/meta/Meta.java
+++ b/core/src/main/java/apoc/meta/Meta.java
@@ -66,12 +66,10 @@ import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import static apoc.util.MapUtil.map;
-import static java.util.Arrays.asList;
-import static java.util.Collections.singletonMap;
 import static org.neo4j.internal.kernel.api.TokenRead.ANY_LABEL;
 import static org.neo4j.internal.kernel.api.TokenRead.ANY_RELATIONSHIP_TYPE;
 
-public class    Meta {
+public class Meta {
 
     @Context
     public Transaction tx;
@@ -238,8 +236,6 @@ public class    Meta {
         }
     }
 
-    static final int SAMPLE = 100;
-
     @Deprecated
     @UserFunction
     @Description("apoc.meta.type(value) - type name of a value (INTEGER,FLOAT,STRING,BOOLEAN,RELATIONSHIP,NODE,PATH,NULL,UNKNOWN,MAP,LIST)")
@@ -319,7 +315,6 @@ public class    Meta {
         }
     }
 
-
     @UserFunction("apoc.meta.cypher.types")
     @Description("apoc.meta.cypher.types(node-relationship-map)  - returns a map of keys to types")
     public Map<String,Object> typesCypher(@Name("properties") Object target) {
@@ -338,7 +333,6 @@ public class    Meta {
 
         return result;
     }
-
 
     public static class MetaStats {
         public final long labelCount;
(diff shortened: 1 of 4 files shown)


core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -132,11 +127,11 @@ public static boolean testDBCallEquivalence(GraphDatabaseService db, String test
...
-        System.out.println("COMPARE TO:");
-        System.out.println(toCSV(compareTo.get()));
-
-        System.out.println("TEST SET:");
-        System.out.println(toCSV(testSet.get()));
+//        System.out.println("COMPARE TO:");
+//        System.out.println(toCSV(compareTo.get()));
+//
+//        System.out.println("TEST SET:");
+//        System.out.println(toCSV(testSet.get()));
```
Lojjs, line 134: I think out-commented code is just slightly less bad than printlines in the code... Maybe at least add a comment saying something like `uncomment this to debug` so it is clear they are there on purpose.

core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -808,198 +803,457 @@ public void testMetaGraphExtraRelsWithSample() throws Exception {
...
+                assertEquals( List.of( "Long", "String" ), rec.get( "propertyTypes" ) );
+                assertEquals( 2L, rec.get( "propertyObservations" ) );
+                assertEquals( 2L, rec.get( "totalObservations" ) );
+                assertEquals( false, rec.get( "mandatory" ) );
+            }
+        } );
+    }
+
+    @Test
+    public void testNodeTypePropertiesWithSpecialSampleSize()
```
Lojjs, line 1157: Even if the behaviour is a bit odd in this cases I think it is nice with this test that show the current behaviour!

Lojjs (approved):
Good tests and cleanup!

core/src/main/java/apoc/meta/Meta.java
```diff
@@ -498,40 +492,45 @@ private void collectStats(SubGraph subGraph, Collection<String> labelNames, Coll
...
-        try {
-            return collectTables4LabelsProfile(metaConfig).asRelStream();
-        } catch (Exception e) {
-            log.debug("meta.relTypeProperties(): Failed to return stream", e);
-            throw new RuntimeException(e);
+    @Description( "apoc.meta.relTypeProperties()" )
+    public Stream<Tables4LabelsProfile.RelTypePropertiesEntry> relTypeProperties( @Name( value = "config", defaultValue = "{}" ) Map<String,Object> config )
+    {
+        MetaConfig metaConfig = new MetaConfig( config );
+        try
```
fbiville, line 526: nit: Neo4j formatting rules do not apply to the APOC project

fbiville (approved):

core/src/main/java/apoc/meta/Meta.java
SaschaPeukert: Question: Do you even have consistent formatting rules in APOC?

fbiville: Screenshot 2021-05-28 at 17 45 46

Only 2 results are returned for this coding style, so I'd at least avoid extra spaces within parentheses and newlines before curling braces.

We just filed https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1936 to make this clearer for future contributors.

PR #1907, opened 2021-05-17
Fixes #1824: apoc.nodes.link if not exists

### LINKED ISSUE #1824: apoc.nodes.link if not exists
## Guidelines

To avoid getting duplicate relationships created between nodes when calling apoc.nodes.link, I'd like to be able to have a "if not exists" option.
The idea is to get new relationships created between the head of the linked list and the newly added nodes, without creating duplicate relationships in the linked list.

## Considered alternatives
1. a unique relationship constraint
2. delete all relationships before re-running the script.
3. use a different script that doesn't call apoc.nodes.link when new nodes are added in the db

vga91:
Fixes #1824

Added boolean parameter `avoidDuplicates`, default `false`.


core/src/test/java/apoc/nodes/NodesTest.java
```diff
@@ -62,6 +64,21 @@ public void link() throws Exception {
         long len = TestUtil.singleResultFirstColumn( db,"MATCH (n:Foo {id:1})-[r:BAR*9]->() RETURN size(r) as len");
         assertEquals(9L, len);
     }
+
+    @Test
+    public void linkWithAvoidDuplicateTrue() {
+        db.executeTransactionally("CREATE (n:Foo {id:1}), (m:Foo {id:2}) WITH [n,m] as nodes CALL apoc.nodes.link(nodes,'BAR') RETURN 1");
+        TestUtil.testCall(db, "MATCH (n:Foo)-[r]->() RETURN count(r) as count", row -> assertEquals(1L, row.get("count")));
```
fbiville, line 71: nit: maybe assert the relationship type is `BAR`?

fbiville (approved):

PR #1915, opened 2021-05-24
Fixes #1749: In version 4 of the database there is property dbms.directories.neo4j_home without unsupported. prefix

### LINKED ISSUE #1749: In version 4 of the database there is property "dbms.directories.neo4j_home" without "unsupported." prefix
In FileUtils.getLogDirectory() the home dir of the neo4j installation is retrieved via the database configuration properties. It uses the 'home directory' as a fail save if the dbms.directories.logs value is empty. 

It uses the property key "unsupported.dbms.directories.neo4j_home", but in Neo4j version 4 this is replaced by "dbms.directories.neo4j_home".

something like this would work
```    
public static File getLogDirectory(Log log) {
        String neo4jHome = apocConfig().getString("unsupported.dbms.directories.neo4j_home", "");
        if (neo4jHome == null || neo4jHome.isEmpty()) {
            neo4jHome = apocConfig().getString("dbms.directories.neo4j_home", "");
        }
        String logDir = apocConfig().getString("dbms.directories.logs", "");

        File logs = logDir.isEmpty() ? new File(neo4jHome, "logs") : new File(logDir);
        if (logs.exists() && logs.canRead() && logs.isDirectory()) {
            return logs;
        }

        return null;
    }
```

vga91:
Fixes #1749

- Added `Neo4jLogStreamTest`

### DIFF
--- a/core/src/test/java/apoc/log/Neo4jLogStreamTest.java
+++ b/core/src/test/java/apoc/log/Neo4jLogStreamTest.java
@@ -1,0 +1,39 @@
+package apoc.log;
+
+import apoc.util.TestUtil;
+import org.junit.Before;
+import org.junit.Test;
+import org.neo4j.configuration.GraphDatabaseSettings;
+import org.neo4j.dbms.api.DatabaseManagementService;
+import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.internal.helpers.collection.Iterators;
+import org.neo4j.test.TestDatabaseManagementServiceBuilder;
+
+import java.nio.file.Paths;
+import java.util.stream.Collectors;
+
+import static apoc.ApocConfig.apocConfig;
+import static apoc.util.TestUtil.testResult;
+import static org.junit.Assert.assertTrue;
+
+public class Neo4jLogStreamTest {
+    
+    private GraphDatabaseService db;
+
+    @Before
+    public void setUp() throws Exception {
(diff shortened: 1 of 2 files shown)


core/src/test/java/apoc/log/Neo4jLogStreamTest.java
```diff
@@ -0,0 +1,39 @@
...
+        DatabaseManagementService dbManagementService = new TestDatabaseManagementServiceBuilder(
+                Paths.get("target").toAbsolutePath()).build();
+        apocConfig().setProperty("dbms.directories.logs", "");
+        db = dbManagementService.database(GraphDatabaseSettings.DEFAULT_DATABASE_NAME);
+        TestUtil.registerProcedure(db, Neo4jLogStream.class);
+    }
+    
+    @Test
+    public void testLogStream() {
+        testResult(db, "CALL apoc.log.stream('debug.log')", res -> {
```
fbiville, line 34: Would the test fail without the change in `FileUtils#getLogDirectory`?
Or is the intention to just improve test coverage regardless?

fbiville (approved):

core/src/test/java/apoc/log/Neo4jLogStreamTest.java
vga91: Yes, it fails without the change

PR #1918, opened 2021-05-26
Fixes #1900: apoc.convert.toJson throws NullPointerException for null values in maps

### LINKED ISSUE #1900: apoc.convert.toJson throws NullPointerException for null values in maps
Up until v4.2.0.2 (before the merge of #1795) the following query worked fine:

```cypher
RETURN apoc.convert.toJson({ foo: null })
```
But since v4.2.0.3 it throws a `NullPointerException` because of the [use of `Collectors.toMap`](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/468a208b45e313abf4c2760ba2be0d0ac76698a2/core/src/main/java/apoc/convert/Json.java#L44) which calls `Objects.requireNonNull` for every Map Value.

**EDIT**: After reading the Oracle Docs for Java11 I realize, that this might be not the same for all JDKs. The [docs only talk about the NullPointerException for `toUnmodifiableMap`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Collectors.html#toUnmodifiableMap(java.util.function.Function,java.util.function.Function)), [but not for `toMap`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Collectors.html#toUnmodifiableMap(java.util.function.Function,java.util.function.Function)). I tested this with the current openjdk-11-jdk on debian testing (version 11.0.11+4-1) which I assume is built from [OpenJDK's official source tagged with 11+11](https://github.com/openjdk/jdk/tree/jdk-11+11).

vga91:
Fixes #1900

### DIFF
--- a/core/src/main/java/apoc/convert/Json.java
+++ b/core/src/main/java/apoc/convert/Json.java
@@ -41,8 +41,9 @@ public class Json {
             case MAP:
                 return ((Map<String, Object>) value).entrySet()
                         .stream()
-                        .collect(Collectors.toMap(Map.Entry::getKey,
-                                e -> writeJsonResult(e.getValue())));
+                        .collect(HashMap::new, 
+                                (mapAccumulator, entry) -> mapAccumulator.put(entry.getKey(), writeJsonResult(entry.getValue())), 
+                                HashMap::putAll);  // workaround for https://bugs.openjdk.java.net/browse/JDK-8148463
             default:
                 return value;
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/convert/Json.java
```diff
@@ -41,8 +41,9 @@ private Object writeJsonResult(Object value) {
             case MAP:
                 return ((Map<String, Object>) value).entrySet()
                         .stream()
-                        .collect(Collectors.toMap(Map.Entry::getKey,
-                                e -> writeJsonResult(e.getValue())));
+                        .collect(HashMap::new, // workaround for https://bugs.openjdk.java.net/browse/JDK-8148463
```
fbiville, line 44: @conker84 you can fix this JDK bug in hack-commit-push with Rémi Forax if you want :P

fbiville (approved):

PR #1935, opened 2021-05-28
Update load directory filters to match only files

MishaDemianenko:
With an updated set of dependencies in the latest version, we have LoadDirectories tests that are failing. This PR updates filters for tests to pass, but not entirely sure if thats the right fix for that.


conker84 (approved):

conker84:
Thank you @MishaDemianenko

PR #1942, opened 2021-05-29
fix CI false negative

conker84:
The CI returns green even if there are failing tests

### DIFF
--- a/full/src/main/java/apoc/load/LoadDirectoryHandler.java
+++ b/full/src/main/java/apoc/load/LoadDirectoryHandler.java
@@ -72,7 +72,6 @@ public class LoadDirectoryHandler extends LifecycleAdapter {
     }
 
     public void remove(String name) {
-
         final LoadDirectoryItem loadDirectoryItem = new LoadDirectoryItem(name);
         remove(loadDirectoryItem);
     }
--- a/full/src/main/java/apoc/load/LoadDirectoryHandler.java
+++ b/full/src/main/java/apoc/load/LoadDirectoryHandler.java
@@ -8,13 +8,6 @@ import org.neo4j.graphdb.Transaction;
 import org.neo4j.kernel.lifecycle.LifecycleAdapter;
 import org.neo4j.logging.Log;
 
-import static apoc.util.FileUtils.getDirImport;
-import static apoc.util.FileUtils.isImportUsingNeo4jConfig;
-import static java.nio.file.StandardWatchEventKinds.ENTRY_CREATE;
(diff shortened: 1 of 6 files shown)


full/src/main/java/apoc/load/LoadDirectoryHandler.java
```diff
@@ -146,6 +143,10 @@ private Runnable createListener(LoadDirectoryItem item) {
                     Thread.sleep(config.getInterval());
                 }
             } catch (Exception e) {
+                if (e instanceof InterruptedException) {
+                    return;
+                }
+                System.out.println("ERROR: " + ExceptionUtils.getStackTrace(e));
```
fbiville, line 149: what about `log` instead of `System.out.println`? (it should at least be `System.err.println`)

full/src/test/java/apoc/load/LoadDirectoryTest.java
```diff
@@ -48,6 +50,7 @@
 import static org.junit.Assert.assertTrue;
 import static org.neo4j.test.assertion.Assert.assertEventually;
 
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
fbiville, line 53: is that required?

full/src/main/java/apoc/load/LoadDirectoryHandler.java
conker84: that was just my debug message because with the test database there are no logs, I have to clean up the code, and remove it for sure!

full/src/test/java/apoc/load/LoadDirectoryTest.java
conker84: I had to put them in order to find the one(s) that caused the first error

full/src/test/java/apoc/load/LoadDirectoryTest.java
```diff
@@ -227,11 +231,23 @@ public void testAddFolderListenerWithAllEventTypes() throws IOException {
...
+                Map.of("name", name), (r) -> {
+                    Map<String, Object> result = r.next();
+                    return LoadDirectoryItem.Status.RUNNING.name().equals(result.get("status"));
+                }),
+                value -> value, 30L, TimeUnit.SECONDS);
+    }
+
     @Test
     public void testAddFolderListenerWithCreateEventType() throws InterruptedException, IOException {
         apocConfig().setProperty(APOC_IMPORT_FILE_USE_NEO4J_CONFIG, true);
```
JMHReif, line 247: Wasn't this done above? So this line can be removed here?

JMHReif (approved):

PR #1973, opened 2021-06-04
Fixes #1911: apoc.load.directory path to directory not read correctly

### LINKED ISSUE #1911: apoc.load.directory path to directory not read correctly
## Guidelines

CALL apoc.load.directory('*.JSON', 'file:///C:/Users/BESLU/OneDrive - Equinor/NeoData', {recursive: false})
This should be the correct way of pointing to a directory

CALL apoc.load.directory('*.JSON', 'file:///Users/BESLU/OneDrive - Equinor/NeoData', {recursive: false})
This is how it has to be formatted to work.

apoc.load.json takes the full path to a file as file:///c:/    so why should a directory path omit the drive letter in Windows?
Try the above in APOC 4.2.0.4 in Neo4j Desktop 1.4.5

create some .json-files in a directory
```

see above


### Steps (Mandatory)

  1. See above
  1.
  1.

## Specifications (Mandatory)
Unsure what is required here
Currently used versions

### Versions

  - OS:  Win 10
  - Neo4j: 1.4.5 desktop
  - Neo4j-Apoc: 4.2.0.4

### PR DESCRIPTION
**Pull request by vga91**
Fixes #1911

- Changed `Paths.get(URI.create(urlDir).getPath());` to `Paths.get(URI.create(urlDir));` in `FileUtils.java`
to avoid `"java c:/ InvalidPathException: Illegal char"` on Windows path

- Changed `encodePath(getDirImport())` to `encodePath(pathImport.toUri().toString())` in `LoadDirectory.java`
to avoid `"java.nio.file.FileSystemNotFoundException: Provider "C" not installed"` on Windows path when an empty string is passed

### DIFF
--- a/full/src/main/java/apoc/load/LoadDirectory.java
+++ b/full/src/main/java/apoc/load/LoadDirectory.java
@@ -28,7 +28,6 @@ import static apoc.ApocConfig.apocConfig;
 import static apoc.load.LoadDirectoryHandler.getPathDependingOnUseNeo4jConfig;
 import static apoc.util.FileUtils.getDirImport;
 import static apoc.util.FileUtils.getPathFromUrlString;
-import static org.apache.commons.lang3.StringUtils.chop;
 import static org.eclipse.jetty.util.URIUtil.encodePath;
 import static org.neo4j.graphdb.QueryExecutionType.QueryType.READ_WRITE;
 import static org.neo4j.graphdb.QueryExecutionType.QueryType.WRITE;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: full/src/main/java/apoc/load/LoadDirectory.java [hunk 1]
```diff
@@ -109,9 +112,11 @@
 
     // visible for test purpose
     public static String checkIfUrlBlankAndGetFileUrl(String urlDir) throws IOException {
-        return StringUtils.isBlank(urlDir)
-                ? encodePath(getDirImport())
-                : FileUtils.changeFileUrlIfImportDirectoryConstrained(urlDir.replace("?", "%3F"));
+        if (StringUtils.isBlank(urlDir)) {
+            final Path pathImport = Paths.get(getDirImport()).toAbsolutePath();
+            return chop(encodePath(pathImport.toUri().toString()));
```
fbiville, line 117: why use chop here?

fbiville (approved):

full/src/main/java/apoc/load/LoadDirectory.java
vga91: The `chop` is used to remove the final `/` from the path (`file:///My/Path/Windows/`).
It still works even without it, but it seems more correct to me.

JMHReif: Would there be something already in Java we could use, rather than importing a separate dependency for it?

vga91: Yeah, of course. Now I've changed it.

fbiville (approved):

PR #1996, opened 2021-06-14
Compare deployed extensions with the one available from source

fbiville:


test-utils/src/test/java/StartupTest.java
```diff
@@ -41,4 +43,8 @@ public void test() {
             fail("Should not have thrown exception when trying to start Neo4j: " + ex);
         }
     }
+
+    private List<String> sorted(List<String> signatures) {
```
fbiville, line 47: this can be removed, I think

PR #2011, opened 2021-06-16
manage gradle ignoreFailures in GH Actions and TC build

conker84:
manage gradle ignoreFailures in GH Actions and TC build

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,7 +87,7 @@ subprojects {
         }
 
         // in TC we don't have the CI env so we can manage it in this way
-        if (System.env.CI == null) {
+        if (System.env.TEAMCITY_VERSION != null) {
             ignoreFailures(true)
         }
         testLogging.showStandardStreams = true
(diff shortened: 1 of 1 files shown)


fbiville (approved):

build.gradle
```diff
@@ -86,7 +86,10 @@ subprojects {
             setFailOnNoMatchingTests(false)
         }
 
-        ignoreFailures(true)
+        // in TC we don't have the CI env so we can manage it in this way
```
fbiville, line 89: that will also affect local builds, but CI on GH actions will at least catch it, so I guess it's not too bad

fbiville (approved):

build.gradle
fbiville: maybe there is a specific envvar for TC? That would be better.

conker84: Thank you so much! About the local build I was aware of it and I agree with, the only thing is that maybe we can try to leverage instead some TC parameter but I preferred to **play it safe**

fbiville: `TEAMCITY_VERSION` seems like a good bet: https://www.jetbrains.com/help/teamcity/predefined-build-parameters.html

conker84: oki

PR #2040, opened 2021-06-28
Fixes #1510: apoc.date.parse gives error when Date is NULL or empty string

### LINKED ISSUE #1510: apoc.date.parse gives error when Date is NULL or empty string
Using `apoc.date.parse` and when date value NULL or empty string is passed it gives the following error

```
with "" as dateStr 
with apoc.date.parse(dateStr,"s","yyyy-MM-dd") as dt return dt
```

Error:

```
Neo.ClientError.Procedure.ProcedureCallFailed: Failed to invoke function `apoc.date.parse`: Caused by: java.text.ParseException: Unparseable date: ""
```

Expected behavior - I was expecting NULL or empty string from the function if the date passed is NULL or empty String. 

Currently used versions: Neo4j version 3.5.17
APOC version 3.5.0.11

##Workaround as suggested by Michael 

with "" as dateStr 
with case dateStr  WHEN ""  then "" else apoc.date.parse(dateStr,"s","yyyy-MM-dd") end as dt return dt

Returns: 

```
╒════╕
│"dt"│
╞════╡
│""  │
└────┘
```

vga91:
Fixes #1510

### DIFF
--- a/core/src/main/java/apoc/date/Date.java
+++ b/core/src/main/java/apoc/date/Date.java
@@ -2,6 +2,7 @@ package apoc.date;
 
 import apoc.util.DateFormatUtil;
 import apoc.util.Util;
+import org.apache.commons.lang3.StringUtils;
 import org.neo4j.graphdb.Label;
 import org.neo4j.graphdb.Node;
 import org.neo4j.procedure.*;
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/date/Date.java
```diff
@@ -162,7 +163,7 @@ public Long fromISO8601(final @Name("time") String time) {
 	@UserFunction
 	@Description("apoc.date.parse('2012-12-23','ms|s|m|h|d','yyyy-MM-dd') - parse date string using the specified format into the specified time unit")
 	public Long parse(@Name("time") String time, @Name(value = "unit", defaultValue = "ms") String unit, @Name(value = "format",defaultValue = DEFAULT_FORMAT) String format, final @Name(value = "timezone", defaultValue = "") String timezone) {
-		Long value = parseOrThrow(time, getFormat(format, timezone));
+		Long value = StringUtils.isBlank(time) ? null : parseOrThrow(time, getFormat(format, timezone));
```
JMHReif, line 166: Should we use isEmpty() instead of isBlank()? Wondering if we should fail when the string is whitespace only. What do you think?

conker84: @vga91 did with `isEmpty`, but I asked to change it to `isBlank` because given my experience as data engineer the data can be very very very messy, so in order to be more lenient I asked to him to do in this way.
@JMHReif please lemme know if you prefer having `isEmpty`

PR #2041, opened 2021-06-28
Fixes #536: make export.csv configurable if it computes property types upfront

### LINKED ISSUE #536: make export.csv configurable if it computes property types upfront
make it possible to disable it

also check against the global list property-keys, if we already have the keys we should not continue to compute

we should also sample across all **existing** labels/rel-types combos.

vga91:
Fixes #536

- Added `sampling` config (default: false). When it's `true`, we use under the hood
the `apoc.meta.nodeTypeProperties` and the `apoc.meta.relTypeProperties` procedure to get the property types

- Fixed wrong test `testExportInvalidQuoteValue`

### DIFF
--- a/core/src/main/java/apoc/meta/tablesforlabels/PropertyTracker.java
+++ b/core/src/main/java/apoc/meta/tablesforlabels/PropertyTracker.java
@@ -39,7 +39,7 @@ public class PropertyTracker {
         return ret;
     }
 
-    private static final Map<String,String> typeMappings = new HashMap<String,String>();
+    public static final Map<String,String> typeMappings = new HashMap<String,String>();
     static {
         typeMappings.put("java.lang.String", "String");
         typeMappings.put("java.lang.String[]", "StringArray");
--- a/core/src/test/java/apoc/export/csv/ExportCsvTest.java
+++ b/core/src/test/java/apoc/export/csv/ExportCsvTest.java
@@ -173,7 +173,7 @@ public class ExportCsvTest {
         assertEquals(EXP_SAMPLE, readFile(fileName));
 
         // quotes: 'none' to simplify header testing
-        TestUtil.testCall(db, "CALL apoc.export.csv.all($file,{sample: 1, quotes: 'none'})", map("file", fileName),
+        TestUtil.testCall(db, "CALL apoc.export.csv.all($file, {samplingConfig: {sample: 1}, quotes: 'none'})", map("file", fileName),
                 (r) -> assertResults(fileName, r, "database", totalNodes, totalRels, totalProps, false));
         
         final String[] s = Files.lines(new File(directory, fileName).toPath()).findFirst().get().split(",");
--- a/core/src/test/java/apoc/export/csv/ExportCsvTest.java
+++ b/core/src/test/java/apoc/export/csv/ExportCsvTest.java
@@ -296,10 +296,10 @@ public class ExportCsvTest {
     }
 
     private void assertResults(String fileName, Map<String, Object> r, final String source, 
-                               Long expectedNodes, Long expectedRelationships, Long expectedProperties, boolean assertEquals) {
+                               Long expectedNodes, Long expectedRelationships, Long expectedProperties, boolean assertPropEquality) {
         assertEquals(expectedNodes, r.get("nodes"));
         assertEquals(expectedRelationships, r.get("relationships"));
-        if (assertEquals) {
+        if (assertPropEquality) {
             assertEquals(expectedProperties, r.get("properties"));
         } else {
             assertTrue((Long) r.get("properties") < expectedProperties);
--- a/core/src/test/java/apoc/export/csv/ExportCsvTest.java
+++ b/core/src/test/java/apoc/export/csv/ExportCsvTest.java
@@ -155,7 +155,7 @@ public class ExportCsvTest {
         assertEquals(EXP_SAMPLE, readFile(fileName));
 
         // quotes: 'none' to simplify header testing
-        TestUtil.testCall(db, "CALL apoc.export.csv.all($file, {samplingConfig: {sample: 1}, quotes: 'none'})", map("file", fileName),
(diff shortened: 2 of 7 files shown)


core/src/main/java/apoc/meta/tablesforlabels/PropertyTracker.java
```diff
@@ -39,7 +39,7 @@ private String assignTypeName(Object value) {
         return ret;
     }
 
-    private static final Map<String,String> typeMappings = new HashMap<String,String>();
+    public static final Map<String,String> typeMappings = new HashMap<String,String>();
```
JMHReif, line 42: why did this change from private to public?

core/src/test/java/apoc/export/csv/ExportCsvTest.java
```diff
@@ -122,6 +142,29 @@ public void testExportAllCsv() throws Exception {
...
+        String fileName = "all.csv";
+        final long totalNodes = 10L;
+        final long totalRels = 3L;
+        final long totalProps = 19L;
+        TestUtil.testCall(db, "CALL apoc.export.csv.all($file, null)", map("file", fileName),
+                (r) -> assertResults(fileName, r, "database", totalNodes, totalRels, totalProps, true));
+        assertEquals(EXP_SAMPLE, readFile(fileName));
+
+        // quotes: 'none' to simplify header testing
+        TestUtil.testCall(db, "CALL apoc.export.csv.all($file, {sampling: true, samplingConfig: {sample: 1}, quotes: 'none'})", map("file", fileName),
```
JMHReif, line 158: What's the purpose of choosing a sampling of 1 here? Lines below this aren't clear about the purpose.

docs/asciidoc/modules/ROOT/partials/usage/apoc.export.csv.all.adoc
```diff
@@ -95,3 +95,43 @@ RETURN file, nodes, relationships, properties, data
 "
 
 |===
+
+You can use the configuration `sampling` (default: `false`).
+With this config, the `apoc.export.csv.all` procedure use under the hood
```
JMHReif, line 100:
```suggestion
With this config, the `apoc.export.csv.all` procedure uses
```

docs/asciidoc/modules/ROOT/partials/usage/apoc.export.csv.all.adoc
```diff
@@ -95,3 +95,43 @@ RETURN file, nodes, relationships, properties, data
...
+the `apoc.meta.nodeTypeProperties` and the `apoc.meta.relTypeProperties` procedure to get the property types.
```
JMHReif, line 101:
```suggestion
the `apoc.meta.nodeTypeProperties` and the `apoc.meta.relTypeProperties` procedures under the hood to get the property types.
```

docs/asciidoc/modules/ROOT/partials/usage/apoc.export.csv.all.adoc
```diff
@@ -95,3 +95,43 @@ RETURN file, nodes, relationships, properties, data
...
+You can customize the configuration of these 2 `apoc.meta.*` procedure, using the `samplingConfig: MAP` configuration, 
+to limit the number of nodes/rels to analyze.
+
+So you can execute with the following dataset
```
JMHReif, line 105:
```suggestion
So you can execute with the following data set
```

docs/asciidoc/modules/ROOT/partials/usage/apoc.export.csv.all.adoc
```diff
@@ -95,3 +95,43 @@ RETURN file, nodes, relationships, properties, data
...
+----
+
+.Results
+[opts="header"]
+|===
+| file         | source                        | format | nodes | relationships | properties    | time | rows | batchSize  | batches    | done | data
+| "movies.csv" | "database: nodes(4), rels(1)" | "csv"  | 4     | 1             | 3             | 4   | 5   | 20000         | 1         | TRUE | NULL
+|===
+
+with a content similar to this (with `sample` the result could change):
```
JMHReif, line 126:
```suggestion
Execution of the above query would output content similar to that below (result could change depending on the `sample`):
```

docs/asciidoc/modules/ROOT/partials/usage/apoc.export.csv.all.adoc
```diff
@@ -95,3 +95,43 @@ RETURN file, nodes, relationships, properties, data
...
+
+So you can execute with the following dataset
+
+[source,cypher]
+----
+CREATE (:User:Sample {`last:Name`:'Galilei'}), (:User:Sample {address:'Universe'}), 
+    (:User:Sample {foo:'bar'})-[:KNOWS {one: 'two', three: 'four'}]->(:User:Sample {baz:'baa', foo: true})
+----
+
+the following query:
```
JMHReif, line 113:
```suggestion
Combined with the following query: 
```

core/src/test/java/apoc/export/csv/ExportCsvTest.java
vga91: To be sure not to analyze all available properties. 
With low sample (with `1` we are as sure as possible),
the header of the csv is not populated with all possible properties (and this is what I check on line 162 `s.length <17` ). 
However, the values "_id", "_labels", "_start", "_end", "_type" must always be present (line 163)

core/src/main/java/apoc/meta/tablesforlabels/PropertyTracker.java
vga91: Because I used the `typeMapping` to create the `REVERSED_TYPE_MAP` in `MetaInformation.java` 
(which, in fact, does the inverse of the `typeMapping`, that is, it gets `Class` from `propertyType`)

PR #2051, opened 2021-06-30
Fixes #832: Add apoc.redis.* procedures for integrating with Redis

### LINKED ISSUE #832: Add apoc.redis.* procedures for integrating with Redis
to e.g. quickly cache data our outsource materialized paths or something else with TTL

vga91:
Fixes #832


full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+                map("uri", URI, "from", from, "to", to, "config", config),
+                r -> assertEquals(true, r.get("value")));
+        
+        TestUtil.testCall(db, "CALL apoc.redis.get($uri, $to, $config)",
+                map("uri", URI, "to", to, "config", config),
+                r -> assertEquals("one", fromCodec(r.get("value"))));
+        
+        TestUtil.testCall(db, "CALL apoc.redis.exists($uri, [$to], $config)",
+                map("uri", URI, "to", to, "config", config),
+                r -> assertEquals(1L, r.get("value")));
```
fbiville, line 294: since `apoc.redis.copy` returns a true boolean, why does `apoc.redis.exists` return 1 here?

full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+        
+        TestUtil.testCall(db, "CALL apoc.redis.pexpire($uri, $to, 9999, false, $config)",
+                map("uri", URI, "to", to, "config", config),
+                r -> assertEquals(true, r.get("value")));
+        
+        TestUtil.testCall(db, "CALL apoc.redis.pttl($uri, $to, $config)",
+                map("uri", URI, "to", to, "config", config),
+                r -> {
+                    final long value = (long) r.get("value");
+                    assertTrue( value <= 9999L && value > 0L);
```
fbiville, line 304:
```suggestion
                    assertTrue( value <= 9999L && value >= 0L);
```
this might get flaky in slow environments otherwise

full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+                });
+
+        TestUtil.testCall(db, "CALL apoc.redis.persist($uri, $to, $config)",
```
fbiville, line 307: might get flaky in slow environments

full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+        TestUtil.testCall(db, "CALL apoc.redis.pttl($uri, $to, $config)",
+                map("uri", URI, "to", to, "config", config),
+                r -> assertEquals(-1L, (long) r.get("value")));
+    }
+
+    @Test
+    public void testServersCommand() {
+        TestUtil.testCall(db, "CALL apoc.redis.info($uri, $config)",
+                map("uri", URI, "config", map("codec", codec)),
+                r -> assertTrue(((String) r.get("value")).contains("redis_version:6.2.3")));
```
fbiville, line 320: maybe we can be a bit less specific about the version to make upgrades easier?

full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+
+    @Parameterized.Parameter
+    public String codec;
+
+    
+    @Test
+    public void testStringsCommands() {
+        Map<String, Object> config = map("codec", codec);
+        Object key = getByCodec("myKey");
+        TestUtil.testCall(db, "CALL apoc.redis.setGet($uri, $key, $myValue, $config)",
```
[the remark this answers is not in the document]
vga91: Yes, the command is `GETSET`, I chose to call it this for consistency with the Java Client name command
https://lettuce.io/core/release/api/io/lettuce/core/api/sync/RedisStringCommands.html#setGet-K-V- ,
but of course, I can change it if you want.

full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+
+    private static GenericContainer redis;
+
+    @ClassRule
+    public static DbmsRule db = new ImpermanentDbmsRule();
+
+
+    @BeforeClass
+    public static void beforeClass() {
+        assumeFalse(isRunningInCI());
```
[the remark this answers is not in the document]
vga91: Yes, in agreement with @conker84 I added this one because, as for `MongoDb` case, the `GenericContainer` tests are very slow.

full/src/main/java/apoc/redis/RedisConnection.java
```diff
@@ -0,0 +1,24 @@
+package apoc.redis;
+
+import io.lettuce.core.ClientOptions;
+import io.lettuce.core.RedisClient;
+
+abstract public class RedisConnection<T> implements IRedisConnection<T> {
```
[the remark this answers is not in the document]
vga91: No in fact. Changed with `default` visibility

full/src/test/java/apoc/redis/RedisTest.java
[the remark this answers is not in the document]
vga91: Ok, I changed this one after consulting with @conker84

full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+        TestUtil.testCall(db, "CALL apoc.redis.scard($uri, $key, $config)",
+                map("uri", URI, "key", key, "config", config),
+                r -> assertEquals(2L, r.get("value")));
+    }
+
+    @Test
+    public void testSortedSetsCommands() {
+        Map<String, Object> config = map("codec", this.codec);
+        Object key = getByCodec("mySortedSetKey");
+        TestUtil.testCall(db, "CALL apoc.redis.zadd($uri, $key, [0, $first, 100, $third, 1, $second], $config)",
```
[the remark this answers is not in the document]
vga91: Certainly, it would be more readable, but unfortunately I cannot transform the parameter into a map because I would have the error `type Map cannot be converted to a Neo4j type: Maps are required to have String keys`, with `byte[]` codec.

full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    @Test
+    public void testHashesCommands() {
+        Map<String, Object> config = map("codec", this.codec);
+        Object key = getByCodec("mapKey");
+        Object numberValue = getByCodec("number");
+        TestUtil.testCall(db, "CALL apoc.redis.hset($uri, $key, $field, $value, $config)",
+                map("uri", URI, "key", key, "field", numberValue, "value", getByCodec("1"), "config", config),
+                r -> assertEquals(true, r.get("value")));
```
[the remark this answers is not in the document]
vga91: Yes, the `HSET` command returns this, but the corresponding Java command returns a boolean (true if the field is new, otherwise false).
To be clearer, I updated the procedure description with this note.
https://lettuce.io/lettuce-4/release/api/com/lambdaworks/redis/api/sync/RedisHashCommands.html#hset-K-K-V-

full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    @Test
+    public void testEvalCommand() {
+        Object keyEval = getByCodec("testEval");
+        TestUtil.testCall(db, "CALL apoc.redis.setGet($uri, $keyEval, $valueEval, $config)",
+                map("uri", URI, "keyEval", keyEval, "valueEval", getByCodec("valueEval"), "config", map("codec", codec)), 
+                r -> assertNull(r.get("value")));
+        
+        TestUtil.testCall(db, "CALL apoc.redis.eval($uri, 'return redis.call(\"get\", KEYS[1])', 'VALUE', [$keyEval], [$keyName], $config)",
```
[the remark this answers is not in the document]
vga91: The `VALUE` here is the `ScriptOutputType`, that is a `String` or a `byte[]` (depending on `codec` type).
This is an `enum` (with values `BOOLEAN`, `INTEGER`, `STATUS`, `VALUE`, `MULTI`)
https://lettuce.io/core/release/api/io/lettuce/core/ScriptOutputType.html

Regarding the second point I updated the test case to be more exhaustive, 
Now the procedure `CALL apoc.redis.eval($uri, 'return redis.call("set", KEYS[1], ARGV[1])', 'VALUE', [$keyEval], [$keyName], $config)` is equivalent to redis-cli command https://redis.io/commands/eval "return redis.call('set', KEYS[1], ARGV[1])" 1 $keyEval $keyName`, where `$keyEval` is the value for `KEYS[1]` value, and `$valueEval` is the value for `ARGV[1]`.

full/src/test/java/apoc/redis/RedisTest.java
```diff
@@ -0,0 +1,375 @@
...
+
+    @Test
+    public void testEvalCommand() {
+        Object keyEval = getByCodec("testEval");
+        TestUtil.testCall(db, "CALL apoc.redis.setGet($uri, $keyEval, $valueEval, $config)",
+                map("uri", URI, "keyEval", keyEval, "valueEval", getByCodec("valueEval"), "config", map("codec", codec)), 
+                r -> assertNull(r.get("value")));
+        
+        TestUtil.testCall(db, "CALL apoc.redis.eval($uri, 'return redis.call(\"get\", KEYS[1])', 'VALUE', [$keyEval], [$keyName], $config)", 
+                map("uri", URI, "keyEval", keyEval, "keyName", getByCodec("key:name"), "config", map("codec", codec)),
```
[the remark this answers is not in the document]
vga91: Changed as above, previously this param represented the `ARGV[n]` params, that wasn't there.
I put it on to show that it still works, but it wasn't clear, in effect...

full/src/test/java/apoc/redis/RedisTest.java
vga91: Also in this case, the java client works differently, because `copy` returns true or false instead of 1 or 0  https://redis.io/commands/COPY command.
I could change the `@Description`, but maybe in this case it is worth changing the return value to make it consistent with the `COPY` command. Wdyt?

full/src/test/java/apoc/redis/RedisTest.java
vga91: Changed with `assertEventually`

full/src/test/java/apoc/redis/RedisTest.java
vga91: Now I extracted the string the `6.2.3` string here and in `GenericContainer` param,
in a `REDIS_VERSION` constant to avoid discrepancies.

full/src/test/java/apoc/redis/RedisTest.java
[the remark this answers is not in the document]
JMHReif: Ok, makes sense. Thanks for clarifying!

full/src/test/java/apoc/redis/RedisTest.java
JMHReif: Let's leave it consistent with the Java client. Otherwise, we should change them all to match Redis client.

full/src/main/java/apoc/redis/Redis.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    @Procedure
+    @Description("apoc.redis.hgetall(uri, key, {config}) | Execute the 'HGETALL key' command")
+    public Stream<MapResult> hgetall(@Name("uri") String uri, @Name("key") Object key, @Name(value = "config", defaultValue = "{}") Map<String, Object> config) {
+        return withConnection(uri, config, connection -> Stream.of(new MapResult(connection.hgetall(key))));
+    }
+
+    @Procedure
+    @Description("apoc.redis.hset(uri, key, field, value, {config}) | Execute the 'HSET key field value' command and returns true if is a new field in the hash or false if the field already exists")
```
JMHReif, line 79:
```suggestion
    @Description("apoc.redis.hset(uri, key, field, value, {config}) | Execute the 'HSET key field value' command and returns true if it is a new field in the hash or false if the field already exists")
```

docs/asciidoc/modules/ROOT/pages/overview/apoc.redis/apoc.redis.eval.adoc
```diff
@@ -0,0 +1,40 @@
+////
+This file is generated by DocsTest, so don't change it!
+////
+
+= apoc.redis.eval
+:description: This section contains reference documentation for the apoc.redis.eval procedure.
+
```
JMHReif, line 7: In the parameters provided to the procedure, `keys` are bound to the `KEYS[n]` like special array of the Lua script and `values` are bound to the `ARGV[n]` like special array of the Lua script.

JMHReif: Might want to add this to the docs to try to make this procedure clearer.....at least to us non-Redis users, it feels a bit complicated. 😅

vga91: You're right, it's not easily understandable. Now I've added this phrase into `@Description` and 2 usage examples in the documentation.

fbiville (approved):

PR #2053, opened 2021-07-01
Fixes #1487: apoc.schema.nodes() fails if a fulltext index (node or relationship) is present on multiple labels

### LINKED ISSUE #1487: apoc.schema.nodes() fails if a fulltext index (node or relationship) is present on multiple labels
apoc.schema.nodes() should succeed without error no matter what indexes or constraints exist.

When a fulltext index (of any kind) is present and it is on multiple labels, `apoc.schema.nodes()` fails with the following error:

```
Neo.ClientError.Procedure.ProcedureCallFailed: Failed to invoke procedure `apoc.schema.nodes`: Caused by: java.lang.IllegalStateException: Index with more than one label
```

Create a fulltext schema index (node or relationship) upon multiple labels or relationship types, then attempt to execute `apoc.schema.nodes()`

```
call db.index.fulltext.createRelationshipIndex('test', ['Blah', 'Moon'], ['weightProp']);
```

### Steps (Mandatory)

  1. Create a fulltext index similar to the call cited in the simple dataset section above
  2. Execute `apoc.schema.nodes()`
  3. Observe error message

## Specifications (Mandatory)

APOC 3.5.0.9

### Versions

  - OS: Mac OS 10.14 (Mojave)
  - Neo4j: 3.5.8
  - Neo4j-Apoc: 3.5.0.9

vga91:
Fixes #1487

- Fix `apoc.schema.nodes` error with rel-constraints and vice-versa

- FullText indexes can have composite labels, so previously `apoc.schema.assert` (with default dropIndex: `true`) failed because of `Iterable.single`


core/src/test/java/apoc/schema/SchemasTest.java
```diff
@@ -585,4 +634,25 @@ public void testConstraintsRelationshipsAndExcludeRelationshipsValuatedShouldFai
         }
 
     }
+
+    @Test
+    public void testIndexesWithMultipleLabels() {
```
fbiville, line 639: would be good to add a test for indices with multi rel types

core/src/main/java/apoc/schema/Schemas.java
```diff
@@ -185,6 +187,18 @@ private AssertSchemaResult createUniqueConstraint(Schema schema, String lbl, Str
         return result;
     }
 
+    private Object getLabelForAssert(IndexDefinition definition, boolean nodeIndex) {
```
fbiville, line 190: what about using `List<String>` as a return type, and return either the list as is (multi-label / multi-type), or a list of 1 element (single label / single type)?

fbiville (approved):

core/src/main/java/apoc/schema/Schemas.java
vga91: This would cause a breaking change,
because I would have as field `label`, no longer `"Foo"` but `["Foo"]` for example.
I don't know if it's worth it, wdyt?

core/src/test/java/apoc/schema/SchemasTest.java
vga91: Yes the name was confusing, now I've changed it.
There is already a multi-relType index, `CALL db.index.fulltext.createRelationshipIndex ('fullIdxRel', ['TYPE_1', 'TYPE_2'], ['alpha', 'beta'])`.
To maintain the same behavior as before, `apoc.schema.nodes` shows both indexes and constraints,
while `apoc.schema.relationships` only the constraints,
so in this test-case it must return an empty list, without errors unlike before (effectively this check was missing, I added it on line `658`)

PR #2054, opened 2021-07-01
Fixes #1445: apoc.json.path() returns nulls where it should not

### LINKED ISSUE #1445: apoc.json.path() returns nulls where it should not
## Guidelines

When using the  attached sourceTree on https://jsonpath.curiousconcept.com/ with the expression
     `$..columns`
I get the correct list of results.

## Expected Behavior 
apoc.json.path(tree,'$..columns') should return correct results.

## Actual Behavior 
apoc.json.path(tree,'$..columns') inserts nulls. See attached files for input and output

##Attachments
[jsonPathResult.txt](https://github.com/neo4j-contrib/neo4j-apoc-procedures/files/4323980/jsonPathResult.txt)
[sourceTree.txt](https://github.com/neo4j-contrib/neo4j-apoc-procedures/files/4323981/sourceTree.txt)

### Versions

  - OS: Linux
  - Neo4j: 3.5.4
  - Neo4j-Apoc: 3.5.0.4

vga91:
Fixes #1445

- Added `path-options` in every function/procedure that can use `json-path` as parameter


core/src/test/java/apoc/convert/ConvertJsonTest.java
```diff
@@ -21,20 +23,104 @@
...
+    @Test
+    public void testJsonPathWithMapFunctions() {
+        // apoc.convert.getJsonPropertyMap and apoc.convert.fromJsonMap must fail with "ALWAYS_RETURN_LIST" because should return a Map.
+        final Map<String, String> expectedMap = Map.of("_id", "772col2");
+        final String expectedError = "com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.util.LinkedHashMap<java.lang.Object,java.lang.Object>` out of START_ARRAY token";
+
+        testCall(db, "RETURN apoc.convert.fromJsonMap($json, '$.columns.col2') AS path", Map.of("json", JSON),
+                (row) -> assertEquals(expectedMap, row.get("path")));
+        try {
+            testCall(db, "RETURN apoc.convert.fromJsonMap($json, '$.columns.col2', ['ALWAYS_RETURN_LIST']) AS path", Map.of("json", JSON),
```
JMHReif, line 108: Could we catch the `ALWAYS_RETURN_LIST` option early before it runs? Then we could give a better error message that this option doesn't make sense with a map.

core/src/test/java/apoc/convert/ConvertJsonTest.java
```diff
@@ -21,20 +23,104 @@
...
+                    (row) -> fail("Should fail because of MismatchedInputException"));
+        } catch (QueryExecutionException e) {
+            assertTrue(e.getMessage().contains(expectedError));
+        }
+        
+        db.executeTransactionally("CREATE (n:JsonPathNode {prop: $prop})", Map.of("prop", JSON));
+        testCall(db, "MATCH (n:JsonPathNode) RETURN apoc.convert.getJsonPropertyMap(n, 'prop', '$.columns.col2') AS path", Map.of("json", JSON),
+                (row) -> assertEquals(expectedMap, row.get("path")));
+        try {
+            testCall(db, "MATCH (n:JsonPathNode) RETURN apoc.convert.getJsonPropertyMap(n, 'prop', '$.columns.col2', ['ALWAYS_RETURN_LIST']) AS path", Map.of("json", JSON),
```
JMHReif, line 118: Same comment as above.

JMHReif (approved):

core/src/test/java/apoc/convert/ConvertJsonTest.java
vga91: Surely, added in `JsonUtil.parse()`

core/src/test/java/apoc/convert/ConvertJsonTest.java
vga91: Added as above

fbiville (approved):

PR #2055, opened 2021-07-01
Fixes #1157: apoc.atomic.add and related procedures do not work with Integer properties

### LINKED ISSUE #1157: apoc.atomic.add and related procedures  do not work with Integer properties

When a Node has a property with an Integer value (which is allowed by Neo4J PropertyContainer.setProperty()), the apoc.atomic.add() procedure should work.

An IllegalArgumentException is thrown :

<pre>
2019-04-09 03:22:01.901+0000 ERROR org.neo4j.graphdb.QueryExecutionException: Failed to invoke procedure `apoc.atomic.add`: Caused by: java.lang.IllegalArgumentException: [null] is not a supported property value
	at [...]
Caused by: java.lang.IllegalArgumentException: [null] is not a supported property value
	at org.neo4j.values.storable.Values.unsafeOf(Values.java:594)
	at org.neo4j.values.storable.Values.of(Values.java:516)
	at org.neo4j.kernel.impl.core.NodeProxy.setProperty(NodeProxy.java:242)
	at apoc.atomic.Atomic.lambda$add$0(Atomic.java:47
	at apoc.atomic.Atomic.lambda$retry$6(Atomic.java:226)
	at org.neo4j.helpers.TransactionTemplate.lambda$execute$2(TransactionTemplate.java:179)
	at org.neo4j.helpers.TransactionTemplate.execute(TransactionTemplate.java:201)
	... 82 more
</pre>

Create a node with an int property (use node.setProperty with an int argument), and try and invoke apoc.atomic.add on the node for that property.

### Steps (Mandatory)

  1. Create Node
  2. Set a property to an int value
  3. Commit
  3. Retrieve the node and invoke apoc.atomic.add on it

### Root Cause

The AtomicUtils class in apoc.atomic.util returns <code>null</code> if the type of the property is not one of the 4 types it expects (Long, Double, Short, Byte). Integer, Float and Short should be supported too since they can all occur as types of a Property value in a PropertyContainer.

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: OSX
  - Neo4j: 3.5.3
  - Neo4j-Apoc: 3.5.0.2

vga91:
Fixes #1157

### DIFF
--- a/core/src/test/java/apoc/atomic/AtomicTest.java
+++ b/core/src/test/java/apoc/atomic/AtomicTest.java
@@ -35,6 +35,40 @@ public class AtomicTest {
 	}
 
 	@Test
+	public void testAddAndSubIngeger(){
+		db.executeTransactionally("CREATE (p:Person {name:'Tom'})");
+		try(Transaction tx = db.beginTx()){
+			final Node node = tx.getAllNodes().stream().findFirst().orElse(null);
+			node.setProperty("age", 1);
+			tx.commit();
+		}
+		testCall(db, "MATCH (n:Person {name:'Tom'}) CALL apoc.atomic.add(n,$property,$value) YIELD container RETURN count(*)",map("property","age","value",10), (r) -> {});
+		int age = TestUtil.singleResultFirstColumn(db, "MATCH (n:Person {name:'Tom'}) RETURN n.age as age");
+		assertEquals(11, age);
+		testCall(db, "MATCH (n:Person {name:'Tom'}) CALL apoc.atomic.subtract(n,$property,$value) YIELD container RETURN count(*)",map("property","age","value",5), (r) -> {});
+		int ageSub = TestUtil.singleResultFirstColumn(db, "MATCH (n:Person {name:'Tom'}) RETURN n.age as age");
+		assertEquals(6, ageSub);
+	}
+
+	@Test
+	public void testAddAndSubFloat() {
+		db.executeTransactionally("CREATE (p:Person {name:'Tom'})");
+		try(Transaction tx = db.beginTx()){
+			final Node node = tx.getAllNodes().stream().findFirst().orElse(null);
+			node.setProperty("age", 2.4F);
+			tx.commit();
+		}
+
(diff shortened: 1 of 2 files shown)


core/src/test/java/apoc/atomic/AtomicTest.java
```diff
@@ -34,6 +34,44 @@
 		TestUtil.registerProcedure(db, Atomic.class);
 	}
 
+	@Test
+	public void testAddAndSubInteger(){
+		db.executeTransactionally("CREATE (p:Person {name:'Tom'})");
+		try(Transaction tx = db.beginTx()) {
```
fbiville, line 40: maybe it's easier to set the age property in the first query directly?

core/src/test/java/apoc/atomic/AtomicTest.java
```diff
@@ -34,6 +34,44 @@
...
+				map("property","age","value",5));
+		int ageSub = TestUtil.singleResultFirstColumn(db, "MATCH (n:Person {name:'Tom'}) RETURN n.age as age");
+		assertEquals(6, ageSub);
+	}
+
+	@Test
+	public void testAddAndSubFloat() {
+		db.executeTransactionally("CREATE (p:Person {name:'Tom'})");
+		try(Transaction tx = db.beginTx()) {
+			final Node node = tx.getAllNodes().stream().findFirst().orElse(null);
```
fbiville, line 59: same comment as above

fbiville (approved):

core/src/test/java/apoc/atomic/AtomicTest.java
vga91: I created a new `setProperty` to "force" the data type to be an Integer,
because with `"CREATE (p:Person {name:'Tom', age: 1})"` in `executeTransactionally`,
the `age` property will be under-the-hood turned into `Long`.

core/src/test/java/apoc/atomic/AtomicTest.java
vga91: As above

PR #2062, opened 2021-07-01
Fixes #1949: The procedure apoc.export.xls.graph fails with 100+ nodes

### LINKED ISSUE #1949: The procedure `apoc.export.xls.graph` fails with 100+ nodes

The procedure should work correclty.

```
Failed to invoke procedure `apoc.export.xls.graph`: Caused by: java.lang.NullPointerException
```

```
UNWIND range(1,100) as range CREATE (n:Test)

```

### Steps (Mandatory)

```
CALL apoc.graph.fromDB('test',{}) yield graph
CALL apoc.export.xls.graph(graph, "export.xls", {})
YIELD file
RETURN file;
```

### Versions

  - OS: Mac OSX
  - Neo4j: 4.2.6
  - Neo4j-Apoc: 4.2.0.4

vga91:
Fixes #1949

Improved test assertions to also check the Excel header.

---

Unfortunately, this is not the optimal solution.
We should find a way to create the Excel header without resorting to the `sheet.getRow(0)` of line 172, 
because without the modification of this pr it is null with more than one hundred lines.
On the other hand, with this solution we don't take advantage of the streaming functionality that Apache POI's`SXSSFWorkbook` offers.


full/src/test/java/apoc/export/csv/ExportXlsTest.java
```diff
@@ -71,6 +77,39 @@ public void testExportGraphXls() throws Exception {
...
+        String endNodeId = "customEnd";
+        db.executeTransactionally("UNWIND range(1,200) as range CREATE (n:Test)");
+        String fileName = "graph.xlsx";
+        TestUtil.testCall(db, "CALL apoc.graph.fromDB('test',{}) yield graph " +
+                        "CALL apoc.export.xls.graph(graph, $file, $conf) " +
+                        "YIELD nodes, relationships, properties, file, source,format, time " +
+                        "RETURN *",
+                map("file", fileName, "conf", map("headerNodeId", nodeId,
+                        "headerRelationshipId", relId, "headerStartNodeId", startNodeId, "headerEndNodeId", endNodeId)),
+                (r) -> assertResults(fileName, r, "graph", 208L, 2L, 206));
```
fbiville, line 108: can you explain these numbers?
Shouldn't we get 200 nodes, 0 rels?
Not sure what 206 counts, here.

fbiville (approved):

full/src/test/java/apoc/export/csv/ExportXlsTest.java
vga91: The number `206` is due to `setUp()` method, where 6 additional nodes are created (the same for relationships).
The `208L` instead it is due to having 2 nodes with double labels (also in `setUp()`).
So, this numbers are the sum of `assertResults` on line `177` plus 200 new nodes.

PR #2072, opened 2021-07-07
Fixes #914: virtual paths and graphs

### LINKED ISSUE #914: virtual paths and graphs
create proper, independent, virtual paths/graphs perhaps with a config option in

`apoc.graph.fromXXX`

add a procedure and function: 

- apoc.create.vPath(path)
- apoc.create.vPaths(paths)

alternatively we can also allow object as parameter and allow relationship,node, or lists of both.

vga91:
Fixes #914

Added `apoc.create.vPath(path)` and `apoc.create.vPaths(paths)`


core/src/main/java/apoc/create/Create.java
```diff
@@ -220,6 +221,31 @@ public Relationship vRelationshipFunction(@Name("from") Node from, @Name("relTyp
         return Stream.of(new VirtualPathResult(from, rel, to));
     }
 
+    @Procedure
+    @Description("apoc.create.vPath")
```
fbiville, line 225: what's the difference with `apoc.create.virtualPath`?

vga91: The `apoc.create.vPath` create a virtual Path from another real Path,
through a `PathImpl.Builder` that create everything I need.
With `apoc.create.virtualPath` we have to pass several parameters (`labels`, `relType`, `props`) to create our parameter,
so if I already have a path, I should model it to fit the `apoc.create.virtualPath`.

fbiville: what about renaming this procedure to something like `clonePathToVirtual`?

core/src/main/java/apoc/create/Create.java
```diff
@@ -220,6 +221,31 @@ public Relationship vRelationshipFunction(@Name("from") Node from, @Name("relTyp
...
+    @Procedure
+    @Description("apoc.create.vPaths")
+    public Stream<PathResult> vPaths(@Name("paths") List<Path> paths) {
+        return paths.stream().map(this::createVPath);
+    }
+
+    private PathResult createVPath(Path path) {
+        final Iterable<Relationship> relationships = path.relationships();
+        final Node first = path.startNode();
+        PathImpl.Builder builder = new PathImpl.Builder(new VirtualNode(first, Iterables.asList(first.getPropertyKeys())));
```
fbiville, line 239: I'd rather avoid using this class, even though it's currently public.
What about introducing a `VirtualPath` class that implements the stable `Path` interface, similar to what has been done with `VirtualNode` and `VirtualRelationship`?

core/src/test/java/apoc/create/CreateTest.java
```diff
@@ -228,5 +235,66 @@ public void testVirtualFromNodeFunction() throws Exception {
...
+    @Test
+    public void testVirtualPath() {
+        db.executeTransactionally("CREATE p=(a:Test {foo: 7})-[:TEST]->(b:Baa:Baz {a:'b'})<-[:TEST_2 {aa:'bb'}]-(:Bar {one:'www'}), \n" +
+                "q=(:Omega {alpha: 'beta'})<-[:TEST_3 {aa:'ccc'}]-(:Bar {one:'jjj'})");
+        testCall(db, "MATCH p=(a:Test {foo: 7})-[:TEST]->(b:Baa:Baz {a:'b'})<-[:TEST_2 {aa:'bb'}]-(:Bar {one:'www'}) WITH p \n" +
+                        "CALL apoc.create.vPath(p) YIELD path RETURN path",
+                (row) -> assertionsFirstVPath((Path) row.get("path")));
+
+        testResult(db, "MATCH p=(a:Test {foo: 7})-[:TEST]->(b:Baa:Baz {a:'b'})<-[:TEST_2 {aa:'bb'}]-(:Bar {one:'www'}), \n" +
+                        "q=(:Omega {alpha: 'beta'})<-[:TEST_3 {aa:'ccc'}]-(:Bar {one:'jjj'}) WITH collect(p)+q as paths \n" +
```
fbiville, line 248:
```suggestion
                        "q=(:Omega {alpha: 'beta'})<-[:TEST_3 {aa:'ccc'}]-(:Bar {one:'jjj'}) WITH [p, q] as paths \n" +
```
wouldn't this work instead?

fbiville (dismissed):

core/src/main/java/apoc/create/Create.java
vga91: Ok, changed with `VirtualPath` class

core/src/main/java/apoc/result/VirtualPath.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public VirtualPath(Node start) {
+        Objects.requireNonNull(start);
+        this.start = start;
+    }
+
+    public void addRel(Relationship relationship) {
+        Objects.requireNonNull(relationship);
+        this.relationships.add(relationship);
+    }
```
JMHReif, line 33: optional: maybe add a check to ensure the rel is either connected to the last end node or to the path's start node? Would ensure we don't get disconnected paths.

JMHReif (approved):

PR #2075, opened 2021-07-08
Fixes #2074: apoc.import.csv failed for NULL values when using INT datatype with error [null] is not a supported property value

### LINKED ISSUE #2074: `apoc.import.csv` failed for NULL values when using `INT` datatype with error "[null] is not a supported property value"
## Guidelines

CSV should get imported by ignoring the null value or may be a config to ignore null/empty  values can be added.

apoc.import.csv  fails with below error:
`
Failed to invoke procedure `apoc.import.csv`: Caused by: java.lang.IllegalArgumentException: [null] is not a supported property value
`

The issue can be easily replicated using :

`CALL apoc.import.csv([{fileName: "file:/fail.csv", labels: ["entity"]}], [], {})`

Empty string value works fine in this case but error is thrown in case of integer property. I had a look at the documentation, the available config parameters does not contain any parameter to avoid nulls. 

The attachment pass.csv contains nulls in string and works fine.
fail.csv fails with above mentioned error as it contains null for Integer property.

fail.csv
:ID(node_space_1),:LABEL,str_attribute:STRING,int_attribute:INT
n1,thing,once upon a time,1
n2,thing,there was a boy,2
n3,thing_variant,,3

pass.csv
:ID(node_space_1),:LABEL,str_attribute:STRING,int_attribute:INT
n1,thing,once upon a time,1
n2,thing,there was a boy,2
n3,thing_variant,,3

```

CALL apoc.import.csv([{fileName: "file:/fail.csv", labels: ["entity"]}], [], {})


```

### Steps (Mandatory)

  1. Make sure compatible APOC version is installed  
  1. copy fail.csv and pass.csv in import folder 
  1.CALL apoc.import.csv([{fileName: "file:/fail.csv", labels: ["entity"]}], [], {})

## Specifications (Mandatory)

String values having nulls got imported successfully . The issue exists with Integer property values. Available config parameters doesn't contain any parameter to skip nulls from properties.

Also interested to know if the string attribute can be interpreted as having no value instead of being a literal empty string?

Currently used versions

### Versions

  - OS:MacOS
  - Neo4j: 4.3.1
  - Neo4j-Apoc: 4.2.0.2

vga91:
Fixes #2074
- Ignore null properties during the import
- Added `ignoreBlankString` config

In addition to the fix, I've added in this pr the handling of `INT[]` and `FLOAT[]` (before they failed due to `Cast Exeption`)

### DIFF
--- a/core/src/main/java/apoc/export/csv/CsvLoaderConfig.java
+++ b/core/src/main/java/apoc/export/csv/CsvLoaderConfig.java
@@ -15,6 +15,7 @@ public class CsvLoaderConfig {
     private static final String SKIP_LINES = "skipLines";
     private static final String BATCH_SIZE = "batchSize";
     private static final String IGNORE_DUPLICATE_NODES = "ignoreDuplicateNodes";
+    private static final String IGNORE_EMPTY_STRING = "ignoreEmptyString";
 
     private static char DELIMITER_DEFAULT = ',';
     private static char ARRAY_DELIMITER_DEFAULT = ';';
@@ -23,6 +24,7 @@ public class CsvLoaderConfig {
     private static int SKIP_LINES_DEFAULT = 1;
     private static int BATCH_SIZE_DEFAULT = 2000;
     private static boolean IGNORE_DUPLICATE_NODES_DEFAULT = false;
+    private static boolean IGNORE_EMPTY_STRING_DEFAULT = false;
 
     private final char delimiter;
     private final char arrayDelimiter;
@@ -31,6 +33,7 @@ public class CsvLoaderConfig {
     private final int skipLines;
     private final int batchSize;
     private final boolean ignoreDuplicateNodes;
+    private final boolean ignoreEmptyString;
 
     private CsvLoaderConfig(Builder builder) {
         this.delimiter = builder.delimiter;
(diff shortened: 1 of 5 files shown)


fbiville (approved):

core/src/main/java/apoc/export/csv/CsvLoaderConfig.java
```diff
@@ -102,7 +110,7 @@ public static CsvLoaderConfig from(Map<String, Object> config) {
         if (config.get(SKIP_LINES) != null) builder.skipLines((int) config.get(SKIP_LINES));
         if (config.get(BATCH_SIZE) != null) builder.batchSize((int) config.get(BATCH_SIZE));
         if (config.get(IGNORE_DUPLICATE_NODES) != null) builder.ignoreDuplicateNodes((boolean) config.get(IGNORE_DUPLICATE_NODES));
-
+        if (config.get(IGNORE_EMPTY_STRING) != null) builder.ignoreEmptyString((boolean) config.get(IGNORE_EMPTY_STRING));
```
fbiville, line 113: nit: s/IGNORE_EMPTY_STRING/IGNORE_BLANK_STRING/ ?

fbiville (approved):

core/src/main/java/apoc/export/csv/CsvLoaderConfig.java
vga91: Sorry, I didn't get it 😅.
You mean another config `ignoreBlankString()`?

fbiville: My bad, I meant, rename `IGNORE_EMPTY_STRING` to `IGNORE_BLANK_STRING` (but you're right about renaming `builder#ignoreEmptyString` too!)

JMHReif (approved):

PR #2097, opened 2021-07-20
Fixes #1910: apoc.meta.stats return false relationship count when having multi-labeled nodes

### LINKED ISSUE #1910: apoc.meta.stats return false relationship count when having multi-labeled nodes
If you have multi-labeled nodes in a graph, the `apoc.meta.stats` procedure return false information about relationship count.

Example dataset:

```
CREATE (:Node:Test)-[:REL]->(:Node)<-[:REL]-(:Node:Test)
```

We can then run the apoc.meta.stats procedure

```
call apoc.meta.stats YIELD relCount, relTypesCount, relTypeCount
```

The results are actually quite weird:

results_stats

I guess the global number of relCount is correct as it return a value of 2. The relTypesCount duplicates the relationship count when you are dealing with a multilabeled node. OTOH I don't really know what relTypeCount of 7 is really doing

vga91:
Fixes #1910

The problem is that `TokenRead.labelCount()` and `TokenRead.relationshipTypeCount()` returns all labels/relTypes (even those deleted).
For consistency with the other results, we should return only labels/relTypes in use.

### DIFF
--- a/core/src/test/java/apoc/meta/MetaTest.java
+++ b/core/src/test/java/apoc/meta/MetaTest.java
@@ -1092,22 +1092,41 @@ public class MetaTest {
 
     @Test
     public void testMetaStatsWithLabelAndRelTypeCountInUse() {
-        db.executeTransactionally("CREATE (:Node:Test)-[:REL {a:'b'}]->(:Node {c: 'd'})<-[:REL]-(:Node:Test)");
+        db.executeTransactionally("CREATE (:Node:Test)-[:REL {a: 'b'}]->(:Node {c: 'd'})<-[:REL]-(:Node:Test)");
         db.executeTransactionally("CREATE (:A {e: 'f'})-[:ANOTHER {g: 'h'}]->(:C)");
         
         TestUtil.testCall(db, "CALL apoc.meta.stats()", row -> {
-            assertEquals(4L, row.get("labelCount"));
-            assertEquals(2L, row.get("relTypeCount"));
+            assertEquals(map("A", 1L, "C", 1L, "Test", 2L, "Node", 3L), row.get("labels"));
             assertEquals(5L, row.get("nodeCount"));
+            assertEquals(4L, row.get("labelCount"));
+            
+            assertEquals(map("REL", 4L, "ANOTHER", 1L), row.get("relTypesCount"));
+            assertEquals(2L, row.get("relTypeCount"));
             assertEquals(3L, row.get("relCount"));
+            Map<String, Object> expectedRelTypes = map("(:A)-[:ANOTHER]->()", 1L,
+                    "()-[:REL]->(:Node)", 2L, 
+                    "(:Test)-[:REL]->()", 2L,
+                    "(:Node)-[:REL]->()", 2L, 
+                    "()-[:ANOTHER]->(:C)", 1L,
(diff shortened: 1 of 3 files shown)


core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -1089,4 +1089,45 @@ public void testMetaStatsWithTwoDots() {
...
+    public void testMetaStatsWithLabelAndRelTypeCountInUse() {
+        db.executeTransactionally("CREATE (:Node:Test)-[:REL {a: 'b'}]->(:Node {c: 'd'})<-[:REL]-(:Node:Test)");
+        db.executeTransactionally("CREATE (:A {e: 'f'})-[:ANOTHER {g: 'h'}]->(:C)");
+        
+        TestUtil.testCall(db, "CALL apoc.meta.stats()", row -> {
+            assertEquals(map("A", 1L, "C", 1L, "Test", 2L, "Node", 3L), row.get("labels"));
+            assertEquals(5L, row.get("nodeCount"));
+            assertEquals(4L, row.get("labelCount"));
+            
+            assertEquals(map("REL", 4L, "ANOTHER", 1L), row.get("relTypesCount"));
```
fbiville, line 1103:
```suggestion
            assertEquals(map("REL", 2L, "ANOTHER", 1L), row.get("relTypesCount"));
```
shouldn't it be 2?

core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -1089,4 +1089,45 @@ public void testMetaStatsWithTwoDots() {
...
+            assertEquals(expectedRelTypes, row.get("relTypes"));
+        });
+        
+        db.executeTransactionally("match p=(:A)-[:ANOTHER]->(:C) delete p");
+        TestUtil.testCall(db, "CALL apoc.meta.stats()", row -> {
+            assertEquals(map("Test", 2L, "Node", 3L), row.get("labels"));
+            assertEquals(3L, row.get("nodeCount"));
+            assertEquals(2L, row.get("labelCount"));
+            
+            assertEquals(map("REL", 4L), row.get("relTypesCount"));
```
fbiville, line 1122: same question as above

core/src/test/java/apoc/meta/MetaTest.java
vga91: The 4 is due to double label nodes `:Node:Test`.
In this case, the `apoc.meta.stats` count separately 2 relationship `(:Test)-[:REL]-(:Node)` and 2 `(:Node)-[:REL]-(:Node)`.
Instead, using a single labels, `REL` count in 2.

core/src/test/java/apoc/meta/MetaTest.java
vga91: As above

core/src/test/java/apoc/meta/MetaTest.java
JMHReif: Could we add something in the docs for this to mention that it counts unique patterns where multi-label nodes will be counted as separate paths?

vga91: Of course, I added the test case example in usage doc page

fbiville (approved):

PR #2098, opened 2021-07-20
Fixes #1691: Document (and list) which APOC procs start their own transaction

### LINKED ISSUE #1691: Document (and list) which APOC procs start their own transaction
When recently trying to make use of `apoc.uuid.install` inside a transaction, it caused the client connection to the database to hang (no error message).

image

As described above, @jexp was able to inform me a couple of days later that the procedure starts its own transaction.

It would be a very good idea for procedures which start their own transaction to be documented as such, and a list of all procs which start a transaction to be collected in the APOC docs, too.

## Considered alternatives
N/A

By making it easier to understand a crucial under-the-hood detail that only maintainers or those familiar with the APOC source code would currently know about.

vga91:
Fixes #1691

To extract these, 
I found the classes that call `Transaction.java` class through a `beginTx()`
and from these I obtained the procedures / functions from which they were used.

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/transaction/index.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/transaction/index.adoc
@@ -1,0 +1,34 @@
+[[transaction]]
+= List of procedures with its own transaction
+:description: This chapter describes a list of procedures with its own transaction in the APOC library.
+
+The list of procedures that start their own transaction:
+
+* apoc.trigger.add
+* apoc.trigger.remove
+* apoc.trigger.removeAll
+* apoc.trigger.pause
+* apoc.trigger.resume
+* apoc.uuid.install
+* apoc.uuid.remove
+* apoc.uuid.removeAll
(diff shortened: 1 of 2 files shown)


docs/asciidoc/modules/ROOT/pages/transaction/index.adoc
```diff
@@ -0,0 +1,35 @@
+[[transaction]]
+= List of procedures with its own transaction
+:description: This chapter describes a list of procedures with its own transaction in the APOC library.
```
fbiville, line 3:
```suggestion
:description: This chapter describes the list of procedures that start their own transaction in the APOC library.
```

fbiville (approved):
shall we file another issue to try to automate that list? WDYT?

vga91:
> shall we file another issue to try to automate that list? WDYT?

Yes, makes sense. I just created it (#2148)

JMHReif (approved):

PR #2107, opened 2021-07-20
Introduce annotation processor to list existing Neo4j extensions

conker84:
* Move `apoc.data.email` to APOC full

`apoc.data.email` relies on a third-party dependency for any
JDK >= 9 (which is the baseline of Neo4j 4.x and APOC 4.x).

* Add missing @Extended annotation to full procedures

* Temporarily ignore flaky `UtilIT`


conker84:
@fbiville can we merge this?

PR #2108, opened 2021-07-20
Fixes #2025: Check transaction guard in apoc.import.json

### LINKED ISSUE #2025: Check transaction guard in `apoc.import.json`
Try with this in order to reproduce it
```
create constraint on (g:Stream) assert g.neo4jImportId is unique;
create constraint on (g:User) assert g.neo4jImportId is unique;
create constraint on (g:Team) assert g.neo4jImportId is unique;
create constraint on (g:Language) assert g.neo4jImportId is unique;
create constraint on (g:Game) assert g.neo4jImportId is unique;
​
/* alternatively this could be a bit faster
create index on :Stream(neo4jImportId);
create index on :Language(neo4jImportId);
create index on :Team(neo4jImportId);
create index on :Game(neo4jImportId);
create index on :User(neo4jImportId);
*/
​
call dbms.setConfigValue('dbms.transaction.timeout','0');
​
CALL apoc.import.json("https://devrel-data-science.s3.us-east-2.amazonaws.com/twitch_all.json", 
  {txBatchSize:100000, unwindBatchSize:10000});
​
call dbms.listQueries() yield queryId, query, elapsedTimeMillis;
```

Check constraints, at least it should output the necessary constraints and abort if they don't exist

vga91:
Fixes #2025

- added `@Context TerminationGuard`
- added `checkConstraints` to verify that the all labels have a constraint with  property defined by `importIdName` config

To be noted that it is not possible to create constraints directly in the import process 
because the transaction cannot write data and modify the DB schema at once.


core/src/main/java/apoc/export/json/JsonImporter.java
```diff
@@ -130,11 +141,28 @@ private void manageNode(Map<String, Object> param) {
...
     }
 
+    private void checkConstraints(List<String> labels) {
+        if (labels.isEmpty()) {
+            return;
+        }
+        try (final Transaction tx = db.beginTx()) {
+            final Schema schema = tx.schema();
+            final String importIdName = importJsonConfig.getImportIdName();
+            final String label = labels.get(0);
```
JMHReif, line 157: Why are we only looking at the first label?

core/src/main/java/apoc/export/json/JsonImporter.java
```diff
@@ -130,11 +141,28 @@ private void manageNode(Map<String, Object> param) {
...
         }
     }
 
+    private void checkConstraints(List<String> labels) {
+        if (labels.isEmpty()) {
+            return;
+        }
+        try (final Transaction tx = db.beginTx()) {
+            final Schema schema = tx.schema();
+            final String importIdName = importJsonConfig.getImportIdName();
```
JMHReif, line 156: what does this config do? We looked at the doc, but the description isn't super clear. https://neo4j.com/labs/apoc/4.2/overview/apoc.import/apoc.import.json/#_config_parameters

core/src/test/java/apoc/export/json/ImportJsonTest.java
```diff
@@ -165,15 +188,110 @@ public void shouldImportNodesWithoutLabels() throws Exception {
...
+                    .<Node>columnAs("n")
+                    .next();
+            Assert.assertNotNull("node should be not null", node);
+            final double[] actual = (double[]) node.getProperty("bbox");
+            Assert.assertArrayEquals(expected, actual, 0.05D);
+        }
+    }
+    
+    @Test
+    @Ignore("It should only be execute offline")
```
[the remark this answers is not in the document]
vga91: Of course, now I've changed it with a `call dbms.killQuery(...)`

core/src/test/java/apoc/export/json/ImportJsonTest.java
```diff
@@ -165,15 +188,110 @@ public void shouldImportNodesWithoutLabels() throws Exception {
...
+            Map<String, Object> first = res.next();
+            assertEquals("call dbms.listQueries", first.get("query"));
+            assertFalse(res.hasNext());
+        });
+
+        restartDb(Duration.ZERO);
+    }
+
+    @Test
+    public void shouldImportAllNodesAndRels() {
```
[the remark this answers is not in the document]
vga91: This test is used to verify that the import is done only if there are all the constraints.
In effect, the name `big.json` is confusing, now I changed it.

core/src/main/java/apoc/export/json/JsonImporter.java
vga91: Sorry my bad. Effectively it makes much more sense to check all the labels. Changed.

core/src/main/java/apoc/export/json/JsonImporter.java
vga91: This is the name of the property to be populated with the "id" field present into the json. 
For example `{"type":"node", "labels":["Language"], "id":"10"}`, with importIdName:`foo`, will create a `(:User {foo: "10"})`.
(I changed the documentation as well)

core/src/test/java/apoc/export/json/ImportJsonTest.java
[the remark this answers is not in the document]
JMHReif: Do we still need to ignore it then? Since we kill the query, then the test shouldn't take long, right? :)

JMHReif (approved):

PR #2150, opened 2021-08-06
generating adoc files for aura docs integration

conker84:
We now generate 2 files one per each type:
- `core.adoc` for core procedure/functions
- `full.adoc` for full procedure/functions


conker84:
this is not necessary anymore it'll manage via Aura side

PR #2157, opened 2021-08-06
Add virtual relationship check

JMHReif:
Signed-off-by: Florent Biville <florent.biville@neo4j.com>


JMHReif:
@conker84, should we update the equalsHashCode of virtual nodes and virtual relationships? Does the compare work when dealing with virtual nodes/rels?

conker84:
@JMHReif @fbiville I have a proposal for this, why we don't get rid of VirtualPath and use the `PathImpl` class? So we can delegate any kind of business logic to it? I did a quick research and seems that `VirtualPath` is used only one point so the change should be very quick

core/src/main/java/apoc/result/VirtualPath.java
```diff
@@ -113,9 +113,30 @@ public void remove() {
...
+		Node previousStartNode;
+		Node previousEndNode;
+		Relationship previousRelationship = lastRelationship();
+		if (previousRelationship != null) {
+			previousStartNode = previousRelationship.getStartNode();
+			previousEndNode = previousRelationship.getEndNode();
+		} else {
+			previousStartNode = startNode();
+			previousEndNode = endNode();
+		}
```
conker84, line 132: Can you please fix the indentation?

conker84 (approved):

conker84:
@JMHReif can you please rebase it

PR #2195, opened 2021-09-02
Trigger replication failing

### LINKED ISSUE #2193: Trigger Replication Problem
Pasted form internal trello card

Neo4j Enterprise 4.2.9
APOC 4.2.0.2

Customer reported and I have since tested what seems to be a failure to replicate triggers created via apoc.trigger.add(). I’m aware of a bug in apoc that throws a server no longer accepts writes error unless you connect directly via bolt:// to the System db leader and add the trigger there (related cards and GH issue attached). But now, that trigger doesn’t appear anywhere else on the cluster, apart from that one core that I created it on. Is raft somehow being selective in what transactions its replicating across to members??  Since if I repeat the exercise with a simple create/merge node (also having connected directly via bolt to a user db) the merged node appears on all cores straightaway. I don’t think this is down to replication of updates in the system db, since other updates to system are replicated just fine. 

So the question is: What could possibly be preventing that trigger from appearing on nodes other than the one it was directly created on (system leader) using bolt://??

Interestingly, the trigger appears just fine on all cores, on a 3.5.29 cluster, which adds to my suspicion about how the System db updates in 4.x are being replicated, differently, to how graph.db updates were handled in 3.5.x

Attaching logs and relevant files from both test clusters (4.2.9 and 3.5.29). Please let me know if any additional details are required.

Repro steps:

1- connect via neo4j:// to a 4.x cluster (any 4.x version). Ensure beforehand to have the appropriate apoc plugin under /plugins directory and to whitelist apoc as follows in neo4j.conf:

```
apoc.trigger.enabled=true
dbms.security.procedures.whitelist=apoc.*
dbms.security.procedures.unrestricted=apoc.*
```

2- run `CALL apoc.trigger.add("trigger1", "MERGE (p:Person {name: 'invalid'})",{phase:'before'})`

unless one connects to instance that just happens to be the system db leader at the time, the following error will be thrown:

`Server at mydomainl:7687 no longer accepts writes`

This is a known bug with apoc, reported on the attached cards/GH.

3- CALL dbms.cluster.overview() and identify leader for the system db. Then connect via bolt:// directly (either via browser or cypher-shell) to that instance.

4- run `CALL apoc.trigger.add("trigger1", "MERGE (p:Person {name: 'invalid'})",{phase:'before'})`. This succeeds and one can view trigger1 as a listed trigger via call apoc.trigger.list().

5- Connect to any other core via bolt:// that is NOT the leader for system db and execute `call apoc.trigger.run`. Result: Trigger1 is NOT listed there, or on any other cluster members. This appears just fine on a 3.5.x cluster, but not in 4.x.

Thanks

conker84:
Fixes #2193

Fixes the replication problem 

  - Fixed the bug
  - Improved docs
  - Added tests in cluster env

### DIFF
--- a/core/src/main/java/apoc/trigger/TriggerHandler.java
+++ b/core/src/main/java/apoc/trigger/TriggerHandler.java
@@ -21,19 +21,27 @@ import org.neo4j.kernel.api.procedure.Context;
 import org.neo4j.kernel.api.procedure.GlobalProcedures;
 import org.neo4j.kernel.lifecycle.LifecycleAdapter;
 import org.neo4j.logging.Log;
+import org.neo4j.scheduler.Group;
+import org.neo4j.scheduler.JobHandle;
+import org.neo4j.scheduler.JobScheduler;
 
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static apoc.ApocConfig.APOC_TRIGGER_ENABLED;
+import static apoc.ApocConfig.apocConfig;
 
 public class TriggerHandler extends LifecycleAdapter implements TransactionEventListener<Void> {
 
     private enum Phase {before, after, rollback, afterAsync}
+
+    public static final String TRIGGER_REFRESH = "apoc.trigger.refresh";
 
     private final ConcurrentHashMap<String, Map<String,Object>> activeTriggers = new ConcurrentHashMap();
     private final Log log;
@@ -41,6 +49,11 @@ public class TriggerHandler extends LifecycleAdapter implements TransactionEvent
     private final DatabaseManagementService databaseManagementService;
     private final ApocConfig apocConfig;
     private final Pools pools;
+    private final JobScheduler jobScheduler;
+
+    private long lastUpdate;
+
+    private JobHandle restoreTriggerHandler;
 
     private final AtomicBoolean registeredWithKernel = new AtomicBoolean(false);
 
@@ -50,13 +63,14 @@ public class TriggerHandler extends LifecycleAdapter implements TransactionEvent
 
     public TriggerHandler(GraphDatabaseService db, DatabaseManagementService databaseManagementService,
                           ApocConfig apocConfig, Log log, GlobalProcedures globalProceduresRegistry,
-                          Pools pools) {
+                          Pools pools, JobScheduler jobScheduler) {
         this.db = db;
         this.databaseManagementService = databaseManagementService;
         this.apocConfig = apocConfig;
         this.log = log;
         transactionComponentFunction = globalProceduresRegistry.lookupComponentProvider(Transaction.class, true);
         this.pools = pools;
+        this.jobScheduler = jobScheduler;
     }
 
     private boolean isEnabled() {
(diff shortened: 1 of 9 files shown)


core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -138,26 +157,28 @@ private synchronized void reconcileKernelRegistration() {
...
+        withSystemDb(tx -> {
             tx.findNodes(SystemLabels.ApocTrigger,
-                    SystemPropertyKeys.database.name(), db.databaseName(),
-                    SystemPropertyKeys.name.name(), name)
+                            SystemPropertyKeys.database.name(), db.databaseName(),
+                            SystemPropertyKeys.name.name(), name)
                     .forEachRemaining(node -> node.setProperty(SystemPropertyKeys.paused.name(), paused));
-            tx.commit();
-        }
+            setLastUpdate(tx);
```
fbiville, line 179: why not call `setLastUpdate` in `withSystemDb` directly?

conker84: becuse we use `withSystemDb` also for reading ops like here:

https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/c1b34daf825933881ca91691dcb2752d3b914e20/core/src/main/java/apoc/trigger/TriggerHandler.java#L91-L105

conker84:
@fbiville do another check I fixed a broken test

docs/asciidoc/modules/ROOT/partials/triggers.adoc
```diff
@@ -5,4 +5,13 @@ We can enable them by setting the following property in `apoc.conf`:
...
\ No newline at end of file
+apoc.trigger.refresh=60000
+----
+
+.Description
+[opts=header,cols="3"]
+|===
+| Option Key | Value | Description
+| apoc.trigger.enabled | true/false, default false | Enable/Disable the feature
+| apoc.trigger.refresh | number, default 60000 | Value in ms for trigger replication check across cluster nodes
```
fbiville, line 16:
```suggestion
| apoc.trigger.refresh | number, default 60000 | Interval in ms after which a replication check is triggered across all cluster nodes
```

full/src/test/java/apoc/trigger/TriggerClusterTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @Before
+    public void before() {
+        cluster.getSession().run("CALL apoc.trigger.removeAll()");
+        cluster.getSession().run("MATCH (n) DETACH DELETE n");
+    }
+
+
+    @Test
+    public void testTimeStampTriggerForUpdatedProperties() throws Exception {
+        cluster.getSession().run("CALL apoc.trigger.add('timestamp','UNWIND apoc.trigger.nodesByLabel($assignedNodeProperties,null) AS n SET n.ts = timestamp()',{})");
```
fbiville, line 53: wouldn't `$assignedLabels` make more sense here (and below)?

fbiville: Also, maybe for another PR, the short documentation of apoc.trigger.add still shows the old parameter syntax (https://neo4j.com/labs/apoc/4.0/overview/apoc.trigger/apoc.trigger.nodesByLabel/)

full/src/test/java/apoc/trigger/TriggerClusterTest.java
```diff
@@ -0,0 +1,109 @@
...
+        TestContainerUtil.testCall(cluster.getSession(), "MATCH (f:Foo) RETURN f", (row) -> {
+            assertEquals(true, ((Node)row.get("f")).containsKey("ts"));
+        });
+    }
+
+    @Test
+    public void testReplication() throws Exception {
+        cluster.getSession().run("CALL apoc.trigger.add('timestamp','UNWIND apoc.trigger.nodesByLabel($assignedNodeProperties,null) AS n SET n.ts = timestamp()',{})");
+
+        Thread.sleep(1000);
```
fbiville, line 64: do we have utilities such as `assertEventually` or something like that?

fbiville (approved):

full/src/test/java/apoc/trigger/TriggerClusterTest.java
conker84: done

full/src/test/java/apoc/trigger/TriggerClusterTest.java
conker84: oki

docs/asciidoc/modules/ROOT/partials/triggers.adoc
conker84: done

fbiville (approved):

full/src/test/java/apoc/trigger/TriggerClusterTest.java
fbiville: do you still need the call to Thread.sleep, then?

fbiville (approved):

full/src/test/java/apoc/trigger/TriggerClusterTest.java
conker84: I think that you're looking at the old code because I removed it

PR #2201, opened 2021-09-10
fixes #2200: apoc.export allows writes anywhere in the host filesystem

### LINKED ISSUE #2200: apoc.export allows writes anywhere in the host filesystem
`apoc.export.*` allows writes anywhere in the host filesystem, we want to restrict these situations:

```
"apoc.export.csv.all"
"apoc.export.csv.data"
"apoc.export.csv.graph"
"apoc.export.csv.query"
"apoc.export.cypher.all"
"apoc.export.cypher.data"
"apoc.export.cypher.graph"
"apoc.export.cypher.query"
"apoc.export.cypher.schema"
"apoc.export.cypherAll"
"apoc.export.graphml.all"
"apoc.export.graphml.data"
"apoc.export.graphml.graph"
"apoc.export.graphml.query"
"apoc.export.json.all"
"apoc.export.json.data"
"apoc.export.json.graph"
"apoc.export.json.query"
```

conker84:
Fixes #2200

This PR restrict the cases when procedures are allowed to write data into the file system

  - Fixed the bug
  - Added tests

### DIFF
--- a/core/src/test/java/apoc/export/ExportCoreSecurityTest.java
+++ b/core/src/test/java/apoc/export/ExportCoreSecurityTest.java
@@ -1,6 +1,7 @@
 package apoc.export;
 
 import apoc.ApocConfig;
+import apoc.ApocSettings;
 import apoc.export.csv.ExportCSV;
 import apoc.export.cypher.ExportCypher;
 import apoc.export.graphml.ExportGraphML;
@@ -42,7 +43,8 @@ public class ExportCoreSecurityTest {
 
     @ClassRule
     public static DbmsRule db = new ImpermanentDbmsRule()
-            .withSetting(GraphDatabaseSettings.load_csv_file_url_root, directory.toPath().toAbsolutePath());
+            .withSetting(GraphDatabaseSettings.load_csv_file_url_root, directory.toPath().toAbsolutePath())
+            .withSetting(ApocSettings.apoc_export_file_enabled, false);
 
     @BeforeClass
     public static void setUp() throws Exception {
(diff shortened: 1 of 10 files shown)


core/src/test/java/apoc/export/ExportCoreSecurityTest.java
```diff
@@ -0,0 +1,189 @@
...
+public class ExportCoreSecurityTest {
+
+    private static File directory = new File("target/import");
+
+    static {
+        directory.mkdirs();
+    }
+
+    @ClassRule
+    public static DbmsRule db = new ImpermanentDbmsRule()
```
fbiville, line 44: maybe worth adding `apoc.export.file.enabled=false` just to be explicit?

core/src/test/java/apoc/export/ExportCoreSecurityTest.java
```diff
@@ -0,0 +1,189 @@
...
+            fail(message);
+        } catch (Exception e) {
+            assertExportError(e);
+        }
+        try {
+            db.executeTransactionally("CALL apoc.export." + exportMethod + ".query(\"RETURN 'hello' as key\", './hello', {stream:true})", Map.of(),
+                    Result::resultAsString);
+            fail(message);
+        } catch (Exception e) {
+            assertExportError(e);
```
fbiville, line 79: has something changed in this PR that makes the export fail? initial (internal) issue report stated that this export would (wrongly) succeed and we don't see anything in `checkWriteAllowed` that would prevent the export when `stream` is `true`

conker84: This case should be allowed only if `apoc.export.file.enabled=true` but in those test cases we only test `apoc.export.file.enabled=false`

fbiville (approved):

JMHReif (approved):

PR #2259, opened 2021-10-25
Fixes #2037: apoc.export.cypher.all to generate constraints with IF NOT EXISTS

### LINKED ISSUE #2037: apoc.export.cypher.all to generate constraints with IF NOT EXISTS
Add an option (or make default) the generation of IF NOT EXISTS when outputting the cypher to generate the constraints.
Currently if the loading of the cypher file is interrupted it requires to remove the constraints to be running without error.

## Considered alternatives
N/A

This allows to re-run a generated cypher (in case it did not complete) specially if MERGE is the strategy used.

vga91:
Fixes neo4j-contrib#2037
Fixes neo4j-contrib#2073

Changed index creation with new syntax

### DIFF
--- a/core/src/main/java/apoc/export/util/ExportConfig.java
+++ b/core/src/main/java/apoc/export/util/ExportConfig.java
@@ -24,6 +24,7 @@ public class ExportConfig {
     public static final String DEFAULT_ARRAY_DELIM = ";";
     public static final String DEFAULT_QUOTES = ALWAYS_QUOTES;
     private final boolean streamStatements;
+    private final boolean ifNotExists;
 
     private int batchSize;
     private boolean silent;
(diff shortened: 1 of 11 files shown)


core/src/main/java/apoc/export/util/ExportConfig.java
```diff
@@ -97,6 +98,7 @@ public ExportConfig(Map<String,Object> config) {
         this.config = config;
         this.streamStatements = toBoolean(config.get("streamStatements")) || toBoolean(config.get("stream"));
         this.writeNodeProperties = toBoolean(config.get("writeNodeProperties"));
+        this.ifNotExists = toBoolean(config.get("ifNotExists"));
```
jexp, line 101: should the default be true? so it's more idempotent

jexp: and then folks that have older versions can disable it.
or we introduce it here with default false and then change it in the next release (4.4)?

jexp (approved):

havardthom:
Can this also be merged to 4.4?

conker84:
@havardthom done: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2512

PR #2262, opened 2021-10-25
Fixes #2015: Various bugs with Neo4j 4.3 indexes

### LINKED ISSUE #2015: Various bugs with Neo4j 4.3 indexes
With relationship and lookup indexes, there are some problems:

### Steps (Mandatory)

1. `apoc.meta.data` the relations give `indexed = false` even if there are indexes
2. `apoc.meta.schema` there isn't the `indexed` field with relationships
3. `apoc.schema.nodes` displays both indexes on nodes and relations, while `apoc.schema.relationships` does not display indexes
4. `apoc.schema.assert` does not delete rel indexes
5. The `LOOKUP` indexes are never considered

### Versions

  - OS: Mac Big Sur
  - Neo4j: 4.3
  - Neo4j-Apoc: 4.3.0.0

vga91:
Fixes #2015

- Implemented `getIndexes(RelationshipType type)` in `SubGraph.java`
- Added `relIndexes` in `Meta.java`
- Added lookup index handling

### DIFF
--- a/core/src/main/java/org/neo4j/cypher/export/CypherResultSubGraph.java
+++ b/core/src/main/java/org/neo4j/cypher/export/CypherResultSubGraph.java
@@ -206,6 +206,13 @@ public class CypherResultSubGraph implements SubGraph
     }
 
     @Override
+    public Iterable<IndexDefinition> getIndexes(RelationshipType type) {
+        return indexes.stream()
+                .filter(idx -> StreamSupport.stream(idx.getRelationshipTypes().spliterator(), false).anyMatch(lb -> lb.equals(type)))
+                .collect(Collectors.toSet());
+    }
+
+    @Override
     public Iterable<RelationshipType> getAllRelationshipTypesInUse() {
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/neo4j/cypher/export/CypherResultSubGraph.java
```diff
@@ -205,6 +205,13 @@ public boolean contains( Relationship relationship )
                 .collect(Collectors.toSet());
     }
 
+    @Override
+    public Iterable<IndexDefinition> getIndexes(RelationshipType type) {
+        return indexes.stream()
+                .filter(idx -> StreamSupport.stream(idx.getRelationshipTypes().spliterator(), false).anyMatch(lb -> lb.equals(type)))
```
jexp, line 211: should probably do equals on the name() to be sure, because RelationshipType is an interface so there is no symmetric equals

jexp (approved):
Great PR! Thanks a lot

PR #2272, opened 2021-10-28
Fixes #2093: Add check of the number of connections in couchbase test

### LINKED ISSUE #2093: Add check of the number of connections in couchbase test
Add check of the number of connections in `CouchbaseIT.java` test, similarly to `MongoDbTest.java`.

vga91:
Fixes #2093

### DIFF
--- a/full/src/test/java/apoc/couchbase/CouchbaseIT.java
+++ b/full/src/test/java/apoc/couchbase/CouchbaseIT.java
@@ -7,8 +7,10 @@ import com.couchbase.client.core.error.UnambiguousTimeoutException;
 import com.couchbase.client.java.codec.RawBinaryTranscoder;
 import com.couchbase.client.java.json.JsonObject;
 import com.couchbase.client.java.kv.InsertOptions;
+import org.junit.After;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.junit.AfterClass;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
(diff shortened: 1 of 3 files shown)


full/src/test/java/apoc/couchbase/CouchbaseIT.java
```diff
@@ -49,6 +53,18 @@ public static void tearDown() {
...
+    @Before
+    public void before() {
+        numberConnections = getNumConnections();
+    }
+
+    @After
+    public void after() {
+        // the connections active before must be equal to the connections active after
+        long numConnectionsAfter = getNumConnections();
+        assertEquals(numberConnections, numConnectionsAfter);
```
jexp, line 66: what happens if not and what would be the reason for not?

jexp (approved):
Looks good but I don't fully understand the why

full/src/test/java/apoc/couchbase/CouchbaseIT.java
vga91: For example if I don't wrap correctly the procedure in a `try-res`, and so I don't call the `apoc.java.CouchbaseConnection.close()` method to disconnect the `Cluster` and shutdown the `ClusterEnvironment`, 
I receive a `numberConnections` equals to `numConnectionsAfter + 1`.
So, this assert is useful to verify that you have put the `try-res` wherever it's needed

conker84:
As we discussed in the last meeting, is for checking that we close any connection to couchbase properly as we did for the mongo procedures

PR #2274, opened 2021-11-02
Fixes #2225: Missing documentation for the input parameter of apoc.meta.schema

### LINKED ISSUE #2225: Missing documentation for the input parameter of apoc.meta.schema()
The documentation is missing for the input parameter of apoc.meta.schema, it should be the same as for apoc.meta.graph/subgraph

vga91:
Fixes #2225

Only `sample` config. Other `MetaConfig` are not used.

### DIFF
--- a/docs/asciidoc/modules/ROOT/partials/usage/apoc.meta.schema.adoc
+++ b/docs/asciidoc/modules/ROOT/partials/usage/apoc.meta.schema.adoc
@@ -46,19 +46,15 @@ RETURN key, value[key] AS value;
 |===
 
 
+Because the count stores return an incomplete picture of the data, we have to cross check the results with the actual data to filter out false positives.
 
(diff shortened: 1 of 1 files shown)


jexp:
I think this update misses the point,
the question was about: what does sample do, not that we can pass it in.

so how do we treat it

- percentage or amounts of data we look at
- how do we pick the sample of the data we look at (don't take the first SAMPLE elements but the first of each bach of total/SAMPLE)
- why do we have to look at the actual data in the first place (because the count stores return an incomplete picture of the data and just a set-union of the results needs to be cross checked againts actual data to filter out false positives)

jexp (changes requested):
see my comment, the actual PR change is not helpful about the why

jexp (approved):

PR #2276, opened 2021-11-03
Fixes #494: add a sigmoid UDF

### LINKED ISSUE #494: add a sigmoid UDF
`sigmoid(x) = return 1.0 / (1.0 + exp(-x) )`

vga91:
Fixes #494

- Added functions `sigmoid`, `sigmoidPrime` and hyperbolic `tangent`, `cotangent`, `sin`, `cosin`, `secant` and `cosecant`


core/src/main/java/apoc/math/Maths.java
```diff
@@ -69,5 +69,60 @@ public Long minByte(){
         return Long.valueOf(Byte.MIN_VALUE);
     }
 
+    @UserFunction
```
jexp, line 72: not sure if we should NaN (e.g division by zero) just bubble up or convert them to null?

jexp (changes requested):
please don't check in all those generated documentation pages.
but instead add a docs page why we added these functions (for ML purposes)
and please have a look at my NaN comment

vga91:
> please don't check in all those generated documentation pages. but instead add a docs page why we added these functions (for ML purposes) and please have a look at my NaN comment

@jexp 
Ok, I added a page with all these functions `mathematical/sigmoid-hyperbolic-operations.adoc`. 
Sorry but i don't understand if you want to get rid off all `overview/apoc.math/apoc.math.*.adoc`, because they are invoked in`docs/asciidoc/modules/ROOT/partials/generated-documentation/nav.adoc` for example. 
Or maybe in this page, should I add the link to `sigmoid-hyperbolic-operations.adoc` instead of `overview/apoc.math/apoc.math.*.adoc`?

Regarding the `NaN` thing, a division by zero should never happen in `sigmoid` function because `Math.exp(-value)` can't be less than 0, or am I wrong? Anyway, personally I should retrieve a `NaN`.

core/src/main/java/apoc/math/Maths.java
conker84: @jexp but in this case the solution for `1.0 + Math.exp(-x)=0` is non-real, how should we manage it?

PR #2277, opened 2021-11-03
Fixes #577: return apoc.nodes.count(['User'])

### LINKED ISSUE #577: return apoc.nodes.count(['User'])
Can we get a count of all nodes from the stats store? As we cannot currently do:
`RETURN count((:User)) AS uc, count((:User)-[:POLICY_HOLDER]->()) / count((:User)) as avg_policies`

vga91:
Fixes #577


docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.meta.nodes-lite.csv
```diff
@@ -0,0 +1,2 @@
+¦signature
+¦apoc.meta.nodes.count(nodes = [] :: LIST? OF STRING?, config = {} :: MAP?) :: (INTEGER?)
```
jexp, line 2: aren't these files generated and should not be checked in? everything in generated documentation.

docs/asciidoc/modules/ROOT/partials/usage/apoc.meta.nodes.count.adoc
```diff
@@ -0,0 +1,66 @@
...
+We can return all nodes with a label `MyCountLabel` or a label `ThirdLabel`
+
+[source,cypher]
+----
+RETURN apoc.meta.nodes.count(['MyCountLabel', 'ThirdLabel']) AS count;
+----
+
+.Results
+[opts="header"]
+|===
```
jexp, line 19: I don't think the implementation is what I expected it to be, I expected one result (row) per label

jexp (changes requested):
I don't think the implementation is what I expected it to be, I expected one result (row) per label

vga91:
> I don't think the implementation is what I expected it to be, I expected one result (row) per label

@jexp Sorry, I thought I answered 😅
I changed the implementation (1 row per label) and deleted some extra auto-generated files.

PR #2335, opened 2021-11-17
MultiThreadedGlobalGraphOperations is storage-agnostic

tinwelint:
Storage engine agnostic MultiThreadedGlobalGraphOperations implementation

  - MultiThreadedGlobalGraphOperations was accessing record storage specific classes and methods, which won't work with other storage engines going forward. Changed it to use the cursor scan concept and tries to keep the same multi-threaded functionality of it.


jexp (approved):
Looks good, thanks a lot

PR #2379, opened 2021-11-27
neo4j-contrib#2378 Support for UUID encoded in Base64

### LINKED ISSUE #2378: Support for UUID encoded in Base64 to reduce significantly indices size
## Feature description
APOC uses standard text format of UUID identifiers which is a hex representation (e.g. `290d6cba-ce94-455e-b59f-029cf1e395c5`). The problem is that such format use twice more space than the binary identifier itself - it consumes 32 characters while UUID is 16 bytes long (128 bits). As a result, the size of indices in Neo4j are much bigger than they could be.

## Considered alternatives
The proposed solution is to allow to use Base64 encoded UUID (e.g. `KQ1sus6URV61nwKc8eOVxQ`). Without loosing on a value diversity, it produces 22 characters long identifiers (not including two alignment characters i.e. ==). 

We can provide a function that allows to generate Base64 encoded identifiers. It would be valuable to provide two other functions that allow two ways migration between both formats.

Just one small note, for clarity, is that the Base64 conversion should be performed on UUID source buffer, i.e. 16 bytes, not on the generated HEX presentation.

My experiment shows that using Base64 encoded UUID size of indices is reduced by 24-34%. The benefit is that it decreases significantly size of occupied memory by indices. Then such space can be used by other cached data or just reduce the required amount of memory. My blog article about this improvement (unfortunately in Polish: https://appga.pl/2021/11/23/funkcja-warta-8-tysiecy-dolarow/)

grzegorz-aniol:
Fixes #2378

Support for UUID encoded with Base64

## Proposed Changes

1. New functions:
- adding new function `apoc.create.uuidBase64()` - create an UUID encoded with Base64
- adding new function `apoc.create.uuidBase64ToHex()` - convert an UUID between Base64 and HEX format (useful in case of ids migration)
- adding new function `apoc.create.uuidHexToBase64()` - convert an UUID between HEX and Base64 format (useful in case of ids migration)

2. Support for automatic UUID generation (transaction handler) 
- introducing new configuration option `apoc.uuid.format` with two possible options allowed:
    - `hex` (default value) - generate well know UUID in HEX format (32 characters long)
    - `base64` - generate UUID in Base64


tomasonjo:
Updating the documentation will help in merging this PR

conker84:
Hi @grzegorz-aniol and @tomasonjo, why simply just combine `apoc.create.uuid()` and `apoc.text.base64Encode/hexValue`?
I mean maybe we can combine these three procedures in order to achieve the same result?
ie. `RETURN apoc.text.base64Encode(apoc.create.uuid())` is the same as `REUTRN apoc.create.uuidBase64()`
and if you need a shortcut maybe you can define your own cypher function `custom.uuidBase64()` that under the hood:
`CALL apoc.custom.declareFunction('uuidBase64() :: STRING', 'RETURN apoc.text.base64Encode(apoc.create.uuid())')`

Wdyt?

grzegorz-aniol:
Hi @conker84, thank you for your feedback.

Maybe I was not clear enough with my proposal. Let me clarify where is the benefit. 

Usually we identify UUID with a textual hex representation e.g. `290d6cba-ce94-455e-b59f-029cf1e395c5` which is 32 characters long. However UUID originally, as a byte array is 16 bytes long. This array value is so far available only in Java UUID object. 
The `apoc.create.uuid()` function returns 32 characters and encoding this to Base64 gives as 48 characters (`MjkwZDZjYmEtY2U5NC00NTVlLWI1OWYtMDI5Y2YxZTM5NWM1`) - so that's obviously is not a benefit. 
The idea is to convert directly 16 bytes array into Base64. This produce 22 characters long unique identifiers (e.g.  `KQ1sus6URV61nwKc8eOVxQ`). 
And here is the benefit. Key value can be ~30% shorter (and the same way the whole index) with keeping exactly the same diversity of values. As a result we may significantly reduce amount of RAM memory required by an index, which should be fully in RAM, right?
The perfect way would be to use 16 bytes of UUID already as an UUID key. For example PostgreSQL has native support for such values. But Neo4j doesn't have it (yet?).

JMHReif (approved):

JMHReif (approved):

PR #2382, opened 2021-12-01
distance has been replaced with point.distance

pontusmelke:
The `distance` function has been moved to `point.distance` in 5.0. This changes all usages to reflect the change.


jexp (approved):

jexp:
@conker84 we can merge this one.

PR #2412, opened 2022-01-03
Fixes #2239: Provide query statistics for apoc.merge functions

### LINKED ISSUE #2239: Provide query statistics for apoc.merge functions
## Feature description 

Currently `apoc.merge.node` and `apoc.merge.relationship` functions do not return any query statistics.  These functions are often used in `UNWIND $data ..` queries where the user wants the ability to dynamically create the node and/or relationship label. 

These functions generate and execute cypher statements from their inputs, so it should be possible to provide query statistics for them.

Related https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1815

## Considered alternatives

None that I am aware of

It will provide accurate statistics of the queries run.  The current statistics returned are 0 for every SummaryCounter field e.g.:

```
  "counters": {
    "_stats": {
      "nodesCreated": 0,
      "nodesDeleted": 0,
      "relationshipsCreated": 0,
      "relationshipsDeleted": 0,
      "propertiesSet": 0,
      "labelsAdded": 0,
      "labelsRemoved": 0,
      "indexesAdded": 0,
      "indexesRemoved": 0,
      "constraintsAdded": 0,
      "constraintsRemoved": 0
    },
    "_systemUpdates": 0
  },
  "updateStatistics": {
    "_stats": {
      "nodesCreated": 0,
      "nodesDeleted": 0,
      "relationshipsCreated": 0,
      "relationshipsDeleted": 0,
      "propertiesSet": 0,
      "labelsAdded": 0,
      "labelsRemoved": 0,
      "indexesAdded": 0,
      "indexesRemoved": 0,
      "constraintsAdded": 0,
      "constraintsRemoved": 0
    },
    "_systemUpdates": 0
  },
```

vga91:
Fixes #2239

- Added `merge` versions with `stats` column.

Other changes:
- changed wrong `apoc.merge.node.eager` with `apoc.merge.node`


core/src/main/java/apoc/merge/Merge.java
```diff
@@ -31,12 +35,37 @@
...
+                                          @Name("identProps") Map<String, Object> identProps,
+                                          @Name(value = "props",defaultValue = "{}") Map<String, Object> props,
+                                          @Name(value = "onMatchProps",defaultValue = "{}") Map<String, Object> onMatchProps) {
+        final Result nodeResult = getNodeResult(labelNames, identProps, props, onMatchProps);
+        return nodeResult.columnAs("n").stream().map(node -> new NodeResult((Node) node));
+    }
+
+    @Procedure(value="apoc.merge.nodeWithStats.eager", mode = Mode.WRITE, eager = true)
+    @Description("apoc.merge.nodeWithStats.eager - same as apoc.merge.node.eager providing queryStatistics into result")
+    public Stream<NodeResultWithStats> nodeWithStatsEager(@Name("label") List<String> labelNames,
```
JMHReif, line 49: Should we make the stats and eager as config we can set on the procedure? Or does that complicate/overload each proc and testing too much?

i.e. apoc.merge.node(['Label'], identProps:{key:value, ...}, onCreateProps:{key:value,...}, onMatchProps:{key:value,...}, config: {eager: true, stats: true})

JMHReif (approved):

core/src/main/java/apoc/merge/Merge.java
vga91: Regarding the `eager`, the procedure must start in a method with a`@Procedure(... eager = true)`.
Perhaps we could change somehow the annotation parameter at runtime, but I don't think it's worth it.

With `stats`, I should change the `apoc.merge.node` result to `Stream<NodeResultWithStats>`,
but in this case the `apoc.merge.node` (without start) will yield also a column `stats` with value null.
Maybe it is better to leave it unchanged?

JMHReif: Ah, ok. Never mind. Thanks for clarifying!

JMHReif (approved):

PR #2419, opened 2022-01-05
Fixes schema enterprise test 4.4

vga91:
Fixes schema enterprise test 4.4

- Added `filter(isRelConstraint)` and `filter(isNodeConstraint)` to avoid Exceptions in case of constraints of both types.
- Added `@Before` with drop schema, coherently with SchemasTest
- Changed `List` to `Set` in `testKeptNodeKeyAndUniqueConstraintIfExistsAndDropExistingIsFalse` test, not to consider order


conker84:
LGTM thanks @vga91

PR #2417, opened 2022-01-04
Remove usage of CoordinateReferenceSystem

parnmatt:
`CoordinateReferenceSystem` is not listed as `@PublicAPI`. This usage was found in an internal Neo4j failing build.
A pending internal PR is changing some constant names to fit with internal naming conventions. Rather than fix it to use the new (still internal) name; this change explicitly checks for the string, minimising direct use, and assumptions, of internals.

PR #2422, opened 2022-01-06
Fix CoordinateReferenceSystem naming

parnmatt:
`CoordinateReferenceSystem` has changed it's internal naming. #2417 fixed the internal Neo4j build error; however, it was masking these additional usages. I should have double checked if there were more usages when investigating previously.


parnmatt:
@Lojjs as name change has been merged into the internal dev branch; dev, and I presume a lot of other builds, are probably failing because of this oversight of mine.

Hopefully you'll see this otherwise trivial change, and freely merge it if you are happy with it.
Thanks.

---

I have grepped the repo, this _should_ be all the usages fixed; and in these cases, I believe it _does_ make sense to use the internal names (at least until we can figure out a better "public" way)

MishaDemianenko (approved):

PR #2467, opened 2022-01-18
Adds load json ssrf mitigation

ncordon:
Companion to https://github.com/neo-technology/neo4j/pull/14015 (although that one is for 5.0)

## What

Uses Neo4j internal option `unsupported.dbms.cypher_ip_blocklist`, which consists on a list of ip addresses in CDIR notation, i.e. ip + mask for subnets, e.g. `192.167.2.3/8, 192.234.3.123/16`. When using the `apoc.load.json(url)` command, the url will be looked up in a DNS resolver and we will fail if it is on the list of blocked ip addresses.

## Why

Users can try to load data from any url (including internal ones), which could expose files the user does not have permissions to see from outside.


build.gradle
```diff
@@ -109,7 +109,7 @@ subprojects {
 
 ext {
     // NB: due to version.json generation by parsing this file, the next line must not have any if/then/else logic
-    neo4jVersion = "4.4.1"
+    neo4jVersion = "4.4.4-SNAPSHOT"
```
ncordon, line 112: A new 4.4 version of the database needs to be released and substituted here.

ncordon:
[Configuration setting for IP block/allow list](https://trello.com/c/M7mJwCvA/2787-configuration-setting-for-ip-block-allow-list)

core/src/main/java/apoc/util/JsonUtil.java
```diff
@@ -78,20 +85,30 @@ private static Configuration getJsonPathConfig(List<String> options) {
...
-        return loadJson(urlOrBinary, headers, payload, path, failOnError, null, options);
+    public static Stream<Object> loadJson(Object urlOrBinary, Map<String,Object> headers, String payload, String path, boolean failOnError, List<String> options, ApocConfig apocConfig) {
+        return loadJson(urlOrBinary, headers, payload, path, failOnError, null, options, apocConfig);
     }
     
-    public static Stream<Object> loadJson(Object urlOrBinary, Map<String,Object> headers, String payload, String path, boolean failOnError, String compressionAlgo, List<String> options) {
+    public static Stream<Object> loadJson(Object urlOrBinary, Map<String,Object> headers, String payload, String path, boolean failOnError, String compressionAlgo, List<String> options, ApocConfig apocConfig) {
         try {
             if (urlOrBinary instanceof String) {
                 String url = (String) urlOrBinary;
```
jakewins, line 99: Shouldn't this check be done in `checkReadAllowed`, where the other access rules as processed? With the check here, you can still make calls that violate the blocklist using `apoc.load.xml`, `apoc.xml.import`, `apoc.load.xmlSimple`, `apoc.load.xls`, `apoc.cypher.runFiles` and so on. 

Similarly, we should process the blocklist in `checkWriteAllowed`, otherwise people can violate the blocklist by sending POST requests and other output network connections?

jexp: yes we should not pass the full apoc config through, either just what we need or better like Jake said handle it in the single place where we control the loading.

jexp: https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.3/core/src/main/java/apoc/util/FileUtils.java#L142-L241

jexp: Also you don't need to pass `apocConfig` through but use `ApocConfig.getConfig().methodCall()`

there are already some methods for checking, we can handle the address as part of that.
so you would rather do something like `ApocConfig#isIpAllowed(ip)` or `ApocConfig#isHostNameAllowed(hostname)`
need to make sure though to resolve the final hostname properly to ip-address (after all the http-redirects are handled).

https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.3/core/src/main/java/apoc/ApocConfig.java#L250-L262

build.gradle
conker84: @ncordon If we don't use a non snapshot version we aren't able to test the PR in GH can you please change it to 4.4.3?

conker84 (changes requested):
this should be not necessary

core/src/main/java/apoc/ApocConfig.java
```diff
@@ -218,6 +218,10 @@ private void addDbmsDirectoriesMetricsSettings() {
         }
     }
 
+    public Config getNeo4jConfig() {
+        return neo4jConfig;
+    }
+
```
conker84, line 224: this should be not necessary

core/src/main/java/apoc/util/JsonUtil.java
conker84: @ncordon please look at the comments above

conker84 (changes requested):

build.gradle
ncordon: The code we need is in 4.4.5, and to release 4.4.5 we are going to need this to be merged on red. The CI of the database will compile the APOC code overriding the version.

So even bumping this, don't expect the CI to pass.

After the database has been released, we can bump here to 4.4.5 and release a new version of APOC.

core/src/main/java/apoc/util/JsonUtil.java
ncordon: @jexp I have addressed the concern for redirects in a pending PR: https://github.com/neo-technology/neo4j/pull/14171. The logic for the blocking lives in the kernel of the database and here I am reusing the same logic. So once that's merged we should be safe with respect to redirects (I hope).

@conker84 I get the need for `checkReadAllowed`, but `checkWriteAllowed` works only with filenames, which this blocklist doesn't include, judging by its signature?

Do we have any situation where we can do a POST to a path in a remote server?

core/src/main/java/apoc/util/FileUtils.java
```diff
@@ -42,6 +42,7 @@
     public enum SupportedProtocols {
         http(true, null),
         https(true, null),
+        ftp(true, null),
```
ncordon, line 45: This change is for consistency with the database. At the moment we don't let import a csv from an ftp server, we consider the url a file instead and that's blocked in Aura for example.

Any concerns doing this?

core/src/main/java/apoc/Neo4jSettings.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        @Override
+        public Class<IPAddressString> getType()
+        {
+            return IPAddressString.class;
+        }
+    };
+
+    public static final Setting<List<IPAddressString>> cypher_ip_blocklist = newBuilder( CYPHER_IP_BLOCKLIST, listOf( CIDR_IP ), List.of() ).build();
```
ncordon, line 54: I have had to add this setting here to avoid binary incompatibilities with other versions of Neo4j 4.xx.yy. At the moment the setting lives in `neo4j.conf` (even if the apoc version is newer than the neo4j one). Should we move it to `apoc.conf`? The downside of doing that is that we would end up with a duplicated setting, which I was trying to avoid.

core/src/main/java/apoc/util/JsonUtil.java
ncordon: This has been addressed. 

With respect to `checkWriteAllowed`, we've deemed that change is not necessary as it only affects files. POST requests (load json ones for example) go through the `checkReadAllowed`.

core/src/main/java/apoc/util/FileUtils.java
ncordon: Unfortunately `ftps` is not recognised by java URL, so we need to remove it from here

conker84 (approved):
Thank you so much @ncordon

build.gradle
ncordon: This was done without the need to depend on 4.4.4 in the end

core/src/main/java/apoc/util/JsonUtil.java
jexp: write list also works with urls e.g. for s3 URLs

core/build.gradle
```diff
@@ -60,6 +60,7 @@ dependencies {
         exclude module: "snakeyaml"
     }
     compile group: 'org.yaml', name: 'snakeyaml', version: '1.26'
+    compile group: 'com.github.seancfoley', name: 'ipaddress', version: '5.3.3'
```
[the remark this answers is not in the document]
conker84: @jexp My idea is to keep in 4.4.x so we can cover also older Neo4j versions, and remove it from 5.0

core/src/main/java/apoc/util/JsonUtil.java
ncordon: `checkWriteAllowed` is only used with things that require a physical file I believe:
checkWriteAllowed

PR #2492, opened 2022-02-01
fixes #2372: apoc.any.isDeleted(node/rel)

### LINKED ISSUE #2372: apoc.any.isDeleted(node/rel)
For situations where a node or relationship is already deleted it would be good to be able to check that before acessing it and causing an error.

There used to be a method on Node/Relationship for that if not anymore there is a KernelAPI method.

danielemansillo:
Fixes neo4j-contrib#2372

Added method to fix the issue and add the new functionality

Add a function (apoc.any.isDeleted) to fullfill the request

Added method which executes a query to check if the Node/Rel is still present;
Added test method which checks that the apoc before the node/rel is removed returns false and after the remove returns true;
To not allow the user to use the method improperly I added a runtime IllegalArgumentException to check that the input type is correct.


JMHReif (approved):

JMHReif:
@danielemansillo, could you sign our CLA (if you haven't already)? https://neo4j.com/developer/cla/

PR #2493, opened 2022-02-02
Fixes #1617: Logging for apoc.periodic.*

### LINKED ISSUE #1617: Logging for apoc.periodic.*

vga91:
Fixes #1617 (deleted in 4.1)

Fix adoc `<<< HEAD`

### DIFF
--- a/core/src/main/java/apoc/periodic/Periodic.java
+++ b/core/src/main/java/apoc/periodic/Periodic.java
@@ -82,7 +82,10 @@ public class Periodic {
         Map<String,Long> commitErrors = new ConcurrentHashMap<>();
         AtomicInteger failedBatches = new AtomicInteger();
         Map<String,Long> batchErrors = new ConcurrentHashMap<>();
-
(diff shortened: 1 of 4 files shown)


magaton:
Hello @vga91 ad @conker84 
Shouldn't :
```
log.info("Starting periodic iterate from `%s` operation using iteration `%s` in separate thread with id: `%s`", cypherIterate,cypherAction, periodicId);
```

be at `debug` level,
wrapped with `isDebugEnabled()`

I have lots of my own logging with `INFO` level inside triggers, and the `INFO` logging from batch is messing up my logs that I want to index into Elastic.

Thanks

Issue #1471, opened 2020-04-02
Support spatial and temporal datatypes in CSV import (e.g. point, datetime)


szarnyasg:
## Feature description

We should support spatial and date/datetime formats similarly to the [APOC CSV importer](https://neo4j.com/docs/labs/apoc/current/import/import-csv/).

## Considered alternatives

Users can currently use the offline [import tool](https://neo4j.com/docs/operations-manual/current/tools/import/) distributed as part of the `neo4j-admin` command-line tool to load CSV data with datetime fields.

## How this feature can improve the project?

This feature would be important to reach feature-parity with the offline import tool (regarding the supported syntax). However, I expect this to require a significant development effort as there are many types and cases to handle:

* `point`
* `date`
* `localtime`
* `time`
* `localdatetime`
* `datetime`
* `duration`

The [CSV file header format](https://neo4j.com/docs/operations-manual/current/tools/import/file-header-format/) page summarizes these formats with references to the [temporal](https://neo4j.com/docs/cypher-manual/4.0/syntax/temporal/#cypher-temporal-instants) and [duration](https://neo4j.com/docs/cypher-manual/4.0/syntax/temporal/#cypher-temporal-durations) syntaxes.

Fortunately, there are some temporal and spatial functions in APOC that we could build on:
- https://neo4j.com/docs/labs/apoc/current/misc/spatial/
- https://neo4j.com/docs/labs/apoc/current/temporal/

What do you think @jexp?

PR #2505, opened 2022-02-03
Fix import tool URLs

szarnyasg:
Followup of issue #1471 and PR #2497 - I noticed that the links to the import tool documentation are broken so I fixed them.


conker84:
@szarnyasg thank you!

PR #2523, opened 2022-02-08
Apoc.schema.assert should drops only indexes not included in the 1st parameter

vga91:
Related Trello card: https://trello.com/c/rmXko1UQ/914-s4-enedis-hdr-project-behaviour-of-apocschemaassert-in-42

Drop existing should drops only OTHER indexes, consistent with the documentation and with constraint behaviour.
Instead currently drop also indexes included in the 1st parameter of `apoc.schema.assert()`, and subsequently recreates them.


jexp (approved):
Looks good to me.

PR #2525, opened 2022-02-08
Added slotted=runtime to apoc.cypher.runMany

vga91:
Added `slotted=runtime` to `apoc.cypher.runMany`.

Reused `slottedRuntime()` method present in `Periodic.java` (now moved in `Util.java`)

----

I'm not sure it's the best way to proceed,
I opened the pr to discuss it together.
Should it be done via config or is it okay to force `cypher runtime=slotted` as a prefix, if not present?

### DIFF
--- a/core/src/main/java/apoc/cypher/Cypher.java
+++ b/core/src/main/java/apoc/cypher/Cypher.java
@@ -33,6 +33,7 @@ import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import static apoc.util.MapUtil.map;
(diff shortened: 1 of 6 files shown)


vga91:
This pr is about the Trello card with id `Xjj0rEkC`, which has been archived for a long time, 
with the wording:
```
We don't think there are any immediate actions from Engineering at this point in time, but the cypher runtime team is prioritizing this card to improve performance for these types of workloads.
```

So it's not apoc related, and it's about a version that is no longer supported in any case.

PR #2534, opened 2022-02-11
Deprecating apoc.warmup.run procedure

PetrJanouch:
Also changing tests of the procedure not to assert
the exact number of loaded pages as this is extremely
kernel-internals specific

### DIFF
--- a/core/src/main/java/apoc/warmup/Warmup.java
+++ b/core/src/main/java/apoc/warmup/Warmup.java
@@ -62,6 +62,7 @@ public class Warmup {
         return sb.toString();
     }
 
+    @Deprecated
     @Procedure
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/warmup/Warmup.java
```diff
@@ -62,6 +62,7 @@ private String subPath(File file, String fromParent) {
         return sb.toString();
     }
 
+    @Deprecated
```
OliviaYtterbrink, line 65: It would be nice to add a note about why it is being deprecated. There is a `deprecatedBy` field in the `@Procedure` annotation.

OliviaYtterbrink (changes requested):

OliviaYtterbrink (approved):

PR #2535, opened 2022-02-14
Switch index type from default to not break when default type is changed

Hunterness:


Lojjs (approved):
Have not set up apoc locally yet so cannot run it, but looking at the code this seems good

Hunterness:
[Change default index type from btree to range](https://trello.com/c/elVGcSei/2840-change-default-index-type-from-btree-to-range)

PR #2589, opened 2022-03-01
Fixes #2402: See if we can make the aStar algorithm support spatial points

### LINKED ISSUE #2402: See if we can make the aStar algorithm support spatial points

vga91:
Fixes #2402

Added config 'pointPropName'  in `apoc.algo.aStarConfig` procedure.
Extracted code from org.neo4j.graphalgo.impl.util.GeoEstimateEvaluator

### DIFF
--- a/core/src/main/java/apoc/algo/PathFinding.java
+++ b/core/src/main/java/apoc/algo/PathFinding.java
@@ -75,24 +75,6 @@ public class PathFinding {
     public Transaction tx;
 
     @Procedure
-    @Description("apoc.algo.aStarWithPoint(startNode, endNode, 'relTypesAndDirs', 'distance','pointProp') - " +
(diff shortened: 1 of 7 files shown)


core/src/main/java/apoc/algo/PathFinding.java
```diff
@@ -19,12 +20,78 @@
...
+    }
+
     @Context
     public GraphDatabaseService db;
 
     @Context
     public Transaction tx;
 
+    @Procedure
+    @Description("apoc.algo.aStarWithPoint(startNode, endNode, 'relTypesAndDirs', 'distance','pointProp') - " +
```
Lojjs, line 78: We have started to be stricter about what will be added into APOC core. Would it be possible to add this into APOC full instead?

vga91: Yes sure, I just moved it in full

Lojjs (approved):
As far as I can judge, this looks OK

PR #2590, opened 2022-03-01
apoc.refactor.mergeNodes is not producing desired output

vga91:
Trello issue: https://trello.com/c/7yO7mniS/924-s2cast-softwareapocrefactormergenodes-is-not-producing-desired-output

Added clearer documentation and fixed `mergeRels:true` bug (it should combine also rel-prop):

If properties is a map, properties with a name not included in parameter map are overridden (with an empty map too).

Instead, if properties is null, relationship properties are combined and node properties are overridden (coherently with https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/1051)


Lojjs (approved):
This looks good I think. I can merge it for you if you like @vga91

PR #2591, opened 2022-03-01
Fixes #2384: Add Scraper procedures to apoc.load.html

### LINKED ISSUE #2384: Add Scraper procedures to apoc.load.html
https://github.com/szenyo/neo4jscraperproc

as we use jsoup in apoc.load.html already we can also add some more user friendly functions there.

@rvanbruggen demonstrates some of them here from a user's perspective.

via https://gist.github.com/rvanbruggen/90db8bcca2cf7b8d7a5f15c95be3be70#file-congoholdup-md

vga91:
Fixes #2384

- Added documentation to explain how to emulate the functions provided by https://github.com/szenyo/neo4jscraperproc with the custom jsoup selectors
- Added `htmlText` configuration to pass an html string as 1st parameter
- Added procedure `apoc.load.htmlPlainText`, with the same syntax as apoc.load.html, to retrieve the plain text similarly to `scraper.getPlainText()` in https://github.com/szenyo/neo4jscraperproc

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/import/html.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/import/html.adoc
@@ -34,6 +34,7 @@ Config param is optional, the default value is an empty map.
 |===
 |charset | Default: UTF-8
 |baserUri | Default: "", it is use to resolve relative paths
+|htmlString | Default: false, to use an HTML string instead of an url as 1st parameter
 |===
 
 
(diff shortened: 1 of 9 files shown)


JMHReif (approved):

JMHReif:
@vga91, could you take a look at the failing CI? It looks like there's a test that looks at APOC full and core that is failing with these new procedures.

JMHReif (approved):

PR #2593, opened 2022-03-01
Fixes #2325: added property filters

### LINKED ISSUE #2325: Symmetric export-import for JSON documents
## Feature description

Calling `apoc.export.json.all` results in a JSON file with numeric IDs embedded. This is fine except when you then import with `apoc.import.json` and those IDs appear as properties in your nodes.

I would like an option for these IDs not to appear as node properties on import. A more general solution would be to allow importing to ignore specified property keys.

## Considered alternatives
There are workarounds for this, but they are not always applicable or pleasant.

* Use unique constraints from the application domain instead of IDs, but not all applications have these.
* Clean up afterwards, which is time-consuming for larger databases.

vga91:
Fixes #2325

Added `nodePropFilter` and `relPropFilter`, 
with key the label / rel-type consistently with `nodePropertyMappings` and `relPropertyMappings`.

The values are the property keys to be filtered.

Resolve the linked issue together with the #2592

### DIFF
--- a/core/src/main/java/apoc/export/json/ImportJsonConfig.java
+++ b/core/src/main/java/apoc/export/json/ImportJsonConfig.java
@@ -6,9 +6,13 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 
 public class ImportJsonConfig extends CompressionConfig {
+    public static final String WILDCARD_PROPS = "_all";
+    private final Map<String, List<String>> nodePropFilter;
+    private final Map<String, List<String>> relPropFilter;
 
     private final Map<String, Map<String, String>> nodePropertyMappings;
     private final Map<String, Map<String, String>> relPropertyMappings;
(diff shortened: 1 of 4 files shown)


JMHReif (approved):

JMHReif:
@vga91, looks like we have conflicts with the previous Config boolean (isCleanup) with the additions in this PR. We want both, correct?

```
<<<<<<< issue_2325_filter_props
        this.nodePropFilter = (Map<String, List<String>>) config.getOrDefault("nodePropFilter", Collections.emptyMap());
        this.relPropFilter = (Map<String, List<String>>) config.getOrDefault("relPropFilter", Collections.emptyMap());
=======
        this.cleanup = Util.toBoolean(config.get("cleanup"));
>>>>>>> 4.3
```

vga91:
> @vga91, looks like we have conflicts with the previous Config boolean (isCleanup) with the additions in this PR. We want both, correct?
> 
> ```
> <<<<<<< issue_2325_filter_props
>         this.nodePropFilter = (Map<String, List<String>>) config.getOrDefault("nodePropFilter", Collections.emptyMap());
>         this.relPropFilter = (Map<String, List<String>>) config.getOrDefault("relPropFilter", Collections.emptyMap());
> =======
>         this.cleanup = Util.toBoolean(config.get("cleanup"));
> >>>>>>> 4.3
> ```

Exactly. I just rebased it with both of them.

JMHReif (approved):

PR #2598, opened 2022-03-02
Use CALL {} IN TRANSACTIONS instead of PERIODIC COMMIT in test

tobias-johansson:
Fixes test that will start failing when `PERIODIC COMMIT` is removed in neo4j


tobias-johansson:
new pr: #2599

PR #2614, opened 2022-03-08
Checks full API listings and archives more test results. No AUTO cherry-pick

ncordon:
## What
Checks full API listings and archives more test results.

## Why
There's a file called `full/src/main/resources/extended.txt` that contains the procedures and functions included in apoc full. It's important this file is up to date in the PRs cause it will be used to package apoc, and if not updated, the `apoc.help` method indicate something is in core while it's not (cause it uses that `extended.txt` file internally).

It is bad practice as that file is kept up to date by a test: `./gradlew test --tests HelpExtendedTest`, but that's a different problem.

This PR ensures we archive all tests from all projects but also that the aforementioned file is kept up to date.


.github/workflows/CI.yaml
```diff
@@ -35,10 +35,16 @@ jobs:
       - name: Clean neo4j from the cache
         if: success() || failure() # run this step even if previous step failed
         run: (find ~/.gradle/caches -name "*neo4j*" -exec rm -rf {} \;) || echo "All neo4j files cleaned"
+      - name: Check procedures included in apoc full
+        run: git diff --exit-code full/src/main/resources/extended.txt || (echo "extended.txt file does not contain all changes" && exit -1)
```
ncordon, line 39: The build re-generates the `extended.txt` file. If it contains changes with respect to the one the repo contains, the build will tell us it is incomplete and the changes we have to make (via `git diff`)

full/src/main/resources/extended.txt
```diff
@@ -144,3 +124,20 @@ apoc.redis.zadd
...
+apoc.static.set
+apoc.systemdb.execute
+apoc.systemdb.graph
+apoc.ttl.expire
+apoc.ttl.expireIn
+apoc.uuid.install
+apoc.uuid.list
+apoc.uuid.remove
+apoc.uuid.removeAll
+apoc.data.email
```
Lojjs, line 138: Note sure if it matters for when this is re-generated, but everything above this line seems to be in alphabetical order, the rest is not

.github/workflows/CI.yaml
```diff
@@ -35,10 +35,16 @@ jobs:
...
       - name: Archive test results
         uses: actions/upload-artifact@v2
+        if: ${{ always() }}
```
Lojjs, line 42: I googled to understand the meaning of this, and Stack overflow tells me it can probably be simplified to just 
`if: always()`

Lojjs (approved):
Few minor things to take into consideration but I think they should not block you from merging really.

full/src/main/resources/extended.txt
ncordon: Yep, but that's the order the build generates in, we cannot do much about that, although I don't have an explanation for that, but seems to be a stable order

.github/workflows/CI.yaml
ncordon: I had to add this line to archive the test results also in case of test failures

PR #2620, opened 2022-03-09
Adds a CI. No AUTO cherry-pick

ncordon:
Runs the CI in the 3.5 branch. Deletes some legacy stuff from having used a Travis CI in the past.

Note 3.5 is different to the 4.x versions, as the package is not split into `core` or `full`.


AzuObs:
[Links between github PR and Trello card](https://trello.com/c/FkbVLczP/24-links-between-github-pr-and-trello-card)

build.gradle
```diff
@@ -264,13 +264,11 @@ test {
...
-
-    if (System.env.TRAVIS !=null) {
-        minHeapSize = "128m"
-        maxHeapSize = "512m"
-    } else {
-        maxHeapSize = "4g"
+    reports {
+        junitXml.enabled = false
+        html.enabled = true
     }
```
Lojjs, line 270: I'm a bit unsure about the additions and removals here. As you are not intending to auto-cherry-pick this one, I compared with how it looks in dev branch.  I couldn't really find anything around xml vs. html and I still found some of the forks and heap setting in the dev build.gradle file, but separated from TRAVIS, like below:

```
        maxHeapSize = "8G"
        forkEvery = 50
        maxParallelForks = 1 //Runtime.runtime.availableProcessors().intdiv(2) + 1

        if (System.env.CI != null) {
            minHeapSize = "128m"
            maxHeapSize = "512m"
        }
```

What do you think? Should we try to align it more between versions?

Lojjs (commented):
I have one change I'm unsure about and would like to discuss, see comment. Otherwise, it looks good

build.gradle
ncordon: >  I couldn't really find anything around xml vs. html 

That's due to the gradle version being used in this branch. It's way older, so it produces reports in xml by default. The newer versions produce reports in html by default. We can upgrade gradle in a separate bit of work

With regards to the CI options, you are right, I'll align this with the other branches

ncordon: Done!

build.gradle
```diff
@@ -264,6 +264,15 @@ test {
         }
     }
 
+    maxHeapSize = "8G"
+    forkEvery = 50
```
Lojjs, line 268: We actually have another `forkEvery 100` further down in this file, which you might want to remove now

Lojjs (approved):

PR #2621, opened 2022-03-09
Uses dev docker container in the dev branch. No AUTO cherry-pick

ncordon:
## What
Fetches the dev neo4j container from TeamCity.

## Why
Some tests rely on the neo4j docker container, which is not published until the database is released. Those tests were being ignored in the CI. This way we can develop more reliably.

This container will be available only for PRs coming from internal contributors.


full/build.gradle
```diff
@@ -6,12 +6,11 @@ plugins {
     id 'maven-publish'
     id 'antlr'
     id "org.sonarqube"
-    id "org.jetbrains.kotlin.jvm" version "1.5.31"
+    id "org.jetbrains.kotlin.jvm" version "1.6.0"
```
ncordon, line 9: We need to update this in the other builds

extra-dependencies/xls/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.2.1-bin.zip
\ No newline at end of file
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-bin.zip
```
ncordon, line 6: We need to update this in the other builds

extra-dependencies/xls/build.gradle
```diff
@@ -3,8 +3,8 @@ plugins {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_11
-    targetCompatibility = JavaVersion.VERSION_11
+    sourceCompatibility = JavaVersion.VERSION_17
+    targetCompatibility = JavaVersion.VERSION_17
```
ncordon, line 7: We need to update this in the other maintained branches

full/build.gradle
```diff
@@ -20,7 +19,7 @@ jar {
 }
 
 compileKotlin {
-    kotlinOptions.jvmTarget = "1.8"
+    kotlinOptions.jvmTarget = "17"
```
ncordon, line 22: Update this in the other branches

full/build.gradle
```diff
@@ -153,6 +152,7 @@ dependencies {
 
     configurations.all {
         exclude group: 'org.slf4j', module: 'slf4j-nop'
+        exclude group: 'ch.qos.logback', module: 'logback-classic'
     }
```
ncordon, line 156: Add this in the other branches

build.gradle
```diff
@@ -100,6 +100,9 @@ subprojects {
         filter {
             setFailOnNoMatchingTests(false)
         }
+
+        testLogging.showStandardStreams = true
+        testLogging.exceptionFormat = 'full'
```
ncordon, line 105: Add this lines in other branches

.github/workflows/CI.yaml
```diff
@@ -24,19 +42,16 @@ jobs:
...
-          TEAMCITY_DEV_URL: ${{ secrets.TEAMCITY_DEV_URL }}
-          TEAMCITY_USER: ${{ secrets.TEAMCITY_USER }}
-          TEAMCITY_PASSWORD: ${{ secrets.TEAMCITY_PASSWORD }}
-        run: ./gradlew build
-      - name: Clean neo4j from the cache
-        if: success() || failure() # run this step even if previous step failed
-        run: (find ~/.gradle/caches -name "*neo4j*" -exec rm -rf {} \;) || echo "All neo4j files cleaned"
-      - name: Check procedures included in apoc full
-        run: git diff --exit-code full/src/main/resources/extended.txt || (echo "extended.txt file does not contain all changes" && exit -1)
+        run: chmod +x gradlew && ./gradlew build
```
ncordon, line 47: Simplify these lines and make sure there's no `build --continue` in other branches.

test-startup/src/test/java/StartupTest.java
```diff
@@ -31,50 +32,58 @@
...
+                // if Testcontainers wasn't able to retrieve the docker image we ignore the test
+                if (TestContainerUtil.isDockerImageAvailable(ex)) {
+                    ex.printStackTrace();
+                    fail("Should not have thrown exception when trying to start Neo4j: " + ex);
+                } else if (!TestUtil.isRunningInCI()) {
+                    fail( "The docker image " + TestContainerUtil.neo4jEnterpriseDockerImageVersion + " should be available in the CI");
+                }
             }
         }
     }
```
ncordon, line 60: Refactor this in other branches to also have tests for community version (to avoid binary incompatibilities ever again)

full/build.gradle
```diff
@@ -48,7 +47,7 @@ dependencies {
 //    implementation group: 'commons-codec', name: 'commons-codec', version: '1.14'
     implementation group: 'com.jayway.jsonpath', name: 'json-path', version: '2.4.0'
     implementation group: 'org.hdrhistogram', name: 'HdrHistogram', version: '2.1.9'
-    implementation group: 'org.neo4j.driver', name: 'neo4j-java-driver', version: '4.0.0'
+    compileOnly group: 'org.neo4j.driver', name: 'neo4j-java-driver-slim', version: '4.4.3'
```
ncordon, line 50: Make the driver compile only in other branches to avoid overwriting the one in the database when we plug in apoc

core/src/test/java/apoc/StartupTest.java
```diff
@@ -20,7 +20,7 @@
     public void test() {
         try {
             Neo4jContainerExtension neo4jContainer = createEnterpriseDB(!TestUtil.isRunningInCI())
-                    .withNeo4jConfig("dbms.transaction.timeout", "5s");
+                    .withNeo4jConfig("dbms.transaction.timeout", "60s");
```
ncordon, line 23: Port this to other branches

extra-dependencies/xls/build.gradle
Lojjs: No I don't think so, as Neo 4.x is supposed to be run with Java 11?

full/build.gradle
Lojjs: Yes, but in the other branches 4.2, 4.3 and 4.4, I guess you should update it to 11 rather than 17.

full/src/main/java/apoc/bolt/Bolt.java
```diff
@@ -23,6 +23,7 @@
 import org.neo4j.driver.types.Node;
 import org.neo4j.driver.types.Path;
 import org.neo4j.driver.types.Relationship;
+import org.neo4j.driver.internal.async.connection.EventLoopGroupFactory;
```
Lojjs, line 26: Here it looks like you have added an import without adding any code, so I suspect it is unused.

full/build.gradle
Lojjs: What do you mean with 'only in other branches'? I'm not sure I follow...

test-startup/src/test/java/StartupTest.java
```diff
@@ -31,50 +32,58 @@
...
-            assertTrue(procedureCount > 0);
-            assertTrue(functionCount > 0);
-            assertTrue(coreCount > 0);
-        } catch (Exception ex) {
-            // if Testcontainers wasn't able to retrieve the docker image we ignore the test
-            if (TestContainerUtil.isDockerImageAvailable(ex)) {
-                ex.printStackTrace();
-                fail("Should not have thrown exception when trying to start Neo4j: " + ex);
+        for (var version: Neo4jVersion.values()) {
+            try (Neo4jContainerExtension neo4jContainer = createDB(version, APOC_FULL, !TestUtil.isRunningInCI())
```
Lojjs, line 36: Why did you switch this back to `!TestUtil.isRunningInCI()` instead of true?

full/src/main/java/apoc/bolt/Bolt.java
ncordon: Good catch

test-startup/src/test/java/StartupTest.java
ncordon: I'm not sure whether true helps somehow. I wanted more logs in the tests that run docker containers and I assumed we weren't getting any logs in the CI because of this setting, but it turns out it was because of other misconfigurations in the build.

Reverting these settings as we seem to be getting more thorough logs now. I'm not actually sure the setting makes any difference

Lojjs (approved):

PR #2638, opened 2022-03-15
Rewrite exists(property) to property IS NOT NULL. No AUTO cherry-pick

Lojjs:
Update APOC syntax to adapt to removal of exists(property) in Neo4j.

### DIFF
--- a/core/src/test/java/apoc/schema/SchemasTest.java
+++ b/core/src/test/java/apoc/schema/SchemasTest.java
@@ -554,7 +554,7 @@ public class SchemasTest {
     @Test
     public void testSchemaRelationshipsExclude() {
         ignoreException(() -> {
-            db.executeTransactionally("CREATE CONSTRAINT FOR ()-[like:LIKED]-() REQUIRE like.day I");
+            db.executeTransactionally("CREATE CONSTRAINT FOR ()-[like:LIKED]-() REQUIRE like.day IS NOT NULL");
             testResult(db, "CALL apoc.schema.relationships({excludeRelationships:['LIKED']})", (result) -> assertFalse(result.hasNext()));
         }, QueryExecutionException.class);
     }
--- a/core/src/test/java/apoc/schema/SchemasTest.java
+++ b/core/src/test/java/apoc/schema/SchemasTest.java
@@ -5,6 +5,12 @@ import org.assertj.core.api.Assertions;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+
+import java.util.Collections;
+import java.util.List;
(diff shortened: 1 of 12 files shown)


core/src/test/java/apoc/schema/SchemasTest.java
```diff
@@ -551,7 +554,7 @@ public void testIndexesMoreLabels() {
     @Test
     public void testSchemaRelationshipsExclude() {
         ignoreException(() -> {
-            db.executeTransactionally("CREATE CONSTRAINT FOR ()-[like:LIKED]-() REQUIRE exists(like.day)");
+            db.executeTransactionally("CREATE CONSTRAINT FOR ()-[like:LIKED]-() REQUIRE like.day I");
```
Hunterness, line 557: I feel like some part is missing here?

Hunterness (changes requested):

Hunterness (approved):

docs/asciidoc/modules/ROOT/partials/usage/apoc.schema.relationships.adoc
```diff
@@ -16,6 +16,6 @@ CALL apoc.schema.relationships();
 .Results
 [opts="header"]
 |===
-| name                                                         | type                              | properties | status
-| "CONSTRAINT ON ()-[liked:LIKED]-() ASSERT exists(liked.day)" | "RELATIONSHIP_PROPERTY_EXISTENCE" | ["day"]    | ""
+| name                                                             | type                              | properties | status
+| "CONSTRAINT ON ()-[liked:LIKED]-() ASSERT liked.day IS NOT NULL" | "RELATIONSHIP_PROPERTY_EXISTENCE" | ["day"]    | ""
```
Hunterness, line 20: Actually this should be updated to use `FOR .. REQUIRE` as well

Hunterness (changes requested):

docs/asciidoc/modules/ROOT/partials/usage/apoc.schema.relationships.adoc
Lojjs: Yes. I get over 100 matches on `CONSTRAINT ON` on apoc dev though so I think we need a separate PR for that though if you don't have one open already?

AzuObs (approved):

docs/asciidoc/modules/ROOT/partials/usage/apoc.schema.relationships.adoc
Hunterness: Sure lets do it separately then :)

PR #2651, opened 2022-03-20
feat: add chinese converter

huyuanxin:

Add  a converter that can convert Chinese words to Chinese Pinyin, like `中国 ` to `zhong guo`

  -
  -
  -


conker84:
@huyuanxin thank you! Did you sign the [CLA](https://neo4j.com/developer/cla/)?

huyuanxin:
> 

yeah

Lojjs (changes requested):
If possible it would be good if we could add this to the full instead of the core module, as it is a bit more specialized functionality.

Lojjs:
@huyuanxin Sorry for the late feedback. We are currently not extending APOC core with any new functionality. If we were to add this to APOC extended, we would also probably go for a more generic solution, covering more dialects. The best route for you to be able to use this is to instead create your own User Defined function, see https://neo4j.com/docs/java-reference/4.4/extending-neo4j/functions/#extending-neo4j-functions

PR #2652, opened 2022-03-21
Fixes #2223: Add config option for schema exports to retain the index and/or constraint names

### LINKED ISSUE #2223: Add config option for schema exports to retain the index and/or constraint names
Some users explicitly name their indexes and constraints, and can use these in various scenarios. While APOC has the option to export schema statements (either alone, as part of export.all()), they do not export the schema element names.

## Considered alternatives
No other alternatives exist. For implementation, we could have a single option controlling export of names for both indexes and constraints, or two separate configs for indexes and constraints separately. At this time perhaps only a single option makes sense. If more fine work is needed, it can be applied to the exported file before import.

This helps users who require that schema names be preserved.

vga91:
Fixes #2223

Added configs `saveIndexNames` and `saveConstraintNames`

### DIFF
--- a/core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
+++ b/core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
@@ -264,7 +264,7 @@ public class MultiStatementCypherSubGraphExporter {
                         return new AbstractMap.SimpleEntry<>(FULL, fullTextIndex);
                     }
                     // "normal" schema index
-                    String idxName = getIdxName(name, config.isSaveIndexNames());
+                    String idxName = getIdxName(name, config.shouldSaveIndexNames());
                     String tokenName = tokenNames.get(0);
                     return new AbstractMap.SimpleEntry<>(NORMAL, this.cypherFormat.statementForIndex(tokenName, props, exportConfig.ifNotExists(), idxName));
 
@@ -308,7 +308,7 @@ public class MultiStatementCypherSubGraphExporter {
         return StreamSupport.stream(graph.getIndexes().spliterator(), false)
                 .filter(index -> index.isConstraintIndex())
                 .map(index -> {
-                    String name = getIdxName(index.getName(), config.isSaveConstraintNames());
+                    String name = getIdxName(index.getName(), config.shouldSaveConstraintNames());
                     String label = Iterables.single(index.getLabels()).name();
                     Iterable<String> props = index.getPropertyKeys();
                     return this.cypherFormat.statementForConstraint(label, props, exportConfig.ifNotExists(), name);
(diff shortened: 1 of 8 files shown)


core/src/main/java/apoc/export/util/ExportConfig.java
```diff
@@ -217,4 +221,12 @@ public boolean isSampling() {
     public boolean ifNotExists() {
         return ifNotExists;
     }
+
+    public boolean isSaveIndexNames() {
```
ncordon, line 225: Nitpick: `shouldSaveIndexNames` and `shouldSaveContraintNames` maybe as better names?

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -224,7 +224,7 @@ private void exportSchema(PrintWriter out, ExportConfig config) {
             out.println(index);
         }
         if (artificialUniques > 0) {
-            String cypher = this.cypherFormat.statementForConstraint(UNIQUE_ID_LABEL, Collections.singleton(UNIQUE_ID_PROP), config.ifNotExists());
+            String cypher = this.cypherFormat.statementForConstraint(UNIQUE_ID_LABEL, Collections.singleton(UNIQUE_ID_PROP), config.ifNotExists(), StringUtils.EMPTY);
```
ncordon, line 227: Why empty constraint names here?

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -329,7 +335,7 @@ private void exportCleanUp(PrintWriter out, int batchSize) {
                 artificialUniques -= batchSize;
             }
             begin(out);
-            String cypher = this.cypherFormat.statementForConstraint(UNIQUE_ID_LABEL, Collections.singleton(UNIQUE_ID_PROP), false)
+            String cypher = this.cypherFormat.statementForConstraint(UNIQUE_ID_LABEL, Collections.singleton(UNIQUE_ID_PROP), false, StringUtils.EMPTY)
```
ncordon, line 338: Why empty constraint name here?

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
vga91: Because this statement is about the `UNIQUE IMPORT LABEL`,
which is used internally in order to guarantee the uniqueness.

When the import is completed, this constraint will be removed, so retain the constraint name is unnecessary.

As said in doc:
```
When exporting nodes, 
if a node label does not contain a unique constraint the exporter will add a `UNIQUE IMPORT LABEL` label and `UNIQUE IMPORT ID` property to those nodes to ensure uniqueness of nodes when the export script is executed on a new database.
```

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
vga91: As above

PR #2656, opened 2022-03-24
Make PathExpander return ResourceIterable in readiness for Core API changes

docrozza:
The public API for PathExpanders will be changing to return a `ResourceIterable` instead of a simple `Iterable`.

This will allow the path expansions to be closed and the underlying resources to be released. Currently, any iterators
created during the traversal are not closed and resources could potentially be leaked. The `ResourceIterable` returned
in this commit will track any iterators created and close them when it is itself closed.

The change needs to be made in APOC first to allow the main neo4j build to proceed.


AzuObs:
This looks OK but will need to be resubmitted so that the build passes!

docrozza:
Changes moved to APOC branch rather from fork - see #2720

PR #2671, opened 2022-03-29
Fixes junit compilation errors

ncordon:


Lojjs (approved):

full/src/main/resources/extended.txt
```diff
@@ -157,7 +151,6 @@ apoc.static.set
 apoc.systemdb.execute
 apoc.systemdb.export.metadata
 apoc.systemdb.graph
-apoc.systemdb.export.metadata
```
ncordon, line 160: This is needed cause there was an oversight in this PR: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2631/files adding them when they are already in the list

PR #2700, opened 2022-04-05
Fixes #558: Check for neo4j version being compatible with neo version

### LINKED ISSUE #558: Add automatic sanity checks for Neo4j/APOC version combination
Some rough sanity check would be good. Perhaps as an extra shell script to be run.

see also the old thread issue #461

vga91:
Fixes #558

Added a check between neo4j and apoc version. If versions are different, a log warn will be returned.

reg

### DIFF
--- a/core/src/main/java/apoc/RegisterComponentFactory.java
+++ b/core/src/main/java/apoc/RegisterComponentFactory.java
@@ -1,11 +1,10 @@
 package apoc;
 
-import apoc.trigger.TriggerHandler;
+import org.neo4j.annotations.service.ServiceProvider;
 import org.neo4j.kernel.api.procedure.GlobalProcedures;
 import org.neo4j.kernel.extension.ExtensionFactory;
 import org.neo4j.kernel.extension.ExtensionType;
 import org.neo4j.kernel.extension.context.ExtensionContext;
-import org.neo4j.kernel.internal.Version;
 import org.neo4j.kernel.lifecycle.Lifecycle;
 import org.neo4j.kernel.lifecycle.LifecycleAdapter;
 import org.neo4j.logging.Log;
@@ -18,14 +17,14 @@ import java.util.concurrent.ConcurrentHashMap;
 /**
  * NOTE: this is a GLOBAL component, so only once per DBMS
  */
+@ServiceProvider
 public class RegisterComponentFactory extends ExtensionFactory<RegisterComponentFactory.Dependencies> {
 
     private Log log;
     private GlobalProcedures globalProceduresRegistry;
 
     public RegisterComponentFactory() {
-        super(ExtensionType.GLOBAL,
-                "ApocRegisterComponent");
+        super(ExtensionType.GLOBAL, "ApocRegisterComponent");
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


conker84:
@vga91 please add a statement in the docs where you describe the behaviour of the log

core/src/main/java/apoc/RegisterComponentFactory.java
```diff
@@ -5,6 +5,7 @@
 import org.neo4j.kernel.extension.ExtensionFactory;
 import org.neo4j.kernel.extension.ExtensionType;
 import org.neo4j.kernel.extension.context.ExtensionContext;
+import org.neo4j.kernel.internal.Version;
```
MishaDemianenko, line 8: this is internal class for a reason and should not be used here and there for any convience

MishaDemianenko (changes requested):

core/src/main/java/apoc/RegisterComponentFactory.java
vga91: Oh I see. Therefore, please could you tell me what to use as an alternative to `Version.getKernelVersion()`?

MishaDemianenko: so there is a dbms.components procedure that will return neo4j version; also here you are expecting that version will also have dot separator and that will be major.minor - there is no guarantee that its the case

MishaDemianenko: so there is a dbms.components procedure that will return neo4j version; also here you are expecting that version will also have dot separator and that will be major.minor - there is no guarantee that its the case

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -75,6 +98,14 @@ public void available() {
         }).start();
     }
 
+    private String getMajorMinVersion(String completeVersion) {
+        if (completeVersion == null) {
+            return null;
+        }
+        final String[] split = completeVersion.split("\\.");
+        return split[0] + "." + split[1];
+    }
```
Lojjs, line 107: Talked to @MishaDemianenko and there is no guarantee this is always on the major.minor format. Can we make it a bit more flexible and also add some sort of testing for this?

vga91: I changed the handling. Now we check if `<KERNEL_VERSION> ` starts with `<APOC_VERSION>`,
because expect that APOC versioning is always consistent to Neo4j versioning,
e.g. if we have Neo4j `5_0_0`, the Apoc version will be `5_x_x_x` as well. 

And I added some tests too.

Lojjs (commented):
There also seems to be some test failures in the build, not sure if they are related to the changes or not

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -98,12 +91,9 @@ public void available() {
         }).start();
     }
 
-    private String getMajorMinVersion(String completeVersion) {
-        if (completeVersion == null) {
-            return null;
-        }
-        final String[] split = completeVersion.split("\\.");
-        return split[0] + "." + split[1];
+    // only for testing purpose
```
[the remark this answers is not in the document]
vga91: Yes exactly. Added it and fixed the failed test as well

Lojjs (approved):
Looks good to me now, but would like @MishaDemianenko to have another look as well to see if he thinks his concerns are fixed

MishaDemianenko (approved):

conker84:
Thank you @vga91 and thank you @Lojjs and @MishaDemianenko for the review!

PR #2703, opened 2022-04-05
Fixes #2645: csv import fails when parse an empty string as a date

### LINKED ISSUE #2645: csv import fails when  parse an empty string as a date
## Guidelines

 `apoc.import.csv`  fails when  parse an empty string as a date
But empty string and int process fine the same procedure .

Empty or null date in the csv should import like empty string and empty int 

Failed to invoke procedure `apoc.import.csv`: Caused by: org.neo4j.exceptions.TemporalParseException: Text cannot be parsed to a Date
""
 ^
pass.csv

id:ID,:LABEL,str:STRING,int:INT,date:DATE
1,lab,hello,1,2020-01-01
2,lab,world,2,2020-01-01
3,lab,,,2020-01-01

fail.csv

id:ID,:LABEL,str:STRING,int:INT,date:DATE
1,lab,hello,1,2020-01-01
2,lab,world,2,2020-01-01
3,lab,,,

### Steps (Mandatory)

1 .CALL apoc.import.csv([{fileName: 'pass.csv', labels:[]}], [], {nullValues: ['']})`
works fine in ignoring the blank strings skipping the attribute creation.

However,
2 . `CALL apoc.import.csv([{fileName: 'fail.csv', labels:[]}], [], {nullValues: ['']})`
will fail with

Failed to invoke procedure `apoc.import.csv`: Caused by: org.neo4j.exceptions.TemporalParseException: Text cannot be parsed to a Date
""
 ^

## Specifications (Mandatory)

4.4.4

### Versions

  - OS:ubuntu
  - Neo4j:4.4.4
  - Neo4j-Apoc:apoc-4.4.0.1-all.jar

vga91:
Fixes #2645

Related Trello issue: https://trello.com/c/9kk5cLwh/958-s4govtech-singapore-csv-import-fails-when-parse-an-empty-string-as-a-date

### DIFF
--- a/core/src/test/java/apoc/export/csv/ImportCsvTest.java
+++ b/core/src/test/java/apoc/export/csv/ImportCsvTest.java
@@ -102,6 +102,11 @@ public class ImportCsvTest {
                             ":ID(MultiType-ID)|date1:datetime{timezone:Europe/Stockholm}|date2:datetime|foo:string|joined:date|active:boolean|points:int\n" +
                             "1|2018-05-10T10:30|2018-05-10T12:30|Joe Soap|2017-05-05|true|10\n" +
                             "2|2018-05-10T10:30[Europe/Berlin]|2018-05-10T12:30[Europe/Berlin]|Jane Doe|2017-08-21|true|15\n"),
+                    new AbstractMap.SimpleEntry<>("emptyDate",
+                            "id:ID,:LABEL,str:STRING,int:INT,date:DATE\n" +
+                                    "1,Lab,hello,1,2020-01-01\n" +
+                                    "2,Lab,world,2,2020-01-01\n" +
+                                    "3,Lab,,,\n"),
                     new AbstractMap.SimpleEntry<>("relMultiTypes",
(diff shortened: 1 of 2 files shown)


core/src/test/java/apoc/export/csv/ImportCsvTest.java
```diff
@@ -430,6 +435,26 @@ public void testCustomRelationshipTypes() {
         Assert.assertEquals("John Jane", TestUtil.singleResultFirstColumn(db, "MATCH (p1:Person)-[:FRIENDS_WITH]->(p2:Person) RETURN p1.name + ' ' + p2.name AS pair ORDER BY pair"));
         Assert.assertEquals("Jane John", TestUtil.singleResultFirstColumn(db, "MATCH (p1:Person)-[:KNOWS]->(p2:Person) RETURN p1.name + ' ' + p2.name AS pair ORDER BY pair"));
     }
+
+    @Test
+    public void testEmptyDate() {
```
Lojjs, line 440: I tried it locally, and this test is passing even without the fix. Would it be possible to tweak it to actually show the issue?

vga91: Mmh this is really weird. I just tried this test without my fix and it fails (I also rebased it to be sure).
I really don't know why it doesn't fail you too 😅

Lojjs: Never mind, I run the wrong test 🤦‍♀️

PR #2704, opened 2022-04-05
Fixes #2612: Confusing behavior with map as a custom functions

### LINKED ISSUE #2612: Confusing behavior with map as a custom functions
See https://community.neo4j.com/t/why-do-functions-created-with-apoc-custom-declarefunction-or-asfunction-behave-differently-when-returning-a-map-value-vs-other-types/52095

With `CALL apoc.custom.declareFunction('answer(map :: MAP) :: MAP','RETURN $map as answer', false)`
the result is `{ "answer": { "foo": 1 } }` because of wrapping of the result with key "column result" (answer) and value the map.
This is a wanted behaviour https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.1/full/src/main/java/apoc/custom/CypherProceduresHandler.java#L398

But if we want to return only {foo: 1}, I have to execute for example:
```
CALL apoc.custom.declareFunction('funWithMap(map :: MAP) :: STRING' , 'RETURN {a: 1}' )
```

and this is of course wrong because of `string` as a return type.
A solution would be, not to break change, to add an option to "force" the MAP result without wrap the result.

vga91:
Fixes #2612

Related community issue: https://community.neo4j.com/t/why-do-functions-created-with-apoc-custom-declarefunction-or-asfunction-behave-differently-when-returning-a-map-value-vs-other-types/52095/2

The user ask why the map is wrapped with an `answer` key and the only way is to do this change.

However, this is a breaking-change so it is to be understood if it might be worth.

### DIFF
--- a/full/src/main/java/apoc/custom/CypherProceduresHandler.java
+++ b/full/src/main/java/apoc/custom/CypherProceduresHandler.java
@@ -396,17 +396,10 @@ public class CypherProceduresHandler extends LifecycleAdapter implements Availab
                             List<String> cols = result.columns();
                             if (cols.isEmpty()) return null;
                             if (!forceSingle && outType instanceof Neo4jTypes.ListType) {
-                                Neo4jTypes.ListType listType = (Neo4jTypes.ListType) outType;
-                                Neo4jTypes.AnyType innerType = listType.innerType();
-                                // We wrap the result only if we have a "true" map, and not NodeType or RelationshipType that extends MapType
-                                if (innerType.getClass().equals(Neo4jTypes.MapType.class))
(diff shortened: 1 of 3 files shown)


jexp:
Yes I think that was a bug when implementing it first and and oversight. Sorry for that.

But **now it's a breaking change** (which can't be easily accommodated for as it's not in the parameter list but return type).

So my preferred option would be to pick a different return type name for returning a map directly and add it to the documentation.

e.g. `MAPRESULT`

/cc @vga91

conker84:
we're closing this as 4.3 is out of support. we plan to do a new one with @jexp 's feedback in 5.x

PR #2720, opened 2022-04-07
Make PathExpander return ResourceIterable in readiness for Core API changes

docrozza:
The public API for PathExpanders will be changing to return a ResourceIterable instead of a simple Iterable.

This will allow the path expansions to be closed and the underlying resources to be released. Currently, any iterators
created during the traversal are not closed and resources could potentially be leaked. The ResourceIterable returned
in this commit will track any iterators created and close them when it is itself closed.

The change needs to be made in APOC first to allow the main neo4j build to proceed.

### DIFF
--- a/core/src/main/java/apoc/path/RelationshipSequenceExpander.java
+++ b/core/src/main/java/apoc/path/RelationshipSequenceExpander.java
@@ -12,7 +12,8 @@ import org.neo4j.graphdb.RelationshipType;
 import org.neo4j.graphdb.ResourceIterable;
 import org.neo4j.graphdb.ResourceIterator;
 import org.neo4j.graphdb.traversal.BranchState;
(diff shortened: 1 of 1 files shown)


docrozza:
So ... there are 3 expansion tests that use a limit and these were failing. Depending on what tests ran after these, they would **ALSO** fail due to the previous state of the DB being out (the leaking resource would be affecting the memory tracker checked in the setup)

Tweaked the code to materialise all the results again (as before this change) so that any calls with a limit (i.e. that lead to a call of `Stream.limit` in `PathExplorer`) will release all resources

Will look into how to **not** materialise all the relationships but these changes would be in core-API rather than in APOC

AzuObs (approved):
Great work Rory, thanks for the PR

Issue #2714, opened 2022-04-07
apoc.load.jsonParams has Problems with Azure Blob signed URL's


leonpierre:
## Expected Behavior (Mandatory)

Similar to AWS S3 Url's, they are working fine btw., it should be possible to use an Azure Blob URL with apoc Load methods like apoc.load.jsonParams(url, null, null).

## Actual Behavior (Mandatory)

The method returns a 403 error code: 
Failed to invoke procedure `apoc.load.jsonParams`: Caused by: java.io.IOException: Server returned HTTP response code: 403 for URL: https://[ACCOUNTNAME].blob.core.windows.net/[CONTAINER]/file.json?sv=2021-04-10&se=2022-04-07T14%253A18%253A42Z&sr=b&sp=r&sig=OPo6CIIK8EdyEAqBEjKXfCWWylAJAcoCTrxBQHTQxlc%253D

=> your code encode '%' characters again with %25 instead of using the original Azure Blob Url.

**Original Azure Blob Url:** https://[ACCOUNTNAME].blob.core.windows.net/[CONTAINER]/file.json?sv=2021-04-10&se=2022-04-07T14%3A18%3A42Z&sr=b&sp=r&sig=OPo6CIIK8EdyEAqBEjKXfCWWylAJAcoCTrxBQHTQxlc%3D

## How to Reproduce the Problem

CALL apoc.load.jsonParams("https://[ACCOUNTNAME].blob.core.windows.net/[CONTAINER]/file.json?sv=2021-04-10&se=2022-04-07T14%3A18%3A42Z&sr=b&sp=r&sig=OPo6CIIK8EdyEAqBEjKXfCWWylAJAcoCTrxBQHTQxlc%3D", null, null) YIELD value AS item

### Steps (Mandatory)

  1. Just execute the Cypher query from "How to Reproduce the Problem" and use a valid Azure Blob signed URL

Currently used versions

### Versions

  - OS: Windows 10
  - Neo4j: 4.4.1 Community
  - Neo4j-Apoc: 4.4.0.3

vga91:
@leonpierre 

I think this problem can be fixed by this pr https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2588 ,
merged in 4.3 branch but not yet in 4.4.
Can you try using the jar contained in this attached zip, instead of your current jar and then, restart the db?

[...]

PR #2725, opened 2022-04-08
Fixes #2724: Broken handling of AWS S3 urls

### LINKED ISSUE #2724: Broken handling of AWS S3 urls
## Guidelines

When using AWS S3 url with credentials embedded in the url (eg: `s3://accessKey:secretKey[:sessionToken]@endpoint:port/bucket/key`), procedures such as `apoc.load.json` should be able to read the data if secret key contains url-escaped characters.

With a secret key contains a `/` (which they sometimes do), I tried two approaches:

1. Passing the secret UNESCAPED in the url string (eg [not a real secret]: `ABCDefGHikbWUyzkbsIKapTBr/fk4VkGsGWDOS/2XYZ`).  This fails with:

```
Caused by: java.lang.NumberFormatException: Error at index 1 in: "ABCDefGHikbWUyzkbsIKapTBr"
	at java.lang.NumberFormatException.forCharSequence(NumberFormatException.java:81) ~[?:?]
	at java.lang.Integer.parseInt(Integer.java:735) ~[?:?]
	at java.net.URLStreamHandler.parseURL(URLStreamHandler.java:223) ~[?:?]
	at java.net.URL.<init>(URL.java:674) ~[?:?]
	at java.net.URL.<init>(URL.java:541) ~[?:?]
	at java.net.URL.<init>(URL.java:488) ~[?:?]
	at apoc.util.FileUtils$SupportedProtocols.from(FileUtils.java:119) ~[apoc-4.4.0.2.jar:4.4.0.2]
	at apoc.util.FileUtils.isFile(FileUtils.java:243) ~[apoc-4.4.0.2.jar:4.4.0.2]
	at apoc.ApocConfig.checkReadAllowed(ApocConfig.java:263) ~[apoc-4.4.0.2.jar:4.4.0.2]
	at apoc.util.FileUtils.inputStreamFor(FileUtils.java:163) ~[apoc-4.4.0.2.jar:4.4.0.2]
	at apoc.util.JsonUtil.loadJson(JsonUtil.java:95) ~[apoc-4.4.0.2.jar:4.4.0.2]
	at apoc.load.LoadJson.loadJsonStream(LoadJson.java:71) ~[apoc-4.4.0.2.jar:4.4.0.2]
	at apoc.load.LoadJson.jsonParams(LoadJson.java:60) ~[apoc-4.4.0.2.jar:4.4.0.2]
	at apoc.load.LoadJson.json(LoadJson.java:49) ~[apoc-4.4.0.2.jar:4.4.0.2]
	at org.neo4j.kernel.impl.proc.GeneratedProcedure_json21076469570321.apply(Unknown Source) ~[?:?]
	at org.neo4j.procedure.impl.ProcedureRegistry.callProcedure(ProcedureRegistry.java:235) ~[neo4j-procedure-4.4.3.jar:4.4.3]
	at org.neo4j.procedure.impl.GlobalProceduresRegistry.callProcedure(GlobalProceduresRegistry.java:352) ~[neo4j-procedure-4.4.3.jar:4.4.3]
	at org.neo4j.kernel.impl.newapi.AllStoreHolder.callProcedure(AllStoreHolder.java:1092) ~[neo4j-kernel-4.4.3.jar:4.4.3]
	at org.neo4j.kernel.impl.newapi.AllStoreHolder.procedureCallRead(AllStoreHolder.java:1004) ~[neo4j-kernel-4.4.3.jar:4.4.3]
	at org.neo4j.cypher.internal.runtime.interpreted.CallSupport$.$anonfun$callReadOnlyProcedure$1(CallSupport.scala:47) ~[neo4j-cypher-interpreted-runtime-4.4.3.jar:4.4.3]
	at org.neo4j.cypher.internal.runtime.interpreted.CallSupport$.callProcedure(CallSupport.scala:70) ~[neo4j-cypher-interpreted-runtime-4.4.3.jar:4.4.3]
	at org.neo4j.cypher.internal.runtime.interpreted.CallSupport$.callReadOnlyProcedure(CallSupport.scala:47) ~[neo4j-cypher-interpreted-runtime-4.4.3.jar:4.4.3]
	at org.neo4j.cypher.internal.runtime.interpreted.TransactionBoundReadQueryContext.callReadOnlyProcedure(TransactionBoundQueryContext.scala:1134) ~[neo4j-cypher-interpreted-runtime-4.4.3.jar:4.4.3]
	... 87 more
```

2. Passing the secret UNESCAPED in the url string (eg [this not a real secret]): `ABCDefGHikbWUyzkbsIKapTBr%2Ffk4VkGsGWDOS%2F2XYZ`).  This fails with:

```
Caused by: com.amazonaws.services.s3.model.AmazonS3Exception: The AWS Access Key Id you provided does not exist in our records.
```

So [S3ParamsExtractor](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.4/core/src/main/java/apoc/util/s3/S3ParamsExtractor.java#L28) doesn't url-decode the secret key and [apoc.util.FileUtils$SupportedProtocols.from](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.4/core/src/main/java/apoc/util/FileUtils.java#L122) expects the url to be encoded, so they are in conflict with each other.

1. Construct an S3 URL with a `/` in the secret key and call a procedure such as `apoc.load.csv`.
2. Construct an S3 URL with a `%2F` in the secret key and call a procedure such as `apoc.load.csv`.

### Simple Dataset (where it's possible)

```

```

### Steps (Mandatory)

  1. Obtain an AWS secret key with a `/` in the value
  2. Try to load any S3 url (doesn't have to exist) while passing the credentials as part of the URL (either url-encoded or not)

## Specifications (Mandatory)

- The values extracted from S3 URL should be url-decoded

Currently used versions

### Versions

  - OS: macOS 11.2.3
  - Neo4j: 4.4.4
  - Neo4j-Apoc: 4.4.0.2

hindog:
Fixes #2724

- Fixes issue where secretKey/sessionToken are not url-decoded when extracted from S3 url (by converting the URL to `java.net.URI` instead of `java.net.URL`, which returns the values decoded)
- Fixes issue when parsing the bucket and key.  **PLEASE REVIEW and let me know if I'm missing something here.**

- Modify `S3ParamExtractor` to construct a `java.net.URI` from the S3 url so when we call `getUserInfo`, we get a decoded value (`java.net.URL#getUserInfo` returns the original encoded value).
- Change the bucket+key extraction logic to split on the **first** `/` (after the leading `/`) rather than **last**


hindog:
After testing some more, there is another "double url-encoding" issue that I believe is caused by a call to `apoc.util.FileUtils#changeFileUrlIfImportDirectoryConstrained`, which may be the same issue reported in https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/2714

leonpierre:
@hindog I'm not sure if this will fix our issue (#2714) with AWS Blob Url's, because S3 is working for us, and as far as I understand the fix, you only changed S3 code?!

vga91:
> After testing some more, there is another "double url-encoding" issue that I believe is caused by a call to `apoc.util.FileUtils#changeFileUrlIfImportDirectoryConstrained`, which may be the same issue reported in #2714

I think the same, this issue is fixed in 4.3 but not yet in 4.4.
Can you try to build your project by adding this changes https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2740 ?

hindog:
@vga91 thanks! I tested locally and confirmed the double-encoding issue is fixed after cherry-picking that commit.

@leonpierre that's correct.  After fixing a couple of S3-specific issues, I then hit the double-encoding issue as well, and started to investigate.  As @vga91 mentioned, there is already a fix for that issue in 4.3 that needs to be [pulled into 4.4](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2740/commits).  I pulled that change into this PR and confirmed the double-encoding issue is fixed for me now.

core/src/main/java/apoc/util/s3/S3ParamsExtractor.java
```diff
@@ -45,17 +52,17 @@ public static S3Params extract(URL url) throws  IllegalArgumentException {
         }
 
         // endpoint
-        String endpoint = url.getHost();
+        String endpoint = uri.getHost();
 
-        Integer slashIndex = url.getPath().lastIndexOf("/");
+        Integer slashIndex = uri.getPath().indexOf("/", 1);
```
ncordon, line 57: 👍

core/src/main/java/apoc/util/s3/S3ParamsExtractor.java
```diff
@@ -45,17 +52,17 @@ public static S3Params extract(URL url) throws  IllegalArgumentException {
         }
 
         // endpoint
-        String endpoint = url.getHost();
+        String endpoint = uri.getHost();
```
ncordon, line 56: There's a problem with this (and that's why the `LoadS3Test` tests are failing). The example I give is with http addresses, but it would be similar for S3 ones:

```scala
import java.net.{URL, URI}

new URL("http://us-east-1.127.0.0.1:55220/test-bucket/test.csv?accessKey=accesskey&secretKey=secretkey").getHost 
// us-east-1.127.0.0.1

new URI("http://us-east-1.127.0.0.1:55220/test-bucket/test.csv?accessKey=accesskey&secretKey=secretkey").getHost
// null
```

ncordon: I experimented with a fix for that in 803800d. Feel free to discard it if you have a better suggestion

PR #2728, opened 2022-04-11
Schema tests

AzuObs:
We fixed some tests and then enabled the file to run in CI.

- fixed tests
- enabled tests to run in CI

### DIFF
--- a/core/src/test/java/apoc/schema/SchemasEnterpriseFeaturesTest.java
+++ b/core/src/test/java/apoc/schema/SchemasEnterpriseFeaturesTest.java
@@ -1,7 +1,6 @@
 package apoc.schema;
 
 import apoc.util.Neo4jContainerExtension;
-import apoc.util.TestUtil;
 import org.apache.commons.lang3.StringUtils;
 import org.junit.AfterClass;
 import org.junit.Before;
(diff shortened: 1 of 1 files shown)


core/src/test/java/apoc/schema/SchemasEnterpriseFeaturesTest.java
```diff
@@ -39,23 +34,15 @@ public class SchemasEnterpriseFeaturesTest {
 
     @BeforeClass
     public static void beforeAll() {
-        assumeFalse(isRunningInCI());
```
conker84, line 42: Can you please just remove only this line?
We want to skip Container tests in case the environment does not have Docker up and running

conker84 (changes requested):
@AzuObs thank you so much, please look at my comment

core/src/test/java/apoc/schema/SchemasEnterpriseFeaturesTest.java
AzuObs: Hi Andrea, it's my first PR in this repo so apologies if I didn't know how it is supposed to work. 

I saw this line in the file but I didn't really understand why we would not want these tests to run in the CI. What is the reasoning behind that? 

It looks like the Github job has passed and the TeamCity job should pass too (not sure when it's run though). One thing which you may or may not be aware of is that recently Nacho has added docker to TeamCity/Github I believe.

ncordon (approved):

AzuObs:
[SchemasEnterpriseFeaturesTest not working in Apoc](https://trello.com/c/9f4mu0CF/78-schemasenterprisefeaturestest-not-working-in-apoc)

core/src/test/java/apoc/schema/SchemasEnterpriseFeaturesTest.java
AzuObs: Hi Andrea, 

Sorry to bother you with this but if I would like to see if I can merge this today. I'm thinking that if I don't hear back from you before tomorrow then I will assume that you are OK with this PR and that it's fine for these tests to run in CI 🙇

PR #2760, opened 2022-04-19
NO AUTO Ignores test failures only in TeamCity

ncordon:
Cherry-picks #2743 

## What
It ignores the test failures in TeamCity

## Why
Because we may have tests muted there and we don't want to see `Process exited with code 1 (Step: gen-artifacts (Gradle))` there in those cases.


nadja-muller:
[Whitespace handling Javacc](https://trello.com/c/2RxmsH00/19-whitespace-handling-javacc)

PR #2778, opened 2022-04-20
Relationship/Label Filters properties are comma-separated lists

mauvo:
Relationship Filters and Label Filters properties are both comma-separated sequences. This is explained, but only in the sequences section, at the very end of the page. I've tried to make the reader aware of sequences earlier, while still allowing them to initially focus on the simpler use-cases.

  -
  -
  -


AzuObs (approved):
LGTM. Thank you for your contribution!

PR #2779, opened 2022-04-20
AUTO: Update setting namespace to be internal instead of unsupported (#2771)

neo4j-oss-build:


Lojjs:
No need for these changes in older versions

PR #2784, opened 2022-04-21
Reverting path expander changes NO AUTO

docrozza:
This reverts commit 645a9e219f3aa02fdeae0a4d06380ac1651acc3b.

Builds are failing so reverting whilst investigating a full fix


docrozza:
Mmmmm

A PR build from yesterday in core repository also failed with the same errors I was seeing in APOC, i.e. before the changes about to be reverted were actually committed

Not sure this change is actually the culprit now so looking at what else might have changed recently for the test errors to surface

PR #2816, opened 2022-04-25
Runs all tests in CI

ncordon:
## What

* Removes the check `assumeFalse(isRunningInCI())` from everywhere.
* Cleans up associated resources in the tests.

## Why
* Our CI systems contain the docker containers, so the tests should be capable of running there.
* The database management services weren't being shutdown in most of the tests, which means we are probably consuming resources like threads and not cleaning them up.


core/src/test/java/apoc/algo/PathFindingTest.java
```diff
@@ -58,6 +56,10 @@ public void setUp() throws Exception {
    		TestUtil.registerProcedure(db, PathFinding.class);
    	}
 
+    @After
```
jexp, line 59: should this be AfterClass instead as we don't create/start the db per test method. and doing it per method would explode the test runtime.

core/src/test/java/apoc/convert/ConvertTest.java
```diff
@@ -28,6 +29,11 @@ public class ConvertTest {
     @ClassRule
     public static DbmsRule db = new ImpermanentDbmsRule();
 
+    @After
```
jexp, line 32: should this be AfterClass instead as we don't create/start the db per test method. and doing it per method would explode the test runtime.

core/src/test/java/apoc/create/CreateTest.java
```diff
@@ -40,6 +41,11 @@ public class CreateTest {
         TestUtil.registerProcedure(db,Create.class, Paths.class);
     }
 
+    @After
```
jexp, line 44: should this be AfterClass instead as we don't create/start the db per test method. and doing it per method would explode the test runtime.

core/src/test/java/apoc/cypher/CypherInitializerTest.java
```diff
@@ -29,20 +29,25 @@
...
         // we need at least on APOC proc being registered in finished CypherInitializers
-        TestUtil.registerProcedure(dbmsRule, Utils.class);
+        TestUtil.registerProcedure(db, Utils.class);
 
         waitForInitializerBeingFinished(SYSTEM_DATABASE_NAME);
         waitForInitializerBeingFinished(DEFAULT_DATABASE_NAME);
     }
 
+    @After
+    public void tearDown() {
```
jexp, line 47: should this be AfterClass instead as we don't create/start the db per test method. and doing it per method would explode the test runtime.

core/src/test/java/apoc/data/ExtractTest.java
```diff
@@ -23,6 +24,11 @@ public void setUp() throws Exception {
         TestUtil.registerProcedure(db, Extract.class);
     }
 
+    @After
+    public void tearDown() {
```
jexp, line 28: should this be AfterClass instead as we don't create/start the db per test method. and doing it per method would explode the test runtime.

core/src/test/java/apoc/example/ExamplesTest.java
```diff
@@ -23,6 +24,11 @@ public void setUp() throws Exception {
         TestUtil.registerProcedure(db,Examples.class);
     }
 
+    @After
+    public void tearDown() {
```
jexp, line 28: should this be AfterClass instead as we don't create/start the db per test method. and doing it per method would explode the test runtime.

ncordon:
This work was superseded by #2891

PR #2832, opened 2022-05-02
Fixes #2739: Allow apoc.spatial.geoCodeOnce to also receive config via procedure call

### LINKED ISSUE #2739: Allow apoc.spatial.geoCodeOnce to also receive config via procedure call, not just config file, especially for AuraDB

vga91:
Fixes #2739

- Added config map to geocodeOnce, geocode, reverseGeocode procedures
- Updated docs
- Created `getSupplierEntry` to mock tests

Other changes:
- Changed always-ignored test, even with assertion/other errors (`try{} catch() { Assume.assumeNoException("out of quota", e); `})
- Fixed openCage ignored tests (updated `spatial.json` file)
- Adjusted queries returned in`testCallCount`s (in `SpatialTest.java`) not to return false positives.


core/src/test/java/apoc/spatial/GeocodeTest.java
```diff
@@ -37,6 +41,38 @@ public void setUp() throws Exception {
...
+    public void testGeocodeOpenCageViaConfigMap() throws Exception {
+        Assume.assumeNotNull(OPENCAGE_KEY);
+        // overwrite ApocConfig provider
+        testGeocode("osm", 1000, false, 
+                map("provider", "opencage", "url", "https://api.opencagedata.com/geocode/v1/json?q=PLACE&key=KEY", "key", OPENCAGE_KEY));
+    }
+    
+    @Test
+    public void testReverseGeocodeOpenCageViaConfigMap() throws Exception {
+        Assume.assumeNotNull(OPENCAGE_KEY);
```
ncordon, line 69: These tests are always ignored in CI. How do we know they are doing what they are supposed to be doing? Can we mock or something?

ncordon: Also this particular test doesn't do anything with the key, and I don't know why. If this line gets commented out, the test passes

ncordon (changes requested):
I'm hesitant to approve something when it's not going to be run in the CI, nor we have information about how to run it

core/src/test/java/apoc/spatial/GeocodeTest.java
vga91: I added/updated some tests in the existing `SpatialTest.java` which mocks the `spatial.*` procedures

core/src/test/java/apoc/spatial/GeocodeTest.java
```diff
@@ -37,6 +42,38 @@ public void setUp() throws Exception {
         assumeRunningInCI();
```
ncordon, line 42: Sorry, but again, I wouldn't have tests that don't run in the CI nor locally and we don't have any instructions to run. To the best of my knowledge the OPENCAGE_KEY is not set in the CIs (both GitHub Actions or TeamCity). Are there any plans to set that there?

vga91: Ok sorry, I thought you just wanted me to add mocked tests.

However, I don't have access to TeamCity and therefore I don't know how to set / write the instructions to activate the `OPENCAGE_KEY`.

In any case, I have now added the mocked tests (`SpatialTest.java`).
So, could I just leave (in `GeocodeTest.java`) the osm tests, which don't require a key, and revert the opencage ones?

Although, being hard-coded present:
```
apocConfig().setProperty("apoc.spatial.geocode.opencage.key", "<YOUR_API_KEY>");
```
and 
```
Assume.assumeTrue(!"<YOUR_API_KEY>".equals(apocConfig().getString(Geocode.PREFIX +"." + provider + ".key")));
```
the tests regarding opencage are not performed in any case.

ncordon: First of all, so sorry for the delay reviewing.

The problem for me is precisely having tests in that are never run in the CI. There's an ongoing effort by @AzuObs to enable as many tests as possible in the CI, so I don't think this aligns with that. I'd prefer to have only mocked tests for this feature if you don't mind.

vga91: I updated the pr by deleting the additional ignored tests. @ncordon

ncordon: Thanks 👍

ncordon (approved):

PR #2838, opened 2022-05-03
Move security policy to .github

AzuObs:
Move security policy to `.github` so that the Github integration works


AzuObs:
Closing the PR as the Security Policy is working

PR #2841, opened 2022-05-03
Fixes #2723: apoc.load.xml on large file generates OoM Errors

### LINKED ISSUE #2723: apoc.load.xml on large file generates OoM Errors
## Feature description 
apoc.load.xml should be able to stream during loading of large xml files.

Currently a relatively small xml of 1GB can generate OoM heap errors even when used with a simple cypher statement. For example:

```
call apoc.load.xml("file:///var/lib/neo4j/import/small.xml") 
             yield value as releases
             unwind releases._children as release
             with release.id as releaseID,
               [item in release._children where item._type = "title"][0] as title
             MERGE (r:Release {id: releaseID})
             SET r.title = title._text;
```

Discussed with @jexp over slack. This is in contrast to `apoc.import.xml(file,config)` which does stream.

vga91:
Fixes #2723

Added more doc and tests about xpath, in order to use streaming

### DIFF
--- a/docs/asciidoc/modules/ROOT/partials/usage/apoc.load.xml.adoc
+++ b/docs/asciidoc/modules/ROOT/partials/usage/apoc.load.xml.adoc
@@ -277,6 +277,101 @@ RETURN author;
 | "Ralls, Kim"
 |===
 
+
+=== Avoid OOM using Xpath
+
+Generally, to avoid Heap Space Errors,
+to handle large files you should always try to return the result as a stream, and not as a unique result, to avoid `java.lang.OutOfMemoryError: Java heap space`, if possible
+For example, with a file like this:
+.book.xml
+[largeFile,xml]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- <graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd"> -->
+<graphml name="databases">
+<key id="name" for="node" attr.name="name"/>
+<key id="tagline" for="node" attr.name="tagline"/>
+<key id="title" for="node" attr.name="title"/>
+<key id="labels" for="node" attr.name="labels"/>
+<key id="summary" for="edge" attr.name="summary"/>
+<key id="label" for="edge" attr.name="label"/>
+<graph id="G" edgedefault="directed">
+  <node id="n0" labels=":Movie"><data key="labels">:Movie</data><data key="title">The Matrix</data><data key="tagline">Welcome to the Real World</data><data key="released">1999</data></node>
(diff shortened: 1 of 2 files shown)


docs/asciidoc/modules/ROOT/partials/usage/apoc.load.xml.adoc
```diff
@@ -277,6 +277,101 @@ RETURN author;
...
+[source,cypher]
+----
+CALL apoc.load.xml('largeFile.xml', '/graphml/*', {})
+----
+
+.Results
+[options="header"]
+|===
+| value
+| {"_children":[{"_type":"data","_text":":Movie","key":"labels"},{"_type":"data","_text":"The Matrix","key":"title"},{"_type":"data","_text":"Welcome to the Real World","key":"tagline"},{"_type":"data","_text":"1999","key":"released"}],"_type":"node","id":"n0","labels":":Movie"}
```
nadja-muller, line 331: This should be '/graphml/graph/*', shouldn't it?

docs/asciidoc/modules/ROOT/partials/usage/apoc.load.xml.adoc
```diff
@@ -277,6 +277,101 @@ RETURN author;
...
+| {"_type":"foo","id":"id2","_text":"foo2"}
+| {"_type":"foo","id":"id3","_text":"foo3"}
+|===
+
+
+Or if you want to include only `node` tag:
+
+[source,cypher]
+----
+CALL apoc.load.xml('largeFile.xml', '/graphml/node', {})
```
nadja-muller, line 346: This should be /graphml/graph/node as well, I believe.

nadja-muller (changes requested):
I think some of queries in the documentation need to be updated. Otherwise it looks good to me :+1:

docs/asciidoc/modules/ROOT/partials/usage/apoc.load.xml.adoc
nadja-muller: Also, I get slightly different results for my (outdated) 4.3 apoc version. I wonder if we might need different documentation there? :thought_balloon:

nadja-muller: So for clarification, the result I get from the first statement here is
`{"_type":"node","_children":[{"_type":"data","_text":":Movie","key":"l│
│abels"},{"_type":"data","_text":"The Matrix","key":"title"},{"_type":"│
│data","_text":"Welcome to the Real World","key":"tagline"},{"_type":"d│
│ata","_text":"1999","key":"released"}],"id":"n0","labels":":Movie"}`
So, I think the structure is inherently different. I neither updated my Neo4j version (4.3.2) nor apoc that I installed some time ago, so it might be because of that. But it's definitely interesting :smile:

nadja-muller (approved):

docs/asciidoc/modules/ROOT/partials/usage/apoc.load.xml.adoc
vga91: I see, yes it should depend on the version, thanks for the clarification

PR #2845, opened 2022-05-03
Fixes #2797 and #2780: Graph Refactoring procedures internally handle errors instead of throwing them, leaving half-created results

### LINKED ISSUE #2780: apoc.refactor.cloneNodes() not compatible with (existence + unique) or node key constraints

Execution of `apoc.refactor.cloneNodes()` (and other clone operations) should be able to succeed even when existence and node key constraints are on the nodes in question, provided that the outcome of the refactoring should not violate the constraints

The refactoring fails, producing an error indicating a constraint violation. This appears to be because the refactoring first creates empty nodes with the given labels and then later populates their properties, tripping on the existence requirements of the constraints.

Note that when ONLY an existence constraint is present the refactoring will succeed, but if either existence + unique constraints are present on the same properties, or a NODE KEY constraint is present, the attempt will fail.

We may need to reorder the operations such that the labels are applied after properties are set, instead of before.

### Steps (Mandatory)

  1. Use `:play movies` from the browser to create the sample graph in an empty db.
  1. Execute `CREATE CONSTRAINT ON (p:Movie) ASSERT (p.title, p.released) IS NODE KEY`
  1. Execute the following:

```
MATCH (m:Movie) 
WITH m 
LIMIT 5
CALL apoc.refactor.cloneNodes([m], true) YIELD input, output 
RETURN output
```

The error will occur.

Alternately, create both a unique constraint and an existence constraint on the same property and attempt the clone operation, the same kind of error should occur

## Specifications (Mandatory)

Currently used versions:

### Versions

  - Neo4j: 4.3.10, 4.4.4
  - Neo4j-Apoc: 4.3.x.x, 4.4.0.3

### LINKED ISSUE #2797: Graph Refactoring procedures internally handle errors instead of throwing them, leaving half-created results

Usage of `apoc.refactor.cloneNodes()`, when it encounters an error because of a uniqueness constraint, should throw an exception, which should cause the entire transaction to roll back.

At the least, even if this handles errors internally, it should leave results in a valid state: either fully created according to the refactoring, or rolled back.

The procedure handles the error internally with its own try/catch, and produces records that do reflect the error in the `error` yielded variable. 

However, it produces nodes that are in a half-applied state, which is messy and unexpected. For example, if running into a uniqueness constraint violation, it will produce a node with the desired label, but with no properties (or half-applied properties). 

On an empty database, create this uniqueness constraint:

```
CREATE CONSTRAINT Schema ON (n:Schema) ASSERT n.name IS UNIQUE;
```

Create some nodes, one of which abides by the constraint:

```
MERGE (n:Schema {name:"test1", updated:1234})-[r:SomeRel]-(m:Target) RETURN n
```

Attempt to clone the node. 

```
MATCH (n:Schema) 
CALL apoc.refactor.cloneNodes([n], true) YIELD input, output, error 
RETURN output, error
```

Note that while we do get an error related to the constraint violation, an empty :Schema node is still created.

At the least, we would expect that an error should cancel (or delete) the half-created node instead of leaving it in its half-created state.

It feels more correct that this should throw an exception instead of attempting to handle it itself.

Other refactoring procedures also seem to be handling their own errors instead of throwing exceptions, perhaps a review should happen on whether the current behavior here is correct or if it should change.

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: Any
  - Neo4j: 4.4.5
  - Neo4j-Apoc: 4.4.0.3

Likely applies to all prior versions of APOC.

vga91:
Fixes #2797
Fixes #2780

Tested in the procedures with `tx.createNode()` and `tx.createNode(label)`, 
excluding those concerning the system db and virtual entities.
The error is not present in `apoc.create. *`,, `apoc.generate. *`
`apoc.nodes.link`,` apoc.refactor. * `. `apoc.generate.`
The error is instead present, as well as in the `apoc.refactor. *` also in `apoc.import.graphml / csv` (created separated issue for export procs).

Regarding the refactor procs the error is caused by the catch that silently fails the procedure:
```
        try {
            final Node node = tx.createNode(Util.labels(labelNames));
.....
            return Stream.of(new NodeResult(setProperties(node, props)));
        } catch (Exception e) {
            return result.withError(e);
        }
```

Note that regarding the issue `2797` case, even with this pr `cloneNodes` cannot work due to `ConstraintException`. 
However, currently But it gives "wrong error", that is `must have the property XX` instead of` ConstraintException`, so it make sense to include it on this pr.

### DIFF
--- a/core/src/main/java/apoc/refactor/GraphRefactoring.java
+++ b/core/src/main/java/apoc/refactor/GraphRefactoring.java
@@ -19,6 +19,7 @@ import java.util.*;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
@@ -26,6 +27,7 @@ import java.util.stream.StreamSupport;
 import static apoc.refactor.util.PropertiesManager.mergeProperties;
 import static apoc.refactor.util.RefactorConfig.RelationshipSelectionStrategy.MERGE;
 import static apoc.refactor.util.RefactorUtil.*;
+import static apoc.util.Util.withTransactionAndRebind;
 
 public class GraphRefactoring {
     @Context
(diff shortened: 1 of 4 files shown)


AzuObs (approved):
LGTM. Nice work!

PR #2853, opened 2022-05-05
Fixes #2659: Apoc.import.graphml doesn't work for edges

### LINKED ISSUE #2659: Apoc.import.graphml doesn't work for edges
## Problem Statement
I am trying to export some data using the `apoc.export.csv.data` and then import using `apoc.import.graphml`.
When I run `apoc.export.csv.data` for extracting both nodes and rels in a single file and then import that single file using `apoc.import.graphml`, it works fine.

But for performance and parallelize it I am trying to run for nodes and rels separately.
So we get 2 files - one for nodes and other for relationships.
Then import each the nodes file first and then the relationship file. 
However the problem is `apoc.import.graphml` works fine for the nodes but then fails for rels giving error as  java null pointer exception.
The Relationship file does have the source and target node ids so it should be able to load the data. 

I can easily reproduce this using the following queries.
## Reproducible steps

//Extract node files
MATCH (person:Person)-[actedIn:ACTED_IN]->(movie:Movie)
WHERE person.name starts with "K"
with collect(person)+collect(movie) as node
CALL apoc.export.graphml.data(node, [], "/Users/rohankharwar/Downloads/movies_nodes.graphml", {stream: false})
YIELD file, nodes, relationships, properties, data
RETURN file, nodes, relationships, properties, data;

//Extract Relationship files
MATCH (person:Person)-[actedIn:ACTED_IN]->(movie:Movie)
WHERE person.name starts with "K"
with collect(person)+collect(movie) as node, collect(actedIn) as rels
CALL apoc.export.graphml.data([], rels, "/Users/rohankharwar/Downloads/movies_rels.graphml", {stream: false})
YIELD file, nodes, relationships, properties, data
RETURN file, nodes, relationships, properties, data;

//graphml file import nodes
CALL apoc.import.graphml("file:///Users/rohankharwar/Downloads/movies_nodes.graphml", {readLabels:true})

//graphml file import rels
CALL apoc.import.graphml("file:///Users/rohankharwar/Downloads/movies_rels.graphml", {readLabels:true})

We get the following error (attached image)

I just used the movies dataset. 

### Versions

  - OS: Linux
  - Neo4j: 4.3.10 , 4.4.4
  - Neo4j-Apoc: 4.3.0.5, 4.4.0.3

nullpointerexception

vga91:
Fixes #2659

Added 2 configs `source` and `target` with keys `label` and (optional) `id` to import rel-only files, e.g.:

```
call apoc.import.graphml("rels.graphml", {startLabel: 'Person', endLabel: 'Movie'})
```

Added same config for export graphml procs (currently with only `id` key) to export a custom property instead of current "n<internalId>" as a source / target attribute value

### DIFF
--- a/core/src/main/java/apoc/export/graphml/ExportGraphML.java
+++ b/core/src/main/java/apoc/export/graphml/ExportGraphML.java
@@ -69,6 +69,8 @@ public class ExportGraphML {
             XmlGraphMLReader graphMLReader = new XmlGraphMLReader(db, tx).reporter(reporter)
                     .batchSize(exportConfig.getBatchSize())
                     .relType(exportConfig.defaultRelationshipType())
+                    .source(exportConfig.getSource())
+                    .target(exportConfig.getTarget())
                     .nodeLabels(exportConfig.readLabels());
 
             if (exportConfig.storeNodeIds()) graphMLReader.storeNodeIds();
--- a/core/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
+++ b/core/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
@@ -1,6 +1,7 @@
 package apoc.export.graphml;
 
 import apoc.export.util.BatchTransaction;
+import apoc.export.util.ExportConfig;
 import apoc.export.util.Reporter;
 import apoc.util.JsonUtil;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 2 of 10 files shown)


vga91:
@ncordon 
The CI fails because there seems to be a problem with `apoc.load.html` tests using `geckodriver`,
i fixed it here https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2859.
Perhaps, locally works to me because i downloaded the driver some time ago.

vga91:
Well now it fails for another test I fixed here: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2869
I hope it's the last one to fix 😅

ncordon (approved):

PR #2859, opened 2022-05-06
Forced geckodriver version

vga91:
Forced geckodriver version for selenium,
otherwise a `No proper candidate URL to download geckodriver 0.31.0` will throws.

There seems to be some problem possibly related to this https://github.com/mozilla/geckodriver/issues/2009.

Indeed, on the release page https://github.com/mozilla/geckodriver/releases there is 0.30.0 as latest (instead of 0.31.0), this  appears to be suspicious.


AzuObs (approved):

AzuObs:
Good spot!

conker84:
@vga91 thanks and @AzuObs thanks for the review!

PR #2864, opened 2022-05-06
AUTO: Fixes #558: Check for neo4j version being compatible with neo version (#2700)

### LINKED ISSUE #558: Add automatic sanity checks for Neo4j/APOC version combination

neo4j-oss-build:


vga91:
Merged here: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2929

PR #2885, opened 2022-05-13
Fixes #2878: apoc 4.4.0.4 is reported as incompatible with neo4j-community-4.4.6 in neo4j.log

### LINKED ISSUE #2878: apoc 4.4.0.4 is reported as incompatible with neo4j-community-4.4.6 in neo4j.log
```
2022-05-09 10:06:02.935+0000 WARN  [neo4j/14c22383] The apoc version (4.4.0.4) and the Neo4j DBMS versions [4.4.6] are incompatible. 
See the compatibility matrix in https://neo4j.com/labs/apoc/4.4/installation/ to see the correct version
```

### Versions

  - OS: All
  - Neo4j: 4.4.6-community
  - Neo4j-Apoc: 4.4.0.4

vga91:
Fixes #2878

### DIFF
--- a/core/src/main/java/apoc/cypher/CypherInitializer.java
+++ b/core/src/main/java/apoc/cypher/CypherInitializer.java
@@ -4,6 +4,7 @@ import apoc.ApocConfig;
 import apoc.util.Util;
 import apoc.version.Version;
 import org.apache.commons.configuration2.Configuration;
+import org.apache.commons.lang3.StringUtils;
 import org.neo4j.common.DependencyResolver;
 import org.neo4j.configuration.Config;
 import org.neo4j.configuration.GraphDatabaseSettings;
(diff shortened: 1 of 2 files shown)


vga91:
@Lojjs Sorry about this annoying bug. Unluckily I tried it on a version where it worked 🤦.
Please, can you (or someone else of surface) review it?

Lojjs (commented):
This definitely looks more correct, but I wonder could we be certain that both Neo4j and APOC always will have the versioning separator `.`? 

If yes, I guess we can simplify and just hard-code it. 
If no, we can do the comparison independent of the separator so that e.g.  `4_4_5` is shown as compatible with `4.4.0.4` based on that the first two numbers are the same. 
What do you think?

vga91:
> This definitely looks more correct, but I wonder could we be certain that both Neo4j and APOC always will have the versioning separator `.`?
> 
> If yes, I guess we can simplify and just hard-code it. If no, we can do the comparison independent of the separator so that e.g. `4_4_5` is shown as compatible with `4.4.0.4` based on that the first two numbers are the same. What do you think?

Yes, the second option (4_4_5 = 4.4.0.4) probably makes more sense.
I just changed it in this way

vga91:
[[S4][Pfizer - Manufacturing Excellence Solutions]APOC reports mismatched version](https://trello.com/c/faqZs3Ja/993-s4pfizer-manufacturing-excellence-solutionsapoc-reports-mismatched-version)

PR #2887, opened 2022-05-13
Fixes #2824: Extend the error message for export procedures to also point out stream:true

### LINKED ISSUE #2824: Extend the error message for export procedures to also point out stream:true
Currently if exporting to files is disabled we get 

`Failed to invoke procedure `apoc.export.csv.query`: Caused by: java.lang.RuntimeException: Export to files not enabled, please set apoc.export.file.enabled=true in your neo4j.conf`

We should add to this:

```
Or if you are running in a cloud environment without filesystem access, use the `{stream:true}` parameter to stream the export back to your client.
```

vga91:
Fixes #2824

Changed `apoc.export.file.enabled=true in your neo4j.conf` with `apoc.export.file.enabled=true in your apoc.conf` in `.adoc`

Added note about export.xls: `Note that the stream mode cannot be used with the apoc.export.xls.* procedures.`

### DIFF
--- a/core/src/main/java/apoc/ApocConfig.java
+++ b/core/src/main/java/apoc/ApocConfig.java
@@ -93,7 +93,9 @@ public class ApocConfig extends LifecycleAdapter {
     ));
     private static final String DEFAULT_PATH = ".";
     private static final String CONFIG_DIR = "config-dir=";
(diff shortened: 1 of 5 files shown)


AzuObs (approved):
LGTM. Nice work!

PR #2891, opened 2022-05-16
Enable APOC Full Testing in CI

AzuObs:
In this PR we remove any kind of code which tries to detect whether we are running locally or in the CI. Once we remove this type of logic we can see that the code was not meant to be run in the CI to begin with, and so we begin fixing the tests by adding missing dependencies. There are some tests which can't fix right now because they contain individual errors and require investigations, those tests are marked with `@Ignore`

The PR is structured to be read commit by commit. Each commit tries to provide a useful and meaningful description message, and it also tries to be atomic.

  - Remove code which ignores tests specifically in the CI
  - Fix Neo4j docker image by exposing the right port
  - Add MySQL, Gephi, and GoogleCloudStorage dependencies
  - Delete NLP tests which don't use the mocks
  - Identify and ignore broken tests, we will create a ticket to fix these later


AzuObs:
[Make apoc-full tests run in CI](https://trello.com/c/EAg1inwP/143-make-apoc-full-tests-run-in-ci)

AzuObs:
[Make apoc-core tests run in CI](https://trello.com/c/6uBw3vO1/985-make-apoc-core-tests-run-in-ci)

AzuObs:
[Make apoc-full tests run in CI](https://trello.com/c/XgYQ5Py0/982-make-apoc-full-tests-run-in-ci)

full/src/test/java/apoc/couchbase/CouchbaseTestUtils.java
```diff
@@ -155,7 +148,6 @@ protected static void createCouchbaseContainer() {
         Cluster cluster = Cluster.connect(seedNodes, clusterOptions(USERNAME, PASSWORD).environment(environment));
 
         boolean isFilled = fillDB(cluster);
-        assumeTrue("should fill Couchbase with data", isFilled);
```
ncordon, line 158: Why was this particular line removed?

full/src/test/java/apoc/custom/CypherProceduresClusterTest.java
```diff
@@ -14,26 +14,23 @@
...
 
     private static TestcontainersCausalCluster cluster;
 
     @BeforeClass
     public static void setupCluster() {
-        TestUtil.ignoreException(() ->  cluster = TestContainerUtil
-                .createEnterpriseCluster(3, 1, Collections.emptyMap(), MapUtil.stringMap("apoc.custom.procedures.refresh", "100")),
-                Exception.class);
-        Assume.assumeNotNull(cluster);
-        assumeTrue("Neo4j Cluster should be up-and-running", cluster.isRunning());
```
ncordon, line 29: This line should have been left as well, right?

full/src/test/java/apoc/custom/CypherProceduresClusterTest.java
```diff
@@ -14,29 +13,27 @@
...
```
ncordon, line 29: I'm worried about this type of lines. Is spanning the docker container synchronous? Cause normally it isn't, and that would mean we need to bring back this line with retries. Something in the lines of:

```
eventually(cluster.isRunning(), backoff = 10s, maxTime = 1min)
```

test-utils/src/main/java/apoc/util/Neo4jContainerExtension.java
```diff
@@ -45,6 +45,9 @@ public Neo4jContainerExtension(String dockerImage) {
         // http on 4.0 seems to deliver a 404 first
         setDockerImageName(dockerImage);
 
+        addFixedExposedPort(7474, 7474);
+        addFixedExposedPort(7687, 7687);
+
```
ncordon, line 50: I'm not sure this is right sorry. Normally in a CI you don't want to be spanning a service in a fixed port cause a machine could have something already running there or it means you cannot parallelize several test suites. If you don't map the ports docker will choose one randomly for you and you can see which one was it by doing:
```
neo4jContainer.getMappedPort(7687)
```

test-utils/src/main/java/apoc/util/GoogleCloudStorageContainerExtension.java
```diff
@@ -0,0 +1,18 @@
+package apoc.util;
+
+import org.testcontainers.containers.BindMode;
+import org.testcontainers.containers.GenericContainer;
+
+public class GoogleCloudStorageContainerExtension extends GenericContainer<GoogleCloudStorageContainerExtension> {
+
+    public GoogleCloudStorageContainerExtension() {
+        super("fsouza/fake-gcs-server:latest");
+        this.addFixedExposedPort(4443, 4443);
```
ncordon, line 10: This is a fixed port and I'd rather we do a random port

full/src/test/java/apoc/metrics/MetricsTest.java
```diff
@@ -50,6 +51,7 @@ public static void afterAll() {
     }
 
     @Test
+    @Ignore
```
ncordon, line 54: Can you leave TODOs for these tests please? A TODO somehwere is easily searchable, the @Ignore task alone may not give us enough information on whether we need to un-ignore these tests

full/src/test/java/apoc/mongodb/MongoDBTest.java
```diff
@@ -247,6 +247,7 @@ public void testDelete() throws Exception {
     }
 
     @Test
+    @Ignore
```
ncordon, line 250: Can you leave TODOs for these tests please? A TODO somehwere is easily searchable, the @Ignore task alone may not give us enough information on whether we need to un-ignore these tests

ncordon (commented):
Really nice work, I left a few nitpicks

full/src/test/java/apoc/gephi/GephiMock.java
```diff
@@ -0,0 +1,134 @@
...
+                .map(property -> String.format("\"%s\":\"%s\"", property, property))
+                .collect(Collectors.joining(","));
+
+            return properties.isEmpty() ? "" : String.format(",%s", joined);
+        }
+
+        @Override
+        public String toRegexPattern() {
+            return String.format(
+                "\\{\"ae\":\\{\"%d\":\\{\"label\":\"%s\",\"TYPE\":\"%s\",\"source\":\"%d\",\"target\":\"%d\",\"directed\":true,\"weight\":%s,\"r\":%s,\"g\":%s,\"b\":%s%s\\}\\}\\}",
```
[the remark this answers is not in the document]
AzuObs: It is supposed to be `ae` which stands for "Add edge" in the GephiStream API.

This has made me realise that I forgot to add some documentation about my GephiMock and where the syntax I use comes from. I fixed that and included this link if you want to read more about their API (I wouldn't recommend reading it as there is nothing special about it).

https://github.com/gephi/gephi/wiki/GraphStreaming#supported-formats

full/src/test/java/apoc/custom/CypherProceduresClusterTest.java
AzuObs: Thanks for pointing this out.

There was a way of doing this already in the codebase for Neo4jContainerExtension. If you look at this [constructor method](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.4/test-utils/src/main/java/apoc/util/Neo4jContainerExtension.java#L44) then you'll see that it has a WaitStrategy which will only return once some conditions are met, or it will timeout and throw an error after 2 minutes.

The TestContainerCausalCluster setup we have is instantiating multiple Neo4jContainerExtension under the hood and will thus wait for them to all come up.

I had however forgotten to add a wait Strategy for my new GoogleCloudStorageContainerExtension and MySQLContainerExtension

AzuObs: Since I made the comment above a lot has changed for the Neo4jContainerExtension because we introduced the  .withWaitForNeo4jDatabaseReady function (and I'm removing the old logic from the constructor method in this PR).

test-utils/src/main/java/apoc/util/Neo4jContainerExtension.java
```diff
@@ -132,20 +137,28 @@ private Neo4jContainerExtension withWaitForDatabaseReady(
...
+                    .withReadTimeout(Duration.ofSeconds(3))
                     .withStartupTimeout(timeout));
         }
 
         return this;
     }
 
     public Neo4jContainerExtension withWaitForNeo4jDatabaseReady(String password, Neo4jVersion version) {
-        return withWaitForDatabaseReady("neo4j", password, "neo4j", Duration.ofSeconds(60), version);
+        return withWaitForDatabaseReady("neo4j", password, "neo4j", Duration.ofSeconds(150), version); // daniel
```
ncordon, line 161: I think you've forgotten this comment

ncordon (approved):
Looks good mate, awesome work!

Just remember to remove the comment either in this PR or in another one, please

PR #2898, opened 2022-05-17
Var docs fixes: path.expand, ttl, import.csv and group.nodes

vga91:
Cherry-pick `d12ef6d12e5bd48bb56f3de96a255887dd91a4b2`

I cannot push in `neo4j-contrib`,
anyway it should be not needed to trigger the CI because is a documentation-only pr.


conker84:
@Lojjs I'm merging this even if the CI is red as the changes are related only to doc files

PR #2903, opened 2022-05-18
AUTO: changed show functions/procedures with call dbms.*

neo4j-oss-build:


vga91:
Only for `4.2`. Not needed in this case

vga91:
Only for `4.2`. Not needed in this case

PR #2925, opened 2022-05-23
Ignore failing TeamCity Firefox tests

vga91:
Ignore failing TeamCity Firefox tests

Discussed internally on Slack , `apoc-dev` channel


ncordon (changes requested):
Sorry but this should be done only on TeamCity, we are good on GitHub actions, aren't we?

vga91:
> Sorry but this should be done only on TeamCity, we are good on GitHub actions, aren't we?

Yes, it's right, i just changed it.

ncordon:
This is not the way we should mute those tests in TeamCity. TeamCity has its own functionality so that is transparent what is muted or not. I've muted them manually.

PR #2935, opened 2022-05-25
fixes #2934: `apoc.meta.relTypeProperties` doest not return relationship information

### LINKED ISSUE #2934: `apoc.meta.relTypeProperties` doest not return relationship information

Given the movie database if we run the following query:

```cypher
CALL apoc.meta.relTypeProperties({includeRels: ['REVIEWED']})
```

I would expect the following result

```bash
+-------------------------------------------------------------------------------------------------------------------------------------------+
| relType       | sourceNodeLabels | targetNodeLabels | propertyName | propertyTypes | mandatory | propertyObservations | totalObservations |
+-------------------------------------------------------------------------------------------------------------------------------------------+
| ":`REVIEWED`" | ["Person"]       | ["Movie"]        | "rating"     | ["Long"]      | false     | 1                    | 1                 |
| ":`REVIEWED`" | ["Person"]       | ["Movie"]        | "summary"    | ["String"]    | false     | 1                    | 1                 |
+-------------------------------------------------------------------------------------------------------------------------------------------+
2 rows
```

The actual response from the procedure is:

```bash
+-------------------------------------------------------------------------------------------------------------------------------------------+
| relType       | sourceNodeLabels | targetNodeLabels | propertyName | propertyTypes | mandatory | propertyObservations | totalObservations |
+-------------------------------------------------------------------------------------------------------------------------------------------+
+-------------------------------------------------------------------------------------------------------------------------------------------+
0 rows
```

Just use the movie database

### Steps (Mandatory)

  1. `CALL apoc.meta.relTypeProperties({includeRels: ['REVIEWED']})`

### Versions

  - OS: MacOS
  - Neo4j: AuraDB
  - Neo4j-Apoc: 4.4.0.4

conker84:
Fixes #2934

Fixed the bug

  - Fixed the bug
  - Added a test case

### DIFF
--- a/core/src/main/java/apoc/cypher/CypherInitializer.java
+++ b/core/src/main/java/apoc/cypher/CypherInitializer.java
@@ -65,13 +65,17 @@ public class CypherInitializer implements AvailabilityListener {
                 }
 
                 if (defaultDb.equals(db.databaseName())) {
-                    final List<String> versions = db.executeTransactionally("CALL dbms.components", Collections.emptyMap(),
-                            r -> (List<String>) r.next().get("versions"));
-                    final String apocFullVersion = Version.class.getPackage().getImplementationVersion();
-                    if (isVersionDifferent(versions, apocFullVersion)) {
-                        userLog.warn("The apoc version (%s) and the Neo4j DBMS versions %s are incompatible. \n" +
-                                        "See the compatibility matrix in https://neo4j.com/labs/apoc/4.4/installation/ to see the correct version",
-                                apocFullVersion, versions.toString());
+                    try {
+                        final List<String> versions = db.executeTransactionally("CALL dbms.components", Collections.emptyMap(),
+                                r -> (List<String>) r.next().get("versions"));
+                        final String apocFullVersion = Version.class.getPackage().getImplementationVersion();
+                        if (isVersionDifferent(versions, apocFullVersion)) {
+                            userLog.warn("The apoc version (%s) and the Neo4j DBMS versions %s are incompatible. \n" +
+                                            "See the compatibility matrix in https://neo4j.com/labs/apoc/4.4/installation/ to see the correct version",
+                                    apocFullVersion, versions.toString());
+                        }
+                    } catch (Exception ignored) {
+                        userLog.info("Cannot check APOC version compatibility because of a transient error. Retrying your request at a later time may succeed");
                     }
                 }
                 Configuration config = dependencyResolver.resolveDependency(ApocConfig.class).getConfig();
--- a/core/src/main/java/apoc/meta/Meta.java
+++ b/core/src/main/java/apoc/meta/Meta.java
@@ -640,22 +640,24 @@ public class Meta {
                 long labelCount = countStore.get(labelName);
                 long sample = getSampleForLabelCount(labelCount, config.getSample());
 
-                //System.out.println("Sampling " + sample + " for " + labelName);
-
                 try (ResourceIterator<Node> nodes = tx.findNodes(label)) {
                     int count = 1;
                     while (nodes.hasNext()) {
                         Node node = nodes.next();
-                        if(count++ % sample == 0) {
-                            boolean acceptNode = !node.hasRelationship()
-                                    || Iterables.stream(node.getRelationshipTypes())
-                                    .map(RelationshipType::name)
-                                    .anyMatch(relName -> !excludeRels.contains(relName)
-                                            && (includeRels.isEmpty() || includeRels.contains(relName)));
-
-                            if (acceptNode) {
-                                profile.observe(node, config);
-                            }
+                        if (count++ % sample == 0) {
+                            final Set<Boolean> skips = StreamSupport
+                                    // we analyze the node for each its relationship type
+                                    .stream(node.getRelationshipTypes().spliterator(), false)
+                                    .map(rel -> excludeRels.contains(rel.name()) // we skip a node when the user said that must be excluded
+                                            // or when the user provided and inclusion list, but it's not in the provided list
+                                            || (!includeRels.isEmpty() && !includeRels.contains(rel.name())))
+                                    .collect(Collectors.toSet());
+                            // if the Set has just one element and is true we skip the node
+                            // if there are two elements [true, false] we don't skip it as give it means that
+                            // it have a relationship that satisfies the condition provided
+                            // by the configuration
+                            if (skips.size() == 1 && skips.iterator().next()) continue;
+                            profile.observe(node, config);
                         }
                     }
                 }
--- a/core/src/test/java/apoc/meta/MetaTest.java
+++ b/core/src/test/java/apoc/meta/MetaTest.java
@@ -4,6 +4,7 @@ import apoc.graph.Graphs;
 import apoc.util.MapUtil;
 import apoc.util.TestUtil;
 import apoc.util.Util;
+import org.apache.commons.io.IOUtils;
 import org.assertj.core.api.Assertions;
 import org.junit.Assert;
 import org.junit.Before;
@@ -29,6 +30,7 @@ import org.neo4j.values.storable.LocalTimeValue;
 import org.neo4j.values.storable.TimeValue;
 import org.neo4j.values.storable.Values;
 
+import java.io.InputStreamReader;
 import java.time.Clock;
 import java.time.LocalDate;
 import java.util.ArrayList;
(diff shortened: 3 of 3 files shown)


core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -65,14 +65,16 @@ public void available() {
...
+                    try {
+                        final List<String> versions = db.executeTransactionally("CALL dbms.components", Collections.emptyMap(),
+                                r -> (List<String>) r.next().get("versions"));
+                        final String apocFullVersion = Version.class.getPackage().getImplementationVersion();
+                        if (isVersionDifferent(versions, apocFullVersion)) {
+                            userLog.warn("The apoc version (%s) and the Neo4j DBMS versions %s are incompatible. \n" +
+                                            "See the compatibility matrix in https://neo4j.com/labs/apoc/4.4/installation/ to see the correct version",
+                                    apocFullVersion, versions.toString());
+                        }
+                    } catch (Exception ignored) {}
```
conker84, line 77: this is not related to this PR but I found some exceptions during the shutdown of the test database in the `MetaTest` class.
I can revert it if you want

moxious: ignored exceptions are not a great practice.  Recommend changing to (minimum) logging that exception.  This way you can address the test database issue without swallowing an error someone might need to know about.

moxious (commented):
mostly LGTM, just one comment on the swallowed exception.

core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -1539,4 +1540,36 @@ public void testMetaNodesCount() {
...
 
+    @Test
+    public void testRelTypePropertiesMovies() throws Exception {
+        final String query = IOUtils.toString(new InputStreamReader(Thread.currentThread().getContextClassLoader().getResourceAsStream("movies.cypher")));
+
+        db.executeTransactionally(query);
+        
+        TestUtil.testResult( db, "CALL apoc.meta.relTypeProperties($config)",
+                Map.of("config", Map.of("includeRels", List.of("REVIEWED"))), r -> {
+                    // System.out.println("r.resultAsString() = " + r.resultAsString());
```
Lojjs, line 1551: I think you should clean up this print line

core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -1539,4 +1540,36 @@ public void testMetaNodesCount() {
         
     }
 
+    @Test
+    public void testRelTypePropertiesMovies() throws Exception {
+        final String query = IOUtils.toString(new InputStreamReader(Thread.currentThread().getContextClassLoader().getResourceAsStream("movies.cypher")));
```
Lojjs, line 1545: You could use `getClass().getClassLoader()` instead of `Thread.currentThread().getContextClassLoader()`, I believe. I slightly prefer  it since it is simpler and used in other places, although it is not a big deal.

core/src/main/java/apoc/meta/Meta.java
```diff
@@ -642,27 +642,24 @@ private Tables4LabelsProfile collectTables4LabelsProfile (MetaConfig config) {
...
-                            }
-                            if (skipNode != true) {
-                                profile.observe(node, config);
-                            }
+                        if (count++ % sample == 0) {
+                            final Set<Boolean> skips = StreamSupport
+                                    // we analyze the node for each its relationship type
+                                    .stream(node.getRelationshipTypes().spliterator(), false)
+                                    .map(rel -> excludeRels.contains(rel.name()) // we skip a node when the user said that must be excluded
+                                            // or when the user provided and inclusion list, but it's not in the provided list
```
Lojjs, line 654: Nice with the comments!

core/src/main/java/apoc/meta/Meta.java
```diff
@@ -642,27 +642,24 @@ private Tables4LabelsProfile collectTables4LabelsProfile (MetaConfig config) {
...
+                                            || (!includeRels.isEmpty() && !includeRels.contains(rel.name())))
+                                    .collect(Collectors.toSet());
+                            // if the Set has just one element and is true we skip the node
+                            // if there are two elements [true, false] we don't skip it as give it means that
+                            // it have a relationship that satisfies the condition provided
+                            // by the configuration
+                            if (skips.size() == 1 && skips.iterator().next()) continue;
```
Lojjs, line 661: I think this is not quite right. Can't it happen that we have multiple elements and they are all true, like `[true, true]`? In that case we wouldn't skip now, but I think perhaps we should? Maybe we could add some test around such a case?

core/src/main/java/apoc/cypher/CypherInitializer.java
Lojjs: I agree, that we shouldn't do this

Lojjs (changes requested):
I think there is one case which might not work as it should, please have a look. Also some minor comments.

core/src/main/java/apoc/cypher/CypherInitializer.java
conker84: @moxious @Lojjs 
I ignored it because the stack trace is this one:
```bash
org.neo4j.graphdb.TransactionTerminatedException: The transaction has been terminated. Retry your operation in a new transaction, and you should see a successful result. The database is not currently available to serve your request, refer to the database logs for more details. Retrying your request at a later time may succeed. 
    	at org.neo4j.kernel.impl.api.KernelTransactionImplementation.assertOpen(KernelTransactionImplementation.java:688)
    	at org.neo4j.kernel.impl.newapi.AllStoreHolder.callProcedure(AllStoreHolder.java:1074)
    	at org.neo4j.kernel.impl.newapi.AllStoreHolder.procedureCallDbms(AllStoreHolder.java:1025)
    	at org.neo4j.cypher.internal.runtime.interpreted.CallSupport$.$anonfun$callDbmsProcedure$1(CallSupport.scala:59)
    	at org.neo4j.cypher.internal.runtime.interpreted.CallSupport$.callProcedure(CallSupport.scala:70)
    	at org.neo4j.cypher.internal.runtime.interpreted.CallSupport$.callDbmsProcedure(CallSupport.scala:59)
    	at org.neo4j.cypher.internal.runtime.interpreted.TransactionBoundReadQueryContext.callDbmsProcedure(TransactionBoundQueryContext.scala:1147)
    	at org.neo4j.cypher.internal.planning.ExceptionTranslatingReadQueryContext.callDbmsProcedure(ExceptionTranslatingQueryContext.scala:235)
    	at org.neo4j.cypher.internal.runtime.DbmsCallMode$.callProcedure(ProcedureCallMode.scala:87)
    	at org.neo4j.cypher.internal.runtime.interpreted.pipes.ProcedureCallPipe.call(ProcedureCallPipe.scala:88)
    	at org.neo4j.cypher.internal.runtime.interpreted.pipes.ProcedureCallPipe.$anonfun$internalCreateResultsByAppending$1(ProcedureCallPipe.scala:74)
    	at org.neo4j.cypher.internal.runtime.ClosingIterator$$anon$1.nextCur(ClosingIterator.scala:107)
    	at org.neo4j.cypher.internal.runtime.ClosingIterator$$anon$1.innerHasNext(ClosingIterator.scala:113)
    	at org.neo4j.cypher.internal.runtime.ClosingIterator.hasNext(ClosingIterator.scala:93)
    	at org.neo4j.cypher.internal.runtime.ClosingIterator$$anon$3.innerHasNext(ClosingIterator.scala:152)
    	at org.neo4j.cypher.internal.runtime.ClosingIterator.hasNext(ClosingIterator.scala:93)
    	at org.neo4j.cypher.internal.runtime.interpreted.PipeExecutionResult.serveResults(PipeExecutionResult.scala:85)
    	at org.neo4j.cypher.internal.runtime.interpreted.PipeExecutionResult.request(PipeExecutionResult.scala:73)
    	at org.neo4j.cypher.internal.result.StandardInternalExecutionResult.request(StandardInternalExecutionResult.scala:90)
    	at org.neo4j.cypher.internal.javacompat.ResultSubscriber.fetchResults(ResultSubscriber.java:362)
    	at org.neo4j.cypher.internal.javacompat.ResultSubscriber.materialize(ResultSubscriber.java:92)
    	at org.neo4j.cypher.internal.result.StandardInternalExecutionResult.initiate(StandardInternalExecutionResult.scala:65)
    	at org.neo4j.cypher.internal.result.ClosingExecutionResult.$anonfun$initiate$1(ClosingExecutionResult.scala:58)
    	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
    	at org.neo4j.cypher.internal.result.ClosingExecutionResult.safely(ClosingExecutionResult.scala:99)
```

If we log it we have to be very careful about the message that we return because otherwise it could confuse the users.
What about `Cannot check APOC version compatibility because of a transient error. Retrying your request at a later time may succeed`?

core/src/test/java/apoc/meta/MetaTest.java
conker84: done

core/src/test/java/apoc/meta/MetaTest.java
conker84: ok

core/src/main/java/apoc/meta/Meta.java
conker84: No `skips` collection is a `Set` so it contains only unique values like `[true, false]` or `[false, true]`, the collection will store only one for each boolean type.

Lojjs: Ah, for some reason I thought it was a list and not a set, despite the comment. You are right!

Lojjs (approved):
@conker84 All good now, feel free to merge

PR #2951, opened 2022-05-30
Changing store version type in apoc.monitor.kernel()

PetrJanouch:
Database format (usually referred to as store version)
used to be identified by 8 byte integer. This is no longer
true from 5.0, so the procedure apoc.monitor.kernel()
that unfortunately exposes this kernel detail has to be
modified accordingly.


Lojjs (approved):
Looks good

PR #2956, opened 2022-05-30
Remove legacy neo4j settings.

MishaDemianenko:
Migrate tests to use apoc settings instead.


MishaDemianenko:
@Lojjs please wait with looking into this yet, will have some news about this one in 1-2 days

MishaDemianenko:
It was decided that we do not do any cleanup of apoc namespace and keep it as it is. Closing

PR #2983, opened 2022-06-14
5.0 fix security review issues

gem-neo4j:
Based on the security code review I have:
* Removed unused imports (and added a spotless checker to enforce this/clean up newly introduced ones easier)
* Changed how we used FileStreamInput
* Handled thrown errors or added comments when they are to be ignored
* Removed the return of stack traces in non test cases


gem-neo4j:
[APOC security review - FileInputStream In Use](https://trello.com/c/Au1BH91g/96-apoc-security-review-fileinputstream-in-use)

gem-neo4j:
[APOC security review - Unused Imports](https://trello.com/c/odaCRzm9/99-apoc-security-review-unused-imports)

gem-neo4j:
[APOC security review - Empty Catch Block](https://trello.com/c/b6gFA6Ka/95-apoc-security-review-empty-catch-block)

gem-neo4j:
[APOC security review - Stack Trace Returned](https://trello.com/c/Jya3Drb8/98-apoc-security-review-stack-trace-returned)

core/build.gradle
```diff
@@ -3,6 +3,14 @@ plugins {
...
     id "org.sonarqube"
+    id "com.diffplug.spotless" version "6.7.2"
+}
+
+spotless {
+    java {
+        target 'src/*/java/**/*.java'
+        removeUnusedImports()
+    }
 }
```
ncordon, line 14: This is only affecting to java files, but I believe there's some Kotlin in the database as well right?

full/build.gradle
```diff
@@ -7,6 +7,14 @@ plugins {
...
     id "org.jetbrains.kotlin.jvm" version "1.6.0"
+    id "com.diffplug.spotless" version "6.7.2"
+}
+
+spotless {
+    java {
+        target 'src/*/java/**/*.java'
+        removeUnusedImports()
+    }
 }
```
ncordon, line 18: This is only affecting to java files, but I believe there's some Kotlin in the full codebase

core/src/main/java/apoc/convert/Convert.java
```diff
@@ -214,7 +214,9 @@ private Long parseLongString(String input) {
         } catch (NumberFormatException nfe) {
             try {
                 return (long)Double.parseDouble(input);
-            } catch(NumberFormatException nfe2) {}
+            } catch(NumberFormatException nfe2) {
+                // String was not able to be parsed, return null
```
ncordon, line 218: Maybe do `return null` directly here. As it is, the catch block is still empty as the ticket was saying. Not that I have a big issue with that really 🤷

core/src/main/java/apoc/util/DateParseUtil.java
```diff
@@ -28,7 +28,8 @@ public static TemporalAccessor dateParse(String value, Class<? extends TemporalA
                             return getParse(date, value);
                         }
                     } catch (Exception e) {
-                        continue;
+                        // Ignore here as we are in a loop checking formats
+                        // if all formats fail, a later exception will be called
                     }
```
ncordon, line 33: Same, empty catch block still

full/build.gradle
ncordon: Discussed during standup and we decided we can skip the Kotlin sources

ncordon (approved):

PR #2985, opened 2022-06-15
Fixes #2931: The apoc.import.csv procedure doesn't batch correctly

### LINKED ISSUE #2931: The apoc.import.csv procedure doesn't batch correctly

The import procedure should batch correctly with big files and create multiple transactions.

There are actually two problems.
- `java.lang.ClassCastException` using `batchSize` config

- Even solving the above, the entire file is imported in one transaction

Create a file relatively small, about 6mb.

Set in `neo4j.conf` -> `dbms.memory.heap.initial_size=200m` and `dbms.memory.heap.max_size=200m` .

```
call apoc.import.csv([{fileName: 'largeFile.csv', labels: ["Test"]}], [], {batchSize: 100})
```
this will throw:
```
Failed to invoke procedure `apoc.import.csv`: Caused by: java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')
```

Even after solving the previous problem, the import will be throw a`java.lang.OutOfMemoryError: Java heap space`.

Even considering the different structures (and then thesize differences), the analogous `graphml` procedure works with much larger files, with the same configuration (see [`testImportGraphMLLargeFile` in `[ExportGraphMLTest.java`](https://github.com/neo4j-contrib/neo4j-apoc-procedures/commit/4f3aa1277bd046be700160cdcd8fc1bda8d054dc#diff-9318bb02f88e7b32b5fa30e2bb29a0a69f7c45471ea84e1768883c5f8a517a8fR191))

### Versions

  - OS: MacOs
  - Neo4j: 4.4.5
  - Neo4j-Apoc: 4.4.0.5

vga91:
Fixes #2931

- Changed `csv.readAll()` which read the entire csv without batch it
- Added `btx.increment()` 
- Added `largeFile.csv` test


Lojjs (approved):
Nice!

PR #2987, opened 2022-06-15
Added support for Duration to apoc.coll.avg

vga91:
Add support for Duration to apoc.coll.avg via new procedure `apoc.coll.avgDuration` (added in `full`)

Trello card: https://trello.com/c/2tfco9c9/791-add-support-for-duration-to-apoccollavg

Based on `org.neo4j.cypher.internal.runtime.interpreted.pipes.aggregation.AvgFunction`


vga91:
[Add support for Duration to apoc.coll.avg](https://trello.com/c/2tfco9c9/791-add-support-for-duration-to-apoccollavg)

full/src/test/java/apoc/coll/CollFullTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.neo4j.test.rule.DbmsRule;
+import org.neo4j.test.rule.ImpermanentDbmsRule;
+import org.neo4j.values.storable.DurationValue;
+
+import static org.junit.Assert.assertEquals;
+
+public class CollFullTest {
```
Lojjs, line 13: I would like to see more test coverage, at least the following inputs to apoc.coll.avgDuration:

- empty list - should return null
- null - should return null?
- a list of another type e.g. [1, 2, 3] - should throw some helpful error
- a list with mixed durations and numbers e.g. [duration(''P2DT4H1S'), 2] - should throw some helpful error
- big enough durations and/or long enough list to cause an overflow problem without running avg (see comment on the implementation, I think there is a limit when the sum of all of the durations hit Long.MAX seconds)

Lojjs: It would maybe also make sense to document some of these edge case behaviours.

Lojjs (changes requested):
The happy path works well, but I think it would be good with some updates and tests around corner cases, see comments

full/src/test/java/apoc/coll/CollFullTest.java
vga91: Added

full/src/main/java/apoc/coll/CollFull.java
```diff
@@ -0,0 +1,40 @@
...
+            secondsRunningAvg.addAndGet(duration.get(ChronoUnit.SECONDS));
+            nanosRunningAvg.addAndGet(duration.get(ChronoUnit.NANOS));
+        }
+
+        return DurationValue.approximate(
+                monthsRunningAvg.get() / count,
+                daysRunningAvg.get() / count,
+                secondsRunningAvg.get() / count,
+                nanosRunningAvg.get() / count
+        ).normalize();
```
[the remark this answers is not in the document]
vga91: I've changed the implementation to be totally consistent with the Neo4j one.
Anyway, i think the probability of an overflow is almost 0.
For example, this query `WITH [idx in range(0,99999999) | $dur] AS dur RETURN apoc.coll.avgDuration(dur)`, where `$dur=DurationValue.MAX_VALUE`, the double values sum is obtained is less than  `9.3E26`, so far below the max value.

Lojjs (approved):
Thanks for the amendments, looks fine now

PR #2996, opened 2022-06-17
Fix for #2909.

### LINKED ISSUE #2909: Apoc Procedures can't return empty map
## BUG
Inside APOC procedures is a bug, that prevents you to return an empty map.  
When returning empty map {} the APOC raise exception : 

`Expected positive int value, got 0`

Create APOC procedure that returns an empty map :

```
CALL apoc.custom.asProcedure(
  'test',
  'RETURN {} AS res;',
  'write',
  [['res', 'map']],
  []
);

```

`call custom.test; `

returns:

`Expected positive int value, got 0`

### Versions

  - OS: mac and linux
  - Neo4j: neo4j:4.4.6-community
  - Neo4j-Apoc: apoc-4.4.0.3

tomocesnik:
Fixes #2909

Instead of passing the actual size of the map (which could be 0) to the constructor for MapValueBuilder, the constructor is called without parameters. This functionality is provided by Neo4j and for some reason the constructor with the parameter throws an exception for value 0.

### DIFF
--- a/full/src/main/java/apoc/custom/CypherProceduresHandler.java
+++ b/full/src/main/java/apoc/custom/CypherProceduresHandler.java
@@ -584,7 +584,7 @@ public class CypherProceduresHandler extends LifecycleAdapter implements Availab
                     return VirtualValues.list(objects);
                 } else if (toConvert instanceof Map) {
                     Map<String, Object> map = (Map) toConvert;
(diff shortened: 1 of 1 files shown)


Lojjs (approved):
Thanks for the contribution @tomocesnik, this makes sense to me. I will take care of merging and porting it to the other versions.

PR #3005, opened 2022-06-22
Fixes #3000: apoc.refactor.cloneNodes withRelationships = true - hangs and does not complete call

### LINKED ISSUE #3000: apoc.refactor.cloneNodes withRelationships = true - hangs and does not complete call
When I call apoc.refactor.cloneNodes with the relationships boolean as true, the call simply hangs, I have provided a simple example below.

Return clone of nodes with relationships

Call to apoc.refactor.cloneNodes just hangs forever

CREATE (a:Person {name: "Mark", city: "London"})
CREATE (b:Person {name: "Dan", city: "Hull"})
CREATE (a)-[r:FRIENDS_WITH]->(b)
RETURN a,b

MATCH (p:Person)
WITH collect(p) AS people
CALL apoc.refactor.cloneNodes(people, true)
YIELD input, output
RETURN input, output;

### Steps (Mandatory)

  1.  Execute the cypher above to create person nodes and relationship
  2. Execute the cypher to clone the person nodes with relationships

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: Windows 11
  - Neo4j: 4.4.5
  - Neo4j-Apoc: 4.4.0.6

vga91:
Fixes #3000

The bug appeared with [this pr](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2845/files), in which i've added explicit transaction where needed.
In the issue case, the nodes included in the `apoc.refactor.cloneNodes` are connected to each other,
so we have to close the node creation's transaction first, 
and after we can create "mixed" relationships between the original and cloned nodes.

### DIFF
--- a/core/src/main/java/apoc/refactor/GraphRefactoring.java
+++ b/core/src/main/java/apoc/refactor/GraphRefactoring.java
@@ -55,11 +55,11 @@ public class GraphRefactoring {
 
                     newNode = copyProperties(properties, newNode);
                     copyLabels(node, newNode);
-                    if (withRelationships) {
-                        copyRelationships(node, newNode, false, true);
-                    }
                     return newNode;
                 });
+                if (withRelationships) {
+                    copyRelationships(node, copy, false, true);
+                }
                 return result.withOther(copy);
             } catch (Exception e) {
                 return result.withError(e);
--- a/core/src/test/java/apoc/refactor/GraphRefactoringTest.java
+++ b/core/src/test/java/apoc/refactor/GraphRefactoringTest.java
@@ -16,6 +16,7 @@ import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.QueryExecutionException;
 import org.neo4j.graphdb.Relationship;
 import org.neo4j.graphdb.RelationshipType;
+import org.neo4j.graphdb.ResourceIterator;
 import org.neo4j.internal.helpers.collection.Iterators;
 import org.neo4j.test.rule.DbmsRule;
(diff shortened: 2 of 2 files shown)


Lojjs (approved):
I think this looks correct. Just out of curiosity, is `apoc.refactor.cloneNodesWithRelationships(list)` and `apoc.refactor.cloneNodes(list, true)` doing the same thing? The documentation [here](https://neo4j.com/labs/apoc/4.4/graph-updates/graph-refactoring/clone-nodes/#_clone_nodes_only) doesn't talk about what the `withRelationships` does, but when I tried them out, they seem to be identical

mark-sts:
From a user/developer point of view, the apoc.refactor.cloneNodesWithRelationships(list) was also not working, but with it being depreciated I assumed it had been dropped and I should be using apoc.refactor.cloneNodes(list, true), that is when I tripped across the issue

vga91:
> I think this looks correct. Just out of curiosity, is `apoc.refactor.cloneNodesWithRelationships(list)` and `apoc.refactor.cloneNodes(list, true)` doing the same thing? The documentation [here](https://neo4j.com/labs/apoc/4.4/graph-updates/graph-refactoring/clone-nodes/#_clone_nodes_only) doesn't talk about what the `withRelationships` does, but when I tried them out, they seem to be identical

Yes, the behavior is exactly the same. The documentation is not exhaustive indeed, i created [a pr](https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/3008) for it.

PR #3014, opened 2022-06-27
Fixes neo4j/apoc#22: Handle incorrect JSON documents in apoc.convert.fromJsonMap

vga91:
Fixes https://github.com/neo4j/apoc/issues/22

- Added `failOnError` like `apoc.load.json` for procedures using `JsonPath.parse(...)`, 
basically all of these neo4j-contrib@5262870 

- Added `apoc.json.validate` procedure via a `SilentDeserializer` class, 
which skip in case of errors during the deserializing and returns a result with the failed keys

Other fixes:
- changed wrong `apoc.load.json` and `apoc.load.jsonArray` query with `pathOptions` config


vga91:
[[S4][Citrix Systems - Analytics] - Handle incorrect JSON documents in apoc.convert.fromJsonMap](https://trello.com/c/GFarQTcA/1005-s4citrix-systems-analytics-handle-incorrect-json-documents-in-apocconvertfromjsonmap)

docs/asciidoc/modules/ROOT/pages/overview/apoc.json/apoc.json.path.adoc
```diff
@@ -23,7 +23,8 @@ apoc.json.path(json :: STRING?, path = $ :: STRING?, pathOptions = null :: LIST?
 | Name | Type | Default 
 |json|STRING?|null
 |path|STRING?|$
-|pathOptions|LIST? OF STRING?|[]
+|pathOptions|LIST? OF STRING?|null
+|failOnError|BOOLEAN?|true
 |===
```
AzuObs, line 28: I'm looking at the other json functions and notice that those seem to take failOnError in a config object. Sometimes the config object only has this parameter, such is the case for apoc.load.jsonParams. 

To me it seems like it would be better to have this option in a Config map for consistency. What do you think?

vga91: Yeah, it makes sense.
I've now created a `JsonConfig`, similar e.g. to `CompressionConfig`, to handle any future configurations.

core/src/main/java/apoc/convert/Json.java
```diff
@@ -81,10 +82,19 @@ private static Map<String, Object> mapWithOptionalProps(Map<String,Object> mapEn
...
+    @Procedure("apoc.json.validate")
+    @Description("apoc.json.validate('{json}' [,'json-path' , 'path-options']) - to check if the json is correct (returning an empty result) or not")
+    public Stream<StringResult> validate(@Name("json") String json, @Name(value = "path",defaultValue = "$") String path, @Name(value = "pathOptions", defaultValue = "null") List<String> pathOptions) {
+        return ((List<String>) JsonUtil.parse(json, path, Object.class, pathOptions, false, true))
+                .stream()
+                .map(StringResult::new);
+    }
+
     @UserFunction("apoc.json.path")
     @Description("apoc.json.path('{json}' [,'json-path' , 'path-options'])")
```
AzuObs, line 95: I think the description may need a new "config" input parameter?

core/src/main/java/apoc/util/JsonUtil.java
```diff
@@ -115,19 +119,49 @@ public static <T> T parse(String json, String path, Class<T> type) {
...
+            ObjectMapper objectMapper = getObjectMapper(deserializer);
             final String listOpt = Option.ALWAYS_RETURN_LIST.name();
             if (type == Map.class && options != null && options.contains(listOpt)) {
                 throw new RuntimeException("It's not possible to use " + listOpt + " option because the conversion should return a Map");
             }
             if (path == null || path.isEmpty()) {
-                return OBJECT_MAPPER.readValue(json, type);
+                final T t = (T) objectMapper.readValue(json, Object.class);
+                return getJson(deserializer, t, validation);
             }
```
AzuObs, line 142: Unfortunately, I believe that deserializing any json to an Object is a fairly Jackson vulnerability. I found [this article](https://security.snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-1056418) about it which could explain it better than I could. 

The idea is that an attacker can make you build any class on your classpath that they like. And one of those classes on your class path might take java as a string in their constructor method, and run that java essentially.

If I'm correct and this is vulnerable, then I think there are several solutions that could be considered from the article. I would leave the implementation up to you. But I think that we ought to do what's possible to not introduce this vulnerability if possible.

docs/asciidoc/modules/ROOT/pages/import/includes/jsonpath.adoc
```diff
@@ -66,12 +66,19 @@ or, with custom path options:
 
 [source, cypher]
 ----
-CALL apoc.load.json($url, '$..columns', ['ALWAYS_RETURN_LIST']);
+CALL apoc.load.json($url, '$..columns', {pathOptions: ['ALWAYS_RETURN_LIST']});
```
AzuObs, line 69: Missing "config"

core/src/main/java/apoc/convert/Json.java
```diff
@@ -81,10 +82,19 @@ private static Map<String, Object> mapWithOptionalProps(Map<String,Object> mapEn
...
 
+
+    @Procedure("apoc.json.validate")
+    @Description("apoc.json.validate('{json}' [,'json-path' , 'path-options']) - to check if the json is correct (returning an empty result) or not")
+    public Stream<StringResult> validate(@Name("json") String json, @Name(value = "path",defaultValue = "$") String path, @Name(value = "pathOptions", defaultValue = "null") List<String> pathOptions) {
+        return ((List<String>) JsonUtil.parse(json, path, Object.class, pathOptions, false, true))
+                .stream()
+                .map(StringResult::new);
+    }
+
```
AzuObs, line 93: There is a bit of a discussion in the team currently about adding new methods to APOC. Would you mind allowing us a few days to get back to you about this?

AzuObs:
I am removing myself from this PR for now, but if you wish to make some changes at some point then feel to add me back as "reviewed".

vga91:
If I remember correctly, I waited to move the `apoc.util.validate` procedure to `full/extended`, because I asked the customer if he was using aura or not, but I didn't receive an answer. 
However, considering that there is also a github issue opened by another user, and in order not to waste the work done, I still moved the procedure to extended, presumably there are other users to whom it can be useful.

AzuObs:
Hi Giuseppe, thank you for making the changes requests so far in the PR.

Unfortunately I think that moving the procedure is a breaking change, isn't it? Basically some users are going to only be using core (Aura users for example), but now all of a sudden this procedure is not going to be available.  As such I'm not sure we should move it to Extended.

I would also think that it being in Core would mean that most users are able to use it. Basically I expect every user who has APOC installed has at least APOC Core installed, because APOC Full includes APOC Core.

vga91:
> Hi Giuseppe, thank you for making the changes requests so far in the PR.
> 
> Unfortunately I think that moving the procedure is a breaking change, isn't it? Basically some users are going to only be using core (Aura users for example), but now all of a sudden this procedure is not going to be available. As such I'm not sure we should move it to Extended.
> 
> I would also think that it being in Core would mean that most users are able to use it. Basically I expect every user who has APOC installed has at least APOC Core installed, because APOC Full includes APOC Core.

Hi Daniel, maybe I wasn't totally clear with "i moved the procedure", actually it's a new procedure which I created in this pr. I had initially created it in core, but we had decided to move to full in order not to receive new core functionality.
For this reason we asked the client of the trello card if he was using aura.

AzuObs:
So sorry, I got confused when I checked the APOC docs and Google showed me (apoc.util.validate)[https://neo4j.com/labs/apoc/4.2/overview/apoc.util/apoc.util.validate/]

Good call to move this to Extended :+1:

AzuObs (approved):

full/src/main/java/apoc/json/JsonExtended.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.stream.Stream;
+
+
+@Extended
+public class JsonExtended {
+
+    @Context
+    public org.neo4j.graphdb.GraphDatabaseService db;
+
+    @Procedure
```
jexp, line 21: wouldn't it be better if this was a function that returned a list, so that people could just say  `WHERE size(apoc.json.validate(json)) = 0` ...

jexp (commented):
Can we make it a function instead of a procedure then it is much easier to use it in boolean e xpressions and still get the list of errors strings if needed

otherwise people will always have to use subqueries and collect which is painful for no good reason.

PR #3017, opened 2022-06-27
Use non-deprecated UserFunctionSignature constructor

Lojjs:
The UserFunctionSignature constructor which APOC used was deprecated in Neo4j and will be removed. The allowed field is not used in 5.0 and there is two new fields for Neo4j builtIn and internal functions, which should be false for everything in APOC.

This PR should only go into dev, and not be backported.


Lojjs:
[Update APOC dev to use latest UserFunctionSignature](https://trello.com/c/4ajvjBZP/137-update-apoc-dev-to-use-latest-userfunctionsignature)

nadja-muller (approved):

PR #3022, opened 2022-06-30
Fix testLoadLDAP

vga91:
As can be guessed, the `ldap.forumsys.com` server is not reachable at the moment, as we can see [here](https://www.filestash.app/ldap-test-tool.html?host=ldap%3A%2F%2Fldap.forumsys.com).

In this pr i used another online server.

Otherwise, to avoid possible similar errors in the future, maybe we could use an embedded mock server, such as https://github.com/zapodot/embedded-ldap-junit, or other test solutions.


ncordon (approved):
This is a good temporal fix for the moment, but we do really need to take care of this with dependencies we control. I'll card this on the wall

PR #3035, opened 2022-07-06
Export cypher overwrite relationships in case of merge with multi-rels with the same nodes and type

vga91:
Export cypher overwrite in case of merge with multi-relationships

Added `multipleRelationshipsWithType` config, default `false, to distinguish (with true) multiple relationships with the same start/end nodes and same rel-type.

Related Trello issue: https://trello.com/c/qjfKSOqG/971-s4volvo-apocexportcypherquery-loses-relationships

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/export/cypher.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/export/cypher.adoc
@@ -65,6 +65,7 @@ The procedures support the following config parameters:
 | awaitForIndexes | Long | 300 | Timeout to use for `db.awaitIndexes` when using `format: "cypher-shell"`
 | saveIndexNames | boolean | false | Save name indexes on export
 | saveConstraintNames | boolean | false | Save name constraints on export
+| multipleRelationshipsWithType | boolean | false | In case of multiple relationships of the same type between two nodes, add a `UNIQUE IMPORT ID REL` property in order to distinguish them when using `MERGE`.
 |===
 
 [[export-cypher-file-export]]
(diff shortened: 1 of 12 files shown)


docs/asciidoc/modules/ROOT/pages/export/cypher.adoc
```diff
@@ -706,6 +707,70 @@ RETURN nodes, relationships, properties, nodeStatements, relationshipStatements,
...
+    (pers)-[:WORKS_FOR {id: 2}]->(proj),
+    (pers)-[:WORKS_FOR {id: 2}]->(proj),
+    (pers)-[:WORKS_FOR {id: 3}]->(proj),
+    (pers)-[:WORKS_FOR {id: 4}]->(proj),
+    (pers)-[:WORKS_FOR {id: 5}]->(proj),
+    (pers)-[:IS_TEAM_MEMBER_OF {name: 'aaa'}]->(:Team {name: 'one'}),
+    (pers)-[:IS_TEAM_MEMBER_OF {name: 'eee'}]->(:Team {name: 'two'})
+----
+
+we can see that between `:Person` and `:Project` nodes, there are several relationships with the same type (`WORKS_FOR`).
```
AzuObs, line 727:
```suggestion
We can see that between `:Person` and `:Project` nodes, there are several relationships with the same type (`WORKS_FOR`).
```

docs/asciidoc/modules/ROOT/pages/export/cypher.adoc
```diff
@@ -65,6 +65,7 @@ The procedures support the following config parameters:
 | awaitForIndexes | Long | 300 | Timeout to use for `db.awaitIndexes` when using `format: "cypher-shell"`
 | saveIndexNames | boolean | false | Save name indexes on export
 | saveConstraintNames | boolean | false | Save name constraints on export
+| multipleRelationshipsWithType | boolean | false | In case of multiple relationships of the same type between two nodes, add a `UNIQUE IMPORT ID REL` property in order to distinguish them when using `MERGE`.
```
AzuObs, line 68: I can see how for existing users we wouldn't want to break backwards compatibility and that for them it might be good for this to be off by default. And I can also see that for new users it might be better for this to be enabled by default since an export should include multiple relationships.

Could I ask what your thoughts are about setting this to false by default?

AzuObs (commented):
Great job on this. I had a minor nitpick and a couple of open-ended questions

docs/asciidoc/modules/ROOT/pages/export/cypher.adoc
AzuObs: I just wanted to make sure I understood it right. I think that both approaches have their pros and cons - maybe it's best to leave it like this in dev also for consistency, but if you want to change it in dev then that's OK too.

AzuObs (approved):

docs/asciidoc/modules/ROOT/pages/export/cypher.adoc
vga91: Ok, I'll leave it consistent in dev

vga91:
[[S4][Volvo] apoc.export.cypher.query loses relationships](https://trello.com/c/qjfKSOqG/971-s4volvo-apocexportcypherquery-loses-relationships)

docs/asciidoc/modules/ROOT/pages/export/cypher.adoc
jexp: So the reason for only merging relationships on type is the massive performance impact, it is easily 10x slower to merge on rel-properties. So that's why the default was to focus on fast import as having multiple rels between the same nodes is rarer and the user should be able to configure what's the single id-property on the relationship to minimize the MERGE comparison effort. 
It might benefit from rel-indexes if there are any defined but that's also up to the user.

So from my perspective I would leave the defaults as they were and users that really need this feature should enable it and deal with the the performance impact.

vga91: Yes, it makes totally sense. I left the default as before

PR #3033, opened 2022-07-05
Fixes #2966: apoc.custom.declareProcedure does not accept default float values

vga91:
Fixes #2966

PR #3036, opened 2022-07-06
[NOID] Fixes #2978: The apoc.custom.declareProcedure does not accept default string values

### LINKED ISSUE #2978: The apoc.custom.declareProcedure does not accept default string values

Should create the procedure without errors.

It fails with error:
```
Syntax error(s) in signature definition .....
```

This works:
```
call apoc.custom.declareProcedure('myProc(in ::STRING) :: (out::STRING)', 'return $in as out')
```

But none of these (with the default String parameter) work:
```
call apoc.custom.declareProcedure('myProc1(in = test ::STRING) :: (out::STRING)', 'return $in as out')
```
```
call apoc.custom.declareProcedure('myProc1(in = "test" ::STRING) :: (out::STRING)', 'return $in as out')
```
```
call apoc.custom.declareProcedure("myProc1(in = 'test' ::STRING) :: (out::STRING)", "return $in as out")
```

### Versions

  - OS: OSX
  - Neo4j: 4.4.6
  - Neo4j-Apoc: 4.4.0.5

vga91:
Fixes #2978

- Updated `SIGNATURE_SYNTAX_ERROR`
- Fixed default string handling
- Added `getDefaultParameterValue` to differentiate between plain text and other data types

With these changes the `apoc.custom.declareFunction/Procedure` should be able to accept any kind of default value, except for `float` type fixed [here](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3033).

### DIFF
--- a/full/src/main/java/apoc/custom/Signatures.java
+++ b/full/src/main/java/apoc/custom/Signatures.java
@@ -2,6 +2,7 @@ package apoc.custom;
 
 import apoc.util.JsonUtil;
 import org.antlr.v4.runtime.*;
+import org.jetbrains.annotations.NotNull;
 import org.neo4j.internal.kernel.api.procs.*;
 import org.neo4j.procedure.Mode;
 
@@ -10,6 +11,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import static org.neo4j.internal.kernel.api.procs.Neo4jTypes.*;
@@ -17,7 +20,7 @@ import static org.neo4j.internal.kernel.api.procs.Neo4jTypes.*;
 public class Signatures {
 
     public static final String SIGNATURE_SYNTAX_ERROR = "Syntax error(s) in signature definition %s. " +
-            "\nNote that procedure/function name, input and output names must have at least 2 character:\n";
+            "\nNote that procedure/function name, possible map keys, input and output names must have at least 2 character:\n";
     private final String prefix;
 
     public Signatures(String prefix) {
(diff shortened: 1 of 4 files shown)


jexp:
The reason for originally leaving it out in the parser was that the default signature output doesn't put quotes around strings, which made it too hard to parse for me.

vga91:
> The reason for originally leaving it out in the parser was that the default signature output doesn't put quotes around strings, which made it too hard to parse for me.

@jexp 
Just to be sure, so do you think the changes are ok?
Or should I change them in another way?

Anyway, the pr change the `Signature.g4` placed in `core`, but indeed this file is only called by `Signature.java` (in `full`), 
therefore since we are not changing `core` functionality, I think it may be mergeable.

jexp:
@Lojjs we should still adjust the grammar file in core. (And possible also fix the signature output in cypher for strings)

jexp:
Actually @vga91 @Lojjs we should move the grammar file to extended as it's not used anywhere in core, at least not to my knowledge.

full/src/main/java/apoc/custom/Signatures.java
```diff
@@ -126,23 +127,51 @@ private DefaultParameterValue defaultValue(SignatureParser.DefaultValueContext d
...
+            
+            String text = stringCxt.getText();
+            if (stringCxt.SINGLE_QUOTED_STRING_VALUE() != null || stringCxt.QUOTED_STRING_VALUE() != null) {
+                text = text.substring(1, text.length() - 1);
+            }
+            return DefaultParameterValue.ntString(text);
+        }
+        if (v.INT_VALUE() != null) {
+            final String text = v.INT_VALUE().getText();
+            return getDefaultParameterValue(type, text, () -> DefaultParameterValue.ntInteger(Integer.parseInt(text)));
```
jexp, line 141: Btw. we should fix this as it should be Long.parseLong()

full/src/main/java/apoc/custom/Signatures.java
```diff
@@ -126,23 +127,51 @@ private DefaultParameterValue defaultValue(SignatureParser.DefaultValueContext d
...
+        }
+        if (v.FLOAT_VALUE() != null) {
+            final String text = v.FLOAT_VALUE().getText();
+            return getDefaultParameterValue(type, text, () -> DefaultParameterValue.ntFloat(Float.parseFloat(text)));
```
jexp, line 145: And this Double.parseDouble()

Lojjs:
> Actually @vga91 @Lojjs we should move the grammar file to extended as it's not used anywhere in core, at least not to my knowledge.

I will move it as part of this PR in dev: https://github.com/neo4j/apoc/pull/261. So if you can do your cherry-pick after that goes in, that would be good. Otherwise there is a risk I will write over your changes to the file by moving it

full/src/main/java/apoc/custom/Signatures.java
vga91: Changed

full/src/main/java/apoc/custom/Signatures.java
vga91: Changed

PR #3041, opened 2022-07-08
Fixes #1064: Add procedure to compare graphs

### LINKED ISSUE #1064: Add procedure to compare graphs

vga91:
Cherry pick 1064  

Differences from cherry-picked pr:
- added class `DiffFull` because `Bolt` located in `full` as well.
- in `BoltConnection.java`, changed `toMap(Session session..` , `toCollection(Session session... ` etc to `toMap(Transaction tx..` , `toCollection(Transaction tx... `
- added `systemUpdates` in `Neo4jContainerExtension`


full/src/main/java/apoc/bolt/BoltConnection.java
```diff
@@ -77,7 +81,7 @@ public Stream<RowResult> loadFromLocal(String localStatement, String remoteState
                 String withColumns = "WITH " + localResult.columns().stream()
                         .map(c -> "$" + c + " AS " + c)
                         .collect(Collectors.joining(", ")) + "\n";
-                Map<Long, Object> nodesCache = new HashMap<>();
+                Map<Long, VirtualNode> nodesCache = new HashMap<>();
```
jexp, line 84: Can we use `Node` here?

jexp:
Does this still support diffing two node-sets in the same graph?

And it would be good if there was a way from the Java API to diff to a second multi-db database. But I'm not sure how easily this is possible.

vga91:
> Does this still support diffing two node-sets in the same graph?
> 
> And it would be good if there was a way from the Java API to diff to a second multi-db database. But I'm not sure how easily this is possible.
@jexp 
Yes, we can compare 2 set in the same db, I added a test `shouldFindDifferencesInTheSameDb` for this.

And now I added the possibility to compare 2 graph using bolt protocol (via boltConfig param),
and via type: "DATABASE" (using `DatabaseManagementService`) instead of "URL".

Also, I added some other tests and docs and var. small code improvements.

PR #3044, opened 2022-07-11
Avoid reusing the AmbientCursor in node.getLabels()

AzuObs:
We were calling node.getLabels() severals which is unecessary work, and is also might have correctness implications because it is an AmbientCursor. During the bug, https://trello.com/c/zLF4jj16/194-s4sony-europe-network-map-attack-surface-underlyingstorageexception-access-to-record-node-went-out-of-bounds-of-the-page, we had a conversation with Sergey were he mentions that AmbientCursor reuse should be avoided.

(cherry picked from commit c8a96f3e298d34951882fa2375ce77926258d662)

  -
  -
  -

### DIFF
--- a/core/src/main/java/apoc/path/LabelMatcher.java
+++ b/core/src/main/java/apoc/path/LabelMatcher.java
@@ -1,7 +1,5 @@
 package apoc.path;
 
-import org.neo4j.graphdb.Node;
-
 import java.util.*;
 
 /**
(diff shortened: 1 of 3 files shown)


AzuObs:
[[s4][Sony Europe - Network Map Attack Surface] UnderlyingStorageException: Access to record Node went out of bounds of the page.](https://trello.com/c/zLF4jj16/194-s4sony-europe-network-map-attack-surface-underlyingstorageexception-access-to-record-node-went-out-of-bounds-of-the-page)

Issue #3048, opened 2022-07-13
Unable to bring database up in Neo4j 4.4 using the `expand-commands' with APOC jar in the plugins directory


pecollet:
## Expected Behavior (Mandatory)
Neo4j 4.4, on windows, with APOC 4.4 plugin, using expand-commands should start up with no fuss

## Actual Behavior (Mandatory)
Neo4j 4.4, on windows, with APOC 4.4 plugin, using expand-commands fails to start.

The following error occurs on startup : 

```
2022-07-06 14:35:19 Apache Commons Daemon procrun stderr initialized.
java.lang.NoSuchMethodError: 'java.lang.String org.apache.commons.lang3.SystemUtils.getUserName()'
              at org.neo4j.configuration.Config$Builder.validateFilePermissionForCommandExpansion(Config.java:356)
              at org.neo4j.configuration.Config$Builder.build(Config.java:318)
              at org.neo4j.server.NeoBootstrapper.start(NeoBootstrapper.java:115)
              at org.neo4j.server.BlockingBootstrapper.start(BlockingBootstrapper.java:45)
              at org.neo4j.server.NeoBootstrapper.start(NeoBootstrapper.java:95)
              at com.neo4j.server.enterprise.EnterpriseEntryPoint.start(EnterpriseEntryPoint.java:38)
```

This issue seems similar to issue #2215 (`org.apache.commons.lang3` version conflict between Neo4j dependency and APOC dependency), fixed by #2216 but only on 4.3 apparently.

### Steps (Mandatory)

- Deploy a neo4j 4.4 on windows 
- with APOC 4.4.0.3 plugin
- neo4j.conf : use a command expansion `some.setting=$(echo 123)`
- use `neo4j start --expand-commands` to start neo4j

### Versions

  - OS: windows
  - Neo4j: 4.4.8
  - Neo4j-Apoc: 4.4.0.3

PR #3058, opened 2022-07-15
Forces dependency to especific version to avoid binary incompatibilities

ncordon:
Solves https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/3048

Also checked that this problem is present in dev, so this will be cherry picked there.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -68,6 +68,13 @@ dependencies {
         exclude group: "org.yaml"
         exclude module: "snakeyaml"
     }
+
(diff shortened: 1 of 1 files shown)


AzuObs (approved):

ncordon:
[apoc 4.4 and neo4j 4.4 conflict on the commons-lang3 version](https://trello.com/c/v2T6jtTH/1038-apoc-44-and-neo4j-44-conflict-on-the-commons-lang3-version)

PR #3062, opened 2022-07-18
NO AUTO Bump neo4j and driver version, force cache refresh in CI

ncordon:
## What
* Bumps neo4j and driver version.
* Forces cache refresh in CI, because the current key would not pick up changes in .gradle files for example.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -22,8 +22,6 @@ jobs:
             ~/.gradle/caches
             ~/.gradle/wrapper
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
-          restore-keys: |
-            ${{ runner.os }}-gradle-
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Build with Gradle
--- a/build.gradle
+++ b/build.gradle
@@ -120,7 +120,7 @@ subprojects {
 
 ext {
     // NB: due to version.json generation by parsing this file, the next line must not have any if/then/else logic
-    neo4jVersion = "4.4.5"
+    neo4jVersion = "4.4.8"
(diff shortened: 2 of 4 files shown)


ncordon:
[***APOC Release 4.4.0.7***](https://trello.com/c/MN7Ec9DT/1042-apoc-release-4407)

build.gradle
```diff
@@ -120,7 +120,7 @@ subprojects {
 
 ext {
     // NB: due to version.json generation by parsing this file, the next line must not have any if/then/else logic
-    neo4jVersion = "4.4.5"
+    neo4jVersion = "4.4.8"
```
nadja-muller, line 123: shouldn't this also be 4.4.9?

.github/workflows/CI.yaml
```diff
@@ -22,8 +22,6 @@ jobs:
             ~/.gradle/caches
             ~/.gradle/wrapper
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
-          restore-keys: |
```
nadja-muller, line 25: I guess this is the force cache refresh in CI, right?

nadja-muller (approved):

.github/workflows/CI.yaml
ncordon: Yes. The problem I found (that we have in other branches as well) is that we'll have a revision for the .gradle files (which means dependencies, gradle version, etc). For example for this branch since I touched .gradle files, the hash could be `xyz`, but if the cache does not find that key linux-gradle-xyz, this line was making it default to the latest hash available for `linux-gradle`, which would still have old dependencies

build.gradle
ncordon: We discussed this privately and as said java-driver and neo4j version do not normally have to match.

PR #3069, opened 2022-07-20
Remove usages of db.indexes and db.constraints

Lojjs:
The db.indexes() and db.constraints() procedures will be removed from Neo4j. The replacement is SHOW INDEXES and SHOW CONSTRAINTS, although these are not possible to use from procedures with READ mode.


Hunterness (approved):
LGTM

burqen (approved):
This looks good to me. The export is nog including the index configuration but that was the case already before so not a new problem.

AzuObs (approved):
Great job 🎉

PR #3082, opened 2022-08-01
Fixes #3070: The apoc.export.csv.all with bulkImport config doesn't work with a file name without extension

### LINKED ISSUE #3070: The `apoc.export.csv.all` with `bulkImport` config doesn't work with a file name without extension

Should export multiple files, for each rel-type and label:
- test.nodes.End
- test.nodes.Start
- test.relationships.Rel

Export a single "test" file

Create a dataset:
```
CREATE (:Start)-[:Rel]->(:End)
```

Execute an export, without `.csv` extension:
```
call apoc.export.csv.all("test", {bulkImport: true})
```

The only file `test` exported has a content:
```
:ID,:LABEL
0,Start
```

### Versions

  - OS: OSX
  - Neo4j: 4.4.4
  - Neo4j-Apoc: 4.4.0.7

vga91:
Fixes #3070

If the file has at leas one dots, eg "test.ccc" / "test.foo.bar.ccc", 
via `FilenameUtils.getExtension` we consider the "ccc" part to be appended to `fileName.nodes.LABELNAME`.
If the file has no dots, I simply attach `.nodes.LABELNAME` to it.

- Added tests with multiple points in the filename
- Changed `DbmsRule` from `@ClassRule` to `@Rule` in order to recreate the dataset and avoid conflicts

### DIFF
--- a/core/src/main/java/apoc/export/cypher/FileManagerFactory.java
+++ b/core/src/main/java/apoc/export/cypher/FileManagerFactory.java
@@ -2,8 +2,8 @@ package apoc.export.cypher;
 
 import java.io.OutputStream;
 import apoc.export.util.ExportConfig;
-import apoc.util.CompressionAlgo;
 import apoc.util.Util;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 
 import java.io.PrintWriter;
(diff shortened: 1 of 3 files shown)


AzuObs (approved):
LGTM 👍

PR #3085, opened 2022-08-01
NO AUTO Moves coll procedures to core only

ncordon:
Strips core `coll` procedures and functions from `full` package.

PR #3086, opened 2022-08-02
NO AUTO Moves export procedures to core only

ncordon:
Strips core `export` procedures and functions from `full` package.

To be merged after #3085.


full/build.gradle
```diff
@@ -46,8 +46,11 @@ shadowJar {
 
 dependencies {
     implementation project(":common")
-    apt project(":processor")
+    testImplementation project(":common").sourceSets.test.output
     testImplementation project(':test-utils')
+    testImplementation project(":core")
```
ncordon, line 51: Unfortunately we need to add this for the moment: `full->test` depending on `core`. I don't know whether we are going to be able to take out this dependency easily.

AzuObs (approved):
✅ Looks like I also reviewed this in the bigger PR (this had a few slight difference but still good to go)

PR #3087, opened 2022-08-02
NO AUTO Moves map, math, text, number agg, convert procedures to core only 

ncordon:
Strips core `map`, `math`, `text`, `number`, `agg` and `convert` procedures and functions from full package.

To be merged after #3086.


AzuObs (approved):
✅ I already reviewed these in your other PR

PR #3098, opened 2022-08-02
Fixes #1582: rebind flag in apoc.periodic.iterate

### LINKED ISSUE #1582: rebind flag in apoc.periodic.iterate
- read entity still works, only writing fails
- <entity>.getId() is still transferrable/useable
- with the flag (default to true or false) ??
- try to walk through datastuctures and lookup entities by id in current tx
- fail if it's not possible

- add an apoc.node/relationship.rebind(node/rel) function
-

vga91:
Fixes #1582

- Added `apoc.node.rebind` / `apoc.rel.rebind` and `apoc.any.rebind` functions
- Added `rebind` config to `apoc.periodic.iterate` proc not to rebind manually the query if needed
- Added tests which fails without rebind via `match (n) where id(n) = $id...` or without `apoc.<entity>.rebind` function


vga91:
I just moved the new functionality in full

jexp:
@ncordon We should probably discuss this with Luisa.

Since 4.x this breaking change in neo4j is a big source of problems, as transaction state from nodes that are not rebound will leak into the original transaction and make it blow up.
As you can see from the issue it was already raised July 2020 and should have been long fixed.

ncordon (changes requested):
I don't understand why this is needed, nor we have added documentation in this PR, even though it's for 4.4 and the docs still live with the code.

Could you please explain me what's the problem if we don't have rebind?

vga91:
> I don't understand why this is needed, nor we have added documentation in this PR, even though it's for 4.4 and the docs still live with the code.
> 
> Could you explain me what rebind is for, please?

@ncordon 
I added some docs and put an example to explain it.
For example, I could execute a `CREATE (:Article {content: '1234'});`,
and then:
```
CALL apoc.periodic.iterate('MATCH (art:Article) RETURN art',
'CREATE (node:Category) with art, node call apoc.create.relationship(art, "CATEGORY", {b: 1}, node) yield rel return rel', {});
```

In this case the procedure fails because the `apoc.periodic.iterate` open a new transaction, and inside this I create a node `Category` and try creating a new relationship, via `org.neo4j.graphdb.Node.createRelationshipTo()`, 
that is in this case `nodeWithLabelArticle.createRelationshipTo(nodeWithLabelCategory)`.

This fails (`org.neo4j.graphdb.NotFoundException: Node[7] is deleted and cannot be used to create a relationship": 1`) 
because `nodeWithLabelArticle` was created outside the transaction and hold a reference to that, because of this `createRelationship(..)` does not recognize that node.

So we can rebind the node, via `apoc.node.rebind` or `MATCH (n) WHERE id(n) = idNodeArticle ...` to make sure that the node is recognized inside the new transaction.

ncordon:
Thanks for the explanation 👍. I think this makes sense, but will hold off from approving given this is now out of Cypher's Surface responsibilities as long as it's in the extended part.

docs/asciidoc/modules/ROOT/pages/operational/rebind.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+Unlike versions up to 3.5, 
+in Neo4j 4.x the entities hold a reference to their originating transaction.
+
+This can cause problems when for example, we create a node (called n1) in a transaction, open a new transaction in which we create another node (called n2) and execute
+via `org.neo4j.graphdb.Node.createRelationshipTo()`,
+that is `n1.createRelationshipTo(n2)`.
+
+
+Is what happens when we perform the following operation:
```
jexp, line 17: Actually this is not the main issue, that also didn't work before, it's regular tx-visibility and I think that will also not be resolved by rebinding.

The issue that all the periodic-iterate tx-state is attached to the originating transaction if the node is not rebound to the new updating tx, which then leads to OOM issues. @vga91

full/src/main/java/apoc/nodes/NodesExtended.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @UserFunction("apoc.rel.rebind")
+    @Description("apoc.rel.rebind(rel) - to rebind a rel (i.e. executing a Transaction.getRelationshipById(rel.getId())  ")
+    public Relationship relationshipRebind(@Name("rel") Relationship rel) {
+        return Util.rebind(tx, rel);
+    }
+
+    @UserFunction("apoc.any.rebind")
+    @Description("apoc.any.rebind(Object) - to rebind any rel, node, path, map, list or combination of them (i.e. executing a Transaction.getNodeById(node.getId()) / Transaction.getRelationshipById(rel.getId()))")
```
jexp, line 33: we should have an explicit one for path, as otherwise we lose the type information and cypher doesn't recognize it as a path anymore

PR #3099, opened 2022-08-02
Fixes #2930: The apoc.import.json constraint check should works only when an unique constraint exists

### LINKED ISSUE #2930: The `apoc.import.json` constraint check should works only when an unique constraint exists

Missing constraint error

The import works

Create a file `all.json` with label `User`:
```
{"type":"node","id":"0","labels":["User"],"properties":{"born":"2015-07-04T19:32:24","name":"Adam","place":{"crs":"wgs-84","latitude":33.46789,"longitude":13.1,"height":null},"male":true,"age":42,"kids":["Sam","Anna","Grace"]}}
```

Try to import the file without constraints:
```
CALL apoc.import.json('all.json')
```
this will throw an error message:
```
Missing constraint required for import. Execute this query: 
CREATE CONSTRAINT ON (n:User) assert n.neo4jImportId IS UNIQUE;
```

Create an existence constraint instead of an unique one.
```
CREATE CONSTRAINT ON (n:User) assert n.neo4jImportId IS NOT NULL
```

Retry the import and note that it's work.

### Versions

  - OS: OSX
  - Neo4j: 4.3.11
  - Neo4j-Apoc: 4.3.0.5

vga91:
Fixes #2930

Excluded node key and unique constraints

### DIFF
--- a/core/src/main/java/apoc/export/json/JsonImporter.java
+++ b/core/src/main/java/apoc/export/json/JsonImporter.java
@@ -2,12 +2,13 @@ package apoc.export.json;
 
 import apoc.export.util.Reporter;
 import apoc.util.Util;
-import com.google.common.collect.Iterables;
 import org.apache.commons.lang3.StringUtils;
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.Label;
 import org.neo4j.graphdb.Transaction;
+import org.neo4j.graphdb.schema.ConstraintType;
 import org.neo4j.graphdb.schema.Schema;
+import org.neo4j.internal.helpers.collection.Iterables;
 import org.neo4j.values.storable.DurationValue;
 import org.neo4j.values.storable.PointValue;
 
@@ -24,6 +25,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -174,7 +176,8 @@ public class JsonImporter implements Closeable {
             final String importIdName = importJsonConfig.getImportIdName();
             final String missingConstraint = labels.stream().filter(label -> 
                     StreamSupport.stream(schema.getConstraints(Label.label(label)).spliterator(), false)
-                            .noneMatch(constraint -> Iterables.contains(constraint.getPropertyKeys(), importIdName))
+                            .filter(c -> c.isConstraintType(ConstraintType.UNIQUENESS) || c.isConstraintType(ConstraintType.NODE_KEY))
+                            .noneMatch(constraint -> Iterables.asSet(constraint.getPropertyKeys()).equals(Set.of(importIdName)))
             ).findAny()
             .orElse(null);
             if (missingConstraint != null) {
--- a/core/src/test/java/apoc/export/json/ImportJsonEnterpriseTest.java
+++ b/core/src/test/java/apoc/export/json/ImportJsonEnterpriseTest.java
@@ -1,0 +1,102 @@
+package apoc.export.json;
+
+import apoc.util.CompressionAlgo;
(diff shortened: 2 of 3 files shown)


jexp:
@ncordon do you know where we are on this?

core/src/main/java/apoc/export/json/JsonImporter.java
```diff
@@ -174,7 +176,8 @@ private void checkConstraints(List<String> labels) {
             final String importIdName = importJsonConfig.getImportIdName();
             final String missingConstraint = labels.stream().filter(label -> 
                     StreamSupport.stream(schema.getConstraints(Label.label(label)).spliterator(), false)
-                            .noneMatch(constraint -> Iterables.contains(constraint.getPropertyKeys(), importIdName))
+                            .filter(c -> c.isConstraintType(ConstraintType.UNIQUENESS) || c.isConstraintType(ConstraintType.NODE_KEY))
```
ncordon, line 179: Should this be and condition instead of or?

core/src/test/java/apoc/export/json/ImportJsonEnterpriseTest.java
```diff
@@ -0,0 +1,96 @@
...
+import static apoc.util.BinaryTestUtil.fileToBinary;
+import static apoc.util.CompressionConfig.COMPRESSION;
+import static apoc.util.TestContainerUtil.createEnterpriseDB;
+import static apoc.util.TestContainerUtil.testCall;
+import static apoc.util.Util.map;
+import static java.lang.String.format;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class ImportJsonEnterpriseTest {
```
ncordon, line 23: Why do we need an integration test instead of a unit test? Is it because we are touching the system database?

core/src/test/java/apoc/export/json/ImportJsonEnterpriseTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @AfterClass
+    public static void afterAll() {
+        session.close();
+        neo4jContainer.close();
+    }
+
+    @Test
+    public void shouldFailsWithExistenceConstraintDueToMissingUniqueConstraint() {
```
ncordon, line 44: Nitpick:
```suggestion
    public void shouldFailWithExistenceConstraintDueToMissingUniqueConstraint() {
```

core/src/test/java/apoc/export/json/ImportJsonEnterpriseTest.java
```diff
@@ -0,0 +1,96 @@
...
+        } catch (RuntimeException e) {
+            assertTrue(e.getMessage().contains(MISSING_USER_CONSTRAINT));
+        } finally {
+            session.run("DROP CONSTRAINT userExistence");
+        }
+        
+    }
+
+    @Test
+    public void shouldFailsWithMultipleNodeKeyDueToMissingUniqueConstraint() {
```
ncordon, line 63: Nitpick:
```suggestion
    public void shouldFailWithMultipleNodeKeyDueToMissingUniqueConstraint() {
```

core/src/main/java/apoc/export/json/JsonImporter.java
vga91: I think it is right this way, that is, I verify that there is 
either a `CREATE CONSTRAINT name ON (n:LABEL) assert (n.ID_IMPORT) IS NODE KEY` 
or a `CREATE CONSTRAINT name ON (n:LABEL) assert n.ID_IMPORT IS UNIQUE`

core/src/test/java/apoc/export/json/ImportJsonEnterpriseTest.java
vga91: I created a new enterprise test class because I had to add `"IS NODE KEY"` constraints which are not supported in `ImportJsonTest`

ncordon (approved):

PR #3100, opened 2022-08-02
NO AUTO Moves nodes, create, refactor, trigger tests to core only

ncordon:
Strips core `nodes`, `create`, `refactor` and `trigger` procedures and functions from full package.

To be merged after #3087 .


core/src/test/resources/csv-inputs/forum.csv
```diff
@@ -0,0 +1,2 @@
+id:ID(Forum)|title:STRING|creationDate:LONG
+0|Wall of Mahinda Perera|20100214153220447
```
AzuObs, line 2: Is the file "forum.csv" used? I can do a search and find "personsWithoutIdField.csv" but not this one

AzuObs (approved):
Nice work 💯 

One small comment you may want to check, but feel free to merge even if you think a small change was needed

core/src/test/resources/csv-inputs/forum.csv
ncordon: This is leftover from having built this PR on top of another. It shouldn't be there anymore. This file is generated automatically by some tests.

build.gradle
```diff
@@ -79,7 +79,6 @@ subprojects {
         forkEvery = 50
         maxParallelForks = 1
         minHeapSize = "128m"
-        maxHeapSize = "512m"
```
ncordon, line 82: Funny cause this option is set to 8G just 4 lines above, so I wondered which one prevailed.

PR #3101, opened 2022-08-02
Fixes #2890: The apoc.export.graphml.query export additional unwanted nodes

### LINKED ISSUE #2890: The apoc.export.graphml.query export additional unwanted nodes
If we want to execute something like [this use case](https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/2659),
but via `apoc.export.graphml.query` instead of `apoc.export.graphml.data`,
the relationships file's export will add additional unwanted nodes.

Should import only relationships

Import start and end nodes as well

Create a simple rel:
```
CREATE (:Start {id: 1})-[:REL {foo: 'bar'}]->(:End {id: '1'})
```

Execute this query:
```
call apoc.export.graphml.query("MATCH (start:Start)-[rel:REL]->(end:End) RETURN rel", null, {stream: true})
YIELD  data
```

Note that the `data` result contains `<node...>` tags::
```
"<?xml version="1.0" encoding="UTF-8"?>
<graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
<key id="labels" for="node" attr.name="labels"/>
<key id="foo" for="edge" attr.name="foo"/>
<key id="label" for="edge" attr.name="label"/>
<graph id="G" edgedefault="directed">
<node id="n14" labels=":Start"><data key="labels">:Start</data><data key="id">1</data></node>
<node id="n15" labels=":End"><data key="labels">:End</data><data key="id">1</data></node>
<edge id="e13" source="n14" target="n15" label="REL"><data key="label">REL</data><data key="foo">bar</data></edge>
</graph>
</graphml>
"
```

### Versions

  - OS: OSX
  - Neo4j: 4.4.0.5
  - Neo4j-Apoc: 4.4.5

vga91:
Fixes #2890

The bug regards the export cypher as well, because the same method it's called.

Added config in order to export or not start/end nodes, with default `true`.

----

Other fixes: documented and unit-tested `nodesOfRelationships` config

### DIFF
--- a/core/src/main/java/apoc/export/cypher/ExportCypher.java
+++ b/core/src/main/java/apoc/export/cypher/ExportCypher.java
@@ -96,7 +96,7 @@ public class ExportCypher {
         Result result = tx.execute(query);
         SubGraph graph;
         try {
-            graph = CypherResultSubGraph.from(tx, result, c.getRelsInBetween());
+            graph = CypherResultSubGraph.from(tx, result, c.getRelsInBetween(), c.isAddRelNodes());
         } catch (IllegalStateException e) {
             throw new RuntimeException("Full-text indexes on relationships are not supported, please delete them in order to complete the process");
         }
--- a/core/src/main/java/apoc/export/graphml/ExportGraphML.java
+++ b/core/src/main/java/apoc/export/graphml/ExportGraphML.java
@@ -110,7 +110,7 @@ public class ExportGraphML {
     public Stream<ProgressInfo> query(@Name("query") String query, @Name("file") String fileName, @Name("config") Map<String, Object> config) throws Exception {
         ExportConfig c = new ExportConfig(config);
         Result result = tx.execute(query);
-        SubGraph graph = CypherResultSubGraph.from(tx, result, c.getRelsInBetween());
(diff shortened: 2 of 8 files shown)


AzuObs (changes requested):
Hiya Giuseppe,

Thanks for the amazing work in this PR. Unfortunately it's my understanding that if we wish to introduce new functionality to core then we need to obtain consent first by discussing the change as a team. This is to limit feature-creep for something we'd like to ultimately graduate to Cypher.

AFAIK that consent hasn't been obtained for this piece of work, so I'd like to invite you to bring up the topic in #apoc-dev before proceeding further with this PR.

vga91:
@AzuObs Sorry Daniel, I missed your comment at the time :( 
Because of latest changes (i.e. new repo), maybe should I move the pr to new repo before discussing it?

AzuObs:
Hi Giuseppe, sorry that I also missed your message. Yes, please do feel free to re-open this PR on the other repo if you would like!

PR #3102, opened 2022-08-02
NO AUTO Strips refactor and periodic procedures from full

ncordon:
Moves `refactor` and `periodic` to core only.


full/src/main/java/apoc/uuid/Uuid.java
```diff
@@ -37,12 +37,20 @@ public Stream<UuidInstallInfo> install(@Name("label") String label, @Name(value
...
                             .next()
-            );
+                );
+            } catch (RuntimeException e) {
+                if (e.getMessage().contains( "There is no procedure with the name `apoc.periodic.iterate` registered for this database instance" )) {
+                    throw new RuntimeException("apoc core needs to be installed when using apoc.uuid.install with the flag addToExistingNodes = true");
+                } else {
+                    throw e;
+                }
+            }
```
AzuObs, line 53: Good catch.

I think this is fine for now, but it might be nice to think about how we should handle this type of stuff in the long term.

AzuObs (approved):
Good job 🎉

PR #3104, opened 2022-08-03
NO AUTO Ignores some container tests until their startup is fixed

ncordon:
## What
Ignores newly created tests that require docker.

## Why
The docker container startups are really flaky at the moment.


AzuObs (approved):
Thanks for that 👍

PR #3105, opened 2022-08-03
NO AUTO Strips algo, meta, util, path procs from full

ncordon:
Moves `algo`, `meta`, `util` and `path` procedures to core only.

To be merged after #3102, #3086


AzuObs (approved):
Good job 🙌

PR #3106, opened 2022-08-03
Fixes flaky testParallelTransactionGuard test

ncordon:
Fixes a flaky test that was expeting to find the same number of transactions before and after firing a parallel command. There are no assurances of that, as there could be service transactions. 

The only alternative is to check none of the transactions in `SHOW TRANSACTIONS` correspond to one of the parallel ones.


AzuObs (approved):
LGTM :+1:

PR #3110, opened 2022-08-03
NO AUTO Strips atomic, hashing, temporal, date, index, neighbours, merge from full

ncordon:
Strips `atomic`, `hashing`, `temporal`, `date`, `index`, `neighbours`, `merge` procedures from full

PR #3113, opened 2022-08-04
NO AUTO Strips cypher, lock, graph, lock, path, search procedures from full

ncordon:
Strips `cypher`, `lock`, `graph`, `lock`, `path`, `search` procedures from full.

To be merged after #3110


AzuObs (approved):
Great job :+1:

PR #3114, opened 2022-08-04
NO AUTO Strips load, warmup, stats, bitwise, scoring, log, label from full

ncordon:
Strips `load`, `warmup`, `stats`, `bitwise`, `scoring`, `log`, `label`, `example` core procedures from full.

To be reviewed and merged after #3113


AzuObs (approved):
Nicely done!

PR #3123, opened 2022-08-09
NO AUTO Does a tidy-up and clean-up after splitting most of the code

ncordon:
* Moves more files to full (the ones that are only needed there and not in core)
* Tidy ups dependencies: there were lots of unused things.
* Removes sonarqube plugin. I don't think this was used anymore, this is from the old days when the code was published to Sonar rather than Maven.


full/build.gradle
```diff
@@ -45,65 +44,14 @@ shadowJar {
...
-    testImplementation group: 'com.amazonaws', name: 'aws-java-sdk-s3', version: '1.11.270'
-
-    testImplementation group: 'org.reflections', name: 'reflections', version: '0.9.12'
-
-    testImplementation group: 'org.hamcrest', name: 'hamcrest-library', version: '1.3'
-    testImplementation group: 'org.neo4j.test', name: 'neo4j-harness', version: neo4jVersionEffective
-    testImplementation group: 'org.eclipse.jetty', name: 'jetty-server', version: '9.2.22.v20170606'
-    testImplementation group: 'org.apache.derby', name: 'derby', version: '10.12.1.1'
-
-    testImplementation group: 'org.postgresql', name: 'postgresql', version: '42.1.4'
```
AzuObs, line 106: It looks like a test for the Postgres driver is failing for a genuine reason. Could be that this Postgres dependency is needed?

You may wish to run the tests locally if you haven't already, since many of these types of tests are probably disabled in the CI.

core/src/test/java/apoc/example/ExamplesTest.java
```diff
@@ -1,35 +0,0 @@
...
-    @Test
-    public void testMovies() throws Exception {
-        TestUtil.testCall(db,"CALL apoc.example.movies", r -> {
-            assertEquals("movies.cypher",r.get("file"));
-            assertEquals(169L,r.get("nodes"));
-            assertEquals(250L,r.get("relationships"));
-        });
-    }
-
-}
```
AzuObs, line 35: I think you may have wanted to delete the src/test/Examples.java instead of the src/test/ExamplesTest.java?

full/build.gradle
ncordon: Yep good catch, I've brought back this, but I don't understand how the code compiles if we need it in runtime

core/src/test/java/apoc/example/ExamplesTest.java
ncordon: Nice catch man

build.gradle
```diff
@@ -71,7 +70,7 @@ subprojects {
 
         // neo4jDockerImage system property is used in TestContainerUtil
         systemProperties 'user.language' : 'en' ,
-                'user.country ' : 'US',
+                'user.country' : 'US',
```
ncordon, line 73: This was spotted by @AzuObs, not me

AzuObs (approved):
Great job on this! 🙌

PR #3126, opened 2022-08-11
NO AUTO Renames full to extended

ncordon:
Renames the full package to extended, given it doesn't contain all the procedures anymore and it's a disjunction with core.


AzuObs (approved):
Great work 🎉

PR #3134, opened 2022-08-12
NO AUTO Adds CODEOWNERS

ncordon:
## Why

To control who approves workflows that touch sensible things, like CI workflows, by request of the security team


nadja-muller (approved):
Github complains about the lack of new lines. Not sure if you want to add it :)

PR #3140, opened 2022-08-17
Makes project depend on projects compiled from apoc (core)

ncordon:
Removes core from the repo and makes it depend on the (yet to be) published projects.


AzuObs (approved):
Changes LGTM - pending the build passes! 😄

PR #3145, opened 2022-08-23
Sanitize back-ticks in strings when there is a possibility of Cypher injection

gem-neo4j:
Cherry Picked: https://github.com/neo4j/apoc/pull/2


gem-neo4j:
[APOC Backtick Vulnerability](https://trello.com/c/1zjbkDMC/248-apoc-backtick-vulnerability)

PR #3152, opened 2022-08-29
respect config.writeNodeProperties when writing node details

### LINKED ISSUE #3150: apoc.export.json.data ignores config parameter 'writeNodeProperties'

All APOC procedure to export JSON data like `apoc.export.json.data()` accept a config parameter `writeNodeProperties` to conditionally export properties: https://neo4j.com/labs/apoc/4.4/export/json/#export-json-available-procedures

If this parameter is `false` (default), only the type (node, relationship), id and labels/type should be exported. 

The parameter `writeNodeProperties` is being ignored and node properties are always being exported. This increases the export size enormously. 

```
MATCH (m:Movie)-[r]-()
WITH collect(m) as movies, collect(r) as rels
CALL apoc.export.json.data(movies, rels, null, {stream: true, writeNodeProperties: false })
YIELD file, source, format, nodes, relationships, properties, time, rows, batchSize, batches, done, data
RETURN file, source, format, nodes, relationships, properties, time, rows, batchSize, batches, done, data
```

### Steps (Mandatory)

The export procedure is defined here:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/29051f331c0aa1eb760d15382408ade249ecbb27/core/src/main/java/apoc/export/json/ExportJson.java#L64-L69

The export config defines the parameter correctly:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/29051f331c0aa1eb760d15382408ade249ecbb27/core/src/main/java/apoc/export/util/ExportConfig.java#L127

The actual serialization into JSON starts here:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/29051f331c0aa1eb760d15382408ade249ecbb27/core/src/main/java/apoc/export/json/JsonFormat.java#L170-L181

The actual JSON format is defined here:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/29051f331c0aa1eb760d15382408ade249ecbb27/core/src/main/java/apoc/export/json/JsonFormat.java#L170-L181

The method call in line 22 causes the issue, because it always passes `true` to parameter `withNodeProperties` on method `writeNodeDetails()`

https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/29051f331c0aa1eb760d15382408ade249ecbb27/core/src/main/java/apoc/export/json/JsonFormatSerializer.java#L22
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/29051f331c0aa1eb760d15382408ade249ecbb27/core/src/main/java/apoc/export/json/JsonFormatSerializer.java#L70

The correct value to pass would be `config.writeNodeProperties()` as it is done here:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/29051f331c0aa1eb760d15382408ade249ecbb27/core/src/main/java/apoc/export/json/JsonFormatSerializer.java#L90

Lastly, serialization of relationships ignores the config parameter completely:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/29051f331c0aa1eb760d15382408ade249ecbb27/core/src/main/java/apoc/export/json/JsonFormatSerializer.java#L27-L38

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: macOS
  - Neo4j: 4.2.1
  - Neo4j-Apoc: 4.2.0.12

zirkelc:
Fixes #3150

Passes the actual config parameter `writeNodeProperties` to serialization method `writeNodeDetails()` instead of a static `true`. 

  - Respect config parameter `writeNodeProperties` when exporting JSON data

Further open issues, but not implement yet:
  - Introduce new config parameter `writeRelationshipProperties` to control relationships as well
  - Fix `NullPointerException` when exporting only Nodes or Relationships: currently, an empty List `[]`must be passed to procedure parameters `nodes` or `rels` to avoid the exception. However, the documentation states that null is the default value for both parameters which indicates that they are optional
  - Adapt test cases for export JSON to check on node/relationship properties 

I would continue to work on this PR if this issue is confirmed by the team. Please let me know, if and what should be done.


AzuObs (changes requested):
Pending review

AzuObs:
Hello @zirkelc,

Thank you so much for spotting this issue and creating this PR. I have no idea why the setting has been ignored, but it looks like it has been ignored since the project's birth.

Below are some recommendations I'd like to offer.

**Backwards Incompatible Change**
This setting has been ignored since the beginning of time, and has been hard coded to true. Fixing this config setting by making it respected will default all values to false which will break a lot of existing user queries.

I think we ought to implement the fix you suggest, but I think that the default value of the config map should be set to `true` and that the docs should be updated. WDYT?

**ExportJsonTest** 
Some of these tests are not passing. If we add the default setting to the config map then these tests should pass. Additionally, I think it would be a nice idea to add an extra test to this file which measures the behaviour of this config setting.

Let me know if you have any help with anything and I'll be happy to assist. 

Many thanks,

zirkelc:
Hi @AzuObs,

I agree with all your points and implemented them accordingly. Could you please run the tests on GitHub again? I get a lot of Java issues locally so I wasn't able to run them on my machine.

Here's a short summary of all changes:
- config `writeNodeProperties` defaults to true if not given
- config `writeNodeProperties` will be respected in **JsonFormatSerializer.java** for serialization of Nodes and Relationships 
  - _I thought of a second config `writeRelationshipProperties`, but I think a better approach would be to re-name `writeNodeProperties` to `writeProperties` in the next major release to indicate that this config affects both Nodes and Relationships_
- add tests for `writeNodeProperties` equals true, false and no value (= default value true)
- change the documentation for **apoc.export.json.*** to indicate the default value true
  - _the **apoc.export.cypher.all** and **apoc.export.cypher.data** procedures also have this config `writeNodeProperties`, even though I couldn't find any reference to this config value and I think it is not being considered at all_
- initialize variables `nodes` and `rels` with an empty List in **ExportJson.java** to avoid a NullPointerException if no relationships are given to the procedure (`rels=null`)

Please let me know if you have any remarks.

AzuObs:
Thank you for your continued work on this!

I've ran the tests and some seem to not be passing. I've investigated why some of the tests are not passing, and the main reason seems to be that relationships now include node properties in the tests whereas before they didn't.

The reason for this is that `writeNodeDetails(jsonGenerator, node, ?)` was hardcoded to `true` only sometimes and sometimes it was using `config.writeNodeProperties()` which before this change was defaulted to `false`. Now that we've changed the `config.writeNodeProperties()` to default to `true` some tests have started adding more information.

I think the solution you've come up is good though:
- it fixes the bug where the setting was not being respected
- it publishes MORE information than before which shouldn't really be considered a logical breaking change

Publishing more information can still be problematic though as it can mean a lot more information gets published which could lead to memory and disk space issues. But taking that risk is still probably a good tradeoff because it seems quite unlikely to affect many users, and those affected can remediate by disabling the setting.

It should be possible for you to run the ExportJsonTest locally by using Java 11 with Gradle 6.1.1. If you would like me to take it from here then I'd be happy to do that (credit for the contribution still goes to you). Of course, if you'd prefer to continue working on this then we would welcome your contribution!

zirkelc:
Thank you for the feedback!

I was able to run the tests locally with `gradlew build` and get the same results as the GitHub build. I can fix the failing test cases as well. Do you know if it is possible to run the `ExportJson.java` tests instead of the entire test suite? 

Before is start I would like to clarify a thought.  As you mentioned, some tests are failing now because more information is added than before. For example, the `testExportAllJson` expected the following JSON:

```json
{"type":"node","id":"0","labels":["User"],"properties":{"born":"2015-07-04T19:32:24","name":"Adam","place":{"crs":"wgs-84","latitude":13.1,"longitude":33.46789,"height":null},"age":42,"male":true,"kids":["Sam","Anna","Grace"]}}
{"type":"node","id":"1","labels":["User"],"properties":{"name":"Jim","age":42}}
{"type":"node","id":"2","labels":["User"],"properties":{"age":12}}
{"id":"0","type":"relationship","label":"KNOWS","properties":{"since":1993,"bffSince":"P5M1DT12H"},"start":{"id":"0","labels":["User"]},"end":{"id":"1","labels":["User"]}}
```

But, it failed because the relationship on the last line contained the properties of start and end node:
```json
{"id":"0","type":"relationship","label":"KNOWS","properties":{"since":1993,"bffSince":"P5M1DT12H"},"start":{"id":"0","labels":["User"],"properties":{"born":"2015-07-04T19:32:24","name":"Adam","place":{"crs":"wgs-84","latitude":13.1,"longitude":33.46789,"height":null},"age":42,"male":true,"kids":["Sam","Anna","Grace"]}},"end":{"id":"1","labels":["User"], "properties":{"name":"Jim","age":42}}}
```

As the node properties are already present in the file (if `writeNodeProperties` is true) on the individual nodes, maybe they shouldn't be written again for relationship start and end nodes? The docs and examples didn't include node properties on the relationship start and end node, and the default value for `writeNodeProperties` was false which was respected for relationships.


So in summary, I could either
- disable writing node properties for relationship start and end nodes 
- add the node properties to the relationship on each test file

Both options would be okay for me. What do you think?

AzuObs:
You should be able to compile the project without running any tests by doing `./gradlew clean shadow` and then you can run any particular test by adapting this command `./gradlew :core:test --tests "apoc.export.json.ExportJsonTest"`

For your question about what to do in the logic... It's a little tricky! The problem I see is that it was previously possible to print the relationship's "start" and "end" node properties because the configuration setting was being respected in that position. That the docs didn't explain it doesn't really mean that users aren't relying on the behaviour, although I would agree that it makes it less likely.

I think both options are defendable, but ultimately I feel like I will be in a better position to explain our decision if we try our best to ensure we didn't remove any data. If we add too much data and that ends up breaking a query because of size issues then that's very unfortunate, but at least our bug fix wouldn't have been backwards incompatible on a logical level.

So if you still agree, then I think we ought to choose the "add the node properties to the relationship on each test file" option.

AzuObs (approved):
Approved ✅ 

Great job on this @zirkelc and thank you for your contribution. It is customary for us to credit contributors in the release notes, so please keep an eye open for that :)

AzuObs:
Hi Zirkelc, to thank you for your work! We'd like to know if you'd like some swag from us?

If you would like some then could you please send the following information to me at daniel.leaver@neo4j.com:
- name
- delivery address
- phone number
- t-shirt size

Let me know 😁

zirkelc:
Hi @AzuObs,

yes, that would be great! :-) I just sent the mail.

Thanks a lot!

PR #3154, opened 2022-08-31
Add private method removed from APOC core

Lojjs:
This method has been removed from core, but was still used from this single file,
so re-added it as a private method so it will still compile when we release
a new APOC core.


gem-neo4j (approved):
Thanks! :)

PR #3157, opened 2022-09-01
Move over extended-only configuration from core

Lojjs:
Parts of the APOC config was only used from extended. This PR introduce corresponding config to the extended repo. Should be reviewed and merged together with https://github.com/neo4j/apoc/pull/166

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -2,17 +2,17 @@ name: CI
 
 on:
   push:
-    branches: [ "dev" ]
+    branches: [ "5.0" ]
   pull_request:
-    branches: [ "dev" ]
+    branches: [ "5.0" ]
 
 jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      DOCKER_ENTERPRISE_DEV_URL: ${{ secrets.DOCKER_ENTERPRISE_DEV_URL }}
-      DOCKER_COMMUNITY_DEV_URL: ${{ secrets.DOCKER_COMMUNITY_DEV_URL }}
-      TEAMCITY_DEV_URL: ${{ secrets.TEAMCITY_DEV_URL }}
+      DOCKER_ENTERPRISE_FIVE_URL: ${{ secrets.DOCKER_ENTERPRISE_FIVE_URL }}
+      DOCKER_COMMUNITY_FIVE_URL: ${{ secrets.DOCKER_COMMUNITY_FIVE_URL }}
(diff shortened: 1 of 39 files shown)


extended/src/main/java/apoc/ExtendedApocConfig.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * use apache commons to load configuration
+     * classpath:/apoc-config.xml contains a description where to load configuration from
+     */
+
+    protected void loadConfiguration() {
+        try {
+
+            URL resource = getClass().getClassLoader().getResource("apoc-config.xml");
```
gem-neo4j, line 99: As extended is now separate from core, is there any concern with them using the same config file?

Lojjs: Good question... As extended still use config from the core file, I guess it should be alright. It could be cleaner to have the extended config in a separate file, but on the other hand it will be one more thing to do for people migrating from extended 4.x to extended 5.x. What do you think? Should we ask the others in the team for feedback regarding this?

gem-neo4j: Hmm I think that makes sense remembering it uses the common package actually. I am happy for it to continue to use the same file, but feel free to ask the team for feedback, there might be something else we haven't thought about.

gem-neo4j (approved):
LGTM :)

ncordon:
Depends on #3167

.github/workflows/CI.yaml
```diff
@@ -28,8 +28,8 @@ jobs:
 
       - name: Download neo4j dev docker container
```
Lojjs, line 29: This should probably be changed to `Download neo4j 5 docker container`

build.gradle
```diff
@@ -131,8 +131,8 @@ ext {
     // NB: due to version.json generation by parsing this file, the next line must not have any if/then/else logic
     neo4jVersion = "5.0.0"
     // This should be removed when the version is stable and released
-    neo4jInDevVersion = (System.env.CI != null) ? neo4jVersion + "-dev" : neo4jVersion + "-SNAPSHOT"
+    // neo4jInDevVersion = (System.env.CI != null) ? neo4jVersion + "-dev" : neo4jVersion + "-SNAPSHOT"
```
Lojjs, line 134: Do we still want this commented out, or will it only cause confusion?

ncordon: This is always going to be needed in the dev branch, so I would prefer to leave it commented

PR #3167, opened 2022-09-15
Depends on neo4j 5.0.0

ncordon:
Neo4j 5.0.0 has not been released publicly yet, but it has been upgraded from snapshot dependencies to a release. We should depend on that one here in apoc


ncordon:
Closing given this needs to be merged together with the changes in #3157 to work

PR #3169, opened 2022-09-16
Add core repo as a git submodule

vga91:
Created via `git submodule add  --name apoc-core -b git_remote_as_repo_core https://github.com/neo4j/apoc apoc-core`.
If the `git_remote_as_repo_core` is merged, we should need to change `git_remote_as_repo_core` with `dev` (and/or `5.0` ?)


vga91:
There are some failing tests (it seems due to a wrong path), but they are all part of `core` repository. 
I could try to solve them, but maybe, in order not to weigh down the build, it would be better to skip them. Wdyt?

settings.gradle
```diff
@@ -1,3 +1,23 @@
...
+        ':common:build',
+        ':processor:build',
+        ':test-utils:build',
+        ':core:test',
+        ':common:test',
+        ':processor:test',
+        ':test-utils:test',
+        ':core:publish',
+        ':common:publish',
+        ':test-utils:publish'
```
ncordon, line 22: `processor` does not have a publish task, so it fails if added here

ncordon (approved):
LGTM but we should wait for another review on this, since I've made small modifications

settings.gradle
```diff
@@ -1,3 +1,23 @@
...
+
+startParameter.excludedTaskNames = [
+        ':core:build',
+        ':common:build',
+        ':processor:build',
+        ':test-utils:build',
+        ':core:test',
+        ':common:test',
+        ':processor:test',
+        ':test-utils:test',
```
ncordon, line 19: Without these the code from `core`, `common` would be tested as well, but we don't want to do that from the `extended` repo

settings.gradle
Lojjs, line 22: There is no 'processor:test' and also no mention of 'test-startup' which was included in the CI.yaml. Is that intentional?

readme.adoc
```diff
@@ -278,14 +278,16 @@ docker run \
...
-./gradlew shadow
+./gradlew --init-script init.gradle shadow
 cp build/extended/libs/apoc-<version>-extended.jar $NEO4J_HOME/plugins/
 $NEO4J_HOME/bin/neo4j restart
 ----
 
 // If you want to run embedded or use neo4j-shell on a disk store, configure your `plugins` directory in `conf/neo4j.conf` with `dbms.plugin.directory=path/to/plugins`.
 
-A full build including running the tests can be run by `./gradlew build`.
+A full build including running the tests can be run by `./gradlew --init-script init.gradle build`.
```
Lojjs, line 288: Not sure if it is worth pointing out in the readme, but I had to run this with sudo because otherwise I got permission denied for `rm -r apoc-core`

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "apoc-core"]
+	path = apoc-core
+	url = https://github.com/neo4j/apoc
+	branch = git_remote_as_repo_core
```
Lojjs, line 4: Is this an existing branch on the core repo or how does it work?

extended/build.gradle
```diff
@@ -43,8 +43,15 @@ shadowJar {
     archiveClassifier.set('extended')
 }
 
+// to init the submodule in case of new submodule or refresh it if exists, every time we compile 
+task gitSubmoduleLoad {
+    def execute = ["bash", "-c", "git submodule status apoc-core | grep -q \"^-\" && rm -r apoc-core && git submodule update --init apoc-core || git submodule update --remote apoc-core"]
+            .execute()
+    execute.waitForProcessOutput(System.out, System.err)
```
Lojjs, line 50: Two questions on this:

1. What does the `grep -q \"^-\"` part do?
2. Why do we need `rm -r apoc-core && git submodule update --init apoc-core` both here and in the init script?

Lojjs (commented):
As far as I can judge it looks OK, but I had some questions

extended/build.gradle
Lojjs: @vga91 Nacho couldn't explain this to me, can you?

vga91: @Lojjs 

1. The `git submodule status apoc-core`, initially it has an output like this:
```
-COMMIT_HASH apoc-core
```
where the first `-` sign indicates that the submodule is not initialized.
Therefore, with `grep -q \"^-\"`, I'll check whether the module is initialized (and in this case I'll do `rm -r apoc-core && git submodule update --init apoc-core`) or not (then `git submodule update --remote apoc-core` will be executed).
It's the simplest method I've found to automatically check if the module is initialized and execute the correct command.

2. Because I could execute `Load gradle changes` (e.g. via IntelliJ), and in this case this piece of code will be executed, and the init script would be useless.

And therefore, every time I want to update the core changes, it would be enough to do "Load gradle changes"

Otherwise, with the CI (or by just doing `git clone <repo>` and `./gradlew shadow`), this piece of code would not be executed, so we need to manually run` init.gradle` (via `./gradlew  --init-script init.gradle build/shadow/otherCommand`)

Lojjs: Thanks for the explanation, now I understand :)

Lojjs (approved):
Looks good now! I'm not really sure about the CI test failures, but they feel unrelated, so perhaps you can just try to rerun the CI.

ncordon (approved):

PR #3172, opened 2022-09-21
[JB2or4Qk] apoc.cypher.runFile does not return paths

vga91:
Regards both APOC `core` and `extended` .

----

Due to asynchronous implementation, the `testRunFileWithCreateAndReturnFile` fails in a flaky way.
The `collect(toList())` seems to solve it, but of course impact performance.
I think besides this solution, we have other 2 ways:
- remove the `collect` and document that both CREATE and RETURN are not allowed 
- make the `collect(toList()` configurable, but in this case it would be an additional feature in `core` (because of `runMany`)

----

Moreover, the `Util.anyRebind(..)` created is equivalent to this https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3098/files#diff-f88abb534773494608522b01357e40fd270d873f32d7286d83bf3ac0c120925cR15,
which I placed to full because the other pr doesn't regards core.
In case, could we change the other pr after this one is merged?

### DIFF
--- a/core/src/main/java/apoc/cypher/Cypher.java
+++ b/core/src/main/java/apoc/cypher/Cypher.java
@@ -157,7 +157,9 @@ public class Cypher {
         int queueCapacity = Util.toInteger(config.getOrDefault("queueCapacity",100));
 
         StringReader stringReader = new StringReader(cypher);
-        return runManyStatements(stringReader ,params, false, addStatistics, timeout, queueCapacity);
+        return runManyStatements(stringReader ,params, false, addStatistics, timeout, queueCapacity)
+                .collect(toList()).stream()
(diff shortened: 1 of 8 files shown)


conker84:
@gem-neo4j @nadja-muller @ncordon @AzuObs any thoughts on this?

nadja-muller:
Is this change to fix the test or is there a problem with the behavior?

vga91:
@nadja-muller This is a bug due to incorrect behavior, which is fixed with `Util.anyRebind(tx, rowResult.result))`

vga91:
Closed as the related card is closed and meanwhile the bug has been fixed via another pr: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3631

PR #3177, opened 2022-09-22
Fix double escaping graphml

vga91:
The `javax.xml.stream.XMLStreamWriter` escape some chars, e.g from `&` to `&amp;`, so we cannot use StringEscapeUtils.escapeXml10() because we would escape twice.
We just have to remove the invalid xml chars.

The bug is occurred [here](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3176), and not in 4.4, because of different retrieving order of `node.getPropertyKeys()`.
Therefore, I added more chars in test cases.

----
Also, added missing `@Test` in `testExportGraphGraphMLQueryTinkerPop`


gem-neo4j:
Looks good, out of curiosity, where did you get the list for the invalid characters? are we sure that contains all of them? :)

vga91:
I've extracted the char list from the previous implementation, [StringEscapeUtils.escapeXml10](https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/StringEscapeUtils.java#L233). 

Alternatively we could change the pattern in a negative way, to `^[setOfAllowedChars]`, like [here](https://stackoverflow.com/questions/397250/unicode-regex-invalid-xml-characters). 
However, the results match each other.

gem-neo4j (approved):
LGTM :)

PR #3183, opened 2022-09-26
[gRNAet7B] Fixes trigger procedures in clusters

ncordon:
Fixes trigger procedures by splitting the `Lifecycle + TransactionEventListener` and the utils to write to the system database. It also needs to pass the user `databaseName` as first argument.


vga91:
I added the 3 three points of the [associated card](https://trello.com/c/gRNAet7B/1403-fix-apoc-triggers-for-clusters). Therefore:
- I split `TriggerHandler` (which reads from systemdb, and handles` LifecycleAdapter` and `TransactionEventListener`) and `TriggerUtils` (which writes to system)
- I deprecated the old procedures, and added a warning asking users to use the new ones
- I added a check to make the old procedures fail if they are performed on a FOLLOWER

---

- I removed the `ConcurrentHashMap activeTriggers` as we now read directly from the system db.
So `apoc.trigger.refresh` at this point is just for doing` registerTransactionEventListener() `and` unregisterTransactionEventListener()`.

- `TriggerUtils`, as a Util class, has all the static methods, alternatively you could set it as an annotable class with`@Context`, like `TriggerHandler`, but at that point I think the suffix `Utils` would be misleading (Maybe we can change the 2 class name to `TriggerHandlerRead` / `TriggerHandlerWrite`

- Changed `TriggerTest` making it parameterized, to see that there are no inconsistencies between the old procedures and the new ones.

- Added a `TriggerClusterRoutingTest` and changed `Neo4jContainerExtension` in order to optionally create a cluster with routing 

- Removed `Class.forName("org.neo4j.kernel.ha.HighlyAvailableGraphDatabase"` in `Util.java`, which is no longer present since neo4j 3.5, in fact it gives a  `java.lang.ClassNotFoundException: org.neo4j.kernel.ha.HighlyAvailableGraphDatabase`.

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -50,28 +51,64 @@ public TriggerInfo( String name, String query, Map<String,Object> selector, Map<
...
+    private void preprocessDeprecatedProcedures() {
+        final String msgDeprecation =
+                "use the `apoc.trigger.install`, `apoc.trigger.drop`, `apoc.trigger.dropAll`, `apoc.trigger.stop`, and `apoc.trigger.start` procedures \n" + 
+                "instead of, respectively, `apoc.trigger.add`, `apoc.trigger.remove`, `apoc.trigger.removeAll`, `apoc.trigger.pause`, and `apoc.trigger.resume`.";
+                
+        log.warn("Please note that the current procedure is deprecated, \n" + msgDeprecation);
+        
+        if (!Util.isWriteableInstance(db, GraphDatabaseSettings.SYSTEM_DATABASE_NAME)) {
+            throw new RuntimeException("It's not possible to write into a cluster member with a non-LEADER system database.\n" +
+                    "Either the procedure using the bolt against a core protocol with LEADER system database, \n" +
```
ncordon: The wording of this message needs to be changed

`Either use the procedure executing it with bolt against a LEADER system database`

ncordon: Are we sure the procedure is not redirected by Fabric to a user LEADER even if we ask for system LEADER? I think we should recommend using the other procedures directly, not using these ones at all in this case.

vga91: Changed it

core/src/main/java/apoc/util/Util.java
```diff
@@ -779,18 +778,14 @@ public static String withMapping(Stream<String> columns, Function<String, String
         return with.isEmpty() ? with : " WITH "+with+" ";
     }
 
-    public static boolean isWriteableInstance(GraphDatabaseAPI db) {
+    public static boolean isWriteableInstance(GraphDatabaseService db) {
+        return isWriteableInstance(db, db.databaseName());
+    }
+    
+    public static boolean isWriteableInstance(GraphDatabaseService db, String dbName) {
```
ncordon, line 785: I'd suggest collapsing this into a single method if we are not using the one that takes two parameters anywhere else

core/src/test/java/apoc/trigger/TriggerDisabledTest.java
```diff
@@ -65,4 +65,28 @@ public void testTriggerDisabledResume() {
...
+    
+    @Test
+    public void testTriggerDisabledDropAll() {
+        db.executeTransactionally("CALL apoc.trigger.dropAll('neo4j')");
+    }
+
+    @Test
+    public void testTriggerDisabledStart() {
+        db.executeTransactionally("CALL apoc.trigger.start('neo4j', 'test-trigger')");
+    }
```
ncordon, line 86: What is this testing? It seems like we get no failure at all if triggers are disabled? Is that normal?

core/src/test/java/apoc/trigger/TriggerRestartTest.java
```diff
@@ -48,13 +46,19 @@ private void restartDb() {
 
     @Test
     public void testTriggerRunsAfterRestart() throws Exception {
+        final String query = "CALL apoc.trigger.add('myTrigger', 'unwind $createdNodes as n set n.trigger=true', {phase:'before'})";
+        testTriggerRestartCommon(query);
```
ncordon, line 50: If the triggers are eventually consistent now, this would be flaky

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    @Test
+    public void testInstallTriggerInSystemDb() {
+        try {
+            testCall(db, "CALL apoc.trigger.install('system', 'name', 'RETURN 1',{})", 
+                    r -> fail("Should fail because of unrecognised system procedure"));
+        } catch (QueryExecutionException e) {
+            assertTrue(e.getMessage().contains("Not a recognised system command or procedure"));
```
ncordon, line 52: 👍

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -48,38 +54,61 @@ public TriggerInfo( String name, String query, Map<String,Object> selector, Map<
...
-    @Procedure(mode = Mode.WRITE)
+    @Deprecated
+    @Procedure(mode = Mode.WRITE, deprecatedBy = "apoc.trigger.drop")
     @Description("remove previously added trigger, returns trigger information")
     public Stream<TriggerInfo> remove(@Name("name")String name) {
-        Map<String, Object> removed = triggerHandler.remove(name);
-        if (removed == null) {
+        preprocessDeprecatedProcedures();
+        Map<String, Object> removed = TriggerUtils.remove(db.databaseName(), name);
+        triggerHandler.updateCache();
```
ncordon, line 97: Some bits have changed in the code here: for example the `updateCache` call. My personal preference would be to live all the old procedures as they were (even if we have to duplicate functionality temporarily). What do you think?

core/src/main/java/apoc/util/Util.java
vga91: I've splitted it because we're using an `Util.isWriteableInstance(db)` in `TTLLifeCycle.java` too.

core/src/test/java/apoc/trigger/TriggerDisabledTest.java
vga91: Yes it's a wanted behavior, because this tests actually fail but are catched by the `exceptionRule.expectMessage(TriggerUtils.NOT_ENABLED_ERROR);` (line 40). In fact, with `apocConfig().setProperty(APOC_TRIGGER_ENABLED, true);` they fails with error `Expected test to throw`.

core/src/main/java/apoc/trigger/Trigger.java
vga91: Yes makes sense, I just split the trigger, now the deprecated one are in another class (`TriggerDeprecatedProcedures`  ) with their original code

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -109,26 +79,75 @@ public Stream<TriggerInfo> list() {
...
     }
 
+    // TODO - change with @SystemOnlyProcedure
+    @SystemProcedure
+    @Procedure(mode = Mode.WRITE)
+    @Description("CALL apoc.trigger.install(databaseName, name, statement, selector, config) | add a trigger kernelTransaction under a name, in the kernelTransaction you can use $createdNodes, $deletedNodes etc., the selector is {phase:'before/after/rollback/afterAsync'} returns previous and new trigger information. Takes in an optional configuration.")
+    public Stream<TriggerInfo> install(@Name("databaseName") String databaseName, @Name("name") String name, @Name("kernelTransaction") String statement, @Name(value = "selector")  Map<String,Object> selector, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) {
+        Util.validateQuery(ApocConfig.apocConfig().getDatabase(databaseName), statement);
+
+        Map<String,Object> params = (Map)config.getOrDefault("params", Collections.emptyMap());
```
ncordon, line 89: Should we add a check to the new trigger methods to check that they are executed against a system LEADER?

ncordon:
> Maybe we can change the 2 class name to TriggerHandlerRead / TriggerHandlerWrite

I think this is a good suggestion that would help understand the code better

core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -114,9 +72,11 @@ private void updateCache() {
      * For most deployments this isn't an issue, since you can turn the config flag off, but in large fleet deployments
      * it's nice to have uniform config, and then the memory savings on databases that don't use triggers is good.
      */
-    private synchronized void reconcileKernelRegistration() {
+    public synchronized void reconcileKernelRegistration(boolean hasTriggers) {
+        lastUpdate = System.currentTimeMillis();
```
ncordon, line 76: Why do we still want a `lastUpdate`? That made sense only when there was a cache for the triggers

Also why did we get rid of the cache? Was that necessary?

core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -57,54 +49,20 @@ private enum Phase {before, after, rollback, afterAsync}
...
-                            )
-                    )
-            );
-            return null;
-        });
 
-        reconcileKernelRegistration();
+    public void updateCache() {
+        final boolean hasTriggers = withSystemDb(tx -> getTriggerNodes(db.databaseName(), tx)
+                .hasNext());
```
ncordon, line 64: Maybe the `hasTriggers` should be cached as well so that we don't fire a request to the database each time we call this

core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -302,28 +209,11 @@ public void stop() {
...
     }
 
-    private void setLastUpdate(Transaction tx) {
-        Node node = tx.findNode(SystemLabels.ApocTriggerMeta, SystemPropertyKeys.database.name(), db.databaseName());
-        if (node == null) {
-            node = tx.createNode(SystemLabels.ApocTriggerMeta);
-            node.setProperty(SystemPropertyKeys.database.name(), db.databaseName());
-        }
-        node.setProperty(SystemPropertyKeys.lastUpdated.name(), System.currentTimeMillis());
-    }
```
ncordon, line 327: This is not right. If you get rid of this, then you don't need to do any scheduler job. Basically the condition:
```
if (getLastUpdate() > lastUpdate) {
    updateCache();
}
```

is never going to be true

vga91: I moved this piece of code in `TriggerUtils`, method `setLastUpdate(String databaseName, Transaction tx)`

core/src/main/java/apoc/trigger/TriggerHandler.java
vga91: Yes true, indeed there is no more need currently. Removed

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -109,26 +86,80 @@ public Stream<TriggerInfo> list() {
                 );
     }
 
+    // TODO - change with @SystemOnlyProcedure
+    @SystemProcedure
+    @Procedure(mode = Mode.WRITE)
+    @Description("CALL apoc.trigger.install(databaseName, name, statement, selector, config) | add a trigger kernelTransaction under a name, in the kernelTransaction you can use $createdNodes, $deletedNodes etc., the selector is {phase:'before/after/rollback/afterAsync'} returns previous and new trigger information. Takes in an optional configuration.")
+    public Stream<TriggerInfo> install(@Name("databaseName") String databaseName, @Name("name") String name, @Name("kernelTransaction") String statement, @Name(value = "selector")  Map<String,Object> selector, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) {
+        checkProcedureRouted();
+        Util.validateQuery(ApocConfig.apocConfig().getDatabase(databaseName), statement);
```
AzuObs, line 95:
```java
        Util.validateQuery(ApocConfig.apocConfig().getDatabase(databaseName), statement);
```

Please forgive me if I'm wrong but this type of logic will not work in 5.0.0. Basically what the code is doing is that it is beginning a new transaction against the target database in order to made sure the statement is valid. 

The reason that this will not work in 5.0.0 is because this Java code will be executed on the leader of the system database, and that in 5.0.0 not all machines in the cluster will have the all the databases on them. And so, you have to assume that the leader of the system database doesn't have access to this user database.

It only works in 4.4 because all machines in a cluster contain all databases.

My opinion is that this type of logic should not be necessary. It should be up to our users to make sure their queries work in their dev environments before pushing their queries to production.  We shouldn't be running these types of checks in production which make queries slower. 

If we did however want to have this type of logic in 5.x then we'd probably have to use Fabric to make sure the validation gets routed to the correct database. In 5.x I don't see any other option that to remove entirely this type of logic. I'm not sure if that's possible? We may want to ship a simple version of the code first that can be improved later.

AzuObs:
I would encourage you to write up some documentation for these new functions, and also to modify the previous functions to show to our users that they're now deprecated.

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
```diff
@@ -0,0 +1,153 @@
...
+        if (!isEnabled()) {
+            throw new RuntimeException(NOT_ENABLED_ERROR);
+        }
+    }
+
+    public static Map<String, Object> install(String databaseName, String triggerName, String statement, Map<String,Object> selector, Map<String,Object> params) {
+        checkEnabled();
+        final HashMap<String, Object> previous = new HashMap<>();
+
+        withSystemDb(tx -> {
```
AzuObs, line 51: There are many times where we should already be in a transaction being executed against "system" where we are opening a new nested transaction.

If it's not necessary then I don't think we should be opening nested transactions because these are known to have some issues. It's also quite expensive to open a new transaction, compared to reusing an existing one.

If this code is necessary then that changes things, but I'd like to know why that is the case!

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
```diff
@@ -0,0 +1,153 @@
...
+            Node node = Util.mergeNode(tx, SystemLabels.ApocTrigger, null,
```
AzuObs, line 52: It seems like both the new Triggers functionality and the deprecated Triggers functionality is using the same node label.

I'm not sure that we should really be consider the new triggers equivalent to the old triggers. For example, should the new apoc.triggers.stop procedure be able to stop the old triggers? Should the apoc.trigger.add (deprecated) be able to add the new triggers? 

It doesn't seem to me like these new procedures need to be backwards-compatible with the old ones, and so I'd probably rather keep them separate entirely. I imagine that it will help us and our users debug support issues when users are migrating from the old procedures to the new ones, as will probably help us avoid a great deal of confusion. 

But I'm happy to hear counter-arguments, of course!

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
```diff
@@ -0,0 +1,153 @@
...
+
+    public static void checkEnabled() {
+        if (!isEnabled()) {
+            throw new RuntimeException(NOT_ENABLED_ERROR);
+        }
+    }
+
+    public static Map<String, Object> install(String databaseName, String triggerName, String statement, Map<String,Object> selector, Map<String,Object> params) {
+        checkEnabled();
+        final HashMap<String, Object> previous = new HashMap<>();
```
AzuObs, line 49: I'm not sure that `previous` is a good name for this variable anymore.

It is a good variable name for the deprecated apoc.trigger.add procedure because that procedure returns a copy of the triggers before pre-update. 

The apoc.trigger.install procedure is instead returning a copy of the triggers post-update. I would maybe suggest calling this `result` or `updated` - I don't really mind the exact name so long as it's sensible.

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
```diff
@@ -0,0 +1,153 @@
...
+    public static <T> T withSystemDb(Function<Transaction, T> action) {
+        try (Transaction tx = apocConfig().getSystemDb().beginTx()) {
+            T result = action.apply(tx);
+            tx.commit();
+            return result;
+        }
+    }
+
+    private static void setLastUpdate(String databaseName, Transaction tx) {
+        Node node = tx.findNode(SystemLabels.ApocTriggerMeta, SystemPropertyKeys.database.name(), databaseName);
```
AzuObs, line 145: If we agree that new triggers should be decoupled from the old triggers (see https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183/files#r1005590912) then I think we should also decouple the node label being used to store trigger metadata.

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    public static void checkEnabled() {
+        if (!isEnabled()) {
+            throw new RuntimeException(NOT_ENABLED_ERROR);
+        }
+    }
+
+    public static Map<String, Object> install(String databaseName, String triggerName, String statement, Map<String,Object> selector, Map<String,Object> params) {
+        checkEnabled();
```
AzuObs, line 48: If this class assumes it's being executed against the system database (either because the user open a session against "system" or because Fabric routed to "system") then should we be asserting this assumption and throwing an exception when it's not the case?

Bonus points if we can also assert that it's the leader of the system database.

AzuObs: PS. I quite like the idea that this should be able to work even with Fabric disabled, so long as the user is manually opening a session against "system" himself.

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
ncordon: For me the argument for having the same levels is you want people with triggers in their database to still work. We are recommending to use the new procedures, the old ones will be deprecated and deleted. Unless we provide a migration path, we cannot let people that have triggers installed already without support in a couple years.

core/src/main/java/apoc/trigger/TriggerHandlerRead.java
```diff
@@ -0,0 +1,228 @@
...
+            if(registeredWithKernel.compareAndSet(false, true)) {
+                databaseManagementService.registerTransactionEventListener(db.databaseName(), this);
+            }
+        } else {
+            // This gets called every time triggers update; only unregister if we aren't already
+            if(registeredWithKernel.compareAndSet(true, false)) {
+                databaseManagementService.unregisterTransactionEventListener(db.databaseName(), this);
+            }
+        }
+    }
```
AzuObs, line 91: I thought I had written a comment about this, but I've been jumping in and out of this review all day and can't find it anymore 😢 

I'm not sure that this whole section would work, or perhaps I'm misunderstanding what it's supposed to do. This code seems to be called whenever we install a new trigger, which means it's being run in a transaction against the "system" database. At that point wouldn't this be registered against the system database which might not even have a given user database (in 5.x at least, in 4.x they always will)?

Shouldn't this logic rather be run in the background jobs happening for each user database for which there are triggers?

AzuObs: Oh I see that this code is being called from multiple places... I think I get how it's supposed to work now. The start() method is gonna get called on each DBMS which has APOC installed, and then that's gonna call reconcileKernelRegistration.

What might still not make sense to me is why this method is being called during apoc.trigger.install, apoc.trigger.drop, and apoc.trigger.dropAll when that code is executed in a "system" transaction on a DBMS which in 5.x might not even have a given user database.

core/src/main/java/apoc/trigger/TriggerHandlerRead.java
```diff
@@ -0,0 +1,228 @@
...
+        this.pools = pools;
+        this.jobScheduler = jobScheduler;
+    }
+
+
+    public void reconcileKernelRegistration() {
+        final boolean hasTriggers = withSystemDb(tx -> getTriggerNodes(db.databaseName(), tx)
+                .hasNext());
+        reconcileKernelRegistration(hasTriggers);
+    }
```
AzuObs, line 69: If we agree that reconcileKernelRegistration should be a private method and should not be called from the controller class (see my other comment below) then the only place this is called is from updateCache().

If the only place this called from is updateCache() then this method should not be needed, and will only make things slower. This is because updateCache() can directly call reconcileKernelRegistration(true | false) without having to retrieve the trigger nodes again from the system database.

core/src/main/java/apoc/trigger/TriggerHandlerRead.java
```diff
@@ -0,0 +1,228 @@
...
+                                    "params", Util.fromJson((String) node.getProperty(SystemPropertyKeys.params.name()), Map.class),
+                                    "paused", node.getProperty(SystemPropertyKeys.paused.name())
+                            )
+                    )
+            );
+            return null;
+        });
+
+        reconcileKernelRegistration();
+    }
```
AzuObs, line 216: I struggle to imagine of a use case where a user would want a trigger that could only read data. It would seem utterly pointless to me. 

I can imagine use cases where users might want to trigger some side effect, such as writing data to another database - which is definitely possible with Extended. 

I'm just wondering if it would be a sensible restriction to only triggers to be executed on WRITE transactions to the leader of a database. I think this would be too much of a breaking change though.

However, do we know what happens when a follower of a database finally catches up with the leader? Does that happen in a transaction that would re-execute the trigger?

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,71 @@
...
+                    TestContainerUtil.testCall(container.getSession(), "CALL apoc.trigger.add($name, 'UNWIND $createdNodes AS n SET n.ts = timestamp()',{})",
+                            Map.of("name", "trigger-" + container.getContainerName()),
+                            row -> fail("Should fail because of non leader trigger addition"));
+                } catch (RuntimeException e) {
+                    String errorMsg = e.getMessage();
+                    assertTrue("The actual message is: " + errorMsg, errorMsg.contains(SYS_NON_LEADER_ERROR));
+                }
+            }
+        }
+    }
```
AzuObs, line 70: I would really have liked this test to be written in terms of how a user should structure his queries using the Java driver in production.

Since users don't have access to container information then we shouldn't be using that.

I don't have the exact syntax but I would have liked to basically see us open a new driver session against "system" and create a write transaction which should get automatically routed to the "system" database LEADER because it's a write. This routing should happen automatically in the client if client-side routing enabled, if not it should happen automatically in the backend (in the DBMS) if server-side routing is enabled.

Then, creating another session the "neo4j" database, I would expect the `CREATE (f: Foo)` statement to be wrong.

AzuObs: At the very least, this current implementation of the test has two problems:
- adding trigger is asynchronous but the test ignores that and will be very flaky as a result
- the user query which creates FOO is run against the leader of the system database which will often not be the leader of the user database, which will also be a source of flakiness

core/src/main/java/apoc/trigger/TriggerHandlerRead.java
AzuObs: Actually, it might be worth checking whether TransactionEventListener is even invoked for READ transactions. This [answer](https://neo4j.slack.com/archives/C02JJGEPD/p1666802698640259) from Sergey to a different question has me thinking that might not be the case 

If only LEADERS are allowed to execute transactions because they are the only ones that can handle writes, then our code should reflect that, and we shouldn't be registering TransactionEventListeners on DBMS members which don't have write access to the database for the given trigger.

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,71 @@
...
+                continue;
+            }
+            final String systemRole = TestContainerUtil.singleResultFirstColumn(container.getSession(), "CALL dbms.cluster.role('system')");
+            if ("LEADER".equals(systemRole)) {
+                container.getSession().run("CALL apoc.trigger.add($name, 'UNWIND $createdNodes AS n SET n.ts = timestamp()',{})",
+                        Map.of("name", "trigger-" + container.getContainerName()));
+
+                container.getSession().run("CREATE (f:Foo)");
+                TestContainerUtil.testCall(container.getSession(), "MATCH (f:Foo) RETURN f",
+                        (row) -> assertTrue(((Node) row.get("f")).containsKey("ts")));
```
AzuObs, line 58: This is being executed against the system database leader which may or may not be the user database leader - in any case we probably shouldn't be counting on it as it would be a very flaky test indeed!

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,71 @@
...
+                            Map.of("name", "trigger-" + container.getContainerName()),
+                            row -> fail("Should fail because of non leader trigger addition"));
+                } catch (RuntimeException e) {
+                    String errorMsg = e.getMessage();
+                    assertTrue("The actual message is: " + errorMsg, errorMsg.contains(SYS_NON_LEADER_ERROR));
+                }
+            }
+        }
+    }
+}
```
AzuObs, line 71: I would like to encourage writing these tests the way a user would write his queries in production using the Java Driver. This is because we can be sure that our logic works as expected for our users, and they can also be used as an example code.

Users' application code won't have access to the container information of their deployment. Instead they have to open a session against "system" which will then be routed by the client (if client-side routing is enabled) or by the DBMS (if server-side routing is enabled). The routing is transparent to the user.

Then the user would have to open another session against his "neo4j" database to create nodes, etc. This should also be automatically routed by the drivers.

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,71 @@
...
+                ));
+    }
+
+    @AfterClass
+    public static void bringDownCluster() {
+        if (cluster != null) {
+            cluster.close();
+        }
+    }
+
```
AzuObs, line 40: I think it would be nice to add some of the new procedures to this file too. 

Of particular interest I think would be:
- making sure that a session against "system" can install/drop triggers
- making sure that a session against "neo4j" can't install/drop triggers
- fabric tests perhaps, so that we can see if that works? I'm not sure you can do `USE system` in Fabric for example 🤔 And I'm not SystemOnlyProcedure annotation has been implemented for Neo4j yet

core/src/main/java/apoc/trigger/TriggerHandlerRead.java
AzuObs: Ah thanks, I edited my original answer I think, maybe just before you posted your reply. In any case I think we're now on a similar understanding of how this works. That part of the code is fine and great.

What still doesn't make much sense to me is why reconcileKernelRegistration() is being called in the controller class because in the code in the controller class is basically being executed while we're in a transaction for the system database.

Example of unnecessary/harmful invocations of this method:
- [here in apoc.trigger.install](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183/files#diff-28ed54275ec240e9acd877b090ba248e89e4f67e3083f086fef57b2ddbce5c5aR100)
- [here in apoc.trigger.drop](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183/files#diff-28ed54275ec240e9acd877b090ba248e89e4f67e3083f086fef57b2ddbce5c5aR117)
- [here in apoc.trigger.dropAll](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183/files#diff-28ed54275ec240e9acd877b090ba248e89e4f67e3083f086fef57b2ddbce5c5aR133)

If there is still confusion from either of us after this then maybe a call will be the best way to resolve things.

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
vga91: It's not so evident, but actually this the previous. There is a `previous.putAll(TriggerHandlerWrite.toTriggerInfo(node));` (line 56), which populates the return value before setting new changes. I added a comment to clarify

core/src/main/java/apoc/trigger/TriggerHandlerRead.java
vga91: I've changed the caching logic. Now new and deprecated procedure use the same `TriggerHandler`, which has a private `reconcilceKernelRegistration()`

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
vga91: I've changed the caching logic. Now new and deprecated procedure use the same `TriggerHandler`, so we use the same system labels

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
vga91: I just joined the two checks, in `Trigger.java`

core/src/main/java/apoc/trigger/TriggerHandlerRead.java
AzuObs: Had a call with Nacho. He made a great point that we should be registering the event listeners even on followers, so as to avoid any missed transactions in case the follower gets promoted to leader.

core/src/main/java/apoc/trigger/Trigger.java
vga91: It makes sense. I might be wrong but, given that this piece of code should be useful to prevent additions with wrong syntaxes, I think we could just do `Util.validateQuery(db, statement);`, where `db` is defined as `@Context public GraphDatabaseService db;`. I think it can be ok, but I could be wrong. Wdyt?

ncordon: As discussed this is fine in 4.4, but wouldn't work in 5.x

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
vga91: Yep makes sense, I removed this part. Indeed I should just test the `apoc.trigger.add`

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
vga91: Personally, I think it is probably best to wait for the `@SystemOnlyProcedure` annotation before doing these tests? Just to be sure i create sensible tests.

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
AzuObs: Nicely done!

core/src/main/java/apoc/trigger/TriggerHandlerRead.java
AzuObs: Beautiful, cheers!

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
AzuObs: I think the change you've made to that test is great!

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
AzuObs: For the Fabric tests then I would be happy to wait.

But I'm not sure we have to wait if want to test that we can't install triggers against "neo4j" or whether we want to test that dropping triggers works. You can use apoc.trigger.list and apoc.trigger.show to confirm whether a trigger was dropped or not.

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Test
+    public void testTriggerAddAndInstallAllowedOnlyInLeaderMember() {
```
AzuObs, line 42: It's perhaps a small detail but could we split this test into two separate tests, please? 

Tests should probably only test one thing at a time to keep them simple. I would recommend one test for apoc.trigger.add and another separate test for apoc.trigger.install.

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
AzuObs: This was fixed :+1:

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -123,12 +125,13 @@ public Stream<TriggerInfo> pause(@Name("name")String name) {
     @Procedure(mode = Mode.WRITE)
     @Description("CALL apoc.trigger.resume(name) | it resumes the paused trigger")
     public Stream<TriggerInfo> resume(@Name("name")String name) {
```
ncordon, line 127: I think this needs to be deprecated as well, true?

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
ncordon: This is still a pending point @vga91. I think we check we are executing this against the system database in the functionality that uses it? So the system transaction is not needed

core/src/test/java/apoc/trigger/TriggerRestartTest.java
```diff
@@ -48,19 +56,49 @@ private void restartDb() {
...
-                result -> {
-                    Map<String, Object> single = Iterators.single(result);
-                    System.out.println(single);
-                });
-        db.executeTransactionally("CREATE (p:Person{id:1})");
-        TestUtil.testCallCount(db, "match (n:Person{trigger:true}) return n", Collections.emptyMap(), 1);
+    @Test
+    public void testTriggerViaInstallRunsAfterRestart() {
+        final String query = "CALL apoc.trigger.install('neo4j', 'myTrigger', 'unwind $createdNodes as n set n.trigger = n.trigger + 1', {phase:'before'})";
+        testTriggerRestartCommon(query, this::awaitTriggers);
```
ncordon, line 66: This should be an eventually probably

core/src/test/java/apoc/trigger/TriggerRestartTest.java
```diff
@@ -48,19 +56,49 @@ private void restartDb() {
...
+    }
+
+    @Test
+    public void testTriggerViaBothAddAndInstall() {
+        // executing both trigger add and install with the same name will not duplicate the eventListeners
+        final String query = "CALL apoc.trigger.install('neo4j', 'myTrigger', 'unwind $createdNodes as n set n.trigger = n.trigger + 1', {phase:'before'})";
+        final Runnable runnable = () -> {
+            db.executeTransactionally("CALL apoc.trigger.install('neo4j', 'myTrigger', 'unwind $createdNodes as n set n.trigger = n.trigger + 1', {phase:'before'})");
+            awaitTriggers();
```
ncordon, line 75: This should be an eventually probably

full/src/test/java/apoc/trigger/TriggerClusterTest.java
```diff
@@ -104,4 +107,79 @@ public void testTxIdAfterAsync() throws Exception {
...
+    //
+
+    @Test
+    public void testTimeStampTriggerForUpdatedPropertiesNewProcedures() throws Exception {
+        cluster.getSession().run("CALL apoc.trigger.install('neo4j', 'timestamp','UNWIND apoc.trigger.nodesByLabel($assignedNodeProperties,null) AS n SET n.ts = timestamp()',{})");
+        cluster.getSession().run("CREATE (f:Foo) SET f.foo='bar'");
+        TestContainerUtil.testCall(cluster.getSession(), "MATCH (f:Foo) RETURN f", (row) -> {
+            assertTrue(((Node) row.get("f")).containsKey("ts"));
+        });
+    }
```
ncordon, line 122: This test doesn't wait for the trigger to be installed I think?

full/src/test/java/apoc/trigger/TriggerClusterTest.java
```diff
@@ -104,4 +107,79 @@ public void testTxIdAfterAsync() throws Exception {
...
+
+    @Test
+    public void testReplicationNewProcedures() {
+        cluster.getSession().run("CALL apoc.trigger.install('neo4j', 'timestamp','UNWIND apoc.trigger.nodesByLabel($assignedNodeProperties,null) AS n SET n.ts = timestamp()',{})");
+        // Test that the trigger is present in another instance
+        awaitProcedureUpdated(cluster.getDriver().session(), "timestamp");
```
ncordon, line 128: This name doesn't reflect what we are doing. I'd say we are waiting for the trigger to be installed, not for a procedure to be upated

full/src/test/java/apoc/trigger/TriggerClusterTest.java
```diff
@@ -104,4 +107,79 @@ public void testTxIdAfterAsync() throws Exception {
...
+    }
+
+    @Test
+    public void testLowerCaseNameNewProcedures() {
+        cluster.getSession().run("create constraint on (p:Person) assert p.id is unique");
+        cluster.getSession().run("CALL apoc.trigger.install('neo4j', 'lowercase', 'UNWIND apoc.trigger.nodesByLabel($assignedLabels,\"Person\") AS n SET n.id = toLower(n.name)',{})");
+
+        awaitProcedureUpdated(cluster.getSession(), "lowercase");
+
+        cluster.getSession().run("CREATE (f:Person {name:'John Doe'})");
```
ncordon, line 138: Daniel has touched on this point as well. We'd rather the tests rely on using the driver, which is what users would do when using neo4j normally.

ncordon, line 138: We check the trigger is installed in maybe another instance and then the CREATE could go to a whole different instance? If yes this could lead to another flaky test.

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
vga91: Done

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
vga91: Just to be sure, do you mean using a `@Context Transaction tx` instead?
Because in this case I receive an error `"Creating new node label on database 'system' is not allowed for user '' with FULL overridden by READ overridden by READ. See GRANT CREATE NEW NODE LABEL ON DATABASE system "`

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
vga91: Added test against "neo4j" session.

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
vga91: As below, I think it is better to wait for the annotations `@SystemOnlyProcedure` for "system" session's tests

core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
ncordon: Let's do this as a separate piece of work, there's a ticket in Trello to reduce the number of transactions we use

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
ncordon: I think what @AzuObs meant is that we shouldn't be using `container.getSession())` in these tests because a client code normally doesn't have access to the `container` data structure. He meant opening a session with bolt I think

core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -57,54 +49,20 @@ private enum Phase {before, after, rollback, afterAsync}
...
-            );
-            return null;
-        });
 
-        reconcileKernelRegistration();
+    public void updateCache() {
+        final boolean hasTriggers = withSystemDb(tx -> getTriggerNodes(db.databaseName(), tx)
+                .hasNext());
+        reconcileKernelRegistration(hasTriggers);
     }
```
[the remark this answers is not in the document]
ncordon: This has been resolved by the new structure of the code.

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
ncordon: I'm happy to merge the PR as is and then we can iterate in some bits that can be improved

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+        return new TriggerInfo(name, null, null, false, false);
+    }
+
+    // TODO - change with @SystemOnlyProcedure
+    @SystemProcedure
+    @Procedure(mode = Mode.WRITE)
+    @Description("CALL apoc.trigger.install(databaseName, name, statement, selector, config) | add a trigger kernelTransaction under a name, in the kernelTransaction you can use $createdNodes, $deletedNodes etc., the selector is {phase:'before/after/rollback/afterAsync'} returns previous and new trigger information. Takes in an optional configuration.")
+    public Stream<TriggerInfo> install(@Name("databaseName") String databaseName, @Name("name") String name, @Name("kernelTransaction") String statement, @Name(value = "selector")  Map<String,Object> selector, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) {
+        preprocessProcedures();
```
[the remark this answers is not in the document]
ncordon: Solved

core/src/test/java/apoc/trigger/TriggerRestartTest.java
```diff
@@ -43,24 +56,58 @@ private void restartDb() {
...
+        final String innerQuery = "unwind $createdNodes as n set n.trigger = n.trigger + 1";
+        
+        final String triggerQuery = "CALL apoc.trigger.add($name, $query, {phase:'before'})";
+
+        final Map<String, Object> params = Map.of("name", name, "query", innerQuery);
+        
+        final Runnable runnable = () -> {
+            sysDb.executeTransactionally("CALL apoc.trigger.install('neo4j', $name, $query, {phase:'before'})",
+                    params);
+            awaitProcedureUpdated(db, name, innerQuery);
```
ncordon, line 92: `awaitTriggerDiscovered` or something like that would be a better name

full/src/test/java/apoc/trigger/TriggerClusterTest.java
```diff
@@ -104,4 +111,104 @@ public void testTxIdAfterAsync() throws Exception {
...
+        }
+    }
+
+    @Test
+    public void testReplicationNewProcedures() throws Exception {
+        try (final Session session = cluster.getDriver().session(forDatabase(SYSTEM_DATABASE_NAME))) {
+            session.run("CALL apoc.trigger.install('neo4j', 'timestamp','UNWIND apoc.trigger.nodesByLabel($assignedNodeProperties,null) AS n SET n.ts = timestamp()',{})");
+        }
+        // Test that the trigger is present in another instance
+        awaitProcedureInstalled(cluster.getDriver().session(), "timestamp");
```
ncordon, line 141: How is this testing another instance?

ncordon:
Good job! Just a few items left:

* We need to strengthen the fabric tests and have integration tests using the driver rather than the underlying TestContainers data structure. This can be done in another PR. Those are comments https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183/files#discussion_r1005988470 and https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183/files#discussion_r1005991535.
* Small nitpick: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183/files#discussion_r1018201340
* An explanation of how this works: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183/files#discussion_r1018203756
* The CI is currently throwing java heap space exceptions, this should be solved in the PR if this is the cause before we merge it 😞

full/src/test/java/apoc/trigger/TriggerClusterTest.java
vga91: Because this method calls a `session.readTransaction(..)`, which should send the query to one of the nodes with the `READ` role (follower or readReplica), 
unlike `session.run(apoc.trigger.install ...)` which sends the query in a LEADER member.

The same thing that is done [on line 70](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183/files#diff-5e26a8edc16341c68758787654f5f129cd26b3f23cc8733fdcecb6cf32eb252cR70)

vga91:
> ...
> * The CI is currently throwing java heap space exceptions, this should be solved in the PR if this is the cause before we merge it 😞

The  java heap space problem seems solved

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Test
+    public void testTriggerAddAllowedOnlyInSysLeaderMember() {
+        final String query = "CALL apoc.trigger.add($name, 'RETURN 1',{})";
+        triggerInSysLeaderMemberCommon(query, SYS_NON_LEADER_ERROR);
+    }
+
+    @Test
+    public void testTriggerRemoveAllowedOnlyInSysLeaderMember() {
+        final String query = "CALL apoc.trigger.remove('neo4j', $name)";
```
ncordon, line 62:
```suggestion
        final String query = "CALL apoc.trigger.remove($name)";
```

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,102 @@
...
+        triggerInSysLeaderMemberCommon(query, TRIGGER_NOT_ROUTED_ERROR);
+    }
+
+    @Test
+    public void testTriggerDropAllowedOnlyInSysLeaderMember() {
+        final String query = "CALL apoc.trigger.drop('neo4j', $name)";
+        triggerInSysLeaderMemberCommon(query, SYS_NON_LEADER_ERROR);
```
ncordon, line 69:
```suggestion
        triggerInSysLeaderMemberCommon(query, TRIGGER_NOT_ROUTED_ERROR);
```

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private static void triggerInSysLeaderMemberCommon(String query, String triggerNotRoutedError) {
+        final List<Neo4jContainerExtension> members = cluster.getClusterMembers();
+        assertEquals(4, members.size());
+        for (Neo4jContainerExtension container: members) {
+            // we skip READ_REPLICA members
+            final String readReplica = TestcontainersCausalCluster.ClusterInstanceType.READ_REPLICA.toString();
+            try (final Session session = container.getSession()) {
```
ncordon, line 78: I'm not sure you can do this. Wouldn't this try and close the session, which is a variable inside the container?

I would leave this without the `try`

vga91:
@ncordon I updated the pr with the latest test changes

ncordon:
✅ I can't approve through GitHub functionality because I opened the PR. Looks good to me now. Good work!

PR #3231, opened 2022-10-18
Update testJaroWinklerDistance test expectation

MishaDemianenko:
Before this PR test was expecting the incorrect value of distance. 
As a result, https://issues.apache.org/jira/browse/TEXT-191 and the new commons-text test is updated with the correct expected value.

### DIFF
--- a/core/src/test/java/apoc/text/StringsTest.java
+++ b/core/src/test/java/apoc/text/StringsTest.java
@@ -317,7 +317,7 @@ public class StringsTest {
 
         testCall(db, "RETURN apoc.text.jaroWinklerDistance($a, $b) AS distance",
                 map("a", text1, "b", text2),
-                row -> assertEquals(0.77, (double) row.get("distance"), 0.01));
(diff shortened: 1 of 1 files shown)


AzuObs (approved):

vga91:
This pr causes the failure of this other https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183
```
2022-10-19T09:23:13.2354082Z apoc.text.StringsTest > testJaroWinklerDistance STANDARD_ERROR
2022-10-19T09:23:13.2354473Z     expected:<0.22> but was:<0.7777777777777777>
```

I think we should add these lines in `build.gradle`, or so, or am I missing something?
```
    compileOnly group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
    testCompile group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
```

MishaDemianenko:
@vga91 so commons-text is coming from the core product so i would assume you need to rebase of some sort?

PR #3235, opened 2022-10-20
Update commons-text version

vga91:
Update commons-text version

In order to fix `testJaroWinklerDistance` after [this pr](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3231) ,
because in 4.4, if we don't explicit it, version 1.9` is picked


AzuObs:
I've reached out to Giuseppe on Slack. Basically it could be that this PR is needed, but I'm not sure I know why that's the case. Probably I'm just missing knowledge on my end.

AzuObs (approved):

PR #3245, opened 2022-10-23
[H10zCpAQ] Fix CWE-73: Added check to prevent reading from outside metrics directory

vga91:
Added check to prevent reading from outside metrics directory


Lojjs (commented):
I will ask our security team to have a look as well, but I think it looks ok apart from the printline

Lojjs (approved):

PR #3261, opened 2022-10-31
[SXnVeQi9] change ApocStreamHandlerFactory to a service provider to avoid deadlock on startup

nadja-muller:


Lojjs (approved):
Looks good to me

PR #3273, opened 2022-11-02
[T7j55Aw9] APOC Extended Compilation Failure due to helper methods work

vga91:
Fixed compilation issues
- changed internal methods with `apoc.util.` analogues
- updated `build.gradle`
- small fixes after latest core changes


conker84:
Thank you so much @vga91

PR #3292, opened 2022-11-15
Makes gradle go to Maven Central for extra-dependencies project

ncordon:
Cherry-picked from #3287 

## Why

Because our TeamCity build are failing when doing `./gradlew buildDependencies` inside `extra-dependencies` folder.

It's unclear to me why it started failing now, probably because those dependencies were removed from the gradle-libs repo.


ncordon:
Merging because it's cherry-picked and the CI is red because the urls are outdated

PR #3325, opened 2022-11-22
[gRNAet7B] added documentation about new trigger procedures and deprecated ones

vga91:
[gRNAet7B] added documentation about new trigger procedures and deprecated ones

Related to https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183


docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -6,32 +6,27 @@
 Triggers allow the registration of Cypher queries that are called when data in Neo4j is changed (created, updated, deleted).
 You can run them before or after commit.
 
+[WARNING]
+====
+The listeners are eventually consistent,
```
ncordon, line 11: Normally a user will not have any context of what a listener is. I think it would be better worded as: installing, updating or removing a trigger is an eventually consistent operation.

ncordon (commented):
I can pull @JPryce-Aklundh into this PR as well if he agrees, so that the documentation team can propose their editorial changes as well?

docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.trigger.install.adoc
```diff
@@ -0,0 +1,5 @@
+¦xref::overview/apoc.trigger/apoc.trigger.install.adoc[apoc.trigger.install icon:book[]] +
+
+`CALL apoc.trigger.install(databaseName, name, statement, selector, config) | add a trigger kernelTransaction under a name, in the kernelTransaction you can use $createdNodes, $deletedNodes etc., the selector is {phase:'before/after/rollback/afterAsync'} returns previous and new trigger information. Takes in an optional configuration.`
```
AzuObs, line 3:
```suggestion
`CALL apoc.trigger.install(databaseName, name, statement, selector, config) | eventually adds a trigger for a given database which is invoked when successful write transactions occur.
```

AzuObs: A slight problem I see, is that in the signature we're using an input parameter called "kernelTransaction" whereas in the docs we are interchangeably using either "kernelTransaction" or "statement". If possible, then I think some consistency would be nice. 

PS. I would probably prefer we use "statement" everywhere as "kernelTransaction" is a not really what this input parameter represents. Essentially the input is a "statement" which will be run after each successful "write transaction".

AzuObs: This problem also occurs in the apoc.trigger.install procedure page.

docs/asciidoc/modules/ROOT/pages/overview/apoc.trigger/apoc.trigger.install.adoc
```diff
@@ -0,0 +1,47 @@
...
+This file is generated by DocsTest, so don't change it!
+////
+
+= apoc.trigger.install
+:description: This section contains reference documentation for the apoc.trigger.install procedure.
+
+label:procedure[] label:apoc-core[]
+
+[.emphasis]
+CALL apoc.trigger.install(databaseName, name, statement, selector, config) | add a trigger kernelTransaction under a name, in the kernelTransaction you can use $createdNodes, $deletedNodes etc., the selector is {phase:'before/after/rollback/afterAsync'} returns previous and new trigger information. Takes in an optional configuration.
```
AzuObs, line 11:
```suggestion
`CALL apoc.trigger.install(databaseName, name, statement, selector, config) | eventually adds a trigger for a given database which is invoked when successful write transactions occur.
```

AzuObs: I wonder if information such as "in the kernelTransaction you can use $createdNodes, $deletedNodes etc., the selector is {phase:'before/after/rollback/afterAsync'} returns previous and new trigger information. Takes in an optional configuration" wouldn't be clearer if we put it further down in the document, where the inputs are described.

We could use the same kind of table we use for https://neo4j.com/labs/apoc/4.4/overview/apoc.periodic/apoc.periodic.iterate/ which IMO is a slight improvement.

docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.trigger.drop.adoc
```diff
@@ -0,0 +1,5 @@
+¦xref::overview/apoc.trigger/apoc.trigger.drop.adoc[apoc.trigger.drop icon:book[]] +
+
+`CALL apoc.trigger.drop(databaseName, name) | remove previously added trigger, returns trigger information`
```
AzuObs, line 3:
```suggestion
`CALL apoc.trigger.drop(databaseName, name) | eventually removes an existing trigger for a database, returns the trigger's information````

File: docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.trigger.start.adoc [hunk 5]
```diff
@@ -0,0 +1,5 @@
+¦xref::overview/apoc.trigger/apoc.trigger.start.adoc[apoc.trigger.start icon:book[]] +
+
+`CALL apoc.trigger.start(databaseName, name) | it resumes the paused trigger`
```
AzuObs, line 3:
```suggestion
`CALL apoc.trigger.start(databaseName, name) | eventually resumes the paused trigger`
```

docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.trigger.stop.adoc
```diff
@@ -0,0 +1,5 @@
+¦xref::overview/apoc.trigger/apoc.trigger.stop.adoc[apoc.trigger.stop icon:book[]] +
+
+`CALL apoc.trigger.stop(databaseName, name) | it pauses the trigger`
```
AzuObs, line 3:
```suggestion
`CALL apoc.trigger.stop(databaseName, name) | eventually pauses the trigger`
```

docs/asciidoc/modules/ROOT/examples/generated-documentation/documentation.csv
```diff
@@ -285,12 +285,17 @@
...
 ¦procedure¦apoc.trigger.resume¦apoc.trigger.resume(name :: STRING?) :: (name :: STRING?, query :: STRING?, selector :: MAP?, params :: MAP?, installed :: BOOLEAN?, paused :: BOOLEAN?)¦CALL apoc.trigger.resume(name) | it resumes the paused trigger¦true¦xref::background-operations/triggers.adoc
+¦procedure¦apoc.trigger.start¦apoc.trigger.start(databaseName :: STRING?, name :: STRING?) :: (name :: STRING?, query :: STRING?, selector :: MAP?, params :: MAP?, installed :: BOOLEAN?, paused :: BOOLEAN?)¦CALL apoc.trigger.start(databaseName, name) | it resumes the paused trigger¦true¦xref::background-operations/triggers.adoc
+¦procedure¦apoc.trigger.stop¦apoc.trigger.stop(databaseName :: STRING?, name :: STRING?) :: (name :: STRING?, query :: STRING?, selector :: MAP?, params :: MAP?, installed :: BOOLEAN?, paused :: BOOLEAN?)¦CALL apoc.trigger.stop(databaseName, name) | it pauses the trigger¦true¦xref::background-operations/triggers.adoc
```
AzuObs, line 298: I would like to propose that we add the word "eventually" to each description if you agree?

docs/asciidoc/modules/ROOT/pages/overview/apoc.trigger/index.adoc
```diff
@@ -46,6 +71,16 @@ CALL apoc.trigger.resume(name) \| it resumes the paused trigger
...
+
+
+|label:function[]
+|label:apoc-full[]
+|xref::overview/apoc.trigger/apoc.trigger.toRelationship.adoc[apoc.trigger.toRelationship icon:book[]]
+
+
 |label:function[]
 |label:apoc-full[]
 |===
```
AzuObs, line 86: To make this page clearer, I was wondering if we couldn't split the page into two sections? The top section would contain the new trigger procedures with a brief explanation of the new changes (eg they are on a per-database basis and work in a cluster environment). The bottom section would contain the old trigger procedures with a brief mention that they are now deprecated.

I think this would help users view at a glance what they should and shouldn't be focusing on.

What do you think?

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -141,7 +136,8 @@ with the current date and the property updated, we can do:
 
 [source,cypher]
 ----
-CALL apoc.trigger.add('setLastUpdate',
+USE system 
+CALL apoc.trigger.install('neo4j', 'setLastUpdate',
```
AzuObs, line 140: Does it work with Fabric like this? Or do you mean to say `:use system`? 

`USE system` is not equivalent to `:use system` because the first relies on Fabric routing and the latter relies on drivers routing. 

I couldn't see any tests where we are using Fabric to route to the system database, maybe that's possible but if so then I'd like to be sure. If not then maybe it's better to use `:use system` here instead?

vga91: Yes, both `:use system` and `USE system CALL apoc.trigger.install`, with a `neo4j://` protocol, works.

I didn't add tests for fabric https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183#discussion_r1018198006 

But maybe we can make the documentation more generic: that is, something like putting a `[NOTE]` at the beginning saying that these procedures have to be used via `":use system"`, by opening a session in "system" db with neo4j driver, or via `"USE system ..."` .

And as an example here just put `CALL apoc.trigger.install('neo4j', 'setLastUpdate', ...`.
This way it is up to the user what he wants to use. 
Wdyt?

docs/asciidoc/modules/ROOT/pages/overview/apoc.trigger/index.adoc
vga91: Yes, good idea. I will add it.

docs/asciidoc/modules/ROOT/pages/overview/apoc.trigger/apoc.trigger.install.adoc
vga91: Good point, I "inherited" this quite confusing documentation from the `apoc.trigger.add`. I will update it.

docs/asciidoc/modules/ROOT/pages/overview/apoc.trigger/index.adoc
vga91: Indeed this `xref` links point to `background-operations/triggers.adoc`, so I added the explanation directly on that page, as the user does not necessarily start from this page when looking for the documentation

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
vga91: Changed

docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.trigger.install.adoc
vga91: Changed

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
AzuObs: Ah yes good point I remember that conversation now! 

I think your suggestion is good 👍

docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.trigger.dropAll.adoc
```diff
@@ -0,0 +1,5 @@
+¦xref::overview/apoc.trigger/apoc.trigger.dropAll.adoc[apoc.trigger.dropAll icon:book[]] +
+
+`CALL apoc.trigger.dropAll(databaseName) | eventually removes all previously added trigger, returns triggers' information``
```
AzuObs, line 3:
```suggestion
`CALL apoc.trigger.dropAll(databaseName) | eventually removes all previously added trigger, returns triggers' information`
```

docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.trigger.install.adoc
```diff
@@ -0,0 +1,5 @@
+¦xref::overview/apoc.trigger/apoc.trigger.install.adoc[apoc.trigger.install icon:book[]] +
+
+`CALL apoc.trigger.install(databaseName, name, statement, selector, config) | eventually adds a trigger for a given database which is invoked when a successful transaction occurs..`
```
AzuObs, line 3:
```suggestion
`CALL apoc.trigger.install(databaseName, name, statement, selector, config) | eventually adds a trigger for a given database which is invoked when a successful transaction occurs.`
```

docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.trigger.start.adoc
```diff
@@ -0,0 +1,5 @@
+¦xref::overview/apoc.trigger/apoc.trigger.start.adoc[apoc.trigger.start icon:book[]] +
+
+`CALL apoc.trigger.start(databaseName, name) | eventually pauses the paused trigger`
```
AzuObs, line 3:
```suggestion
`CALL apoc.trigger.start(databaseName, name) | eventually unpauses the paused trigger`
```

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -6,32 +6,40 @@
 Triggers allow the registration of Cypher queries that are called when data in Neo4j is changed (created, updated, deleted).
 You can run them before or after commit.
 
+[NOTE]
+====
+All these `apoc.trigger.*` procedures are implemented to be executed in the system database,
+therefore they have to be used executing first `:use system` (in cypher-shell or Neo4j Browser), 
+by opening a system db session with a neo4j driver, 
+or by prepending the suffix `USE system` to the cypher statement.
```
AzuObs, line 14:
```suggestion
All these `apoc.trigger.*` procedures are intended to be executed in the system database,
therefore they have to be used executed by opening a system database session. There are several ways of doing this:
- when using cypher-shell or Neo4j Browser, one can prefix their Cypher query with `:use system`
- when using fabric, one can prefix their Cypher query with `USE system` 
- when using the drivers, one can open a session directly against the system database
```

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -66,14 +74,29 @@ You can use these helper functions to extract nodes or relationships by label/re
...
 |===
 
+The 3rd parameter of the `apoc.trigger.install()` is a map `{phase: PHASE}`, where `PHASE` is a string which can have one of the following values:
+
+.Trigger Phase Table
+[cols="1m,5"]
+|===
+| PHASE | Description
+| before | The trigger will be activate right `before` the commit. If no phase is specified, it's the default.
+| rollback | The trigger will be activate right after the  `rollback`
```
AzuObs, line 84:
```suggestion
| rollback | The trigger will be activate right after the `rollback`
```

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -66,14 +74,29 @@ You can use these helper functions to extract nodes or relationships by label/re
...
+| after | The trigger will be activate right `after` the commit
```
AzuObs, line 85:
```suggestion
| after | The trigger will be activate right `after` the commit.
```

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -283,24 +309,26 @@ To solve this, we can use `{phase:'afterAsync'}`
 
 .Pause trigger
 
-We have the possibility to pause a trigger without remove it, if we will need it in the future
+We have the possibility to pause a trigger without remove it, if we will need it in the future.
+Note that this procedure and the `apoc.trigger.start` one are eventually consistent,
+so we need to wait a time defined by the configuration `apoc.trigger.refresh`.
```
AzuObs, line 314:
```suggestion
We have the possibility of pausing a trigger without removing it, in case we might need it in the future.
Note that this procedure and the `apoc.trigger.start` ones are eventually consistent,
so we need to wait some amount of time for the changes to propagate. This amount of time is defined by the configuration `apoc.trigger.refresh`.
```

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -365,27 +393,58 @@ because using this one, the rebuild of entities is executed automatically under
...
+CALL apoc.trigger.install('neo4j', 'count-removals', 
+    'MATCH (c:Counter) SET c.count = c.count + size([f IN $deletedNodes WHERE id(f) > 0])', 
+    {})
+----
+
+we can execute (also in this case, after a time defined by the configuration `apoc.trigger.refresh`):
+
+[source,cypher]
+----
+CALL apoc.trigger.list
```
AzuObs, line 439: Should this be changed to apoc.trigger.show() or do you want to do that later? I don't mind either way, this is just a reminder.

docs/asciidoc/modules/ROOT/pages/overview/apoc.trigger/apoc.trigger.install.adoc
```diff
@@ -0,0 +1,47 @@
...
+This file is generated by DocsTest, so don't change it!
+////
+
+= apoc.trigger.install
+:description: This section contains reference documentation for the apoc.trigger.install procedure.
+
+label:procedure[] label:apoc-core[]
+
+[.emphasis]
+CALL apoc.trigger.install(databaseName, name, statement, selector, config) | eventually adds a trigger for a given database which is invoked when a successful transaction occurs..
```
AzuObs, line 11:
```suggestion
CALL apoc.trigger.install(databaseName, name, statement, selector, config) | eventually adds a trigger for a given database which is invoked when a successful transaction occurs.
```

docs/asciidoc/modules/ROOT/pages/overview/apoc.trigger/index.adoc
```diff
@@ -5,24 +5,49 @@ This file is generated by DocsTest, so don't change it!
...
 |label:apoc-core[]
+
 |xref::overview/apoc.trigger/apoc.trigger.pause.adoc[apoc.trigger.pause icon:book[]]
 
 CALL apoc.trigger.pause(name) \| it pauses the trigger
 |label:procedure[]
 |label:apoc-core[]
+|===
+
+== Apoc deprecated procedures
```
AzuObs, line 35:
```suggestion
== APOC deprecated procedures 
```

docs/asciidoc/modules/ROOT/pages/overview/apoc.trigger/index.adoc
```diff
@@ -38,14 +63,40 @@ removes all previously added trigger, returns trigger information
...
+|label:procedure[]
+|label:apoc-core[]
+|xref::overview/apoc.trigger/apoc.trigger.stop.adoc[apoc.trigger.stop icon:book[]]
 
+CALL apoc.trigger.stop(databaseName, name) \| eventually pauses the trigger
+|label:procedure[]
+|label:apoc-core[]
+|===
+
+== APOC full functions
```
AzuObs, line 78:
```suggestion
== APOC Full functions
```

docs/asciidoc/modules/ROOT/pages/overview/apoc.trigger/index.adoc
```diff
@@ -5,24 +5,49 @@ This file is generated by DocsTest, so don't change it!
 = apoc.trigger
 :description: This section contains reference documentation for the apoc.trigger procedures.
 
-[.procedures, opts=header, cols='5a,1a,1a']
+== Apoc new procedures
```
AzuObs, line 8:
```suggestion
== APOC new procedures 
```

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
vga91: Personally I would do it in the other pr :)

PR #3335, opened 2022-11-24
[gRNAet7B] Added trigger show procedure

vga91:
I guess we can't implement `apoc.trigger.show()` in the same way as `apoc.trigger.list`, 
because the latter uses a `triggerHandler` found only in system databases.

So I guess we should fetch information from system nodes directly.

------

~~My doubt is that, in this way, since the triggers are eventually consistent, 
we could have an `apoc.trigger.show` that shows a trigger which is not actually active yet.
One way to fix it would be to do `ApocConfig.apocConfig().getDatabase(databaseName).executeTransactionally(apoc.trigger.list())`, 
but in 5.x it wouldn't work for [this reason](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183#discussion_r1005547491).
Or, as done in this pr, I think we should add an `updated` property that compares the current timestamp with that of `lastUpdated` and the `apoc.trigger.refresh` configuration (so it's true if `lastUpdated - currentType > interval`)
Obviously it is not consistent, but it's useful to give the administrator a rough idea about the trigger.
Wdyt?~~


ncordon (changes requested):
I have doubts on whether the purpose of `trigger.show` should be to indicate whether a trigger has been discovered by the database yet or rather just list the ones the user installed (doesn't mean they've been picked up yet, but that's the spirit of eventually consistent things).

In the first case, it cannot be solved gracefully and I'd rather not merge this PR and make the user depend then on `apoc.trigger.list` against a user database. A trigger can be listed there but doesn't mean for another instance is discovered yet, just for the particular instance `trigger.show` ends up being executed on.

If we add a `trigger.show`, I'd rather just list the triggers that the system database is aware of (just go to the system database and grab the triggers that are there).

vga91:
> I have doubts on whether the purpose of `trigger.show` should be to indicate whether a trigger has been discovered by the database yet or rather just list the ones the user installed (doesn't mean they've been picked up yet, but that's the spirit of eventually consistent things).
> ...

I changed the `trigger.show` implementation by removing the `updated` output

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -138,12 +141,22 @@ public Stream<TriggerInfo> stop(@Name("databaseName") String databaseName, @Name
...
+        Map<String, Object> resume = TriggerHandlerNewProcedures.updatePaused(databaseName, name, false);
 
-        return Stream.of(new TriggerInfo(name,
-                (String)resume.get("statement"),
-                (Map<String,Object>) resume.get("selector"),
-                (Map<String,Object>) resume.get("params"),true, false));
+        return Stream.of(TriggerInfo.from(resume, true));
     }
 
+    // TODO - perhaps change with @SystemOnlyProcedure
```
AzuObs, line 149: ```suggestion```

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -138,12 +141,22 @@ public Stream<TriggerInfo> stop(@Name("databaseName") String databaseName, @Name
...
+    @SystemProcedure
+    @Procedure(mode = Mode.READ)
+    @Description("CALL apoc.trigger.show(databaseName) | it lists all installed triggers")
+    public Stream<TriggerInfo> show(@Name("databaseName") String databaseName) {
+        // TODO - perhaps not needed
```
AzuObs, line 154:
```suggestion
```

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -499,6 +505,35 @@ public void testDeleteRelationships() {
...
+                res -> {
+                    Map<String, Object> row = res.next();
+                    assertEquals(name, row.get("name"));
+                    assertEquals(query, row.get("query"));
+                    row = res.next();
+                    assertEquals(name2, row.get("name"));
+                    assertEquals(query, row.get("query"));
+                    assertFalse(res.hasNext());
+                });
+    }
```
AzuObs, line 536: Could we also add a test where we install a procedure on a database "foo" and then assert that it isn't returned when showing triggers from database "neo4j"?

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -499,6 +505,35 @@ public void testDeleteRelationships() {
     //
     // new test cases
     //
+
+    @Test
+    public void testTriggerShow() throws Exception {
```
AzuObs, line 510:
```suggestion
    public void testTriggerShow() {
```

AzuObs (changes requested):
Hi Giuseppe, I'm sorry but I'm finding this PR a bit hard to review because it contains two different pieces of work:
- the changes to introduce apoc.trigger.show
- the changes to de-flake the trigger procedures

Would it be possible to split this PR into two PRs, one for each unit of work? I feel like this is generally quite a good practice to follow and can save us all some time in the long run.

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -532,4 +567,59 @@ public void testInstallTriggerInSystemDb() {
...
+            
+            // this does nothing, just to test consistency
+            sysDb.executeTransactionally("CALL apoc.trigger.install('neo4j', $name, 'return 1', {phase: 'afterAsync'})",
+                    map("name", UUID.randomUUID().toString()) );
+
+            // drop trigger and check if dropped
+            sysDb.executeTransactionally("CALL apoc.trigger.drop('neo4j', $name)", Map.of("name", name));
+            testCallCountEventually(db, "CALL apoc.trigger.list() yield name where name = $name return name",
+                    Map.of("name", name),
+                    0, TIMEOUT);
```
AzuObs, line 620: This probably ought to be the new show procedure since I don't think we ought to use a mix of new and deprecated procedures in the same test. What do you think?

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -532,4 +567,59 @@ public void testInstallTriggerInSystemDb() {
...
+                    (row) -> assertEquals(i, row.get("count")));
+
+            // stop trigger
+            sysDb.executeTransactionally("CALL apoc.trigger.stop('neo4j', $name)",
+                    map("name", name, "query", query));
+            awaitTriggerDiscovered(db, name, query, true);
+
+            // this does nothing, just to test consistency
+            sysDb.executeTransactionally("CALL apoc.trigger.install('neo4j', $name, 'return 1', {phase: 'afterAsync'})",
+                    map("name", UUID.randomUUID().toString()) );
```
AzuObs, line 605: I'm not sure how this test consistency? Same for the block of code from L.576-578

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -532,4 +567,59 @@ public void testInstallTriggerInSystemDb() {
         }
     }
 
+    @Test
+    public void testEventualConsistency() {
```
AzuObs, line 571: This test is taking > 3 minutes which is very long. We can discuss it in a separate PR if you agree that it's good to split different work threads into different PRs. 

I'm not sure that we should be testing this 100 times, I would have thought that a single time would suffice. Concurrency conflicts are very hard to detect at best of times, but I don't think we ought to slow the whole CI down because of it. 

What do you think?

core/src/main/java/apoc/trigger/TriggerHandlerNewProcedures.java
```diff
@@ -47,59 +49,65 @@ public static void checkEnabled() {
...
+            node[0].setProperty(SystemPropertyKeys.selector.name(), Util.toJson(selector));
+            node[0].setProperty(SystemPropertyKeys.params.name(), Util.toJson(params));
+            node[0].setProperty(SystemPropertyKeys.paused.name(), false);
 
-            setLastUpdate(databaseName, tx);
             return null;
         });
 
+        setLastUpdate(databaseName, node[0]);
+
```
AzuObs, line 70: Could you help me understand why this change is needed?

The way I see it, splitting the trigger creation/update logic in one transaction, and then update logic for the metadata in another transaction is harmful. Because the first transaction could commit and the second transaction could fail for whatever reason. This scenario would leave the dbms in an inconsistent state, and would mean that although a new triggers were installed in "system", the other databases would not update because they the metadata wouldn't update.

They should all happen in the same transaction (not separate and not nested transactions).

And to be even more correct, we should remove all the calls to withSystemDb which opens a new nested transaction which is discussed in this [comment thread](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183#discussion_r1005585858) but we don't have to do that now.

Let's have a discussion Slack or a call if there is any confusion here.

vga91: Maybe it's not needed (I hope), I should have found another way to solve that.
I created a new pr with the flaky stuff, https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3355.

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
vga91: Changed in the related pr

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
vga91: Answered in the other pr too
```
for some reason by adding more triggers, 
the probability of making the test flaky seems higher 
(but anyway, it seems to affect only the embedded test instance)
```

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
vga91: I think it depends on [this](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3335#pullrequestreview-1194406053). 
I mean, the `apoc.trigger.show` currently check if the trigger is picked up, but it may be that it has not actually been updated in the event listener cache (i.e. in `activeTriggers` map, present in `TriggerHandler.java`). 
So the `apoc.trigger.list` in this case is more accurate for these tests, because it verifies that the event listener really has the updated trigger. 
For this reason, the apoc.trigger.list is not deprecated.

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
AzuObs: Did you forget to remove this test from the PR? It looks like the same test in another PR I reviewed yesterday (https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3355) and I can't seem to see how it is testing apoc.triggers.show

core/src/main/java/apoc/trigger/TriggerHandlerNewProcedures.java
```diff
@@ -122,6 +123,14 @@ public static Map<String, Object> dropAll(String databaseName) {
         return previous;
     }
 
+    public static List<Map<String, Object>> getTriggerNodes(String databaseName) {
+        return withSystemDb(tx -> getTriggerNodes(databaseName, tx, null)
+                .stream()
+                .map(TriggerHandlerNewProcedures::toTriggerInfo)
+                .collect(Collectors.toList())
+        );
+    }
```
AzuObs, line 132: Are you sure this need to be executed in new transaction? 

The only place it's called from is the apoc.trigger.show function which we check is already being executed in a system transaction. Also, none of the other getTriggerNodes functions are executed in a new transaction.

If it's not needed then can we remove it?

vga91: Yes, we can eliminate the transaction by leveraging the `@Context Transaction tx`.

Anyway, this method only works for this procedure (which has `@Procedure(mode = Mode.READ)`), the others with Mode.WRITE give an error: `"Creating new node label on database 'system' is not allowed for user '' with FULL overridden by READ overridden by READ. See GRANT CREATE NEW NODE LABEL ON DATABASE system "`

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
vga91: It seems so :(  However I did the rebase and removed it

AzuObs: No worries :) Thanks for cleaning it up!

docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.trigger.show.adoc
```diff
@@ -0,0 +1,5 @@
+¦xref::overview/apoc.trigger/apoc.trigger.show.adoc[apoc.trigger.show icon:book[]] +
+
+`CALL apoc.trigger.show(databaseName) | it lists all installed triggers`
+¦label:procedure[]
+¦label:apoc-core[]
```
AzuObs, line 5: The slight issue I see is that both "apoc.trigger.show" and "apoc.trigger.list" have the same short description "it lists all installed triggers"

I would like to propose
"CALL apoc.trigger.show(databaseName) | it lists all eventually installed triggers for any database"
"CALL apoc.trigger.list() | list all currently installed triggers for all databases for the session's database"

Anything you'd like to suggest is of course welcome!

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -451,6 +451,17 @@ CALL apoc.trigger.list
...
 
+
+[WARNING]
+====
+Please note that, since the trigger operations are eventually consistent (based on `apoc.trigger.refresh` configuration), 
+the `apoc.trigger.show` may return some triggers not yet added/updated/removed.
+
+If you want the list of actually working triggers, 
+you can use the xref::overview/apoc.trigger/apoc.trigger.list.adoc[apoc.trigger.list] against the user database ("neo4j" in the above case)
+====
```
AzuObs, line 462: "actually working triggers" vs "currently working triggers"

I don't mind which word we use in the descriptions, but if possible let's keep using the same one

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -451,6 +451,17 @@ CALL apoc.trigger.list
 SET c.count = c.count + size([f IN $deletedNodes WHERE id(f) > 0])" | {}       | {}     | TRUE      | FALSE
 |===
 
+
+[WARNING]
+====
+Please note that, since the trigger operations are eventually consistent (based on `apoc.trigger.refresh` configuration), 
+the `apoc.trigger.show` may return some triggers not yet added/updated/removed.
```
AzuObs, line 458: "actually working triggers" vs "currently working triggers"

I don't mind which word we use in the descriptions, but if possible let's keep using the same one

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -451,6 +451,17 @@ CALL apoc.trigger.list
...
+
+If you want the list of actually working triggers, 
+you can use the xref::overview/apoc.trigger/apoc.trigger.list.adoc[apoc.trigger.list] against the user database ("neo4j" in the above case)
```
AzuObs, line 461:
```suggestion
you can use the xref::overview/apoc.trigger/apoc.trigger.list.adoc[apoc.trigger.list] against the session database ("neo4j" in the above case)
```

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
vga91: Ok, I'll use currently :)

PR #3355, opened 2022-12-06
[gRNAet7B] Try de-flaky some tests with new procedure triggers

vga91:
I cannot replicate this test with a real instance, so maybe is a `TestDatabaseManagementServiceBuilder` problem.

To solve this flaky test perhaps either we have to increase the `apoc.trigger.refresh` value or we have to separate the transactions in `Trigger` system node updates and `setLastUpdate` ([i.e. in that way]https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3355/commits/ff6e19258c97215ad766111843a1aec947d3fe48).

I've been trying to figure out why these two things seem to de-flaky the tests, but i'm not really sure about it :X

Of course, the 2nd way it's not not recommended because we change the implementation and make the code less optimized.

 In addition, to further stabilize the tests, I added a `waitDbsAvailable`.

Without these edits, with the embedded neo4j test instances, it seems like very rarely the TriggerHandler.updateCache() can't find the installed/updated node, but because of `lastUpdate = System.currentTimeMillis();` (TriggerHandler, line 83 ) the updateCache is no longer called due to `if (getLastUpdate() > lastUpdate) {` (TriggerHandler, line 283).

### DIFF
--- a/core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
+++ b/core/src/main/java/apoc/trigger/TriggerHandlerWrite.java
@@ -6,6 +6,7 @@ import apoc.util.Util;
 import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.ResourceIterator;
 import org.neo4j.graphdb.Transaction;
+import org.neo4j.internal.helpers.collection.Iterators;
 import org.neo4j.internal.helpers.collection.Pair;
 
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


vga91:
About this comment https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3335#discussion_r1034596762.
I know it's weird, but for some reason by adding more triggers, the probability of making the test flaky seems higher (but anyway, it seems to affect only the embedded test instance)

AzuObs (approved):
LGTM. Well done!

PR #3356, opened 2022-12-06
fixes #3349: Can not create custom procedure with void return type using apoc.custom.declareProcedure

### LINKED ISSUE #3349: Can not create custom procedure with void return type using apoc.custom.declareProcedure
## Guidelines

It is not able to create a custom procedure using  apoc.custom.declareProcedure for void return type . 

dbms.setConfigValue() returns void and custom procedure can not be created . 

```
CALL apoc.custom.declareProcedure(
  'setrole(ssorole::STRING) :: (row::MAP)',
  
  'CALL {CALL dbms.listConfig() YIELD name, value WHERE name CONTAINS "dbms.security.oidc.keycloak.authorization.group_to_role_mapping" RETURN value} 
  WITH value
  CALL dbms.setConfigValue("dbms.security.oidc.keycloak.authorization.group_to_role_mapping",value+$ssorole);'
  
);
```
Throws ==> Failed to invoke procedure `apoc.custom.declareProcedure`: Caused by: java.lang.NullPointerException

It should create a custom procedure 

Throws ==> Failed to invoke procedure `apoc.custom.declareProcedure`: Caused by: java.lang.NullPointerException

Run the above code 

Currently used versions

### Versions

  - OS:
  - Neo4j:4.4.11
  - Neo4j-Apoc:4.4.11

conker84:
Fixes #3349

Extended Cypher Procedures in order to support `DBMS` and `SCHEMA` procedures.

  - Added the new modes
  - Added Test

### DIFF
--- a/full/src/test/java/apoc/custom/CypherProceduresTest.java
+++ b/full/src/test/java/apoc/custom/CypherProceduresTest.java
@@ -3,9 +3,12 @@ package apoc.custom;
 import apoc.RegisterComponentFactory;
 import apoc.SystemLabels;
 import apoc.SystemPropertyKeys;
+import apoc.log.Logging;
+import apoc.result.MapResult;
 import apoc.util.StatusCodeMatcher;
 import apoc.util.TestUtil;
 import org.apache.commons.lang.exception.ExceptionUtils;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
@@ -16,6 +19,8 @@ import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.QueryExecutionException;
 import org.neo4j.graphdb.Transaction;
 import org.neo4j.internal.helpers.collection.Iterators;
+import org.neo4j.procedure.Procedure;
+import org.neo4j.procedure.builtin.BuiltInDbmsProcedures;
 import org.neo4j.test.rule.DbmsRule;
 import org.neo4j.test.rule.ImpermanentDbmsRule;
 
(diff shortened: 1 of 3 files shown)


full/src/test/java/apoc/custom/CypherProceduresTest.java
```diff
@@ -839,6 +847,23 @@ public void testIssue2032() {
         assertProcedureFails(String.format(SIGNATURE_SYNTAX_ERROR, procedureSignature), 
                 "call apoc.custom.declareProcedure('" + procedureSignature + "','RETURN $first + $s AS answer')");
     }
+
+    @Test
+    public void testIssue3349() {
+        String procedure = "CALL apoc.custom.declareProcedure(\n" +
```
gem-neo4j, line 853: The issue asks for a void return type from what I can tell, this looks like it is testing not having any parameters to pass in, can a test be added for testing a void return? Or will that be separate?

conker84: @gem-neo4j thanks for the question, the issue has an example with the following signature:
`setrole(ssorole::STRING) :: (row::MAP)`, that triggers an error, then the user said that he also tested with `VOID` return type. So as the issue is not related to the return type but to the fact (as discussed in the trello card) that was not meant to manage `DBMS` and `SCHEMA` query types, I used an example as much similar as possible to the one shared in the GH issue (https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/3349).

Btw if you want I can add a test also with the return type.

Do you also want a test with a custom procedure with parameters? With `VOID` or another return type?

gem-neo4j: It would be nice to have, considering the title of this fix specifies it is for void return type :)

conker84: @gem-neo4j done!

gem-neo4j (approved):

conker84:
I merged it because the CI was red for this test `TriggerNewProceduresTest#testTriggerPause` which I suppose it's flaky

PR #3357, opened 2022-12-07
[qwJFy8pp] Add @Admin to trigger procedures

vga91:
- Added `@Admin` annotation to trigger procedures
- Added `testTriggersAllowedOnlyWithAdmin`

---

Please note that this is a breaking change
that we've introduced in order to improve procedures security.

---

Better to use `@Admin` check instead of `Util.checkAdmin`, like [here](https://github.com/neo4j-contrib/neo4j-apoc-procedures/commit/8a9d0ca3758817ad92dc61b39cc7e54451cfcc44) :
- it will most likely be compatible with future versions
- it doesn't require additional contexts:
- we don't have to specify the procedure name, e.g. `Util.checkAdmin(securityContext, callContext,"apoc.trigger.install");`

```
    @Context
    public SecurityContext securityContext;
    @Context
    public ProcedureCallContext callContext;
```
- similar error, but with more details

```
// with Admin annotation
Executing admin procedure 'apoc.trigger.remove' permission has not been granted for user 'name' with roles [PUBLIC, editor, publisher] restricted to TOKEN_WRITE.
```

```
// with Util.checkAdmin
Failed to invoke procedure `apoc.trigger.add`: Caused by: java.lang.RuntimeException: This procedure apoc.trigger.install is only available to admin users
```

### DIFF
--- a/core/src/main/java/apoc/trigger/Trigger.java
+++ b/core/src/main/java/apoc/trigger/Trigger.java
@@ -103,6 +103,7 @@ public class Trigger {
         return new TriggerInfo(name, null, null, false, false);
     }
 
+    @Admin
     @Procedure(mode = Mode.READ)
     @Description("list all installed triggers")
     public Stream<TriggerInfo> list() {
--- a/core/src/main/java/apoc/trigger/Trigger.java
+++ b/core/src/main/java/apoc/trigger/Trigger.java
@@ -4,6 +4,7 @@ import apoc.util.Util;
 import org.neo4j.configuration.GraphDatabaseSettings;
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.logging.Log;
+import org.neo4j.procedure.Admin;
 import org.neo4j.procedure.Context;
 import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Mode;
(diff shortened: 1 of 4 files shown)


core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -112,6 +116,7 @@ public Stream<TriggerInfo> list() {
                 );
     }
 
+    @Admin
```
nadja-muller, line 119: Should apoc.trigger.list also have @Admin annotation?

nadja-muller (commented):
Looks good. I'm not sure whether we should also add a @Admin annotation to the list procedure.
Also, we should make sure that it is clear:
- This is a breaking change
- We do it for improving security.
Could you write this in the description of the PR?

core/src/main/java/apoc/trigger/Trigger.java
vga91: Well yes, it probably makes more sense for all procedures to be `@Admin` for consistency.
I added the `apoc.trigger.list` annotation.
I'll add `trigger.show` to this pr when the other pr will be merged, or vice versa.

docs/asciidoc/modules/ROOT/pages/background-operations/triggers.adoc
```diff
@@ -18,6 +18,11 @@ Moreover, they accept as first parameter the name of the database towards which
 
 Through this implementation, we can use these procedures in a cluster environment,
 by leveraging the cluster routing mechanism.
+
+Moreover, to use these procedures the user must have the `admin` role, 
+otherwise the procedure throws with an error `permission has not been granted for user 'xxx'`,
+to prevent anyone from adding a trigger that could escalate privileges.
```
nadja-muller, line 24: We discussed that we usually don't explain the background of the privileges in the documentation, so I would remove this section.

nadja-muller (changes requested):

nadja-muller (approved):

PR #3361, opened 2022-12-09
[wokdlhX3] Extract code from APOC core which is only used in APOC extended

Lojjs:
Should go in together with https://github.com/neo4j/apoc/pull/261

### DIFF
--- a/extended/src/main/java/apoc/load/LoadCsv.java
+++ b/extended/src/main/java/apoc/load/LoadCsv.java
@@ -3,6 +3,7 @@ package apoc.load;
 import apoc.Extended;
 import apoc.export.util.CountingReader;
 import apoc.load.util.LoadCsvConfig;
+import apoc.util.ExtendedUtil;
 import apoc.util.FileUtils;
 import apoc.util.Util;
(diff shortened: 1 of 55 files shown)


extended/src/main/java/apoc/load/LoadCsv.java
```diff
@@ -87,7 +88,8 @@ private String[] getHeader(CSVReader csv, LoadCsvConfig config) throws IOExcepti
         Map<String, Mapping> mappings = config.getMappings();
         for (int i = 0; i < headers.length; i++) {
             String header = headers[i];
-            if (ignore.contains(header) || mappings.getOrDefault(header, Mapping.EMPTY).ignore) {
+            Mapping empty = new Mapping("", Collections.emptyMap(), LoadCsvConfig.DEFAULT_ARRAY_SEP, false);
```
conker84, line 91: Maybe adding it as a static final?

conker84 (changes requested):
Thank you so much @Lojjs, just a couple of things that I want to discuss with you

conker84 (approved):
@Lojjs LGTM thank you so much!

PR #3393, opened 2023-01-09
[gRNAet7B] Changes from neo4j/apoc#250

vga91:
Changes from https://github.com/neo4j/apoc/pull/250
- Added tests 
- Simplified implementation by adding `TriggerInfo.java`

Fixed adoc bug by adding `[]` at the end of the includes

### DIFF
--- a/core/src/main/java/apoc/trigger/TriggerHandlerNewProcedures.java
+++ b/core/src/main/java/apoc/trigger/TriggerHandlerNewProcedures.java
@@ -11,7 +11,6 @@ import org.neo4j.internal.helpers.collection.Pair;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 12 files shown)


core/src/main/java/apoc/trigger/TriggerHandlerNewProcedures.java
```diff
@@ -45,89 +34,84 @@ public static void checkEnabled() {
...
             node.setProperty(SystemPropertyKeys.params.name(), Util.toJson(params));
             node.setProperty(SystemPropertyKeys.paused.name(), false);
 
             setLastUpdate(databaseName, tx);
-            return null;
         });
 
-        return previous;
+        return previous.get();
     }
```
AzuObs, line 57: Hi Giuseppe, I didn't notice this in the dev PR but could you explain why you think you need an atomic reference here? 

AtomicReference is a concurrent programming construct, but this function is executed in a single thread and has no mutable state which is accessible from another thread. So the way I see it, it shouldn't be needed.

AzuObs: It's not really doing any harm, but it's just confusing to see here and makes the reader question what is happening and why it's needed, which doesn't help with maintainability. So if there is no use for it, could we please remove it from here and dev too?

AzuObs (changes requested):
Great backport 😸 One small nitpick.

core/src/main/java/apoc/trigger/TriggerHandlerNewProcedures.java
vga91: Hi @AzuObs, I used an `AtomicReference` just to get rid of the `Variable used in lambda expression should be final or effectively final` compilation's error within the lambda.

I just changed them to a one element final array

AzuObs: Apologies, now that you mention, I remember us having a conversation about this before. Thanks for making the change.

AzuObs (approved):

PR #3409, opened 2023-01-17
[qZZ3O2uX] Updates jackson-databind to fix CVE-2020-36518, CVE-2022-42004, CVE-2022-42003  

gem-neo4j:
vuln-fix: Update dependencies which import jackson-databind that had the following vulnerabilities: CVE-2020-36518, CVE-2022-42004, CVE-2022-42003.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -118,8 +118,8 @@ dependencies {
     testCompile 'org.mock-server:mockserver-netty:5.13.0'
     testCompile 'org.mock-server:mockserver-client-java:5.13.0'
 
-    compileOnly group: 'com.amazonaws', name: 'aws-java-sdk-comprehend', version: '1.12.214' , withoutJacksons
-    testImplementation group: 'com.amazonaws', name: 'aws-java-sdk-comprehend', version: '1.12.214' , withoutJacksons
+    compileOnly group: 'com.amazonaws', name: 'aws-java-sdk-comprehend', version: '1.12.353' , withoutJacksons
+    testImplementation group: 'com.amazonaws', name: 'aws-java-sdk-comprehend', version: '1.12.353' , withoutJacksons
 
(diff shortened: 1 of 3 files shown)


core/build.gradle
```diff
@@ -118,8 +118,8 @@ dependencies {
     testCompile 'org.mock-server:mockserver-netty:5.13.0'
     testCompile 'org.mock-server:mockserver-client-java:5.13.0'
 
-    compileOnly group: 'com.amazonaws', name: 'aws-java-sdk-comprehend', version: '1.12.214' , withoutJacksons
-    testImplementation group: 'com.amazonaws', name: 'aws-java-sdk-comprehend', version: '1.12.214' , withoutJacksons
+    compileOnly group: 'com.amazonaws', name: 'aws-java-sdk-comprehend', version: '1.12.353' , withoutJacksons
```
Lojjs, line 121: According to the changelog https://github.com/aws/aws-sdk-java/blob/master/CHANGELOG.md I think this was fixed in 1.12.322, but SNYK suggest to update to the latest version, which as of yesterday is 1.12.388. Any particular reason why you upgraded to specifically 1.12.353? I think in general the newest stable version which has the fix in it and work with our other dependencies would be good

gem-neo4j: Because the absolute latest version will introduce a breaking change for us. In AWS Entity (https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-comprehend/src/main/java/com/amazonaws/services/comprehend/model/Entity.java#L79) a new field was introduced. APOC returns this whole structure in some of the procedures, which means a new field would be added, it didn't feel worth it to update for this tbh especially as it is in extended. So the bump fixes the databind issues, and doesn't introduce new behaviour :)

Lojjs: I see, that makes sense

Lojjs (approved):
Great job, thanks!

PR #3415, opened 2023-01-19
[21wxsujh] Try to fix APOC redirect test

vga91:
Since the flakiness doesn't appear to be present in 5.x 
and there doesn't seem to be substantial differences between the two versions, 
maybe it makes sense to try to fix the problem by updating the testcontainer with the same version as the 5.x.

TODO - try retrigger multiple times


gem-neo4j (approved):
Seems like a good try :) It passed this time at least, if you run it a few more times, then I'd say it is good to go :D

vga91:
Ok I re-triggered the build 7 times, and it only failed once for another reason (S3 tests).
Seems fixed, I hope :)

PR #3419, opened 2023-01-22
Change neo4j reference to 5.3.0

vga91:
In order to compile cherry-picks to 5.4 and https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3397,
waiting for neo4j 5.4 public release


vga91:
It fails because the submodule is `dev`. it has been changed to the correct one here: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3397

PR #3421, opened 2023-01-24
[hDJkPkOQ] Add QueryType validation in apoc.trigger.add

vga91:
Added QueryType validations in validateQuery() method,
so as to fail if the procedure tries to execute schema operations.

See point 2 here: https://github.com/neo4j/apoc/pull/208#pullrequestreview-1145222227

Only for 4.4, since the `apoc.trigger.add` is deprecated and in 5.x the `validateQuery()` is not longer present.

### DIFF
--- a/core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
+++ b/core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
@@ -583,4 +583,18 @@ public class TriggerNewProceduresTest {
                 map("name", UUID.randomUUID().toString()) );
     }
 
+    @Test
+    public void testTriggerInstallWithSchemaOperation() {
+        String query = "CREATE INDEX periodicIdx FOR (n:Bar) ON (n.first_name, n.last_name)";
+        try {
+            testCall(sysDb, "CALL apoc.trigger.install('neo4j', 'triggerSchema', 'CREATE INDEX periodicIdx FOR (n:Bar) ON (n.first_name, n.last_name)', {phase: 'before'})",
+                    Map.of("query", query),
+                    (row) -> fail("Should fail because of unsupported schema operation"));
+        } catch (RuntimeException e) {
+            final String expected = "Failed to invoke procedure `apoc.trigger.install`: " +
+                    "Caused by: java.lang.RuntimeException: Supported query types for the operation are [READ_ONLY, WRITE, READ_WRITE]";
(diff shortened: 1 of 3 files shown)


core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -583,4 +583,18 @@ public void testEventualConsistency() {
...
+    @Test
+    public void testTriggerInstallWithSchemaOperation() {
+        String query = "CREATE INDEX periodicIdx FOR (n:Bar) ON (n.first_name, n.last_name)";
+        try {
+            testCall(sysDb, "CALL apoc.trigger.install('neo4j', 'triggerSchema', 'CREATE INDEX periodicIdx FOR (n:Bar) ON (n.first_name, n.last_name)', {phase: 'before'})",
+                    Map.of("query", query),
+                    (row) -> fail("Should fail because of unsupported schema operation"));
+        } catch (RuntimeException e) {
+            final String expected = "Failed to invoke procedure `apoc.trigger.install`: " +
+                    "Caused by: java.lang.RuntimeException: Supported query types for the operation are [READ_ONLY, WRITE, READ_WRITE]";
```
AzuObs, line 595: Is there anyway that this could be a more user friendly error message? 

`"The trigger statement must contain READ_ONLY, WRITE, or READ_WRITE query."`

AzuObs (changes requested):

vga91:
> Is there anyway that this could be a more user friendly error message?
> `"The trigger statement must contain READ_ONLY, WRITE, or READ_WRITE query."`

1) while I was making this change,
I realized that it would be better to add this thing to the validate too (https://github.com/neo4j/apoc/commit/e5264e06a1423ebf9f9f8eef88e020e84fa83d28),
i.e. I added `Set.of(Mode.WRITE, Mode.READ, Mode.DEFAULT),` and `testTriggerInstallWithSchemaProcedure` .
Basically as it was before it didn't give an error with a `@Procedure(Mode=SCHEMA)` e.g. `apoc.schema.assert`

2) Along with that, I've added two parameters to the `Util.validateQuery()` with custom error messages.

If you think point `1.` is unnecessary / to be separated / another card needs to be created, I'll be happy to change the pr :)

AzuObs (approved):
Thank you for the change! Looks good 🙂

PR #3454, opened 2023-02-16
[TOQo1n9M] apoc-hadoop dependency is conflicting (#3450)

vga91:
Cherry-pick of #3450.

Waiting for core part merge.

Unlike 4.4 version, this is a test only pr because it already worked before.
Differences from cherry pick:
- added `addExtraDependencies()` in ExtendedTestContainerUtil (instead of TestContainerUtil)
- Created StartupExtendedTest.java
- Deleted StartupTest.java


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "apoc-core"]
 	path = apoc-core
 	url = https://github.com/neo4j/apoc
-	branch = dev
+	branch = hadoop-dependency-conflict_core-dev
```
Lojjs, line 4: Make sure not to check in this

PR #3478, opened 2023-02-28
[NOID] fixes #3477: apoc.load.html does not always report href

### LINKED ISSUE #3477: apoc.load.html does not always report href
Original topic from:

https://community.neo4j.com/t/apoc-load-html-cannot-read-relative-href-value/61151

conker84:
Fixes #3477

If the baseUrl is not defined href is not returned.

  - Fixed the bug
  - Added test


jexp (approved):
Looks good to me, perhaps we can replace the external web request in the test with a short html fragment that has the same structure?

conker84:
> Looks good to me, perhaps we can replace the external web request in the test with a short html fragment that has the same structure?

Sure lemme do that

PR #3484, opened 2023-03-10
[ju7btyJy] APOC UUID Install Failing in Cluster

vga91:
About the triggers, we have chosen to change from `apoc.trigger.add` and `apoc.trigger.remove` respectively
to `apoc.trigger.install`  `apoc.trigger.drop`.

About the uuids, with the remove procedure we kept the same convention, 
while with the add procedure we can't be consistent since the deprecated one was already `apoc.uuid.install`.
So we have choosen to name the new procedure `apoc.uuid.create`.

---
## Changes

- Added refresh mechanism, only for new procedures, (via `apoc.uuid.refresh` config and `ApocUuidMeta` label) like triggers and custom procedures,
to make new procedures work like triggers, i.e.:
   - Via UUIDHandlerNewProcedures: writes to the database
   - Via UUIDHandler: the transaction listener is managed

- Created util methods to be used for other procedures that leverage the system db
- Removed `batchComputationResult` from the new `apoc.create.uuid` procedure because we can't write directly from the procedure to the user database (like `neo4j`), but have to go through the `UUIDHandler` (via `refreshAndAdd()`)
- Added various tests
- Also added some tests with Fabric (i.e. using `USE system ...`)
- Deprecated old procedures
- Changed `TestcontainersCausalCluster` by removing the `("causal_clustering.leadership_balancing", "NO_BALANCING")` to replicate the error with `Use system ...` using deprecated procedures (i.e. system-leader and neo4j-leader in different members)
- created some common methods to be reused also with `apoc.custom` and `apoc.dv` (which use the system database)
- Updated documentation


jexp:
I don't think `apoc.uuid.create` is a good name as it conflicts with the uuid create function conceptually.

How about `apoc.uuid.setup` ? or `.add` ?

docs/asciidoc/modules/ROOT/examples/generated-documentation/apoc.uuid.create.adoc
```diff
@@ -0,0 +1,7 @@
+¦xref::overview/apoc.uuid/apoc.uuid.create.adoc[apoc.uuid.create icon:book[]] +
+
+`CALL apoc.uuid.install(databaseName, label, $config) | eventually adds the uuid transaction handler for the provided label and uuidProperty.
```
jexp, line 3: description is wrong

docs/asciidoc/modules/ROOT/pages/graph-updates/uuid.adoc
```diff
@@ -100,46 +109,80 @@ Add the `uuid`:
 
 [source,cypher]
 ----
-CALL apoc.uuid.install('Person')
+CALL apoc.uuid.create('neo4j', 'Person')
```
jexp, line 112: should we default to the "neo4j" database?
we should probably make it the last parameter then?

docs/asciidoc/modules/ROOT/pages/graph-updates/uuid.adoc
```diff
@@ -100,46 +109,80 @@ Add the `uuid`:
...
+To get the list of all of currently installed UUIDs,
+use the xref::overview/apoc.uuid/apoc.uuid.list.adoc[apoc.uuid.list] against the session database ("neo4j" in the above case).
+====
+
+
+==== Remove Automatic UUIDs
 
 [source,cypher]
 ----
-CALL apoc.uuid.remove('Person')
```
jexp, line 142: why didn't we just add the database as the last parameter to remove and default it to neo4j?

docs/asciidoc/modules/ROOT/pages/graph-updates/uuid.adoc
```diff
@@ -157,7 +200,7 @@ You can also remove all the uuid installed call the procedure as:
 
 [source,cypher]
 ----
-CALL apoc.uuid.removeAll()
+CALL apoc.uuid.dropAll('neo4j')
```
jexp, line 203: same as above

docs/asciidoc/modules/ROOT/pages/overview/apoc.uuid/apoc.uuid.create.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+[.emphasis]
+CALL apoc.uuid.install(databaseName, label, $config) | eventually adds the uuid transaction handler for the provided `label` and `uuidProperty`.
+In case the UUID handler is already present it will be replaced by the new one
+
+== Signature
+
+[source]
+----
+apoc.uuid.create(databaseName :: STRING?, label :: STRING?, config = {} :: MAP?) :: (batchComputationResult :: MAP?, label :: STRING?, installed :: BOOLEAN?, properties :: MAP?)
```
jexp, line 18: couldn't we just have added database as the last parameter with a default value?
or alternatively added a `database` entry to the config?

docs/asciidoc/modules/ROOT/partials/generated-documentation/documentation.adoc
```diff
@@ -3552,6 +3552,22 @@ apoc.util.validatePredicate(predicate, message, params) \| if the predicate yiel
 [.procedures, opts=header, cols='5a,1a,1a']
 |===
 | Qualified Name | Type | Release
+|xref::overview/apoc.uuid/apoc.uuid.create.adoc[apoc.uuid.create icon:book[]]
+
+CALL apoc.uuid.install(databaseName, label, $config) \| eventually adds the uuid transaction handler for the provided `label` and `uuidProperty`.
```
jexp, line 3557: wrong description

full/src/main/java/apoc/uuid/UUIDHandlerNewProcedures.java
```diff
@@ -0,0 +1,118 @@
...
+import static apoc.SystemLabels.*;
+import static apoc.util.SystemDbUtil.getSystemNodes;
+import static apoc.util.SystemDbUtil.setLastUpdate;
+import static apoc.util.SystemDbUtil.withSystemDb;
+import static apoc.uuid.UuidHandler.NOT_ENABLED_ERROR;
+
+public class UUIDHandlerNewProcedures {
+    public static boolean isEnabled(String databaseName) {
+        String apocUUIDEnabledDb = String.format(ApocConfig.APOC_UUID_ENABLED_DB, databaseName);
+        return apocConfig().getConfig().getBoolean(apocUUIDEnabledDb, apocConfig().getBoolean(APOC_UUID_ENABLED));
```
jexp, line 32: what happens if apoc.uuid.enabled is not set, will it fail to start the db? like we had recently?

full/src/main/java/apoc/uuid/UUIDHandlerNewProcedures.java
```diff
@@ -0,0 +1,118 @@
...
+        return result[0];
+    }
+
+    public static UuidInfo drop(String databaseName, String labelName) {
+        final UuidInfo[] previous = new UuidInfo[1];
+
+        withSystemDb(tx -> {
+            getUuidNodes(tx, databaseName, Map.of(SystemPropertyKeys.label.name(), labelName))
+                    .forEachRemaining(node -> {
+                        previous[0] = new UuidInfo(node);
```
jexp, line 70: this is not nice code, could we instead return them and then get the first or last from the stream?

full/src/main/java/apoc/uuid/UUIDHandlerNewProcedures.java
```diff
@@ -0,0 +1,118 @@
...
+
+        return previous[0];
+    }
+
+    public static List<UuidInfo> dropAll(String databaseName) {
+        final List<UuidInfo> previous = new ArrayList<>();
+
+        withSystemDb(tx -> {
+            getUuidNodes(tx, databaseName)
+                    .forEachRemaining(node -> {
```
jexp, line 85: why not map()?

full/src/main/java/apoc/uuid/UUIDNewProcedures.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Comparator;
+import java.util.Map;
+import java.util.stream.Stream;
+
+import static apoc.ApocConfig.apocConfig;
+import static apoc.uuid.UuidHandler.APOC_UUID_REFRESH;
+
+@Extended
+public class UUIDNewProcedures {
+    public static final String UUID_NOT_SET = APOC_UUID_REFRESH + " is not set. Please please set it in your apoc.conf";
```
jexp, line 19: shouldn't it default to 60s ?

full/src/main/java/apoc/uuid/UUIDNewProcedures.java
```diff
@@ -0,0 +1,117 @@
...
+        if (!apocConfig().containsKey(APOC_UUID_REFRESH)) {
+            throw new RuntimeException(UUID_NOT_SET);
+        }
+    }
+
+    // TODO - change with @SystemOnlyProcedure
+    @SystemProcedure
+    @Admin
+    @Procedure(mode = Mode.WRITE)
+    @Description("CALL apoc.uuid.install(databaseName, label, $config) | eventually adds the uuid transaction handler for the provided `label` and `uuidProperty`, in case the UUID handler is already present it will be replaced by the new one")
```
jexp, line 53: description wrong

full/src/main/java/apoc/uuid/Uuid.java
```diff
@@ -12,8 +12,13 @@
 import java.util.Map;
 import java.util.stream.Stream;
 
+import static apoc.util.SystemDbUtil.checkWriteAllowed;
+
 @Extended
 public class Uuid {
+    private static final String MSG_DEPRECATION = "Please note that the current procedure is deprecated, \n" +
+            "it's recommended to use the `apoc.uuid.create`, `apoc.uuid.drop`, `apoc.uuid.dropAll` procedures \n" +
+            "instead of, respectively, `apoc.uuid.install`, `apoc.uuid.remove`, `apoc.uuid.removeAll`.";
```
jexp, line 21: add "executed against the 'system' database

full/src/main/java/apoc/uuid/Uuid.java
```diff
@@ -111,7 +103,18 @@ public static class UuidInstallInfo extends UuidInfo {
 
     }
 
-    private String getUuidFunctionName() {
+    public static Map<String, Object> setExistingNodes(GraphDatabaseService db, Pools pools, String label, UuidConfig uuidConfig) {
+        final String uuidFunctionName = getUuidFunctionName();
+
+        return Util.inTx(db, pools, txInThread ->
+                txInThread.execute("CALL apoc.periodic.iterate(" +
+                                "\"MATCH (n:" + Util.sanitizeAndQuote(label) + ") RETURN n\",\n" +
```
jexp, line 111: change this to return id(n) 
also do `WHERE n.uuidProperty IS NULL` 
and lookup the id in the update statement
otherwise it will leak transactions

jexp:
Did we also add support for exporting from the system database with the new commands here?

full/src/test/java/apoc/uuid/UUIDClusterRoutingTest.java
```diff
@@ -0,0 +1,231 @@
...
+    public void testUuidDropAllowedOnlyInSysLeaderMember() {
+        final String query = "CALL apoc.uuid.drop('neo4j', $label)";
+        uuidInSysLeaderMemberCommon(query, PROCEDURE_NOT_ROUTED_ERROR, SYSTEM_DATABASE_NAME,
+                (session, label) -> testCallEmpty(session, query, Map.of("label", label))
+        );
+    }
+
+    @Test
+    public void testUuidShowAllowedOnlyInSysLeaderMember() {
+        final String query = "CALL apoc.uuid.show('neo4j')";
```
jexp, line 179: but show is read-only so it should be possible to run on all members

full/src/test/java/apoc/uuid/UUIDMultiDbTest.java
```diff
@@ -121,4 +125,46 @@ public void testWithDefaultDatabaseWithUUIDEnabled() throws InterruptedException
...
     }
 
+    //
+    // New procedures tests
+    //
+
+    @Test(expected = RuntimeException.class)
+    public void createUUIDWithSpecificDbWithUUIDDisabled() {
+        driver.session(SessionConfig.forDatabase(dbTest))
+                .run("CREATE CONSTRAINT ON (foo:Baz) ASSERT foo.uuid IS UNIQUE");
```
jexp, line 135: shouldn't this be REQUIRE?

full/src/test/java/apoc/uuid/UUIDMultiDbTest.java
```diff
@@ -121,4 +125,46 @@ public void testWithDefaultDatabaseWithUUIDEnabled() throws InterruptedException
...
+                    "Caused by: java.lang.RuntimeException: " + String.format(NOT_ENABLED_ERROR, dbTest);
+            assertEquals(expectedMessage, e.getMessage());
+            throw e;
+        }
+    }
+
+    @Test
+    public void createUUIDWithDefaultDbWithUUIDEnabled() {
+        driver.session()
+                .run("CREATE CONSTRAINT ON (baz:Baz) ASSERT baz.uuid IS UNIQUE");
```
jexp, line 152: shouldn't this be REQUIRE and the new syntax?

full/src/test/java/apoc/uuid/UUIDNewProceduresTest.java
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    //
+    // test cases taken and adapted from UUIDTest.java
+    //
+
+    @Test
+    public void testUUID() {
+        // given
+        db.executeTransactionally("CREATE CONSTRAINT ON (p:Person) ASSERT p.uuid IS UNIQUE");
```
jexp, line 91: new syntax and REQUIRE?

jexp: also in the other test methods (and IF NOT EXISTS)

full/src/test/java/apoc/uuid/UUIDNewProceduresTest.java
```diff
@@ -0,0 +1,494 @@
...
+    @Test(expected = RuntimeException.class)
+    public void testAddWithError() {
+        try {
+            // when
+            sysDb.executeTransactionally("CALL apoc.uuid.create('neo4j', 'Wrong') YIELD label RETURN label");
+        } catch (RuntimeException e) {
+            // then
+            Throwable except = ExceptionUtils.getRootCause(e);
+            assertTrue(except instanceof RuntimeException);
+            assertEquals("No constraint found for label: Wrong, please add the constraint with the following : `CREATE CONSTRAINT ON (wrong:Wrong) ASSERT wrong.uuid IS UNIQUE`", except.getMessage());
```
jexp, line 339: should also have the error message with the new syntax

full/src/main/java/apoc/uuid/UUIDHandlerNewProcedures.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public static void checkConstraintUuid(Transaction tx, String label, String propertyName) {
+        Schema schema = tx.schema();
+        Stream<ConstraintDefinition> constraintDefinitionStream = StreamSupport.stream(schema.getConstraints(Label.label(label)).spliterator(), false);
+        boolean exists = constraintDefinitionStream.anyMatch(constraint -> {
+            Stream<String> streamPropertyKeys = StreamSupport.stream(constraint.getPropertyKeys().spliterator(), false);
+            return streamPropertyKeys.anyMatch(property -> property.equals(propertyName));
+        });
+        if (!exists) {
+            String error = String.format("`CREATE CONSTRAINT ON (%s:%s) ASSERT %s.%s IS UNIQUE`",
```
jexp, line 113: new syntax perhaps even with a constraint name and IF NOT EXISTS

full/src/test/java/apoc/uuid/UUIDNewProceduresTest.java
```diff
@@ -0,0 +1,494 @@
...
+        } catch (RuntimeException e) {
+            // then
+            Throwable except = ExceptionUtils.getRootCause(e);
+            assertTrue(except instanceof RuntimeException);
+            assertEquals("No constraint found for label: Wrong, please add the constraint with the following : `CREATE CONSTRAINT ON (wrong:Wrong) ASSERT wrong.foo IS UNIQUE`", except.getMessage());
+            throw e;
+        }
+    }
+
+    //
```
jexp, line 358: should we also have a test that has UUID installed for two labels, and then we create a node with one label and add the other? so it should not override the previous uuid and handle it gracefully without failing.

for that situation we should probably also make sure to only remove the uuid from the node if has no labels that have  a uuid handler installed?

jexp: Or we need to disallow it altogether, which would be probably hard to enforce.

full/src/test/java/apoc/uuid/UUIDNewProceduresTest.java
```diff
@@ -0,0 +1,494 @@
...
+                    r -> fail("Should fail because of database not found"));
+        } catch (QueryExecutionException e) {
+            assertThat(e.getMessage(), Matchers.containsString(DatabaseNotFoundException.class.getName()));
+        }
+    }
+
+    @Test
+    public void testShowUuidInUserDb() {
+        try {
+            testCall(db, "CALL apoc.uuid.show('neo4j')",
```
jexp, line 419: show should be able to run on all databases

jexp (changes requested):
Looks good, I left a few comments.

The most important one would be to auto-create the constraint if possible. Otherwise it's really bad UX.

And fixing the constraint syntax throughout

docs/asciidoc/modules/ROOT/pages/graph-updates/uuid.adoc
vga91: Changed

full/src/main/java/apoc/uuid/UUIDHandlerNewProcedures.java
vga91: Changed

full/src/main/java/apoc/uuid/UUIDNewProcedures.java
vga91: By default I put `null` not to cause breaking-change with deprecated ones,
since they don't have a refresh handling: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3484/files#diff-24843e788bdf04d9e84495327581832cf31ff985f006cadbe92197ee9a706c12R78.

full/src/test/java/apoc/uuid/UUIDNewProceduresTest.java
vga91: Changed (and removed `@SystemProcedure`)

full/src/main/java/apoc/uuid/UUIDHandlerNewProcedures.java
vga91: Added a test to check it (`testConstraintAutoCreation`)

vga91:
> Did we also add support for exporting from the system database with the new commands here?

Currently there is still the old syntax.

In addition to changing the syntax,
the problem is that the `apoc.systemdb.export.metadata` is intended to be imported with the apoc.cypher.runFiles(...).
I don't think it's possible to run it directly on the system database, 
but probably we could e.g. export a metadata file by prepending  an `USE system`?
For example:
```
USE SYSTEM CALL apoc.uuid.install('One') ... ;
USE SYSTEM CALL apoc.uuid.install('Two') ... ;
```

full/src/main/java/apoc/uuid/Uuid.java
vga91: Changed

full/src/main/java/apoc/uuid/UUIDNewProcedures.java
```diff
@@ -0,0 +1,117 @@
...
+    // TODO - change with @SystemOnlyProcedure
+    @SystemProcedure
+    @Admin
+    @Procedure(mode = Mode.WRITE)
+    @Description("CALL apoc.uuid.drop(databaseName, label) yield label, installed, properties | eventually removes previously added UUID handler and returns uuid information")
+    public Stream<UuidInfo> drop(@Name("databaseName") String databaseName, @Name("label") String label) {
+        checkInSystemLeader(databaseName);
+
+        final UuidInfo uuidInfo = UUIDHandlerNewProcedures.drop(databaseName, label);
+        return Stream.ofNullable(uuidInfo);
```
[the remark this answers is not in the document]
vga91: Yes, it returns an empty stream if the the uuid doesn't not exists, because the `uuidInfo` returns `null` then.

vga91:
> Looks good, I left a few comments.
> 
> The most important one would be to auto-create the constraint if possible. Otherwise it's really bad UX.
> 
> And fixing the constraint syntax throughout

I added the auto-creation handling, and updated the old syntaxes

PR #3494, opened 2023-03-14
Fix loadJsonStream compiler error in dev

vga91:


vga91:
Merged although the build is red, since the build error will be solved with this pr: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3493

PR #3495, opened 2023-03-17
[0L9VOhhh] Bumps apoc version after release

ncordon:


Lojjs (changes requested):
I compared with https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3471/files and it looks like Nadja changed the apoc version in more files

Lojjs (approved):
Looks fine. I guess we will have to bump the Neo4j version in a separate PR when the CI starts failing because the docker version changed?

ncordon:
> Looks fine. I guess we will have to bump the Neo4j version in a separate PR when the CI starts failing because the docker version changed?

I don't think the CI would fail in this case because these are released versions of neo4j, so they are fetched from Maven. As part of the release card the first step is to bump the neo4j version if there's a new one released publicly

PR #3520, opened 2023-03-30
[Fb3ZFQbn] apoc.initializer.system can not find reader role

vga91:
Added wait mechanism which checks that the system db is populated,
by checking the sys-labels existence.


nadja-muller:
This sounds very error prone and I think there should be a better way to do this. I'll try to do some research.

ncordon (changes requested):
Kernel team have suggested using `db.isAvailable(100)` similar to this other PR instead: https://github.com/neo4j/apoc/pull/358

ncordon:
I'm closing this PR, as it's superseded by https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3533

PR #3523, opened 2023-04-05
Add Support for Blocking Label for apoc.path.* functions.

tomerkonf:
A blocking Label means that the graph search stops when reaching a node with the specified label, but include the node in the result graph.

### DIFF
--- a/core/src/main/java/apoc/path/LabelMatcherGroup.java
+++ b/core/src/main/java/apoc/path/LabelMatcherGroup.java
@@ -24,6 +24,7 @@ public class LabelMatcherGroup {
     private LabelMatcher blacklistMatcher = new LabelMatcher();
     private LabelMatcher endNodeMatcher = new LabelMatcher();
     private LabelMatcher terminatorNodeMatcher = new LabelMatcher();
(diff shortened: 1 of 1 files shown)


tomerkonf:
Hi Gem/Lojjs/anyone
Where is the place to document this new (small) feature?

thanks,
tomer.

gem-neo4j:
Hi there, thank you for creating a PR. I am curious to know the use case here, this option already happens so long as the depth is above the minLevel (which by default is -1) from my understanding:

```
        if (terminatorNodeMatcher.matchesLabels(nodeLabels)) {
            return belowMinLevel ? EXCLUDE_AND_CONTINUE : INCLUDE_AND_PRUNE;
        }

```

Where:

` boolean belowMinLevel = depth < minLevel;`

gem-neo4j:
This is also the status of v5 for APOC Core

tomerkonf:
i'm sorry to hear that. ok, thanks.
tomer.

gem-neo4j:
I'll close this for now, and it can always be reopened if we start adding features again :)

PR #3524, opened 2023-04-06
[xEcl2dHs] Fix some flaky tests in UUIDClusterRoutingTest

vga91:
To solve the card with id `xEcl2dHs`:
- moved the `CREATE CONSTRAINT ..` only where needed (i.e. `apoc.uuid.install`) and added an assertEventually to check that the constraint is really propagated to the current core member

----
Attempts to solve the other flaky tests:
- changed number of cores to 3,  to decrease the failure probablity
- in `TestContainerUtil.checkLeadershipBalanced()` added a try-catch to prevent from potential errors (e.g. `NullPointer`)
- in `TestcontainersCausalCluster.java` added `causal_clustering.minimum_core_cluster_size_at_formation` and `causal_clustering.minimum_core_cluster_size_at_runtime` configs

### DIFF
--- a/full/src/test/java/apoc/uuid/UUIDClusterRoutingTest.java
+++ b/full/src/test/java/apoc/uuid/UUIDClusterRoutingTest.java
@@ -34,7 +34,7 @@ import static org.neo4j.configuration.GraphDatabaseSettings.SYSTEM_DATABASE_NAME
 import static org.neo4j.test.assertion.Assert.assertEventually;
 
 public class UUIDClusterRoutingTest {
-    private static final int NUM_CORES = 4;
+    private static final int NUM_CORES = 3;
     private static TestcontainersCausalCluster cluster;
     private static Session clusterSession;
     private static List<Neo4jContainerExtension> members;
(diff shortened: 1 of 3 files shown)


full/src/test/java/apoc/uuid/UUIDClusterRoutingTest.java
```diff
@@ -134,6 +134,14 @@ public void testUuidInstallAllowedOnlyInSysLeaderMember() {
         final String query = "CALL apoc.uuid.install($label)";
         uuidInSysLeaderMemberCommon(query, SYS_NON_LEADER_ERROR, DEFAULT_DATABASE_NAME,
             (session, label) -> {
+                clusterSession.writeTransaction(tx -> tx.run(format("CREATE CONSTRAINT `%1$s` IF NOT EXISTS FOR (n:`%1$s`) REQUIRE n.uuid IS UNIQUE", label)));
```
ncordon, line 137: Why do we have the trailing `$s` in the constraint label now?

vga91: I replaced `%s` with `%1$s` because I also use the latter as the index name into the `String.format`, to verify just below that it propagated correctly

ncordon: Yeah but aren't you creating the wrong contraint? You are also using it in the label of the `IF NOT EXISTS FOR`

vga91: If I understood correctly, I don't think it's a problem if the index name is equal to the label one. It's just to simplify it.
Anyway, the `apoc.uuid.install` shouldn't work if there isn't a correct uuid constraint.

ncordon (approved):

PR #3527, opened 2023-04-06
[NOID] Fix flaky geocode server errors

vga91:
Added `waitForServerResponseOK` to prevent sporadic OpenStreetMap server errors.

See: https://github.com/neo4j-contrib/neo4j-apoc-procedures/actions/runs/4629830341/jobs/8190593821#step:7:1189

### DIFF
--- a/core/src/test/java/apoc/spatial/GeocodeTest.java
+++ b/core/src/test/java/apoc/spatial/GeocodeTest.java
@@ -6,6 +6,7 @@ import inet.ipaddr.IPAddressString;
 import org.junit.*;
 import org.neo4j.graphdb.QueryExecutionException;
 import org.neo4j.configuration.GraphDatabaseInternalSettings;
+import org.neo4j.graphdb.ResultTransformer;
 import org.neo4j.test.rule.DbmsRule;
 import org.neo4j.test.rule.ImpermanentDbmsRule;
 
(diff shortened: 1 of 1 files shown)


core/src/test/java/apoc/spatial/GeocodeTest.java
```diff
@@ -287,6 +297,23 @@ private void testGeocodeAddress(String address, double lat, double lon, Map<Stri
...
+                        params,
+                        objectResultTransformer);
+                return true;
+            } catch (Exception e) {
+                if (e.getMessage().contains("Server returned HTTP response code")) {
+                    return false;
+                }
+                throw e;
+            }
+        }, (value) -> value, 20L, TimeUnit.SECONDS);
```
gem-neo4j, line 317: Just want to make sure I understood this correctly, this will retry to call the procedure if it gets a 502 etc until it gets a 200 for max 20 seconds? otherwise it will throw an error? :)

gem-neo4j: I took a look at the failing tests, and both are to do with geocode, one is failing due to a 502 ;) might wanna check those out as well :)

gem-neo4j (approved):
LGTM :)

PR #3545, opened 2023-04-21
[6R2yEhGU] No procedure with name ... registered for this database instance

vga91:
Deleted duplicated `TestContainerUtil.java` (available in core)


vga91:
It seems that, unlike 4.4, there is some caching problem, or something else, with clusters.

If for example I create a custom procedure, 
in the read-only core(s), the procedure will be detected sometimes after a few ms, other times after more than a minute, either if I execute `call db.clearQueryCaches()`, or if I don't do any operation.

vga91:
Will be replaced by this pr: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3800 for 5.x

PR #3546, opened 2023-04-21
[Fb3ZFQbn] Fixes cypher initializer

ncordon:
## What
In `CypherInitializer` await 1 min at max for the system database to show status `CURRENT` or  `REQUIRES_UPGRADE`. We do not do anything for the other possible status the dbms would not even be able to operate normally in (i.e. `UNSUPPORTED_BUT_CAN_UPGRADE`, `UNSUPPORTED`, `UNSUPPORTED_FUTURE`).

## Why
The reason for that is that the first time we it the built-in roles (and possibly other system structures) will not be populated immediately in 4.4. I.e. system.isAvailable can output true but some of Lifecycles may not have been run yet (`UNINITIALIZED` status)


test-utils/src/main/java/apoc/util/TestContainerUtil.java
```diff
@@ -272,7 +272,7 @@ public static void checkLeadershipBalanced(Session session) {
                         return false;
                     }
                 },
-                (value) -> value, 30L, TimeUnit.SECONDS);
+                (value) -> value, 60L, TimeUnit.SECONDS);
```
ncordon, line 275: This was needed because there seem to be a few flaky tests around `UUIDClusterRoutingTest`

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -49,6 +53,47 @@ public GraphDatabaseAPI getDb() {
...
+           structures) will not be populated immediately in 4.4.
+           I.e. system.isAvailable can output true but some of the
+           Lifecycles may not have been run yet (UNINITIALIZED status)
+         */
+        boolean condition = awaitUntil( () ->
+                                        {
+                                            var components = dependencyResolver.resolveDependency( SystemGraphComponents.class );
+                                            var status = components.detect( db );
+                                            return (status == Status.CURRENT || status == Status.REQUIRES_UPGRADE);
+                                        }, 120000 );
```
nadja-muller, line 89: I guess the fail-save solution for this could be to parametrize this value, so users can extend the time frame if needed. What do you think?

nadja-muller (changes requested):

core/src/main/java/apoc/cypher/CypherInitializer.java
ncordon: 2 minutes should be more than enough for an instance to have run all of the lifecycles. I feel like introducing a new parameter for this is an overkill

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -49,6 +53,46 @@ public GraphDatabaseAPI getDb() {
...
+
+        if ( !condition )
+        {
+            userLog.warn("attempting to execute apoc.initializer without waiting for systemdb to finish initialization, waiting time elapsed. This may result in failing apoc.initializer commands." );        }
```
nadja-muller, line 93:
```suggestion
            userLog.warn("attempting to execute apoc.initializer without waiting for the system database to finish initialization, waiting time elapsed. This may result in failing apoc.initializer commands." );        }
```

nadja-muller (approved):
Nit: I would rewrite the systemdb part as it's not an official term.
Looks good otherwise.

PR #3562, opened 2023-05-05
[NOID] Fix flaky testQueryH2WithConfig

vga91:
WIP


full/src/test/java/apoc/load/LoadHtmlTestParameterized.java
```diff
@@ -101,13 +103,13 @@ public void testQueryH2WithConfig() {
...
-            testResult(db, "CALL apoc.load.html($url, $query, $config)",
+            assertEventually(() -> db.executeTransactionally("CALL apoc.load.html($url, $query, $config)",
                     map("url",new File("src/test/resources/wikipedia.html").toURI().toString(), "query", query, "config", config),
                     result -> {
                         Map<String, Object> row = result.next();
                         assertEquals(map("h2",asList(RESULT_QUERY_H2)).toString().trim(), row.get("value").toString().trim());
-                        assertFalse(result.hasNext());
-                    });
+                        return !result.hasNext();
+                    }, Duration.ofSeconds(3)), (v) -> v, 30, TimeUnit.SECONDS);
```
ncordon, line 112: This retries on all errors but maybe we just want to retry on TimeoutExceptions?

vga91: Yes, makes sense. Just changed, and now the query is re-executed only after a transaction timeout.

Also, even though the error seems present only in `testQueryH2WithConfig`, it could potentially occur for other selenium tests as well. So I added retry in all these tests.

ncordon (changes requested):
The CI keeps failing on these tests, please take a look

full/src/test/java/apoc/load/LoadHtmlTest.java
```diff
@@ -546,7 +546,8 @@ public static void skipIfBrowserNotPresentOrCompatible(GraphDatabaseService db,
                         resultConsumer.accept(result);
                         return true;
                     }, Duration.ofSeconds(5));
-                } catch (TransactionTerminatedException e) {
+                } catch (Exception e) {
+                    System.out.println("e = " + e);
                     // those query should generally take less than a second.
```
ncordon, line 550: I don't agree with this change. This is ignoring all exceptions here

vga91: Yes, the catch capture any exception, 
however further down there is this construct, so it fails correctly in case of other error messages:
```
                    boolean txTimeout = ExceptionUtils.getRootCause(e).getMessage()
                            .contains("The transaction has not completed within the specified timeout");
                    if (txTimeout) {
                        return false;
                    }
                    throw e;
```

PR #3572, opened 2023-05-11
[UNKBKte1] Redacts passwords in CypherInitializer

ncordon:
## What
We should obfuscate passwords when logging information in CypherInitializer

## Why

Because if we used:

```
apoc.initializer.system.1=CREATE USER dummy IF NOT EXISTS SET PASSWORD pass12345
```

we were getting an entry in neo4j.log with the password in plaintext:

```
INFO  [system/00000000] successfully initialized: alter user neo4j SET PASSWORD "pass12345" CHANGE NOT REQUIRED
```

## Expected behaviour

We should get a redacted password instead:

```
INFO  [system/00000000] successfully initialized: alter user neo4j SET PASSWORD "*****" CHANGE NOT REQUIRED
```

### DIFF
--- a/core/src/main/java/apoc/cypher/CypherInitializer.java
+++ b/core/src/main/java/apoc/cypher/CypherInitializer.java
@@ -20,6 +20,8 @@ package apoc.cypher;
 
 import apoc.ApocConfig;
 import apoc.SystemLabels;
+import apoc.util.LogsUtil;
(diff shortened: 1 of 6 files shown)


pontusmelke (approved):
👍

nadja-muller (approved):
Looks good 👍

PR #3573, opened 2023-05-12
[NrNk9YvB] Adds projects to generate AWS and GCS dependencies

ncordon:


full/src/test/java/apoc/load/LoadGoogleCloudStorageTest.java
```diff
@@ -33,37 +35,30 @@
...
-            assertFalse("It should be the last record", r.hasNext());
-        });
+    @AfterClass
+    public static void tearDown() {
+        gcs.close();
+        db.shutdown();
     }
 
     @Test
-    public void testLoadCsvFromFolderWithoutAuth() throws Exception {
```
nadja-muller, line 64: Are the new tests with auth?

test-utils/src/main/java/apoc/util/GoogleCloudStorageContainerExtension.java
```diff
@@ -0,0 +1,45 @@
...
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
+
+import static java.net.HttpURLConnection.HTTP_OK;
+
+public class GoogleCloudStorageContainerExtension extends GenericContainer<GoogleCloudStorageContainerExtension> {
+
+    public GoogleCloudStorageContainerExtension() {
+        super("fsouza/fake-gcs-server:latest");
+        this.withCommand("-scheme http");
```
nadja-muller, line 31: Why did you decide to use the http scheme? Do we still have a test for the gc scheme?

nadja-muller (commented):
Good work. I left some questions for now.

test-utils/src/main/java/apoc/util/GoogleCloudStorageContainerExtension.java
ncordon: This is cherry picked from dev: https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/dev/test-utils/src/main/java/apoc/util/GoogleCloudStorageContainerExtension.java. I don't know, in any case it's not something I should be doing in this pr I think

full/src/test/java/apoc/load/LoadGoogleCloudStorageTest.java
ncordon: I don't think so. Again this is a cherry-pick of https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/dev/extended/src/test/java/apoc/load/LoadGoogleCloudStorageTest.java.

The underlying container https://github.com/fsouza/fake-gcs-server#running-with-http doesn't have any option to be passing authentication and test that part of the functionality unfortunately.

nadja-muller: Please add the cherry picked PR in the PR description here, so we know part of it is cherry picked. Future us might thank us for it.

nadja-muller (approved):

PR #3575, opened 2023-05-15
Add OpenAI (LLM) procedures

### LINKED ISSUE #3574: Add OpenAI (LLM) procedures
Similar to the NLP procedures, we can provide a number of procedures that make working with LLM easier

Starting with OpenAI API, possibly extending it to other providers

- generate embeddings
- text & chat completions

Thanks to @tomasonjo for the initial PR

jexp:
Fixes #3574

Similar to the NLP procedures, we can provide a number of procedures that make working with LLM easier

Starting with OpenAI API, possibly extending it to other providers

- generate embeddings
- text & chat completions

Thanks to @tomasonjo for the initial PR

### DIFF
--- a/extended/src/main/java/apoc/openai/OpenAI.java
+++ b/extended/src/main/java/apoc/openai/OpenAI.java
@@ -6,13 +6,19 @@ import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Name;
 import org.neo4j.procedure.Procedure;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.stream.Stream;
 import java.util.stream.Collectors;
 import apoc.result.MapResult;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+
 
 @Extended
 public class OpenAI {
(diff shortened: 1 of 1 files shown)


jexp:
Want to change the procedure results

columsn for completion:

```created, model, id,completion_tokens, prompt_tokens total_tokens, object->type

not sure how to handle:  choices=[{finish_reason="stop", index=0, text="Blue", logprobs=null}]
perhaps we can turn that into one row per choice? and then add the columns for finish_reason, index, text, logprobs```


 (not sure what logprobs is)
 
 same for chat-completion

**Comment by conker84** [comment]
> created, model, id,completion_tokens, prompt_tokens total_tokens, object->type

I would leave it as is as currently we don't have a clear view about how the developer will use it; @tomasonjo wdyt?

**Comment by tomasonjo** [comment]
I would also leave it as we return full responses and allow users to do whatever they want.

I am guessing in 98%, the developer usage will be:

```
YIELD value
WITH value.choices[0].text AS response
do something
```

**Comment by jexp** [comment]
Ok then we can release it as is.

File: extended/src/main/java/apoc/ml/OpenAI.java [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+        {
+          "object": "embedding",
+          "embedding": [ 0.0023064255, -0.009327292, .... (1536 floats total for ada-002) -0.0028842222 ],
+          "index": 0
+        }
+      ],
+      "model": "text-embedding-ada-002",
+      "usage": { "prompt_tokens": 8, "total_tokens": 8 } }
+    */
+        Stream<Object> resultStream = executeRequest(apiKey, configuration, "embeddings", "text-embedding-ada-002", "input", texts, "$.data");
```
conker84, line 74: @jexp @tomasonjo does make sense to make this and the other models configurable, via `configuration`?

conker84 (commented):
@jexp @tomasonjo tank you so much, I just have one question

extended/src/main/java/apoc/ml/OpenAI.java
tomasonjo: not sure exactly if I understand what you mean, but we could add model parameter under the configuration map. I would leave the prompts as a separate parameter.

conker84: Yeah I mean something like this
```suggestion
        String model = configuration.getOrDefault("model", "text-embedding-ada-002")
        Stream<Object> resultStream = executeRequest(apiKey, configuration, "embeddings", model, "input", texts, "$.data");
```

jexp: Not sure we need to do this on every call, as it's already done inside of executeRequest.

conker84 (approved):

conker84:
@jexp @tomasonjo thank you!

PR #3604, opened 2023-05-31
Updated exteded.txt with vertexai procs

vga91:
Updated `exteded.txt`


vga91:
The failing CodeQL job will be fixed in another pr

PR #3608, opened 2023-06-01
DETACH DELETE in TTLLifeCycle.java

### LINKED ISSUE #3607: We should probably use DETACH DELETE in TTL
For the nodes deletion statement, just to be safe if any new relationships have sneaked in since the rel-deletion that they would be caught as well.

https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/dev/extended/src/main/java/apoc/ttl/TTLLifeCycle.java#LL67C90-L67C96

jexp:
Fixes #3607

For the nodes deletion statement, just to be safe if any new relationships have sneaked in since the rel-deletion that they would be caught as well.

### DIFF
--- a/extended/src/main/java/apoc/ttl/TTLLifeCycle.java
+++ b/extended/src/main/java/apoc/ttl/TTLLifeCycle.java
@@ -64,7 +64,7 @@ public class TTLLifeCycle extends LifecycleAdapter {
             );
 
             long nodesDeleted = db.executeTransactionally(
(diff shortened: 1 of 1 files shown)


vga91 (approved):
Lgtm!

vga91:
The build is green.
The CodeQL (java) failed job will be solved [here](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3605).

PR #3639, opened 2023-06-22
[NOID] Add reactive hydration on apoc.export.cypher.query

BennuFire:
Fixes OOM on big graphs generated by apoc.export.cypher.query

This approach will rely on multiple of executions (8 at the moment) of the query in order to lazily create the graph to be exported.

3 tests on flaky state to be solved (testExportQueryCypherPlainFormat, testExportCypherAdminOperationErrorMessage, testExportQueryCypherForNeo4j)

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -41,6 +41,17 @@ jobs:
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v2
 
+  license-checks:
+    runs-on: ubuntu-latest
+    needs: compile
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-jdk
+      - uses: ./.github/actions/setup-gradle-cache
+      - name: Check LICENSE and NOTICE files
+        run: ./gradlew validateLicenses generateLicensesFiles
+
   tests:
     strategy:
       fail-fast: false
--- a/LICENSES.txt
+++ b/LICENSES.txt
@@ -8,14 +8,13 @@ Apache-2.0
   HikariCP-java7-2.4.12.jar
   RoaringBitmap-0.7.17.jar
   WMI4Java-1.6.3.jar
-  accessors-smart-2.4.7.jar
   accessors-smart-2.4.9.jar
   annotations-17.0.0.jar
   apiguardian-api-1.1.0.jar
-  arrow-format-10.0.1.jar
-  arrow-memory-core-10.0.1.jar
-  arrow-memory-netty-10.0.1.jar
-  arrow-vector-10.0.1.jar
+  arrow-format-12.0.1.jar
+  arrow-memory-core-12.0.1.jar
+  arrow-memory-netty-12.0.1.jar
+  arrow-vector-12.0.1.jar
   assertj-core-3.18.1.jar
   audience-annotations-0.5.0.jar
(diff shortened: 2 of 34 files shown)


gem-neo4j:
Hi! Thanks for opening this :) I'm pretty happy in general with the PR, but am mainly curious about how you tested it fixed the OOM issue. Also what is your comment about the flakey state meaning? Are they flakey now?

BennuFire:
Hi @gem-neo4j !

Do you think it will be reasonable to create a different procedure so the nowadays apoc.cypher.export.query remains untouched and this implementation get routed with a different entry point?

I did this as a first draft, so there're still couple of things can be improved in order to reduce query executions (ergo CPU usage). e.g. Labels and Rels types in use can be hydrated on a private Collection in order to reduce query executions.

> but am mainly curious about how you tested it fixed the OOM issue

About ways to check this, I loaded a 4.4.21 instance with random data `call apoc.generate.ba(500000, 8, null, null)` and launching the DB with 800mb heap. Original version of the export get OOM while this version goes through. 

> hat is your comment about the flakey state meaning?

It's not so problematic considering some other tests are on failing state even without this PR. The 3 tests I named failed do to cypher commands being on a different order. We can standardize the output by sorting nodes and rels (the same way it's done on constrains and indexes) but it will just add more CPU usage.

BennuFire:
Hi @gem-neo4j ,

This last version it's not fully lazy but uses less ram than initial version. I'm running some tests in order to find a good compromise between query executions and allocated memory (between db query and the procedure itself).

This can be now tested by apoc team :)

core/src/main/java/org/neo4j/cypher/export/CypherDatabaseSubGraph.java
```diff
@@ -0,0 +1,226 @@
...
+                , false
+        ).collect( Collectors.toList());
+        this.labels = StreamSupport.stream( this.nodes.spliterator(), false )
+                .flatMap(node ->  StreamSupport.stream(node.getLabels().spliterator(), false))
+                .distinct()
+                .collect( Collectors.toList());
+        this.relationshipTypes = StreamSupport.stream( this.relationships.spliterator(), false )
+                .map(relationship ->  relationship.getType())
+                .distinct()
+                .collect( Collectors.toList());
```
conker84, line 81: @BennuFire is this the lazy implementation? generally, I see a lot of `Stream...` and `collect`, meaning that we directly materialize the results, so IMO, this implementation does not guarantee any laziness.

@gem-neo4j wdyt?

conker84 (commented):
@BennuFire thank you so much for the PR. Please look at my comment

core/src/main/java/org/neo4j/cypher/export/CypherDatabaseSubGraph.java
BennuFire: You are fully right. At the beginning I was aiming for a full lazy implementation but I in that case I had to change some parts of the MultiStatementCypherSubGraphExporter impl. At the moment it's not lazy at all but reactive (avoiding the use of the heap for the tx.result and graph object simultaneously) that's already a memory usage improvement. 

I may change the title in order to avoid misleading although it's an ultimate goal.

BennuFire:
New version https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3650

PR #3649, opened 2023-07-06
First LLM prompt for cypher, query and schema in APOC

jexp:
For review

Uses OpenAI with a potentially configured API (`apoc.openai.key`) in `apoc.conf` (e.g. for Sandbox)

Extracts the database schema using `apoc.meta.data`

- Uses a system prompt for base grounding
- Adds a schema prompt
- Provides the user question
- adds `apoc.ml.schema` for describing  the schema of the database
- adds `apoc.ml.cypher` for generating N cypher queries for the user question
- adds `apoc.ml.query` for generating and executing cypher queries (with generation-retries (3 by default) if the query failed)  
- Still needs docs and tests (a bit tricky due to the probabalistic nature)

apoc-llm-prompt

### DIFF
--- a/extended/src/main/java/apoc/ml/Prompt.java
+++ b/extended/src/main/java/apoc/ml/Prompt.java
@@ -40,7 +40,7 @@ public class Prompt {
             """;
 
     static final String SYSTEM_PROMPT = """
-                    You are an expert in the Neo4j graph query language Cypher.
+            You are an expert in the Neo4j graph query language Cypher.
             Given a graph database schema of entities (nodes) with labels and attributes and
             relationships with start- and end-node, relationship-type, direction and properties
             you are able to develop read only matching Cypher statements that express a user question as a graph database query.
--- a/extended/src/main/java/apoc/ml/Prompt.java
+++ b/extended/src/main/java/apoc/ml/Prompt.java
@@ -1,0 +1,194 @@
+package apoc.ml;
+
+import apoc.ApocConfig;
+import apoc.Extended;
+import apoc.result.StringResult;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import org.jetbrains.annotations.NotNull;
+import org.neo4j.graphdb.QueryExecutionException;
+import org.neo4j.graphdb.Result;
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.logging.Log;
+import org.neo4j.procedure.*;
+
+import java.net.MalformedURLException;
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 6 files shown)


extended/src/main/java/apoc/ml/Prompt.java
```diff
@@ -0,0 +1,192 @@
...
+
+    @NotNull
+    private String prompt(String userQuestion, String systemPrompt, String assistantPrompt, String schema, String apiKey) throws JsonProcessingException, MalformedURLException {
+        List<Map<String, String>> prompt = new ArrayList<>();
+        if (systemPrompt != null && !systemPrompt.isBlank()) prompt.add(Map.of("role", "system", "content", systemPrompt));
+        if (schema != null && !schema.isBlank()) prompt.add(Map.of("role", "system", "content", "The graph database schema consists of these elements\n" + schema));
+        if (userQuestion != null && !userQuestion.isBlank()) prompt.add(Map.of("role", "user", "content", userQuestion));
+        if (assistantPrompt != null && !assistantPrompt.isBlank()) prompt.add(Map.of("role", "assistant", "content", assistantPrompt));
+        String result = OpenAI.executeRequest(apiKey, Map.of(), "chat/completions",
+                        "gpt-3.5-turbo", "messages", prompt, "$", apocConfig)
```
conker84, line 146: the model is hardcoded, and as GPT-4 is now in GA maybe we can think about having it as a configuration parameter, as we did in `OpenAI` class

extended/src/main/java/apoc/ml/Prompt.java
```diff
@@ -0,0 +1,192 @@
...
+                // just let it fail so that retries can work if (queryResult.query.isBlank()) return Stream.empty();
+                /*
+                if (queryResult.hasError())
+                    throw new QueryExecutionException(queryResult.error, null, queryResult.type);
+                 */
+                query = queryResult.query;
+                Result result = tx.execute(query);
+                AtomicReference<String> ref = new AtomicReference<>(query);
+                return result.stream()
+                        .map(row -> new PromptMapResult(row, ref.getAndSet(null)))
```
conker84, line 98: maybe instead of doing this we can use the [ProcedureCallContext](https://github.com/neo4j/neo4j/blob/4.4/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/procs/ProcedureCallContext.java) and return the `query` every time but only if requested from the developer

conker84 (commented):
thank you so much @jexp just a couple of nit pick for discussion

conker84 (approved):

PR #3651, opened 2023-07-10
[NOID] Add reactive hydration on apoc.export.cypher.query

BennuFire:
Based on https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3639

Created from an internal branch now.

### DIFF
--- a/core/src/main/java/org/neo4j/cypher/export/CypherDatabaseSubGraph.java
+++ b/core/src/main/java/org/neo4j/cypher/export/CypherDatabaseSubGraph.java
@@ -1,0 +1,223 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.cypher.export;
+
+import org.neo4j.graphdb.*;
+import org.neo4j.graphdb.schema.ConstraintDefinition;
+import org.neo4j.graphdb.schema.IndexDefinition;
+import org.neo4j.internal.kernel.api.TokenRead;
+import org.neo4j.kernel.api.KernelTransaction;
+import org.neo4j.kernel.impl.coreapi.InternalTransaction;
+
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+import static apoc.export.cypher.formatter.CypherFormatterUtils.cypherNode;
+import static apoc.util.Util.quote;
+import static org.neo4j.internal.kernel.api.TokenRead.ANY_LABEL;
+
+public class CypherDatabaseSubGraph implements SubGraph
+{
+    private final Transaction transaction;
+
+    private final List<Label> labels;
+
+    private final List<RelationshipType> relationshipTypes;
+
+    private final Iterable<Node> nodes;
+
+    private final Iterable<Relationship> relationships;
+
+    public CypherDatabaseSubGraph(Transaction transaction, String query )
+    {
+        this.transaction = transaction;
+        this.nodes = StreamSupport.stream(
+                transaction.execute(query).stream().flatMap(
+                                row -> row.values().stream().filter(value -> value instanceof Node))
+                        .map(node -> (Node) node)
+                        .distinct()
+                        .spliterator()
+                , false
+        ).collect( Collectors.toList());
+        this.relationships = StreamSupport.stream(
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/neo4j/cypher/export/CypherDatabaseSubGraph.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+import static apoc.export.cypher.formatter.CypherFormatterUtils.cypherNode;
+import static apoc.util.Util.quote;
+import static org.neo4j.internal.kernel.api.TokenRead.ANY_LABEL;
+
+public class CypherDatabaseSubGraph implements SubGraph
```
conker84, line 40: this class, unfortunately, has some problems. Lemme explain, `CypherResultSubGraph` provides information only for the graph entities involved in the subgraph, unlikely this one. Please check the following comments

core/src/main/java/org/neo4j/cypher/export/CypherDatabaseSubGraph.java
```diff
@@ -0,0 +1,223 @@
...
+        return this.relationshipTypes;
+    }
+
+    @Override
+    public Iterable<Label> getAllLabelsInUse() {
+        return this.labels;
+    }
+
+    @Override
+    public long countsForRelationship(Label start, RelationshipType type, Label end) {
```
conker84, line 162: this method returns the true count even if one of the input elements is not in the provided Cypher query.

core/src/main/java/org/neo4j/cypher/export/CypherDatabaseSubGraph.java
```diff
@@ -0,0 +1,223 @@
...
+        String startNode = cypherNode(start);
+        String endNode = cypherNode(end);
+        String relationship = String.format("[r:%s]", quote(type.name()));
+        return transaction.execute(String.format("MATCH %s-%s->%s RETURN count(r) AS count", startNode, relationship, endNode))
+                .<Long>columnAs("count")
+                .next();
+    }
+
+    @Override
+    public long countsForNode(Label label) {
```
conker84, line 188: same as above, it returns the real count even if the label is not in the cypher query

core/src/main/java/org/neo4j/cypher/export/CypherDatabaseSubGraph.java
```diff
@@ -0,0 +1,223 @@
...
+                .next();
+    }
+
+    private boolean labelNotExists(Label label, int labelId) {
+        // `label == null` means that we want to count all the labels
+        // otherwise if we don't found a not-null label, we want the count to return to zero
+        return label != null && labelId == ANY_LABEL;
+    }
+
+    private Integer getLabelId(Label start, TokenRead tokenRead) {
```
conker84, line 212: same as above

core/src/main/java/org/neo4j/cypher/export/CypherDatabaseSubGraph.java
```diff
@@ -0,0 +1,223 @@
...
+
+    @Override
+    public Iterable<IndexDefinition> getIndexes(Label label) {
+        return transaction.schema().getIndexes(label);
+    }
+
+    @Override
+    public Iterable<IndexDefinition> getIndexes(RelationshipType type) {
+        return transaction.schema().getIndexes(type);
+    }
```
conker84, line 149: these methods return the real information even if the label or the relationship type is not in the subgraph.

conker84 (commented):
@BennuFire thank you so much please check my comments.
In the end, maybe is worth investigating what's the cause of the OOM. If you could provide any info about how to reproduce the problem, we're glad to support you in the resolution.

BennuFire:
> @BennuFire thank you so much please check my comments. In the end, maybe is worth investigating what's the cause of the OOM. If you could provide any info about how to reproduce the problem, we're glad to support you in the resolution.

Hi @conker84 thanks for your comment. If you wanna investigate on the OOM, I have created a random database of (around) 300MB with `call apoc.generate.ba(1000000, 8, null, null)`. 

Then I have evaluated the behavior of 

`CALL apoc.export.cypher.query(
    "//nono 
    MATCH ()-[r]->() RETURN *",
    null, {
        stream : true
    }
) YIELD cypherStatements
RETURN count(*)` 

Around 1GB of heap, the procedure stops OOM'ing but give no statements as a result, and starts progressively giving results (count increase) the more heap you give to the db. So it silently fails for unaware users.

conker84:
> > @BennuFire thank you so much please check my comments. In the end, maybe is worth investigating what's the cause of the OOM. If you could provide any info about how to reproduce the problem, we're glad to support you in the resolution.
> 
> Hi @conker84 thanks for your comment. If you wanna investigate on the OOM, I have created a random database of (around) 300MB with `call apoc.generate.ba(1000000, 8, null, null)`.
> 
> Then I have evaluated the behavior of
> 
> `CALL apoc.export.cypher.query( "//nono MATCH ()-[r]->() RETURN *", null, { stream : true } ) YIELD cypherStatements RETURN count(*)`
> 
> Around 1GB of heap, the procedure stops OOM'ing but give no statements as a result, and starts progressively giving results (count increase) the more heap you give to the db. So it silently fails on unaware users.

@BennuFire I tried with 

APOC 4.4.0.19, Neo4j 4.4.23 with default configuration

in particular these values:

```
dbms.memory.heap.initial_size=512m
dbms.memory.heap.max_size=1G
dbms.memory.pagecache.size=512m
```


with your same procedure calls and I wasn't able to reproduce the problem. 

Do you have any more info that you can share in order to reproduce it?

conker84:
I'm closing this for lack of feedback. Feel free to reopen to start again the discussion.
Thank you so much!

PR #3666, opened 2023-07-18
Solve no suitable method found for register

vga91:
See https://github.com/neo4j-contrib/neo4j-apoc-procedures/actions/runs/5585556876/jobs/10208445586?pr=3605#step:6:199


vga91:
The CodeQL failure will be fixed in another pr

PR #3689, opened 2023-07-25
[uXDd3S9j] Upgrade google-cloud-storage-version version to 2.29.1 to mitigate CVE-2023-2976

Lojjs:


docs/asciidoc/modules/ROOT/pages/import/web-apis.adoc
```diff
@@ -69,7 +69,7 @@ In order to use https://cloud.google.com/storage/[Google Cloud Storage], you nee
 * google-auth-library-oauth2-http-0.17.1.jar
 * google-cloud-core-1.90.0.jar
 * google-cloud-core-http-1.90.0.jar
-* google-cloud-storage-1.90.0.jar
+* google-cloud-storage-2.22.5.jar
```
Lojjs, line 72: I'm not sure if we want to update this one, given that the uber-jar described below is old and does not contain this version.

gem-neo4j: Hmm I would also assume the other dependencies are wrong as well, e.g google-cloud-core-http-1.90.0.jar most likely is also 2.22.5 if it is a dependency of cloud storage

core/build.gradle
```diff
@@ -135,8 +135,8 @@ dependencies {
     testCompile group: 'org.skyscreamer', name: 'jsonassert', version: '1.5.0'
     testCompile group: 'org.assertj', name: 'assertj-core', version: '3.13.2'
 
-    compileOnly group: 'com.google.cloud', name: 'google-cloud-storage', version: '2.22.3'
-    testCompile group: 'com.google.cloud', name: 'google-cloud-storage', version: '2.22.3', {
+    compileOnly group: 'com.google.cloud', name: 'google-cloud-storage', version: '2.22.5'
```
gem-neo4j, line 138: Why did we stop at 2.22.5? There are even newer ones released: https://mvnrepository.com/artifact/com.google.cloud/google-cloud-storage

docs/asciidoc/modules/ROOT/pages/import/web-apis.adoc
gem-neo4j: Yeah, looking at the compile dependencies of google-cloud-storage, these are all different, I would not add this change in this PR but make a card to fix this up along with the uber jar maybe? https://mvnrepository.com/artifact/com.google.cloud/google-cloud-storage/2.22.5

core/build.gradle
gem-neo4j: Ah okay, I normally try to go for newer so we have longer between when we need to fix it next :)

Lojjs: That makes sense, will change it to the newest

docs/asciidoc/modules/ROOT/pages/import/web-apis.adoc
gem-neo4j: Sounds good!

gem-neo4j (approved):
Looks good :D

gem-neo4j (approved):
Great :) Nice to have this in

PR #3697, opened 2023-07-26
[K6URm7vi] only create DatabaseEventListener for system database

nadja-muller:
Fixes performance issues for drop database


core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -142,10 +142,20 @@ public void available() {
                 final boolean isSystemDatabase = db.databaseName().equals(GraphDatabaseSettings.SYSTEM_DATABASE_NAME);
                 Configuration config = dependencyResolver.resolveDependency(ApocConfig.class).getConfig();
                 var initializers = collectInitializers(isSystemDatabase, config);
+
+                // Wait for database to become available, once it is available
+                // procedures are also available by necessity.
```
Lojjs, line 147: In 5.x, this comment was for `while (!db.isAvailable(100));` so I wonder if we really should add it here?

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -142,10 +142,20 @@ public void available() {
...
                     awaitApocProceduresRegistered();
-                } else if (!initializers.isEmpty()) {
-                    waitForSystemDbLifecycles();
+                } else {
+                    if (!initializers.isEmpty()) {
+                        waitForSystemDbLifecycles();
+                    }
+                }
+
+                // Register listener to drop apoc nodes from system databsae once a database is deleted
```
Lojjs, line 156:
```suggestion
                // Register listener to drop apoc nodes from system database once a database is deleted
```

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -142,10 +142,20 @@ public void available() {
...
+                if (isSystemDatabase) {
+                    databaseEventListeners.registerDatabaseEventListener(new SystemFunctionalityListener());
                 }
```
Lojjs, line 159: This can be moved into the else {...} above instead as that is following `if (!isSystemDatabase)`. That way we don't have to do the check again

Lojjs (changes requested):
Looks good in general, just some smaller things to fix up

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -142,10 +142,17 @@ public void available() {
...
                 if (!isSystemDatabase) {
                     awaitApocProceduresRegistered();
-                } else if (!initializers.isEmpty()) {
-                    waitForSystemDbLifecycles();
+                } else {
+                    if (!initializers.isEmpty()) {
+                        waitForSystemDbLifecycles();
+                        // Register listener to drop apoc nodes from system database once a database is deleted
+                        databaseEventListeners.registerDatabaseEventListener(new SystemFunctionalityListener());
+                    }
```
Lojjs, line 155: I was thinking inside the else but not inside the `if (!initializers.isEmpty())`? Or did you have a logical reason for why it is inside the if?

Lojjs (approved):

PR #3699, opened 2023-07-27
[NOID] Upgrades gradle version

ncordon:
## What
Upgrades gradle to latest 8.x version with:

```
./gradlew wrapper --gradle-version 8.4
```

## Why
Because we were getting errors that seem to be a bug in previous 6.x versions:

```
Could not resolve all dependencies for configuration ':core:testCompileClasspath'.
Problems reading data from Binary store in /private/var/folders/w1/990bdn7j5nb37bbydkqct5gh0000gn/T/gradle12884335207698123056.bin offset 28067 exists? true
```


gem-neo4j (approved):
Thanks :D

full/build.gradle
```diff
@@ -18,17 +16,24 @@ spotless {
...
+base {
+    archivesName = "apoc"
+}
 
 jar {
     manifest {
-        attributes 'Implementation-Version': version
+        attributes 'Implementation-Version': archiveVersion
     }
+    archiveClassifier = 'full'
```
ncordon, line 27: Make sure this is not breaking the jar classifiers like https://github.com/neo4j/apoc/issues/259#issuecomment-1403779626

gem-neo4j (approved):
Great, this looks reasonable to me! :D Thank you for doing this! Shall we update 5.x to 8.4 as well? 😁

ncordon:
> Shall we update 5.x to 8.4 as well? 😁

Yep, I'll take care

PR #3703, opened 2023-07-31
[NOID] Fixes failing path traversal TeamCity tests

ncordon:
## What
Checks sequences of `%e%e%f%f%e%e` do not fail but stay on the same directory when trying to do path traversals.

Commit [0421c2b](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3703/commits/0421c2b632e60af3f008496ff85b0808c0ff675f) reproduces the problem, [eec4def](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3703/commits/eec4def43b334fac0ea8793948af61b1e8b4a6de) amends the tests to solve it.

## Why
The tests were failing in our TeamCity CI. This treats them as we were treating sequences like `..//..//`

They've solved an issue with OpenJDK where it wasn't properly detecting double slashes in a path and turning them into a single one: https://github.com/openjdk/jdk11u-dev/commit/18939b76259456e85cfa051dc595e57242ca4ef8, https://nvd.nist.gov/vuln/detail/CVE-2023-22049

This fix has gone in Eclipse Temurin 11.0.20, all distributions of [Zulu](https://docs.azul.com/core/zulu-openjdk/release-notes/july-2023) or [Oracle v17](https://linux.oracle.com/errata/ELSA-2023-4159.html). The rationale to change it is that most of the JDKs should catch up with this fix, but also if they don't we were still protected throwing a failure.

### DIFF
--- a/core/src/test/java/apoc/export/ExportCoreSecurityTest.java
+++ b/core/src/test/java/apoc/export/ExportCoreSecurityTest.java
@@ -155,10 +155,10 @@ public class ExportCoreSecurityTest {
         // non-failing cases, with apoc.import.file.use_neo4j_config=false
         public static final List<String> casesAllowed = Arrays.asList(case03, case04, case05);
 
-        private static final String case08 = "tests/../../" + FILENAME;
-        private static final String case09 = "tests/..//..//" + FILENAME;
+        private static final String case07 = "tests/../../" + FILENAME;
+        private static final String case08 = "tests/..//..//" + FILENAME;
 
-        public static final List<String> casesOutsideDir = Arrays.asList(case01, case02, case03, case04, case05, case08, case09);
+        public static final List<String> casesOutsideDir = Arrays.asList(case01, case02, case03, case04, case05, case07, case08);
 
         /*
          All of these will resolve to a local path after normalization which will point to
--- a/core/src/test/java/apoc/export/ExportCoreSecurityTest.java
+++ b/core/src/test/java/apoc/export/ExportCoreSecurityTest.java
@@ -155,13 +155,10 @@ public class ExportCoreSecurityTest {
         // non-failing cases, with apoc.import.file.use_neo4j_config=false
         public static final List<String> casesAllowed = Arrays.asList(case03, case04, case05);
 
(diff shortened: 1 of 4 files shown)


core/src/test/java/apoc/export/ExportCoreSecurityTest.java
```diff
@@ -155,13 +155,10 @@ public TestIllegalExternalFSAccess(String exportMethod, String exportMethodType,
...
         public static final List<String> casesAllowed = Arrays.asList(case03, case04, case05);
 
-        private static final String case06 = "file:///%2e%2e%2f%2f%2e%2e%2f%2f%2e%2e%2f%2f%2e%2e%2f%2fapoc/" + FILENAME;
-        public static final String case07 = "file:///%2e%2e%2f%2f" + FILENAME;
         private static final String case08 = "tests/../../" + FILENAME;
         private static final String case09 = "tests/..//..//" + FILENAME;
 
-        public static final List<String> casesOutsideDir = Arrays.asList(case01, case02, case03, case04, case05,
-                case06, case07, case08, case09);
+        public static final List<String> casesOutsideDir = Arrays.asList(case01, case02, case03, case04, case05, case08, case09);
```
gem-neo4j, line 161: Maybe rename 8 and 9 to 6 and 7 so we don't skip those numbers? :)

full/src/test/java/apoc/export/ExportFullSecurityTest.java
```diff
@@ -202,13 +202,11 @@ public TestPathTraversalAccess(String exportMethod, String exportMethodType, Str
...
         }
 
-        private static final String case1 = "'file:///%2e%2e%2f%2f%2e%2e%2f%2f%2e%2e%2f%2f%2e%2e%2f%2fapoc/test.txt'";
-        private static final String case2 = "'file:///%2e%2e%2f%2ftest.txt'";
         private static final String case3 = "'../test.txt'";
         private static final String case4 = "'tests/../../test.txt'";
         private static final String case5 = "'tests/..//..//test.txt'";
 
-        private static final List<String> cases = Arrays.asList(case1, case2, case3, case4, case5);
+        private static final List<String> cases = Arrays.asList(case3, case4, case5);
```
gem-neo4j, line 209: Same here with case1 and 2

gem-neo4j (approved):
I think this looks good :)

PR #3710, opened 2023-08-02
[47wOlqLp] Adds option to customize decompression ratio

ncordon:
## Why
To allow for customers to tweak exactly how big the uncompressed files can be.

### DIFF
--- a/core-it/src/test/java/apoc/core/it/LoadCoreEnterpriseTest.java
+++ b/core-it/src/test/java/apoc/core/it/LoadCoreEnterpriseTest.java
@@ -41,10 +41,12 @@ import java.util.Map;
 import java.util.function.Consumer;
 import java.util.stream.LongStream;
 
+import static apoc.ApocConfig.APOC_MAX_DECOMPRESSION_RATIO;
+import static apoc.ApocConfig.DEFAULT_DECOMPRESSION_RATIO;
 import static apoc.export.util.LimitedSizeInputStream.SIZE_EXCEEDED_ERROR;
-import static apoc.export.util.LimitedSizeInputStream.SIZE_MULTIPLIER;
 import static apoc.util.TestContainerUtil.createEnterpriseDB;
 import static apoc.util.TestContainerUtil.testCall;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/apoc/ApocSettings.java
```diff
@@ -107,6 +107,9 @@ public Class<IPAddressString> getType()
 
     public static final Setting<Long> apoc_jobs_scheduled_num_threads = newBuilder(APOC_CONFIG_JOBS_SCHEDULED_NUM_THREADS, LONG, null).build();
 
+    @Description("maximum decompression ratio when using procedures that will decompress a compressed file passed as parameter")
+    public static final Setting<Integer> apoc_decompression_ratio = newBuilder(APOC_MAX_DECOMPRESSION_RATIO, INT, DEFAULT_DECOMPRESSION_RATIO).build();
+
```
ncordon, line 112: If I remove this from the file or even if I add a constraint `.addConstraint(min(1))` before the `.build()` call it has no effect, so I think we are not validating apoc settings at all?

core/src/main/java/apoc/ApocConfig.java
```diff
@@ -96,7 +96,8 @@ public enum UuidFormatType { hex, base64 }
     public static final String APOC_CONFIG_JOBS_QUEUE_SIZE = "apoc.jobs.queue.size";
     public static final String APOC_CONFIG_INITIALIZER = "apoc.initializer";
     public static final String LOAD_FROM_FILE_ERROR = "Import from files not enabled, please set apoc.import.file.enabled=true in your apoc.conf";
-
+    public static final String APOC_MAX_DECOMPRESSION_RATIO = "apoc.max.decompression.ratio";
+    public static final Integer DEFAULT_DECOMPRESSION_RATIO = 100;
     /**
```
ncordon, line 101: I think with apoc config there's currently no way to write constraints like min > 1. With the neo4j config there is. Also I don't know whether we want to allow negative numbers (=no protection against zip bombs). What do you think?

Lojjs: For negative number question I would ask Alex in security team. From a technical standpoint I think either is fine. If we end up not allowing it, I guess we can have a check and fail when it is used, even if that is not as elegant as the Neo4j way.

core-it/src/test/java/apoc/core/it/LoadCoreEnterpriseTest.java
```diff
@@ -210,6 +217,23 @@ public void testLoadShouldPreventZipBombAttackUtilDecompress() {
         });
     }
 
+    @Test
+    public void testLoadWorksWithIncreasedCompressionRatio() {
```
Lojjs, line 221: This is a good test, but I think it would be nice with an additional test, that gets zipbomb issue despite not using the default ratio. E.g. if you change the ratio to 101 instead of 100 it should still fail but with an error message that say 101.

Lojjs (changes requested):
Overall it looks fine, but I would like to see one more test and also discuss the negative number thing with security team

Lojjs (approved):
Looks good now :)

PR #3715, opened 2023-08-09
[NOID] Apoc Config import fix

gem-neo4j:
This was a mistake made when backporting to 4.4, the 5.x version was correct but must have auto imported googles one over Neo's :| This means that if a user tries to use expand config with APOC installed, there is a missing dependency, the workaround is to use just pop the google jar into plugins, but this is the real fix :)


ncordon (commented):
I think this would have been caught by an integration test. Should we add one?

gem-neo4j:
Yeah, that sounds good, will do that :)

ncordon (approved):
We decided to skip the test as it was quite tricky to add it

PR #3752, opened 2023-08-28
[w652gDBp] Update Neo4j and APOC versions

Lojjs:


readme.adoc
```diff
@@ -1,8 +1,8 @@
 :readme:
 :branch: 4.4
 :docs: https://neo4j.com/labs/apoc/4.4
-:apoc-release: 4.4.0.8
-:neo4j-version: 4.4.4
+:apoc-release: 4.4.0.22
```
gem-neo4j, line 4: Hmm this is tricky, for 4.4 I guess it does work with just using the previous released version though as there is only one 4.4 branch, unlike 5.x series 🤔

gem-neo4j (approved):
I am happy with this, just a comment about the README but don't see that as an issue with this PR tbh, so think you can merge and we can deal with that later

PR #3800, opened 2023-10-19
[6R2yEhGU] No procedure with name xxxx registered for this database instance

vga91:
## Steps to reproduce

Create multiple procs/overrides with low refresh config value (e.g.: `apoc.custom.procedures.refresh=10`)
```
CALL apoc.custom.declareProcedure('<procName>() :: (answer::INT)', $query)
call custom.<procName>
```

## Why

Sometimes it can happen that during the refresh the system database has discrepancies with the cache (i.e. `registeredProcedureSignatures` and `registeredUserFunctionSignatures`) and therefore some procedures/functions are mistakenly de-registered.

- Added `synchronized` in `restoreProceduresAndFunctions`, `storeFunction` and `storeProcedure` methods


full/src/test/java/apoc/custom/CypherProceduresStorageTest.java
```diff
@@ -55,7 +58,7 @@
 public class CypherProceduresStorageTest {
     private final static String QUERY_CREATE = "RETURN $input1 + $input2 as answer";
     private final static String QUERY_OVERWRITE = "RETURN $input1 + $input2 + 123 as answer";
-    private final int refreshTime = 3000;
+    private final int refreshTime = 50;
```
conker84, line 61: why this?

vga91: To make the `*WorksCorrectlyWithoutConflicts()` tests more deterministic, since the bug occurs when calling a custom procedure/function during the refresh phase. 
With a high `refreshTime` the tests are almost always green even without the `synchronized` clauses.

conker84 (approved):

PR #3802, opened 2023-10-23
Fixes #3737: Add support for AWS Bedrock for ml procedures

### LINKED ISSUE #3737: Add support for AWS Bedrock for ml procedures
* Bedrock API reference 
* Invoke Model: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_InvokeModel.html
* List foundation models: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_ListFoundationModels.html

* AWS Bedrock workshop: https://github.com/aws-samples/amazon-bedrock-workshop
* 
potentially need to look at langchain code:

* Default Model: amazon.titan-tg1-large ?
* Langchain LLMs https://python.langchain.com/docs/integrations/llms/bedrock
* LangChain Embeddings: https://python.langchain.com/docs/integrations/text_embedding/bedrock

vga91:
Added Bedrock Procedures

- apoc.ml.bedrock.custom: customizable (endpoint, modelId, http method, ....)
- ad-hoc procedures for models, like [for this](https://aws.amazon.com/bedrock/claude/), with a custom endpoint and which return a consistent data type based on the response, instead of `ObjectResult`:
   - apoc.ml.bedrock.jurassic
   - apoc.ml.bedrock.anthropic.claude
   - apoc.ml.bedrock.titan.embedding
   - apoc.ml.bedrock.stability

- Procedure to retrieve the [list of models](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_ListFoundationModels.html):
   - apoc.ml.bedrock.list

If the endpoint is not defined in procedures except the `.list`, the default is `https://bedrock-runtime.us-east-1.amazonaws.com/model/<MODEL_ID>/invoke`.

Used [this library](https://github.com/lucasweb78/aws-v4-signer-java/), to [calculate signature](https://docs.aws.amazon.com/AmazonS3/latest/API /sig-v4-examples-using-sdks.html) avoiding downloading the aws sdk dependencies


conker84 (approved):

extended/src/main/java/apoc/util/ExtendedUtil.java
```diff
@@ -2,16 +2,110 @@
...
+            checkResponseSuccess(response, stream);
+
+            return streamObjetsFromIStream(stream, path, pathOptions);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static void checkResponseSuccess(HttpResponse response, InputStream stream) throws IOException {
+        if (response.getStatusLine().getStatusCode() / 100 != 2) {
```
conker84, line 87: Why this?

docs/asciidoc/modules/ROOT/pages/ml/bedrock.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+Here is a list of all available Aws Bedrock procedures:
+
+
+[opts=header, cols="1, 4", separator="|"]
+|===
+|name| description
+|apoc.ml.bedrock.custom(body, $config)| To create a customizable Bedrock API call
+|apoc.ml.bedrock.list($config)| To get the list of foundation or custom models
+|apoc.ml.bedrock.jurassic(body, $config)| To create an API call to `Jurassic-2` model
```
jexp, line 16: it would be nice if we could mirror (at least in addition) the existing procedures, where we pass the model name as config parameter.
and then have at least the two for embedding generation that returns data and metadata for the embeddings
and one for chat-completion that returns the json for the chat completion

if we want to add stable diffusion i.e. image or binary generation, I suggest that we add a third type that is called `apoc.ml.bedrock.image` (and where you can pass different image models)

docs/asciidoc/modules/ROOT/pages/ml/bedrock.adoc
```diff
@@ -0,0 +1,225 @@
...
+|===
+|  completion| 	stopReason
+| " Hello!"	| "stop_sequence"
+|===
+
+
+.apoc.ml.bedrock.titan.embed
+[source,cypher]
+----
+CALL apoc.ml.bedrock.titan.embed({inputText: "Hello World"})
```
jexp, line 154: should be `apoc.ml.bedrock.embedding(text-or-list-of-texts, {config with optional model and a default})`

docs/asciidoc/modules/ROOT/pages/ml/bedrock.adoc
```diff
@@ -0,0 +1,225 @@
...
+| modelId                           | modelArn                                                             |modelName                   |providerName  |responseStreamingSupported|customizationsSupported|inferenceTypesSupported|inputModalities  |outputModalities
+| "amazon.titan-tg1-large"          |"arn:aws:bedrock:us-east-1::foundation-model/amazon.titan-tg1-large"  |"Titan Text Large"          |"Amazon"      |true                      |["FINE_TUNING"]        |["ON_DEMAND"]          |["TEXT"]         |["TEXT"]        
+| "amazon.titan-e1t-medium"         |"arn:aws:bedrock:us-east-1::foundation-model/amazon.titan-e1t-medium" |"Titan Text Embeddings"     |"Amazon"      |null                      |[]                     |["ON_DEMAND"]          |["TEXT"]         |["EMBEDDING"]
+| ...       |... |...     |...     |null                      |[]                     |...         |...         |...
+|===
+
+
+== Custom AWS API Call
+
+Via the `apoc.ml.bedrock.custom` we can create a customizable Bedrock API Request, by choosing the HTTP Method, the endpoint, the region and the additional headers.
```
jexp, line 205: we can definitely keep the custom, just need the same base procedures as the other providers

and not tie ourselves in procedure names to model names, if they change their names or discontinue models it's much harder to adjust than changing default models in the config.

extended/src/main/java/apoc/get/GetProcedures.java
```diff
@@ -20,13 +21,13 @@ public class GetProcedures {
     @Procedure
     @Description("apoc.get.nodes(node|id|[ids]) - quickly returns all nodes with these id's")
     public Stream<NodeResult> nodes(@Name("nodes") Object ids) {
-        return new Get(tx).nodes(ids);
+        return new Get((InternalTransaction) tx).nodes(ids);
```
jexp, line 24: is this related or unrelated change?

vga91: Yes sorry, it is not related to the PR indeed, it's needed to fix a compilation error due to a 5.13 core change,
I'll put it in a separate PR along with the other errors I'm fixing

extended/src/main/java/apoc/ml/bedrock/BedrockInvokeConfig.java
```diff
@@ -0,0 +1,24 @@
+package apoc.ml.bedrock;
+
+import java.util.Map;
+
+public class BedrockInvokeConfig extends BedrockConfig {
+    public static final String MODEL_ID = "modelId";
```
jexp, line 6: change to `model`

extended/src/main/java/apoc/ml/bedrock/Bedrock.java
```diff
@@ -0,0 +1,123 @@
...
+
+import static apoc.ml.bedrock.BedrockInvokeConfig.MODEL_ID;
+import static apoc.util.JsonUtil.OBJECT_MAPPER;
+import static apoc.ml.bedrock.BedrockInvokeResult.*;
+import static apoc.ml.bedrock.BedrockUtil.*;
+
+
+public class Bedrock {
+    
+    @Procedure
```
jexp, line 27: let's put the explicit procedure name into the `@Procedure` annotation, better and safer for refactorings.

extended/src/main/java/apoc/ml/bedrock/Bedrock.java
```diff
@@ -0,0 +1,123 @@
...
+                .flatMap(i -> ((List<Map<String, Object>>) i).stream())
+                .map(ModelItemResult::new);
+    }
+    
+    @Procedure
+    @Description("To create a customizable bedrock call")
+    public Stream<ObjectResult> custom(@Name(value = "body") Object body,
+                                       @Name(value = "config", defaultValue = "{}") Map<String, Object> config) {
+        
+        return executeCustomRequest(body, config, null)
```
jexp, line 42: would it be possible to have a `MapResult` instead?

Do we need the object only for the image generation? If so I would probably skip image generation for the time being.

Or add a `customBinary` procedure for object response.

extended/src/main/java/apoc/ml/bedrock/Bedrock.java
```diff
@@ -0,0 +1,123 @@
...
+                .map(ObjectResult::new);
+    }
+    
+    @Procedure
+    public Stream<Jurassic> jurassic(@Name(value = "body") Object body,
```
jexp, line 47: We should not tie the result type to the model if we can avoid it but convert it to `MapResult`

extended/src/main/java/apoc/ml/bedrock/Bedrock.java
```diff
@@ -0,0 +1,123 @@
...
+                                     @Name(value = "config", defaultValue = "{}") Map<String, Object> config) {
+        
+        config.putIfAbsent(MODEL_ID, JURASSIC_2_ULTRA);
```
jexp, line 50: the config param might not stay a writeable map in the future,

let's create a new one like here:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/dev/extended/src/main/java/apoc/ml/OpenAI.java#L55

extended/src/main/java/apoc/ml/bedrock/Bedrock.java
jexp: same below for the other procedures, which should be generalized anyhow

extended/src/main/java/apoc/ml/bedrock/Bedrock.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private Stream<Map<String, Object>> executeRequestReturningMap(Object body, Map<String, Object> config, String path) {
+        return executeCustomRequest(body, config, path)
+                .map(i -> (Map<String, Object>) i);
+    }
+
+    private Stream<Object> executeCustomRequest(Object body, Map<String, Object> config, String path) {
+        BedrockConfig conf = new BedrockInvokeConfig(config);
+
+        return executeRequestCommon(body, path, conf);
```
jexp, line 92: shouldn't this also return a `Stream<Map<String, Object>>` ?

extended/src/main/java/apoc/ml/bedrock/Bedrock.java
```diff
@@ -0,0 +1,123 @@
...
+        BedrockConfig conf = new BedrockGetModelsConfig(config);
+        
+        return executeRequestCommon(null, "modelSummaries[*]", conf)
+                .flatMap(i -> ((List<Map<String, Object>>) i).stream())
+                .map(ModelItemResult::new);
+    }
+    
+    @Procedure
+    @Description("To create a customizable bedrock call")
+    public Stream<ObjectResult> custom(@Name(value = "body") Object body,
```
jexp, line 39: we should not use object bodies

either have something structured like a map that we we also document (see VertexAI)
or a string/list of strings for simple calls (like in the other ones)

model parameters, like top-k, top-p, temparature etc. should go into the config parameter and passed through

the main ones should also be documented, and also add links in the docs to AWS docs for the model(parameters)

jexp: same for the other procs, but if we generalize it they should become two (embedding + chat-completion) or three (if you want to add image).

and additionally the custom one.

extended/src/main/java/apoc/ml/bedrock/BedrockInvokeResult.java
```diff
@@ -0,0 +1,45 @@
...
+        public static StabilityAi from(Map<String, Object> map) {
+            String base64 = (String) map.get("base64");
+            
+            return new StabilityAi(base64);
+        }
+    }
+
+    public record AnthropicClaude(String completion, String stopReason) {
+        public static AnthropicClaude from(Map<String, Object> map) {
+            String completion = (String) map.get("completion");
```
jexp, line 17: there should be more fields in there

extended/src/main/java/apoc/ml/bedrock/BedrockInvokeResult.java
```diff
@@ -0,0 +1,45 @@
...
+            String stopReason = (String) map.get("stop_reason");
+            
+            return new AnthropicClaude(completion, stopReason);
+        }
+    }
+
+    public record TitanEmbedding(Long inputTextTokenCount, List<Number> embedding) {
+        public static TitanEmbedding from(Map<String, Object> map) {
+            Long inputTextTokenCount = (Long) map.get("inputTextTokenCount");
+            List<Number> embedding = (List<Number>) map.get("embedding");
```
jexp, line 27: probably use List<Double> instead, as Cypher doesn't really know how to deal with Number type.

extended/src/main/java/apoc/ml/bedrock/BedrockInvokeResult.java
```diff
@@ -0,0 +1,45 @@
...
+            String completion = (String) map.get("completion");
+            String stopReason = (String) map.get("stop_reason");
+            
+            return new AnthropicClaude(completion, stopReason);
+        }
+    }
+
+    public record TitanEmbedding(Long inputTextTokenCount, List<Number> embedding) {
+        public static TitanEmbedding from(Map<String, Object> map) {
+            Long inputTextTokenCount = (Long) map.get("inputTextTokenCount");
```
jexp, line 26: there are more fields

extended/src/main/java/apoc/ml/bedrock/BedrockInvokeResult.java
```diff
@@ -0,0 +1,45 @@
...
+            List<Number> embedding = (List<Number>) map.get("embedding");
+            
+            return new TitanEmbedding(inputTextTokenCount, embedding);
+        }
+    }
+    
+    public record Jurassic(Long id, List<Object> promptTokens, List<Object> completions) {
+        public static Jurassic from(Map<String, Object> map) {
+            Long id = (Long) map.get("id");
```
jexp, line 35: what is the id?

extended/src/main/java/apoc/ml/bedrock/BedrockInvokeResult.java
```diff
@@ -0,0 +1,45 @@
...
+
+            Map prompt = (Map) map.get("prompt");
+            List<Object> promptTokens = (List<Object>) prompt.get("tokens");
+
+            List<Object> completions = (List<Object>) map.get("completions");
```
jexp, line 40: do we have a deeper level of what the completions look like? e.g. `List<Map<String,String>>` or something similar? so that cypher can process it later.

extended/src/main/java/apoc/ml/bedrock/BedrockUtil.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Map;
+
+
+public class BedrockUtil {
+    public static final String SERVICE_NAME = "bedrock";
+    
+    public static final String AUTHORIZATION = "Authorization";
+    public static final String X_AMZ_DATE = "X-Amz-Date";
+    
+    public static final String JURASSIC_2_ULTRA = "ai21.j2-ultra-v1";
```
jexp, line 22: should these be here or in the procedures file or somewhere separate?

extended/src/main/java/apoc/util/ExtendedUtil.java
```diff
@@ -2,16 +2,110 @@
...
+            case HttpOptions.METHOD_NAME -> new HttpOptions(uri);
+            case HttpPatch.METHOD_NAME -> new HttpPatch(uri);
+            case HttpPost.METHOD_NAME -> new HttpPost(uri);
+            case HttpPut.METHOD_NAME -> new HttpPut(uri);
+            default -> throw new RuntimeException("Unknown HTTP method name: " + method);
+        };
+    }
+
+    /**
+     * Similar to JsonUtil.loadJson(..) but works e.g. with GET method as well,
```
jexp, line 60: `loadJSON` should also work with get.

we did a lot of work with the security team on http access and load json.
what happens if we pass in file URLs here etc?
not to fix now due to the time constraint but consider to make it work with `loadJson`

extended/src/main/java/apoc/ml/bedrock/BedrockInvokeResult.java
vga91: It's the ID response obtained from the model: https://docs.ai21.com/reference/j2-complete-api-ref#id

conker84 (approved):

PR #3824, opened 2023-11-01
Fixed the dots

orthogonal3:
As per https://neo4j.slack.com/archives/C136J23GE/p1698854430024199 I've removed the leading dots from the index.

It looks like it maybe came across via copy-paste from the tables bit below? It seems those entries start with a dot for numbered list

  -
  -
  -


jexp (approved):

jexp:
@conker84 we need to cherry pick this into dev

PR #3825, opened 2023-11-01
Fixed table formatting-removed "|" in bot|user

martynscn:
Removed "|" in bot|user which caused the table to be misaligned and not understandable.

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/ml/vertexai.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/ml/vertexai.adoc
@@ -111,7 +111,7 @@ yield value
 [%autowidth, opts=header]
 |===
 |name | description
-| messages | List of maps of instructions with `{author:"bot|user", content:"text"}`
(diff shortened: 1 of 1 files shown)


docs/asciidoc/modules/ROOT/pages/ml/vertexai.adoc
```diff
@@ -111,7 +111,7 @@ yield value
 [%autowidth, opts=header]
 |===
 |name | description
-| messages | List of maps of instructions with `{author:"bot|user", content:"text"}`
+| messages | List of maps of instructions with `{author:"bot user", content:"text"}`
```
vga91, line 114: The correct formatting should be with `\|`, which return it `{author:"bot|user", content:"text"}`, which means `"author": "bot" OR "user"`

```suggestion
| messages | List of maps of instructions with `{author:"bot\|user", content:"text"}`
```

vga91 (changes requested):
Thanks for your submission!
Please, can you see my comment?

PR #3880, opened 2023-12-21
Fixes #3858: Allow configuration of openai base URL as parameter

### LINKED ISSUE #3858: Allow configuration of openai base URL as parameter
Extend the work we've done with MSFT and allow configuration of base URL (with placeholder, if there is none, just append the path)

can we add the base URL as parameter for openai calls in apoc
then we can use all the drop-in replacements, like anyscale or the msft openai.

Examples that we could test.

- https://app.endpoints.anyscale.com/
- https://localai.io/
- LLMatic
- OpenLM

vga91:
Fixes #3858

- Added path configurations to manipulate the url suffix (which by default is /embeddings, /completions and /chat/completions)

- Added tests for https://app.endpoints.anyscale.com/

- Added tests for Local AI

- Added tests for LLMatic (with assertEventually as they are very slow and often go into SocketTimeout)

- Added tests to simulate the behavior of https://github.com/r2d4/openlm, which manipulates APIs to have an OpenAI-compatible format. Tested with Cohere and HuggingFace. As with the library, only test for the Completions API

- An OpenAIRequestHandler.Type has been added for HuggingFace, to transform the input into a format consistent with the HuggingFace Api (i.e. `{"inputs": ["text...."]}`) and change the jsonPath to `$[0]`. With this API, we need to explicit the config `path: ''`
- For Cohere, added note on the documentation to explain to make explicit path: '', otherwise it adds /completions to the url, which is not foreseen by the cohere API

- Changed `System.setProperty` to `apocConfig().setProperty` in `OpenAITest.java` not to interfere with other ml tests, if we run them all

### DIFF
--- a/extended/src/main/java/apoc/util/ExtendedUtil.java
+++ b/extended/src/main/java/apoc/util/ExtendedUtil.java
@@ -4,6 +4,11 @@ import static apoc.export.cypher.formatter.CypherFormatterUtils.formatProperties
 import static apoc.export.cypher.formatter.CypherFormatterUtils.formatToString;
 
 import java.math.BigInteger;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.net.SocketException;
+import java.net.URI;
+import java.net.UnknownHostException;
 import java.time.Duration;
 import java.time.temporal.TemporalAccessor;
 import java.util.Map;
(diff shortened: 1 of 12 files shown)


extended/src/main/java/apoc/util/ExtendedUtil.java
```diff
@@ -43,4 +48,18 @@ public static String toCypherMap( Map<String, Object> map) {
         final StringBuilder builder = formatProperties(map);
         return "{" + formatToString(builder) + "}";
     }
+
+    public static boolean isLocalAddress(String url) {
```
conker84, line 52: Please delete it

conker84 (changes requested):

conker84 (approved):

PR #3889, opened 2024-01-09
Fix HelpExtendedTest not to edit extended.txt when executed

vga91:
The [HelpExtendedTest.indicateNotCore()](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/dev/extended/src/test/java/apoc/help/HelpExtendedTest.java#L68) test currently changes the `extended.txt` file when executed.

Therefore, if some classes have / do not have the `@Extended` annotation, [like in this case](https://github.com/neo4j-contrib/neo4j-apoc-procedures/commit/83199902017ac9e89d699954a6f581147fa7e8b6),
the result of `StartupExtendedTest` (which [leverage the extended.txt](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/dev/extended/src/test/java/apoc/StartupExtendedTest.java#L39))
may change if it is run before (which happens with gh action) or after (with TC) HelpExtendedTest.

Changed the HelpExtendedTest so that doesn't change the txt file, 
but checks that `apoc.help('')` and `SHOW FUNCTION/PROCEDURE` coincide.

---

Maybe the extended.txt file could be generated automatically via reflection, using the @Extended annotation, but it should be done every time we compile / build / run test. So can be hardily feasible and error-prone.
I think it's better to leave it as it is, adding the procedures/functions manually.

### DIFF
--- a/extended/src/test/java/apoc/help/HelpExtendedTest.java
+++ b/extended/src/test/java/apoc/help/HelpExtendedTest.java
@@ -14,9 +14,11 @@ import org.reflections.scanners.SubTypesScanner;
 import org.reflections.scanners.TypeAnnotationsScanner;
 import org.reflections.util.ConfigurationBuilder;
 
-import java.io.*;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static apoc.util.Util.map;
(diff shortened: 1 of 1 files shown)


vga91:
Closed in favor of https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3896

PR #3898, opened 2024-01-12
[SGrPWNEf] Fixes backslashes parsing in csv import / export

ncordon:
## What

Uses `RFC4180Parser` instead of `CSVParser` as suggested [here](https://stackoverflow.com/questions/46139735/why-csvwriter-and-csvreader-uses-different-default-escape-characters)

## Why

Because the case where we have a quote `"` in a value we want to export / import from csv was not working:

```
CREATE (n:Test {name: 'Test', value: '{"new":"4\'10\\\""}'});
CALL apoc.export.csv.all("test.csv", {bulkImport:'true'});
MATCH (n) DETACH DELETE n;
CALL apoc.import.csv([{fileName: 'file:test.nodes.Test.csv', labels: ['Test']}],[],{});
```

```
MATCH (n: Test) RETURN n.value
```

should give us a roundtrip where the result is `"{"new":"4'10\""}"`

### DIFF
--- a/core/src/main/java/apoc/export/csv/CsvEntityLoader.java
+++ b/core/src/main/java/apoc/export/csv/CsvEntityLoader.java
@@ -28,6 +28,8 @@ import apoc.util.FileUtils;
 import com.opencsv.CSVParserBuilder;
 import com.opencsv.CSVReader;
 import com.opencsv.CSVReaderBuilder;
+import com.opencsv.RFC4180ParserBuilder;
+
 import java.io.IOException;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/export/csv/CsvEntityLoader.java
```diff
@@ -99,7 +100,7 @@ public void loadNodes(
             final Map<String, Mapping> mapping = getMapping(fields);
 
             final CSVReader csv = new CSVReaderBuilder(reader)
-                    .withCSVParser(new CSVParserBuilder()
+                    .withCSVParser(new RFC4180ParserBuilder()
```
loveleif, line 103: Will this break any existing use cases? If so it would be nice to update documentation that we parse according to RFC 4180 now.

loveleif (approved):
```
review_status,comment
approved,＼(＾O＾)／
```

core/src/main/java/apoc/export/csv/CsvEntityLoader.java
ncordon: I think it would affect only the backslashes, or at least I couldn't find any information suggesting that it would parse anything else differently.

PR #3932, opened 2024-01-26
Fix flaky MongoDBTest

vga91:
The assertEventually should solve the following flaky TC assertion error.
Replicated locally executing the test ~20/30 times.

Bildschirmfoto 2024-01-26 um 00 34 09


vga91:
Closed in favor of https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3936

Issue #3946, opened 2024-02-14
Create apoc.ml.vertexai.image leveraging vertex ai gemini


vga91:
Create `apoc.ml.vertexai.image`, like the `apoc.ml.bedrock.image` one, using [this API](https://cloud.google.com/vertex-ai/docs/generative-ai/image/generate-images) would be great. 

Currently, it has limited access, 
so we should wait for public access to implement it.

PR #3947, opened 2024-02-14
Fixes #3939: vertexai gemini has a different URL, we should support that in the vertexai procedure

### LINKED ISSUE #3939: vertexai gemini has a different URL, we should support that in the vertexai procedure
see:

```
curl -X POST \
    -H "Authorization: Bearer $(gcloud auth print-access-token)" \
    -H "Content-Type: application/json; charset=utf-8" \
    -d @gemini.json \ 
    "https://us-central1-aiplatform.googleapis.com/v1/projects/$PROJECT_ID/locations/us-central1/publishers/google/models/gemini-pro:streamGenerateContent"
```

currently we use `:predict` as suffix which doesn't seem to work anymore with gemini, also the request body changed a bit

we can probably have a new method called apoc.ml.vertexai.stream 

that also works with the new body.

gemini-pro-vision supports also image payloads to e.g. describe or summarize images, not sure if we need that yet

https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/gemini#request_body

`POST https://{REGION}-aiplatform.googleapis.com/v1/projects/{PROJECT_ID}/locations/{REGION}/publishers/google/models/gemini-pro-vision:streamGenerateContent`

https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/gemini#gemini-pro-vision

We probably need that for gemini ultra anyhow.

vga91:
Fixes #3939

- Now the URL can be configured via the apoc.conf `apoc.ml.vertexai.url` (like OpenAI procedures)
- Added `VertexAIHandler` abstract factory to handle the body, jsonpath and URL based on the procedure 

- added `apoc.ml.vertexai.stream`, which calls gemini AI endpoints (which end with `:streamGenerateContent` instead of `:predict`)
- added `apoc.ml.vertexai.custom`, similar to `apoc.ml.bedrock.custom`
  - we can customize the endpoint, model, resource (eg `:streamGenerateContent` or `:predict`)
- updated documentation with configs and examples
- added test cases with gemini-pro and gemini-pro-vision multimodal APIs: https://cloud.google.com/vertex-ai/docs/generative-ai/embeddings/get-multimodal-embeddings

----
### Other considerations/changes

- Some tests currently fail:
  - `apoc.ml.VertexAIIT#chatCompletion`, due to: class java.util.ArrayList cannot be cast to class java.util.Map 
  - `apoc.ml.VertexAIIT#completion`, due to: assertEquals(true, result.containsKey("recitationResult"));

Most likely because the model is changed. 
Hence, I generalized the assertions, as I did in this PR, by just asserting the `text` that should return.

- As said in the issue, an apoc.ml.vertexai.image, like the `apoc.ml.bedrock.image` one, using [this API](https://cloud.google.com/vertex-ai/docs/generative-ai/image/generate-images) would be great. But currently, it has limited access, so I guess we can wait for public access and implement it later.
I opened another issue for that: https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/3946.


jexp (changes requested):
could we do this with named parameters instead, i.e.g {model} etc. will be easier for users to manage and handle. 

We can do a simple string replacement for these, nothing fancy.

jexp (approved):
Looks good, thanks so much!

PR #3950, opened 2024-02-20
Add apoc.cypher.runFile(s) procedures with READ mode version

vga91:
## Issue case

Based on the Trello card: https://trello.com/c/K86MBPd2/86-s4mattress-firm-recommendations-customer-360apoc-44x-5x-requesting-read-mode-version-of-apoccypherrunfile-and-apoccypherrunfiles, that is:

```
...
We are requesting a READ mode version of the following two procedures:

apoc.cypher.runFile()
apoc.cypher.runFiles()

...
```

---

## Implemented changes

- Created `apoc.cypher.runReadFile` and `apoc.cypher.runReadFiles`, respectively analogous to `apoc.cypher.runFile` and `apoc.cypher.runFiles` , but with [Mode.READ](https://github.com/neo4j/neo4j/blob/5.13/community/procedure-api/src/main/java/org/neo4j/procedure/Mode.java#L31).

  - Unlike write procedures, these ones will have the default configuration `statistics:false`, 
since the `true` option is not so useful, because it returns a map based on [this class](https://github.com/neo4j/neo4j/blob/5.13/community/graphdb-api/src/main/java/org/neo4j/graphdb/QueryStatistics.java), e.g.:
```
 {constraintsRemoved: 0, indexesRemoved: 0, nodesCreated: 1, rows: 0, propertiesSet: 1, labelsRemoved: 0, relationshipsDeleted: 0, constraintsAdded: 0, nodesDeleted: 0, indexesAdded: 0, labelsAdded: 1, relationshipsCreated: 0, time: 0}
```

and being read-only operations, they will always have `labelsRemoved`, `relationshipsDeleted`, etc... equal to 0.

- Added tests similar to write procedures, with a read-only files, which don't return statistics
- Added tests with a write file, which dont't return results

### Other changes

- Added explicit names to `@Procedure` annotation, for more clarity
- Added test cases with write procs and `{statistics:false}`, i.e. [testRunFileWithSetAndResultsAndStatisticsFalse](https://github.com/vga91/neo4j-apoc-procedures/pull/427/files#diff-2a962060018a60f344ba0b4b6197b2eae1ca2f07a2b24972e1b6865101a9fbccR132) and [testRunFileWithResultsAndStatisticsFalse](https://github.com/vga91/neo4j-apoc-procedures/pull/427/files#diff-2a962060018a60f344ba0b4b6197b2eae1ca2f07a2b24972e1b6865101a9fbccR148)


jexp:
Can you rename them to: `apoc.cypher.runFileReadOnly` ?
Makes it easier to understand what they are doing.

jexp (changes requested):
seem my comments on the changes, but otherwise looks good.

jexp (approved):
LGTM, Thank you!

PR #3961, opened 2024-02-26
[4Bece3wd] Use Neo4j home to determine default config location.

Lojjs:
On Windows service, 'sun.java.command' is not supported. The fallback default path to the config folder was set to '.' which is not the actual default. With this solution, the home folder of Neo4j is instead used to calculate the default.

This is the extended version of the APOC core fix from https://github.com/neo4j/apoc/pull/586


loveleif (approved):
Nice! It would be great with an installation test that checks this, perhaps we can create a card for release ops?

loveleif (approved):

PR #3994, opened 2024-03-06
Fixes #1464: Create apoc.temporal.overlap(start1, end1, start2, end2, acceptAdjacentSpans:true) function

### LINKED ISSUE #1464: Create apoc.temporal.overlap(start1, end1, start2, end2, acceptAdjacentSpans:true) function
Feature description (Mandatory)
This function should take two timespans, broken into start and end temporal values for each (and an optional boolean for whether an adjacent timespans are considered overlapping, true by default), and returns a boolean on whether the two timespans overlap.

The general formula for calculating overlap is:

`max(start1, start2) <= min(end1, end2)`
If `acceptAdjacentSpans:false`, then this uses `<` instead.

This overlap formula is proved in this Stack Overflow answer:
https://stackoverflow.com/a/325964

This function acts as a shorthand, as this formula is likely not commonly known.

## Considered alternatives
None considered at this time.

Determining time range overlap is a common use case, and we don't currently have anything native or in APOC to deal with this.

vga91:
Fixes #1464

Created `apoc.temporal.overlap` function.

- **_TO EVALUATE_**: It also refers to temporal spans, but perhaps it is worth considering all possible data types, for example to compare two ranges of longs (e.g. `with 1 as start1, 4 as end1, 3 as start2, 4 as end2 return (start1 < end2) AND (start2 < end1) AS value`).
In that case, we could rename the procedure as `apoc.util.overlap`.

- **_TO EVALUATE_**: Consistently with Cypher, if the data types are not comparable (see [testOverlapWithWrongTypes](https://github.com/vga91/neo4j-apoc-procedures/pull/430/files#diff-eb8939b27a0f33f653a0fcea156f71f8980cd8222bf4273f1989ccc20e55204fR128)), the procedure returns `null`.
Alternatively, we could return for example a `RuntimeException("type not comparable...")`,
or perhaps we can add another config `failSilently: <boolean>` that returns null if the config true.

- The issue suggests using `max(start1, start2) <= min(end1, end2)`, but for simplicity, I think it is better to use the other formula suggested in [stackoverflow](https://stackoverflow.com/a/325964), that is `(Start1 <= End2) and (Start2 <= End1)`.

- The procedure leverages `tx.execute(<query>)` instead of using the Java API, e.g. the [AnyValues.COMPARATOR](https://github.com/neo4j/neo4j/blob/5.13/community/values/src/main/java/org/neo4j/values/AnyValues.java#L71), since this way we are sure that the behavior is consistent with Cypher.


jexp (approved):
I think it is good like it is now. Let's merge it.

PR #4071, opened 2024-05-13
[6AtySekS] Bump version

loveleif:
- Restructure gradle project.
- Bump version.

### DIFF
--- a/LICENSES.txt
+++ b/LICENSES.txt
@@ -201,7 +201,7 @@ Apache-2.0
   parboiled-core-1.2.0.jar
   parboiled-scala_2.12-1.2.0.jar
   picocli-4.6.1.jar
-  reactor-core-3.4.15.jar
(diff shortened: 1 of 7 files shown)


LICENSES.txt
```diff
@@ -201,7 +201,7 @@ Apache-2.0
   parboiled-core-1.2.0.jar
   parboiled-scala_2.12-1.2.0.jar
   picocli-4.6.1.jar
-  reactor-core-3.4.15.jar
+  reactor-core-3.6.5.jar
```
JoelBergstrand, line 204: What are these bumps?

JoelBergstrand (approved):
More than not knowing what reactor-core is it looks good

LICENSES.txt
loveleif: Auto generated. I assume we have wildcard version on some dependencies so they can change on every build :(.

PR #4092, opened 2024-05-29
Fixes #4090: The apoc.vectordb.*.get/query procedures should search for nodes/relationships with mapping config

### LINKED ISSUE #4090: The apoc.vectordb.*.get/query procedures should search for nodes/relationships with mapping config
For example the following procedure should return for a node `(:Test {myId: '<metadata foo result returned by qdrant API>'})`:

```
CALL apoc.vectordb.qdrant.query($hostOrKey, 'test_collection',
    [0.2, 0.1, 0.9, 0.7],
    {},
    5, 
    { mapping: {
            nodeLabel: "Test", 
            entityKey: "myId", 
            metadataKey: "foo" 
        }
    })
```

vga91:
Fixes #4090

The apoc.vectordb.*.get/query procedures show nodes/rels that can be updated with mapping config

### DIFF
--- a/extended/src/main/java/apoc/vectordb/ChromaDb.java
+++ b/extended/src/main/java/apoc/vectordb/ChromaDb.java
@@ -129,7 +129,8 @@ public class ChromaDb {
                                                       @Name("collection") String collection,
                                                       @Name("ids") List<Object> ids,
                                                       @Name(value = "configuration", defaultValue = "{}") Map<String, Object> configuration) throws Exception {
-        return getCommon(hostOrKey, collection, ids, configuration, false);
+        setReadOnlyMappingMode(configuration);
+        return getCommon(hostOrKey, collection, ids, configuration);
     }
 
     @Procedure(value = "apoc.vectordb.chroma.getAndUpdate", mode = Mode.WRITE)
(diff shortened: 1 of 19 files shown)


extended/src/main/java/apoc/vectordb/ChromaDb.java
```diff
@@ -129,7 +129,7 @@ public Stream<EmbeddingResult> get(@Name("hostOrKey") String hostOrKey,
                                                       @Name("collection") String collection,
                                                       @Name("ids") List<Object> ids,
                                                       @Name(value = "configuration", defaultValue = "{}") Map<String, Object> configuration) throws Exception {
-        return getCommon(hostOrKey, collection, ids, configuration, true);
+        return getCommon(hostOrKey, collection, ids, configuration, false);
```
jexp, line 132: in general it would be good to avoid boolean variables, rather use an enum, it's confusing esp. if it changes meaning during development.
and what does "updateMode" mean :) 
don't we also have 3 modes? READ_ONLY, UPDATE_EXISTING, CREATE_IF_MISSING?

extended/src/main/java/apoc/vectordb/Milvus.java
```diff
@@ -133,7 +133,7 @@ public Stream<EmbeddingResult> get(@Name("hostOrKey") String hostOrKey,
                                          @Name("collection") String collection,
                                          @Name("ids") List<Object> ids,
                                          @Name(value = "configuration", defaultValue = "{}") Map<String, Object> configuration) throws Exception {
-        return getCommon(hostOrKey, collection, ids, configuration, true);
```
jexp, line 136: why is this Mode.WRITE? (can you also check the others)?

@Procedure(value = "apoc.vectordb.milvus.get", mode = Mode.WRITE)
    @Description("apoc.vectordb.milvus.get(hostOrKey, collection, ids,

PR #4106, opened 2024-06-19
[NOID] Bump Arrow versions based on Neo netty bump

gem-neo4j:


thelonelyvulpes (approved):
Many thanks

LICENSES.txt
```diff
@@ -3059,6 +3068,7 @@ MIT
   reactive-streams-1.0.4.jar
   slf4j-api-1.7.36.jar
   slf4j-api-2.0.11.jar
+  slf4j-api-2.0.9.jar
```
anderson4j, line 3071: All other versions are bumped, here we include an older instead. What gives?

NOTICE.txt
```diff
@@ -459,6 +468,7 @@ MIT
   reactive-streams-1.0.4.jar
   slf4j-api-1.7.36.jar
   slf4j-api-2.0.11.jar
+  slf4j-api-2.0.9.jar
```
anderson4j, line 471: Same as the other comment, so you see both in case it's a mistake

LICENSES.txt
gem-neo4j: This isn't a mistake, these files are auto-generated, it just means that arrow-vector is transitively dependent on 2.0.9

anderson4j (approved):
Good stuff

PR #4142, opened 2024-07-11
Fixes #4087: Add vector info procedures

### LINKED ISSUE #4087: Add vector info procedures
For the vectordb procedures, add a call to an endpoint [like this one, to retrieve the collection info](https://qdrant.github.io/qdrant/redoc/index.html#tag/collections/operation/get_collection).

The procedure names will be e.g "apoc.vectordb.qdrant.info"

vga91:
Fixes #4087

- Added `methodAndPayloadNull` method as a workaround, since we cannot explicit "method: GET" as [setChunkedStreamingMode](https://github.com/neo4j/apoc/blob/dev/common/src/main/java/apoc/util/Util.java#L348) makes procedures throw a HTTP 405 Method Not Allowed,
and by forcing `payload == null` we make sure that the procedure does not go [here](https://github.com/neo4j/apoc/blob/dev/common/src/main/java/apoc/util/Util.java#L377), and in this way it still has as default method "GET", which is the required method for all of them except the milvus one
- Added `apoc.vectordb.<vectorDbName>.info(hostOrKey, collection, $config)` procedures

### Other changes

Refactoring: added some comments in extended/src/main/java/apoc/vectordb/VectorDbUtil.java


docs/asciidoc/modules/ROOT/pages/database-integration/vectordb/chroma.adoc
```diff
@@ -7,6 +7,7 @@ note that the list and the signature procedures are consistent with the others,
 [opts=header, cols="1, 3"]
 |===
 | name | description
+| apoc.vectordb.chroma.info(hostOrKey, collection, $config) | Get information about specified existing collection
```
RobertoSannino, line 10: Add information about the output when the collection does not exists (return null? throws error?)

docs/asciidoc/modules/ROOT/pages/database-integration/vectordb/pinecone.adoc
```diff
@@ -54,6 +55,13 @@ image::pinecone-index.png[width=800]
 
 The following example assume we want to create and manage an index called `test-index`.
 
+.Get collection info (it leverages https://docs.pinecone.io/reference/api/control-plane/describe_collection[this API])
+[source,cypher]
+----
+CALL apoc.vectordb.pinecone.info(hostOrKey, 'test-collection', {<optional config>})
+----
+
```
RobertoSannino, line 63: Output missing

vga91: Added below

docs/asciidoc/modules/ROOT/pages/database-integration/vectordb/chroma.adoc
vga91: Added, i.e. `throws an error if it does not exist`

PR #4144, opened 2024-07-16
Fixes #4000: Add self-explanation to the model, include the verbal schema description to the flow

### LINKED ISSUE #4000: Add self-explanation to the model, include the verbal schema description to the flow
When we do calls to the LLM model, today we only pass in the schema.

We want to try to see if there are better results if we also include the "human description" of the schema that we get with apoc.ml.schema (as an option)

perhaps we can also just document how to do this with the existing procedures to augment the prompt with the additional schema.

vga91:
Fixes #4000

- Improved schema prompt to retrieve slightly better results
- Added `additionalPrompts` config which can accept list of maps with other request messages
 - we can use it together with apoc.ml.schema to improve the results
 - added examples with Northwind database: https://neo4j.com/docs/getting-started/appendix/tutorials/guide-import-relational-and-etl/


jexp (approved):
LGTM

PR #4150, opened 2024-07-24
[aYPLiTQG] Remove isSchema check to be more performant

gem-neo4j:
RunFile is not allowed to run schema operations, in 4.4.0.9 the validation check for this was a regex that was very basic, then in 4.4.0.11 this was updated to be more accurate, what happens is that every query gets run through an explain which checks if the query is a Schema operation, and if it is, it will ignore it. 

The problem I see with this is that it will now run this potentially time consuming check for every query in the file as well as actually running the query.

This PR removes this check and instead checks the errors thrown, if a schema error is thrown then it will handle this there to keep old behaviour.

### DIFF
--- a/full/src/main/java/apoc/cypher/CypherExtended.java
+++ b/full/src/main/java/apoc/cypher/CypherExtended.java
@@ -272,7 +272,8 @@ public class CypherExtended {
                             return consumeResult(result, queue, addStatistics, tx, fileName);
                         } catch (Exception e) {
                             // APOC historically skips schema operations
-                            if (!e.getMessage().contains("Schema operations on database 'neo4j' are not allowed")) {
+                            if (!(e.getMessage().contains("Schema operations on database")
+                                    && e.getMessage().contains("are not allowed"))) {
                                 collectError(queue, reportError, e, fileName);
                                 return null;
                             }
--- a/full/src/main/java/apoc/cypher/CypherExtended.java
+++ b/full/src/main/java/apoc/cypher/CypherExtended.java
@@ -49,6 +49,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


full/src/main/java/apoc/cypher/CypherExtended.java
```diff
@@ -246,40 +246,46 @@ private void runDataStatementsInTx(
...
-                                    stmt, params, result -> consumeResult(result, queue, addStatistics, tx, fileName));
-                        } catch (Exception e) {
-                            collectError(queue, reportError, e, fileName);
-                            return null;
-                        }
-                    });
-                } else {
+            // Periodic operations cannot be schema operations, so no need to check that here (will fail as invalid
+            // query)
+            if (isPeriodicOperation(stmt)) {
```
loveleif, line 257: This check is not fool proof (it's another query string regex). Could we remove this branch?

full/src/main/java/apoc/cypher/CypherExtended.java
```diff
@@ -246,40 +246,46 @@ private void runDataStatementsInTx(
...
+            } else {
+                AtomicBoolean isSchemaError = new AtomicBoolean(false);
+                try {
                     Util.inTx(db, pools, threadTx -> {
                         try (Result result = threadTx.execute(stmt, params)) {
                             return consumeResult(result, queue, addStatistics, tx, fileName);
                         } catch (Exception e) {
-                            collectError(queue, reportError, e, fileName);
+                            // APOC historically skips schema operations
+                            if (!e.getMessage().contains("Schema operations on database 'neo4j' are not allowed")) {
```
loveleif, line 275: It would be nice if we could use the error code here instead of parsing the message, but I assume it's not possible?

loveleif, line 275: `on database 'neo4j'`, what about databases that has another name?

gem-neo4j: The error code was just for "forbidden" things, which felt kinda broad, I guess once the GQL status codes start we will be able to do that more :D

gem-neo4j: Ah yes 🙈

full/src/main/java/apoc/cypher/CypherExtended.java
gem-neo4j: This is 4.4 extended, so I was just leaving the behaviour as it was to not break anything

gem-neo4j:
cc @loveleif :)

full/src/main/java/apoc/cypher/CypherExtended.java
```diff
@@ -272,7 +272,8 @@ private void runDataStatementsInTx(
                             return consumeResult(result, queue, addStatistics, tx, fileName);
                         } catch (Exception e) {
                             // APOC historically skips schema operations
-                            if (!e.getMessage().contains("Schema operations on database 'neo4j' are not allowed")) {
+                            if (!(e.getMessage().contains("Schema operations on database")
+                                    && e.getMessage().contains("are not allowed"))) {
```
loveleif, line 276: YOLO

loveleif (approved):

PR #4170, opened 2024-08-16
[NOID] Add note about how to install APOC Core

gem-neo4j:
User's are getting confused why APOC core isn't automatically installed with APOC extended, this adds a note to the docs about it.


jexp (approved):
LGTM, thank you @gem-neo4j

PR #4177, opened 2024-08-30
Fixes #4138: Add support for loading Gephi GEXF file format

### LINKED ISSUE #4138: Add support for loading Gephi GEXF file format
It's very similar to GraphML so we can reuse parts of the GraphML loader

There are a number of examples listed here, which we can also use as test-cases.
https://gexf.net/data.html

Only question is how we represent node hierarchies, we can either leave them off or introduce a relationship `__CONTAINS__` from parent node to children. (should be configured)

viz: attributes can optionally (config) be stored as attributes of the node, but in principle we could also write Bloom Perspective nodes.

vga91:
Fixes #4138

- Added procedures for reading GEXF files. 
- Copied and pasted code from APOC Core and put it into `XmlUtil.java`
- Changed parseXML method to be able to handle properties of gexf files. To handle the name of relations we parse the `kind` tag. While for node properties we change from doing the parse of the `data` tag to the `attvalue` tag. Also, to parse the label in the `node` tag you parse `labels` instead of `labels` as compared to how it is done for graphml.
- Multiple label support for nodes. Example `<node id=“0” label=“:Gephi:Webatlas”>`.
- Types tested: `string, integer, long, foat, double, boolean, liststring, listboolean`
- Improved `ExtendedUtil.convertValue` with toLowerCase(), to be used by both Parquet/Arrow and Gefx formats
- Improved ExtendedTestUtil
  - added arrays assertion to `assertMapEquals`
  - added `assertRelationship`

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/import/gexf.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/import/gexf.adoc
@@ -1,0 +1,222 @@
+[[gexf]]
+= Load GEXF (Graph Exchange XML Format)
+:description: This section describes procedures that can be used to import data from GEXF files.
+
+
+
+Many existing applications and data integrations use GEXF to describes a graph with nodes and edges.
+For further information, you should visit the https://gexf.net/[official documentation].
+
+It is possible to load or import nodes and relationship from a GEXF file with the procedures
+ `apoc.load.gexf` and `apoc.import.gexf`. You need to:
+
+* provide a path to a GEXF file
+* provide configuration (optional)
+
+The `apoc.import.gexf` read as the `apoc.load.gexf` but also create nodes and relationships in Neo4j.
+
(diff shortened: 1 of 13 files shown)


docs/asciidoc/modules/ROOT/pages/import/gexf.adoc
```diff
@@ -0,0 +1,222 @@
+[[gexf]]
+= Load GEXF (Graph Exchange XML Format)
+:description: This section describes procedures that can be used to import data from GEXF files.
+
+
+
+Many existing applications and data integrations use GEXF to describes a graph with nodes and edges.
```
RobertoSannino, line 7: nodes and relationships OR vertices and edges

RobertoSannino: add also documentation example for malformed GEXF files

PR #4190, opened 2024-09-18
Fixes #4182: The huggingface examples return strange results

### LINKED ISSUE #4182: The following huggingface example returns strange results
The following huggingface example, placed in `openai.adoc`, 
returns strange results, e.g. `orange` or `事件! asking why should I be thankful! `

```
CALL apoc.ml.openai.completion('What color is the sky? Answer in one word: ', $huggingFaceApiKey, 
{endpoint: 'https://api-inference.huggingface.co/models/gpt2', apiType: 'HUGGINGFACE', model: 'gpt2', path: ''})
```

vga91:
Fixes #4182

With `gpt2` and similar, the answers are not always valid.
Changed test case to `google-bert/bert-base-uncased`

Added `PATH_CONF_KEY` and `METHOD_KEY` as optional configurations, if not present, since some models fail with `method: 'GET'`, 

and the `path: ''` causes a fail or an incorrect result depending on the model.

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/ml/openai.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/ml/openai.adoc
@@ -161,9 +161,11 @@ For the https://huggingface.co/[HuggingFace API], we have to define the config `
 For example:
 [source,cypher]
 ----
-CALL apoc.ml.openai.completion('What color is the sky? Answer in one word: ', $huggingFaceApiKey, 
-{endpoint: 'https://api-inference.huggingface.co/models/gpt2', apiType: 'HUGGINGFACE', model: 'gpt2', path: ''})
+CALL apoc.ml.openai.completion('The sky has a [MASK] color', $huggingFaceApiKey,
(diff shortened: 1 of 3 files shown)


docs/asciidoc/modules/ROOT/pages/ml/openai.adoc
```diff
@@ -161,10 +161,12 @@ For the https://huggingface.co/[HuggingFace API], we have to define the config `
 For example:
 [source,cypher]
 ----
-CALL apoc.ml.openai.completion('What color is the sky? Answer in one word: ', $huggingFaceApiKey, 
-{endpoint: 'https://api-inference.huggingface.co/models/gpt2', apiType: 'HUGGINGFACE', model: 'gpt2', path: ''})
+CALL apoc.ml.openai.completion('The sky has a [MASK] color', $huggingFaceApiKey,
```
RobertoSannino, line 164: [MASK] is the color of the sky

PR #4208, opened 2024-10-16
Cypher25 APOC proc func migration

gem-neo4j:
Migrating procedures to APOC for next major (from APOC core).

Note: This also
* Fixes the build, so it works with latest unpublished Neo
* Updates gradle
* Fixes the kotlin
* Updates some dependencies including testContainers which was the timeout bug

There is another PR to come with docs for this, but that one can't be merged until we get a new branch etc next year


docs/asciidoc/modules/ROOT/pages/ml/openai.adoc
```diff
@@ -18,7 +18,7 @@ All the following procedures can have the following APOC config, i.e. in `apoc.c
 | apoc.ml.openai.url | the OpenAI endpoint base url | `https://api.openai.com/v1` by default, 
     `https://api.anthropic.com/v1` if `apoc.ml.openai.type=<ANTHROPIC>`,
     or empty string if `apoc.ml.openai.type=<AZURE OR HUGGINGFACE>`
-| apoc.ml.azure.api.version | in case of `apoc.ml.openai.type=AZURE`, indicates the `api-version` to be passed after the `?api-version=` url
+| apoc.ml.azure.api.version | in case of `apoc.ml.openai.type=AZURE`, indicates the `api-version` to be passed after the `?api-version=` url |
```
vga91, line 21: Maybe I would add something like `<empty string>` or `""` after the final `|`?

extended-it/src/test/java/apoc/MissingExtraDependenciesTest.java
```diff
@@ -47,25 +47,6 @@ public static void tearDown() {
...
-        assertParquetFails("CALL apoc.export.parquet.query.stream('MATCH (n:ParquetNode) RETURN n')");
-
-        // import and load
-        assertParquetFails("CALL apoc.import.parquet('test.parquet')");
-        assertParquetFails("CALL apoc.load.parquet('test.parquet')");
-    }
-
     private static void assertParquetFails(String query) {
         assertFails(query, PARQUET_MISSING_DEPS_ERROR);
     }
```
vga91, line 52: I think this method is not used anymore

extended/src/test/java/apoc/log/Neo4jLogStreamTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public void teardown() {
+        dbManagementService.shutdown();
+    }
+
+    @Test
+    public void testLogStream() {
+        testResult(db, "CYPHER 25 CALL apoc.log.stream('debug.log')", res -> {
+            final String wholeFile =
+                    Iterators.stream(res.<String>columnAs("line")).collect(Collectors.joining(""));
+            assertTrue(wholeFile.contains("apoc.import.file.enabled=false"));
```
vga91, line 66: Just to understand how it works.

Would this test (and the others too) work without `CYPHER 25`?
And what is the difference between the test without the prefix?

vga91 (changes requested):
Nice work! :)
Just 2 little suggestions and one question and it's OK

extended/src/test/java/apoc/log/Neo4jLogStreamTest.java
gem-neo4j: They don't work without Cypher 25 at the moment because we haven't got Cypher Versions 100% done in Neo4j yet, but soon we want to be able to change the default version, and when the default version is Cypher 25, then the test wouldn't need the prefix. Basically, these moved procedures only exist on Cypher 25 and not Cypher 5 for extended, and they only exist on Cypher 5 and not 25 for Core :)

vga91: Ah ok, got it thanks :)

PR #4223, opened 2024-11-08
Fixes #4197: apoc.monitor.store fails for block stores

### LINKED ISSUE #4197: apoc.monitor.store fails for block stores
The following procedure call : `CALL apoc.monitor.store()` fails when database stores are in block format.

The procedure should return store statistics

```
Failed to invoke procedure `apoc.monitor.store`: Caused by: java.lang.IllegalArgumentException: PlainDatabaseLayout{databaseDirectory=/Users/cwillemsen/dev/_graphs/actro/data/databases/neo4j, transactionLogsDirectory=/Users/cwillemsen/dev/_graphs/actro/data/transactions/neo4j} does not describe a record storage database.
```

1. Run Neo4j 5.23
2. Execute the following query in the browser : `CALL apoc.monitor.store()` 

## Versions

  - OS: N/A
  - Neo4j: 5.23
  - Neo4j-Apoc: 5.23

vga91:
Fixes #4197

The `RecordDatabaseLayout.cast()` method works only for embedded, 
while `RecordDatabaseLayout.convert()` works both with embedded db and with neo4j enterprise docker / desktop.

Therefore, added a dockerized test as well.

### DIFF
--- a/extended-it/src/test/java/apoc/monitor/StoreInfoProcedureEnterpriseTest.java
+++ b/extended-it/src/test/java/apoc/monitor/StoreInfoProcedureEnterpriseTest.java
@@ -1,0 +1,47 @@
+package apoc.monitor;
+
+import apoc.util.Neo4jContainerExtension;
+import apoc.util.TestContainerUtil;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import org.neo4j.driver.Session;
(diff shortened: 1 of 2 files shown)


vga91:
todo -

check that https://github.com/neo4j/apoc/blob/dev/processor/src/main/java/apoc/processor/ApocProcessor.java

PR #4225, opened 2024-11-11
Fix ApocSignatures.PROCEDURES not found in StartupExtendedTest due to Cypher25 introduction

vga91:
TODO: waiting for https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/4230 merge??

Fix the cannot find symbol CI error, see [here](https://github.com/neo4j-contrib/neo4j-apoc-procedures/actions/runs/11707717937/job/32716503891?pr=4222#step:10:1855)

```
/home/runner/work/neo4j-apoc-procedures/neo4j-apoc-procedures/extended-it/src/test/java/apoc/neo4j/docker/StartupExtendedTest.java:119: error: cannot find symbol
        assertEquals(sorted(ApocSignatures.PROCEDURES), procedureNames);
```

The error is caused by the following changes in core, 
- https://github.com/neo4j/apoc/commit/8509584bf5d77efd77dd03f166757777072c7541
- https://github.com/neo4j/apoc/commit/f0c8172dac61db06fd6a02879d8794bbe993b54c

which reflects apoc extended: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/4208#discussion_r1816613431

Therefore, changed tests in a similar way as [StartupCore test  ](https://github.com/neo4j/apoc/commit/f0c8172dac61db06fd6a02879d8794bbe993b54c#diff-1e523af12b4239c89ed9f1656e99912150f81a5fc31e7492beb7fd79a56203b5R159)

----

If we don't change anything we have this error: https://github.com/neo4j-contrib/neo4j-apoc-procedures/actions/runs/11777623730/job/32802388039?pr=4225#step:10:4819

If we put the procedure in extended.txt we have these 2 errors: https://github.com/neo4j-contrib/neo4j-apoc-procedures/actions/runs/11778250428/job/32804296390?pr=4225

This because load.jsonParams currently is in both cypher5 core and cypher25 extended.
So we have remove it manually (waiting for core removal).

PR #4230, opened 2024-11-18
Java 21 and 2024.12.0 bump

gem-neo4j:
Update to Java 21 and Neo 2024.12.0


vga91:
@gem-neo4j 

I don't know how far along this PR is, but I think this commit is also needed to make the CI green: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/4225/commits/505ecd59e23ec549e1233947a9d15326bb3d2f05.

Otherwise, we'll get this error: https://github.com/neo4j-contrib/neo4j-apoc-procedures/actions/runs/11707717937/job/32716503891?pr=4222#step:10:1855.

Or at least, I had this error before the Java 21-related errors, I guess it's still present..

gem-neo4j:
Ah thanks! will take a look :)

vga91:
@gem-neo4j 
To solve CI error, I guess we have to change [here](https://github.com/neo4j/apoc/blob/dev/core/src/main/java/apoc/help/Help.java#L48), 
to handle "extendedCypher25.txt" and "extendedCypher5.txt" instead of / besides "extended.txt" :)

gem-neo4j:
@vga91 yep, I have another PR for that on core :) (was out sick so sorry for late reply), will try to get it all working this week!

extended/src/main/java/apoc/load/LoadJsonExtended.java
```diff
@@ -46,8 +46,7 @@ public class LoadJsonExtended {
     public URLAccessChecker urlAccessChecker;
```
vga91, line 46: `LoadJsonExtended` class needs to have `@Extended`

PR #4236, opened 2024-11-19
[NOID] Fixes #4141: apoc.load.jdbcUpdate inside apoc.periodic.iterate leavesidle connections from 5.19 and forward (#4196)

### LINKED ISSUE #4141: apoc.load.jdbcUpdate inside apoc.periodic.iterate leaves idle connections from 5.19 and forward
This is issue #4074, but since I cannot reopen it, I am opening a new issue.

When running a apoc.load.jdbcUpdate operation as the second statement in an apoc.periodic.iterate, the connections to the database (postgres) should be closed after they have been executed.

This specifically happened when 5.19 was released. It still works in 5.18.

The previous PR suggested that this was a driver problem and not a problem with apoc. But changing the postgres driver doesn't do anything: in 5.18 everything works fine and from 5.19 onwards it does not work. So a change that breaks this functionality was introduced in 5.19. It should be possible to isolate the breaking change.

It does not close the connections, and it leaves an idle connection for each statement. When the postgres database reaches max connections the statement fails.

### Steps (Mandatory)

  1. Create a table in a postgres database:
`CREATE TABLE nodes ( 	id serial PRIMARY KEY, 	my_id integer );`
  2. In you neo4j database, create some nodes:
`WITH range(0, 100) as list UNWIND list as l CREATE (n:MyNode{id: l})`
  3. Load the postgres driver:
CALL apoc.load.driver("org.postgresql.Driver")
  4. Try to use apoc.period.iterate to insert into the postgres table: 
  `CALL apoc.periodic.iterate("MATCH (n:MyNode) return n", "WITH n, apoc.text.format('insert into nodes (my_id) values (\\\'%d\\\')',[n.id]) AS sql CALL apoc.load.jdbcUpdate('jdbc:postgresql://my_user:my_password@localhost:5432/my_database',sql) YIELD row AS row2 return row2,n", {batchsize: 10,parallel: true})  yield batches,total,timeTaken,committedOperations,failedOperations,failedBatches,retries,errorMessages, operations, failedParams, updateStatistics return batches,total,timeTaken, committedOperations,failedOperations,failedBatches,retries,errorMessages,operations,failedParams, updateStatistics`
  5. Check postgres for connections:
`select client_addr, state from pg_stat_activity to see the hanging connections from the neo4j server. They can be killed with select pg_terminate_backend(pid) from pg_stat_activity where state = 'idle' but a pg_cancel_backend will not work.`

## Specifications (Mandatory)

Currently used versions

### Versions
- OS: Ubuntu 22.04.4
- Neo4j: 5.19.0
- Neo4j-Apoc: 5.19.0
- pg driver: postgresql-42.7.3.jar
- Postgres version: 16.2

vga91:
Fixes #4141

Cherry-pick https://github.com/neo4j-contrib/neo4j-apoc-procedures/commit/966221d2f4be3ca2d5b43c344e7fd87c4e4bc473

### DIFF
--- a/LICENSES.txt
+++ b/LICENSES.txt
@@ -1918,7 +1918,7 @@ Eclipse Public License - v 1.0
   eclipse-collections-10.4.0.jar
   eclipse-collections-api-10.4.0.jar
   junit-4.13.2.jar
-  schemacrawler-15.04.01.jar
+  schemacrawler-16.22.2.jar
 ------------------------------------------------------------------------------
 
 Eclipse Public License - v 1.0
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -359,7 +359,7 @@ Eclipse Public License - v 1.0
   eclipse-collections-10.4.0.jar
   eclipse-collections-api-10.4.0.jar
   junit-4.13.2.jar
-  schemacrawler-15.04.01.jar
+  schemacrawler-16.22.2.jar
 
 Eclipse Public License - v 1.0
   logback-classic-1.2.10.jar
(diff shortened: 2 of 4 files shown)


vga91:
It should be valid only for 5.19+

PR #4241, opened 2024-11-20
Fixes #4231: The apoc.vectordb.milvus.query* and apoc.vectordb.weaviate.query* procedures should get the fields config from metadataKey if present

### LINKED ISSUE #4231: The `apoc.vectordb.milvus.query*` and `apoc.vectordb.weaviate.query*` procedures should get the fields config from metadataKey if present.
The `apoc.vectordb.milvus.query*` and `apoc.vectordb.weaviate.query*` procedures should get the fields config from metadataKey if present.

It should get the `fields` config from `mapping.mappingKey` config

It fails with error:

```
You have to define `field` list of parameter to be returned
```

Execute:
```
CALL apoc.vectordb.milvus.queryAndUpdate($host, 'test_collection', [0.2, 0.1, 0.9, 0.7], null, 5, 
{allResults: true, mapping: { relType: 'TEST',  entityKey: 'readID', metadataKey: 'foo' } })
```

vga91:
Fixes #4231

Added new method `addMetadataKeyToFields` in VectorDbUtil: the method extracts fields or creates a new ArrayList, evaluates whether the metadata map is full and extracts the metadataKey string. 
If the string is not empty or null it adds it to the fields list


extended/src/main/java/apoc/vectordb/VectorDbUtil.java
```diff
@@ -130,4 +134,18 @@ public static void methodAndPayloadNull(Map<String, Object> config) {
...
     }
+
+    public static List addMetadataKeyToFields(Map<String, Object> config) {
+        List listFields = (List) config.getOrDefault(FIELDS_KEY, new ArrayList<>());
+
+        Map<String, Object> mapping = (Map<String, Object>) config.get(MAPPING_KEY);
+
+        String metadataKey = MapUtils.getString(mapping, METADATA_KEY);
+
+        if (StringUtils.isNotEmpty(metadataKey)) {
```
RobertoSannino, line 145: Are you sure that the behaviour should be to always add the METADATA_KEY mapping? Or should we add them only when the FIELDS_KEY is empty?

extended/src/main/java/apoc/vectordb/VectorMappingConfig.java
```diff
@@ -15,6 +15,7 @@ enum MappingMode {
     public static final String EMBEDDING_KEY = "embeddingKey";
     public static final String SIMILARITY_KEY = "similarity";
     public static final String MODE_KEY = "mode";
+    public static final String NO_FIELDS_ERROR_MSG = "You have to define `field` list of parameter to be returned";
```
RobertoSannino, line 18: "You need to define either the 'field' list parameter or the 'metadataKey' string parameter"

PR #4250, opened 2024-11-26
Fixes #4242: The Pinecone APOC implementation is misleading

### LINKED ISSUE #4242: The Pinecone APOC implementation is misleading
In Pinecone, a collection is a backup of a (hardware) based index, and the index is what "we" consider to be a collection

but the existing implementation queries for info /collections, which is wrong, it uses apoc extended uses `pinecone/collections/<name>`, but it should be `pinecone/indexes/<name>.`..
What Pinecone calls collections are backups of pod based indexes.
What apoc / we call "collection" are simply indexes in pinecone.

## How to change

Change everything to `index` but keep the procedure names as-is, to be consistent with the other ones

vga91:
Fixes #4242

- Changed collection references to indexe ones, 
except to createCollection and deleteCollection names, to be consistent with other vector db procedures

- Moved `NOTE` about collection naming on the top of the documentation.

The Pinecone APIs are changed meanwhile, therefore we:
- increased `Thread.sleep`s where needed
- changed endpoints
- changed testCall to testRetryCallEventually where needed

### DIFF
--- a/docs/asciidoc/modules/ROOT/pages/database-integration/vectordb/pinecone.adoc
+++ b/docs/asciidoc/modules/ROOT/pages/database-integration/vectordb/pinecone.adoc
@@ -1,24 +1,32 @@
 
 = Pinecone
 
+[NOTE]
+====
+The procedures create/drop/handle an index, instead of a collection like the other vectordb procedures,
+since in Pinecone a collection is a static and non-queryable copy of an index.
+
+Anyway, the create / delete index procedures are named `.createCollection` and `.deleteCollection` to be consistent with the other.
+====
+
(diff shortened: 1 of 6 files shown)


docs/asciidoc/modules/ROOT/pages/database-integration/vectordb/pinecone.adoc
```diff
@@ -1,24 +1,32 @@
 
 = Pinecone
 
+[NOTE]
+====
+The procedures create/drop/handle an index, instead of a collection like the other vectordb procedures,
```
RobertoSannino, line 6: I dont understand the sentence, can we rewrite it to be more clear?

PR #4257, opened 2024-11-29
Update moved procedures to have new class names

gem-neo4j:
This is needed in the 5.26 branch, or if someone has both Core and Extended installed, there will be missing procedures!

### DIFF
--- a/extended/src/main/java/apoc/export/arrow/ExportArrowExtended.java
+++ b/extended/src/main/java/apoc/export/arrow/ExportArrowExtended.java
@@ -45,7 +45,7 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.stream.Stream;
 
-public class ExportArrow {
+public class ExportArrowExtended {
 
     @Context
     public Transaction tx;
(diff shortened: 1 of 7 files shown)


extended/src/main/java/apoc/log/Neo4jLogStreamExtended.java
```diff
@@ -41,7 +41,7 @@
  * @author moxious
  * @since 27.02.19
  */
-public class Neo4jLogStream {
+public class Neo4jLogStreamExtended {
```
vga91, line 44: `@Extended` annotation

extended/src/main/java/apoc/export/arrow/ExportArrowExtended.java
```diff
@@ -45,7 +45,7 @@
 import java.util.Map;
 import java.util.stream.Stream;
 
-public class ExportArrow {
+public class ExportArrowExtended {
```
vga91, line 48: `@Extended` annotation

extended/src/main/java/apoc/load/LoadJsonExtended.java
```diff
@@ -46,8 +46,7 @@ public class LoadJsonExtended {
     public URLAccessChecker urlAccessChecker;
 
     @SuppressWarnings("unchecked")
-    @Procedure(name = "apoc.load.jsonParams", deprecatedBy = "This procedure is being moved to APOC Extended.")
-    @Deprecated
+    @Procedure(name = "apoc.load.jsonParams")
```
vga91, line 49: LoadJsonExtended with `@Extended` as well

vga91 (changes requested):
Very nice :) 
Just some Extended here as well

PR #4277, opened 2024-12-06
Fixes #4246: Add apoc.node.match and apoc.relationship.match procedures

### LINKED ISSUE #4246: apoc.merge.node just for Match-only
See trello card with id `zM6zcdk8`

vga91:
Fixes #4246

Replicated the flow of the `apoc.node.merge` and `apoc.relationship.merge` procedures,
modifying the query to perform a match instead of a merge.


docs/asciidoc/modules/ROOT/pages/misc/match-entities.adoc
```diff
@@ -0,0 +1,213 @@
...
+// end::tabs[]
+
+.Results
+[opts="header"]
+|===
+| node
+| (:Person {name: "Billy Reviewer", lastSeen: 2020-11-24T11:33:39.319Z})
+|===
+
+But this procedure is mostly useful for matching nodes that have dynamic labels or properties.
```
RobertoSannino, line 75: I think it would be also great to mention its advantage in combination with apoc.json procedures, maybe adding another example as well (for rel as well)

vga91:
Changed `import org.apache.commons.collections.MapUtils;` with `import apoc.util.ExtendedMapUtils;` to avoid the ClassNotFound error

PR #4278, opened 2024-12-06
Remove stored map of Procedure Handlers

gem-neo4j:
Instead of storing a map of the custom procedure handler to reference as a listener, which can cause a memory leak, common has been updated to find listeners and then add them. This means we can remove the map.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "apoc-core"]
 	path = apoc-core
 	url = https://github.com/neo4j/apoc
-	branch = 5.26_availabilitylistener_registry_update
+	branch = 5.26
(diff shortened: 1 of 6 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "apoc-core"]
 	path = apoc-core
 	url = https://github.com/neo4j/apoc
-	branch = dev
```
gem-neo4j, line 4: This is temporary, was just me getting the submodules to sync, will remove when/if core is merged

vga91:
Added a commit where I removed `AvailabilityListener` and used `LifecycleAdapter` methods for procedure refresh.
Similar to what was implemented in TriggerHandler and the others.

PR #4339, opened 2025-01-24
Fixes #3610: Add advanced SQL analytics to streaming cypher rows before returning results to the client

### LINKED ISSUE #3610: Add advanced SQL analytics to streaming cypher rows before returning results to the client
Imagine there was an aggregation function, `apoc.agg.analytics("SQL statement", {row-map}, optional SQL parameters) yield map/row`.

By e.g. embedding duckdb JDBC we could create a temporary table/CTE over the inputs and run the provided SQL statement including WINDOW functions, PIVOT etc.

The JDBC driver would ofc, be an optional jar.

We could even have arbitrary JDBC URLs, where duckdb would just be one option that runs in process, to delegate complex analytics to an analytics db.

Or somehow as a subquery / docs?

vga91:
Fixes #3610

# Overview

Fixes `https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/3610`

Created an `apoc.jdbc.analytics(<cypherQuery>, <jdbcUrl>, <sqlQueryOverTemporaryTable>, <paramsList>, $config)` procedure.
Imho is better than use an aggregation function, because it's more customizable, since I can choose what results to get and how to get them, otherwise to do something like below, with movies_count I should put e.g. a parameter `aggKeys` and increase the complexity.
```
            MATCH (p:Person)-[r:ACTED_IN]->(m:Movie)
            RETURN 
                p.name AS actor, 
                m.genre AS genre, 
                r.roles AS roles, 
                COUNT(m) AS movies_count
```

The procedure executes 3 steps:
1.  creates a temporary table
2. populates the table starting from the `<cypherQuery>` param
3. queries the table with the `<sqlQuery>` param (which can be an complex analytics query) and returns data

## Changes implemented

- Created test with `apoc.load.jdbc*` for DuckDB in `extended/src/test/java/apoc/load/DuckDBJdbcTest.java`
  - If we put the same JDBC url for all tests, flakily a random test remains pending for no apparent reason
  - also, created a test like [this one](https://github.com/neo4j-contrib/neo4j-apoc-procedures/commit/f978417fb40085269ca23ed18dc317d6bef7b587#diff-b1f6d33f21197c9f6eb449ef65e6f52c5e94034081d7ec88903fcd4882383941R126) for DuckDB
- Fixed errors related to DuckDB with Time and Date SQL Type: 
  - https://github.com/vga91/neo4j-apoc-procedures/pull/468/files#diff-f8ec24426cb3f7100946579df5aed24135ebe4bb00d59c9b91be48c3545bfd92R235
  - https://github.com/vga91/neo4j-apoc-procedures/pull/468/files#diff-f8ec24426cb3f7100946579df5aed24135ebe4bb00d59c9b91be48c3545bfd92R263
  - with DuckDB we can create an in-memory database, where no data is persisted to disk , via `jdbc:duckdb:` URL. Therefore added fallback [here](https://github.com/vga91/neo4j-apoc-procedures/pull/468/files#diff-0171f05c81a6bab620195ef5ea254d27211706815ef8b59cbb63b90f2f61b310R31) to allow it.
- [Improved executeUpdate and executeQuery](https://github.com/vga91/neo4j-apoc-procedures/pull/468/files#diff-f8ec24426cb3f7100946579df5aed24135ebe4bb00d59c9b91be48c3545bfd92R71), in order to pass also the Connection object and get that one if is not null, otherwise the temporary table will be destroyed if we create a new Connection.
- Created `apoc.load.jdbc.analytics` tests for:
  - DuckDb
  - MySQL
  - PostgreSQL
- Added docs
- config getOrDefault("temp_table", "neo4j_tmp_table")

#### EDIT: Changes after review
- added DuckDB, Postgres and MySql mapping to handle different types of numbers, dates and duration
- improved tests with the missing data types
- added escape chars to create table, based on Provider conf, to handle reserved keyword like `time` in MySQL
- refactoring: moved all jdbc stuff to `apoc.load.jdbc` package

- [added blob and binary handling and improved fallback conversion](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/4339/commits/14ccf7b03b747b27ed0e2e8a774819676e5ec2bb)

## To evaluate

- (maybe in a different PR/Issue): Improve handling: some SQL databases like Microsoft SQL Server create temp table in a different way


docs/asciidoc/modules/ROOT/pages/database-integration/load-jdbc.adoc
```diff
@@ -525,3 +525,167 @@ CALL apoc.load.jdbcUpdate('jdbc:derby:derbyDB','UPDATE PERSON SET NAME = ? WHERE
 ----
 CALL apoc.load.jdbc('jdbc:derby:derbyDB', 'PERSON',[],{credentials:{user:'apoc',password:'Ap0c!#Db'}})
 ----
+
+== Load JDBC - Analytics
+
+You can use the `apoc.jdbc.analytics(<cypherQuery>, <jdbcUrl>, <sqlQueryOverTemporaryTable>, <paramsList>, $config)`
+to create a temporary table starting from a Cypher query
+and delegate complex analytics to the database defined JDBC URL.
```
RobertoSannino, line 533: defined by the JDBC URL

docs/asciidoc/modules/ROOT/pages/database-integration/load-jdbc.adoc
```diff
@@ -525,3 +525,167 @@ CALL apoc.load.jdbcUpdate('jdbc:derby:derbyDB','UPDATE PERSON SET NAME = ? WHERE
...
+      country,
+      name,
+      year,
+      population,
+      RANK() OVER (PARTITION BY country ORDER BY year DESC) rank
+  FROM 'neo4j_tmp_table'
+  ORDER BY rank, country, name;",
+ $params,
+ { provider: "POSTGRES" })
+----
```
RobertoSannino, line 691: Lets add a full example using an in-memory DuckDb to perform analytics

extended/src/main/java/apoc/load/Analytics.java
```diff
@@ -0,0 +1,164 @@
...
+    private static String formatSqlValue(Object x) {
+        final String stringValue = x.toString();
+        if (x instanceof Number) return stringValue;
+        return String.format("'%s'", stringValue.replace("'", "''"));
+    }
+
+    private String mapSqlType(Provider provider, Object value) {
+        return switch (provider) {
+            case MYSQL, POSTGRES -> {
+                if (value instanceof Number) yield "INTEGER";
```
RobertoSannino, line 155: what if the Number is a Float?

PR #4439, opened 2025-05-20
Bump antora yml

gem-neo4j:
Bump to be new latest release

### DIFF
--- a/docs/asciidoc/antora.yml
+++ b/docs/asciidoc/antora.yml
@@ -1,11 +1,11 @@
 name: apoc
-version: 5.0
+version: 2025.0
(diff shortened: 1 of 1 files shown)


docs/asciidoc/antora.yml
```diff
@@ -1,11 +1,11 @@
 name: apoc
-version: 5.0
+version: 2025.0
```
gem-neo4j, line 2: This is following how the 5.x series docs look, not sure if it is the best way, but keeps the root name at 2025 and then the actual release is correct. This is just to unblock the team publishing, so the extended team can make this better if they want :)

PR #4485, opened 2025-09-24
Make jdbc connection string sensitive

loveleif:
Annotate jdbc connection string as sensitive to not leak passwords in the query log.


jexp (approved):
Thank you lgtm

jexp (approved):

PR #4572, opened 2026-07-23
Add commit SHA's to the github actions versions

gem-neo4j:
Pin the most recent SHA version for these actions instead of just relying on the less secure versions.


JoelBergstrand (approved):
Okidoki!

hvub (approved):
Nothing strikes my fury

PR #4573, opened 2026-08-04
[NOID] Add full shas and remove trello traceability

gem-neo4j:
Security update of full SHA values, and removing outdated traceability


hvub (approved):
Exciting throwback. I have not looked at the failing CI.

PR #4574, opened 2026-08-05
[NOID] Add SHAs to github actions

gem-neo4j:
Security update of full SHA values


hvub (approved):
Exciting. I have not checked all the hashes but only two, as part of understanding what is going on. :-)

